Issue #117, opened 2026-02-12
Self-hosted sema doesn't attach instantiated generic symbols to call nodes


octalide:
## Summary

The self-hosted compiler's bootstrap (`make mach`) produces a binary with hundreds of unresolved symbols — all generic function calls. The root cause is that sema doesn't attach instantiated generic symbols to call AST nodes, forcing lowering to guess at mangled names incorrectly.

## Symptoms

`make mach` output shows two categories of unresolved symbols:

1. **Module-qualified generic calls** emit non-generic mangled names:
   - `allocator.allocator_alloc[u8](...)` → `_M3std9allocatorN15allocator_alloc` (missing `I2u8E` suffix)
   - Affects: `allocator_alloc`, `allocator_free`, `allocator_resize`, `allocator_zalloc`

2. **Unqualified generic calls** emit names without module prefix:
   - `ok[*Value, u8](x)` → `okIP5ValueK2u8E` (missing `_M` module prefix)
   - Affects: `ok`, `err`, `allocator_free` (when called unqualified)

## Root Cause

**cmach (correct)**: Sema resolves generic call → instantiates → attaches `inst_sym` to `call_node->symbol`. Lowering reads `func->symbol` and calls `symbol_linkage_name()`. Done.

**imach (broken)**: Sema instantiates the generic but **never sets `node.symbol`** on the call expression. Lowering extracts the base function name from AST text and attempts manual `I...E` name construction — but:
- Module-qualified calls (`NODE_EXPR_FIELD`) skip the `I...E` block entirely (guarded by `!is_module_qualified`)
- Unqualified calls construct `I...E` names but symbol lookup fails, leaving bare names without `_M` prefix

### 1. Sema: attach instantiated symbols to call nodes

At the two sites where `instantiate_generic_function()` is called:

**`resolve_expr_type()` (~line 3705)** — currently only handles `NODE_EXPR_IDENT` callees:
```mach
val mangled: str = instantiate_generic_function(sema, sym, type_args);
# MISSING: look up inst_sym by mangled name, set node.symbol = inst_sym::ptr
```
Also needs a new branch for `NODE_EXPR_FIELD` callees (module-qualified generic calls like `allocator.allocator_alloc[u8]()`) — resolve the module alias, find the generic symbol, instantiate, and attach.

**`instantiate_nested_generic_calls()` (~line 3371)** — same pattern: after instantiation, set `node.symbol`.

### 2. Lowering: use `node.symbol` instead of manual name construction

[...]

PR #121, opened 2026-02-12
fix: Sema doesn't resolve all generic call sites — 14 stdlib generi...

### LINKED ISSUE #119: Sema doesn't resolve all generic call sites — 14 stdlib generics still emit base names
## Summary

After #117 fixed the generic symbol attachment for most call sites, 14 stdlib generic functions still emit non-instantiated mangled names during `make mach`. These calls bypass the new `node.symbol` path because sema's generic resolution has gaps.

## Unresolved base names (no type params)

```
_M3std5types6optionN13option_unwrap
_M3std5types6optionN14option_is_none
_M3std5types6optionN14option_is_some
_M3std5types6optionN4some
_M3std5types6resultN12result_is_ok
_M3std5types6resultN13result_is_err
_M3std5types6resultN16result_unwrap_ok
_M3std5types6resultN17result_unwrap_err
_M3std5types6resultN2ok
_M3std5types6resultN3err
_M3std9allocatorN14allocator_free
_M3std9allocatorN15allocator_alloc
_M3std9allocatorN16allocator_resize
_M3std9allocatorN16allocator_zalloc
```

## Likely causes

The generic resolution in `resolve_expr_type()` for `NODE_EXPR_CALL` only triggers when:
1. `callee_type.kind == ty.TY_FUN` — the callee type resolves as a function type
2. `callee.kind == ast.NODE_EXPR_IDENT` — the call is a bare identifier
3. `sym.is_generic` — the symbol has the generic flag set
4. `sym.decl_node.info.fun_decl.ret_type != nil` — the declaration has a return type

Possible gaps:
- `is_generic` flag may not be set on imported stdlib symbols
- `ret_type` check excludes generics with no explicit return type annotation
- Callee type may not resolve to `TY_FUN` for some imported generics
- Some calls may be module-qualified but through a path not handled by `lookup_module_function`

## Fix

Investigate which of the 14 functions fail the generic resolution checks and why. The fix is likely relaxing the conditions or ensuring the `is_generic` flag propagates correctly through imports.

octalide:
Closes #119


octalide:
Closing: #119 was a stale-build artifact, resolved by #117. No code changes needed.

PR #124, opened 2026-02-12
fix: Self-hosted lowering uses `ret_size = 24` fallback for unresol...

### LINKED ISSUE #123: Self-hosted lowering uses `ret_size = 24` fallback for unresolved generic return types, causing stack corruption
## Summary

The `mach` binary (compiled by `imach`) crashes on `mach build .` with TOML parse errors because `file_size` is silently overwritten on the stack. The root cause is a `ret_size = 24` "conservative estimate" in `lower.mach` that produces wrong-sized struct copies when sema fails to resolve generic return types.

## Reproduction

```
mkdir /tmp/mach-test && cd /tmp/mach-test
echo -e 'id = "test"\nname = "test"' > mach.toml
./out/linux/bin/mach build .
# -> "expected '='" (should parse fine)
```

`strace` shows `read(3, "id ", 3)` — the file is 26 bytes but only 3 are read.

### Proximate: `ret_size = 24` fallback in `lower.mach`

In `src/compiler/masm/lower.mach` around lines 3646 and 3671, when a call's return type is `TY_GENERIC` (unresolved), the lowering falls back to:

```mach
ret_size = 24;  # conservative estimate as last resort
```

For `allocator_alloc[u8]` which returns `Result[*u8, str]` (16 bytes: 8-byte bool-padded tag + 8-byte union), this causes:
- 24-byte sret buffer allocation instead of 16
- 24-byte struct copy (3 qwords) instead of 2
- The 3rd qword overwrites the adjacent `file_size` local variable

### Upstream: Sema doesn't fully resolve generic return types

When `resolve_generic_call_return_type()` returns nil, `sema.mach:3853` falls through to:

```mach
node.ty = callee_type.ret_type;
```

This assigns the **uninstantiated** generic return type (`TY_GENERIC`) instead of the concrete `TY_RECORD` with the correct size. Lowering then hits the `TY_GENERIC` branch and uses the wrong 24-byte fallback.

The C bootstrap (`cmach`) has none of this complexity — it simply uses `expr->type->size` because sema always resolves types fully.

## Fix

1. **Remove** both `ret_size = 24` fallback lines — these mask bugs and produce incorrect codegen
2. **Fix sema** to always attach the concrete return type to generic call nodes so `node.ty` is never `TY_GENERIC` at lowering time
3. If a type is still `TY_GENERIC` at lowering, treat it as an error rather than guessing

## GDB Evidence

- `fstat` returns `st_size = 0x1a` (26) ✓
- `st.st_size` loaded correctly as `0x1a` at instruction `0x6414a9` ✓
- Hardware watchpoint on `file_size` catches overwrite: `26 → 3` at `0x6415d2` (`mov %rax, (%rcx)`) — this is the 3rd qword of a 24-byte `Result` copy writing to `rbp-0x450`
- `imach` (compiled by `cmach`) copies only 2 qwords for the same `Result` type — correct
- `mach` (compiled by `imach`) copies 3 qwords — wrong, matches `ret_size = 24`

octalide:
Closes #123


octalide:
Superseded by #125 and #126.

Issue #189, opened 2026-02-16
printf/variadic calls corrupt Writer fd on stack


octalide:
## Summary

When calling functions with variadic-style arguments (like `printf`), the argument handling corrupts nearby stack data. Specifically, the `Writer` struct's `fd` field gets overwritten with garbage, causing `write()` syscalls to use invalid file descriptors.

## Steps to Reproduce

1. Use `std.print` with `printf`-style formatting that involves multiple arguments
2. Run the program under `strace`
3. Observe `write()` calls with corrupted fd values (e.g., `write(779749048, ...)` instead of `write(1, ...)`)

## Observed via strace

```
write(779749048, "...", 26) = -1 EBADF (Bad file descriptor)
```

The fd `779749048` is clearly stack garbage overwriting the `Writer.fd` field.

## Current Workaround

Use separate `print.print()` and `print.u64()` calls instead of combined format strings with multiple arguments.

PR #239, opened 2026-02-16
fix: Port va_list / variadic fixes to cmach (C bootstrap)

### LINKED ISSUE #201: Port va_list / variadic fixes to cmach (C bootstrap)
The two variadic/printf fixes from #189 only apply to the self-hosted compiler source. The C bootstrap (`cmach`) still has the old broken behavior:

1. **va_list sized as pointer (8 bytes)** instead of the correct SysV x86_64 size (24 bytes) — causes `va_start` to overflow into adjacent stack slots
2. **va_named_gp off-by-one for sret functions** — subtracts 1 from the GP register count when sret is present, causing `gp_offset` to point at the wrong save-area slot

This only affects `imach` (the intermediate stage compiled by `cmach`). Since `imach` is just a stepping stone to `smach`/`mach`, this is low priority — but cleaning it up avoids confusing output during `make full` and keeps the bootstrap consistent.

### Files to update (in the C bootstrap)
- va_list type sizing (equivalent of `sema/expr.mach` change)
- `va_named_gp` / `va_named_stack_slots` calculation (equivalent of `decl.mach` change)

octalide:
Closes #201


octalide:
Both fixes from #189 are already present in the C bootstrap:

1. **va_list type sizing**: cmach's `type_get_builtin_va_list()` creates a proper 24-byte struct via `type_create_struct` (4 fields: 2x u32 + 2x ptr). Never had the pointer-sizing bug.
2. **va_named_gp sret off-by-one**: Fixed in #212 (commit 396b229) — removed `- arg_shift` subtraction.
3. **va_named_stack_slots**: Already uses actual `stack_i` count.

Closing as already resolved.

PR #314, opened 2026-02-17
fix: codegen: add is_addr to IRValue to distinguish aggregate addre...

### LINKED ISSUE #313: codegen: add is_addr to IRValue to distinguish aggregate addresses from scalar values
## Problem

`IRValue` uses `VAL_REG` for both "register holds a scalar value" and "register holds an address to aggregate data." Callers of `lower_expr` cannot distinguish the two, leading to:

1. **`emit_store`** uses `addr.size > ptr_size` as a heuristic to detect aggregate addresses — wrong for aggregates ≤ 8 bytes
2. **`lower_var_decl`** has an `is_lit_init` hack (checks AST node kind) to work around emit_store for literal initializers
3. **`lower_field_access`** has a `lower_lvalue` workaround specifically for union bases because `lower_identifier` loads small unions as scalars
4. **Assignment handler** has its own `size > ptr_size` heuristic for aggregate detection

All four are independently trying to recover type information that was lost when `lower_expr` returned an untyped `VAL_REG`.

## Fix

Add `is_addr: bool` to `IRValue`. Set it at the source, check it at the sink.

**Producers** (set `is_addr = true`):
- `lower_identifier` — aggregate path (TY_RECORD, TY_ARRAY, TY_UNION)
- `lower_array_lit` — returns address of stack array
- `lower_field_access` — global symbol aggregate path

**NOT set** (returns actual values):
- `lower_call` — function returns are values, not addresses
- `lower_struct_lit` — returns VAL_MEM (already unambiguous)

**Consumers** (check `src.is_addr`):
- `emit_store` — replaces `addr.size > ptr_size` heuristic
- Assignment handler — replaces `size > ptr_size` heuristic
- `lower_var_decl` — `is_lit_init` hack becomes unnecessary
- `lower_field_access` — `lower_lvalue` workaround becomes unnecessary

Closes #310 regression (TY_UNION in is_aggregate was reverted because emit_store couldn't handle small unions).

octalide:
Closes #313

### DIFF
--- a/src/compiler/masm/ir.mach
+++ b/src/compiler/masm/ir.mach
@@ -220,8 +220,11 @@ pub val CC_AE:   u8 = 10;
 # scale:      index scale (1, 2, 4, 8)
 # label:      label name (for label values)
 # size:       operand size in bytes
+# is_addr:    true when this register holds a pointer to aggregate data
+#             (set by lower_identifier, lower_array_lit for aggregate types)
 pub rec IRValue {
     kind:      IRValueKind;
+    is_addr:   bool;
     reg:       i32;
     imm:       i64;
     offset:    i64;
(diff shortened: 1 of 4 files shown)


octalide:
Superseded — issue #313 is closed and the workaround approach is no longer needed after #315/#318.

PR #581, opened 2026-02-22
fix: Inline asm immediates emitted as register operands after compa...

### LINKED ISSUE #580: Inline asm immediates emitted as register operands after compact IR changes
## Summary

Stage 3 build (smach→mach) segfaults at `_start` because inline assembly immediate operands are being emitted as register-register instructions instead of immediate-register.

## Reproduction

`make clean && make` — stages 1-2 pass, stage 3 crashes:
```
smach -> mach
Segmentation fault (core dumped)
```

## Root cause

`_start` in `mach-std/src/runtime/linux/x86_64.mach` uses inline asm:
```
and rsp, -16      # immediate operand
sub rsp, 8        # immediate operand
```

imach compiles this correctly (cmach codegen). But smach (compiled by imach) produces:
```asm
# expected:
and $0xfffffffffffffff0, %rsp
sub $0x8, %rsp

# actual:
and %rax, %rsp    # register instead of immediate
sub %rax, %rsp    # register instead of immediate → rsp = 0 → crash
```

## Likely location

Inline asm lowering (`src/compiler/lower/iasm.mach`) or x86_64 ISel/encoder (`src/compiler/isa/x86_64/`). The compact IR format changes from #546 likely broke how immediates are packed/unpacked for inline asm instructions.

octalide:
Closes #580

### DIFF
--- a/src/compiler/lower.mach
+++ b/src/compiler/lower.mach
@@ -338,6 +338,8 @@ pub fun emit(ctx: *LowerContext, op: ir.Operator, dst: ir.Operand,
     # --- CALL ---
     if (op == ir.OP_CALL) {
         # pre-ABI CALL: dst=callee (label/reg), src1=ret_meta, src2=result
+        # note: src2 is the result destination, not a value — it should always
+        # be REG or MEM (never IMM). No materialization needed here.
         if (dst.kind == ir.OPK_LABEL) {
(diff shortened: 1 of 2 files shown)


octalide:
Closing — will re-investigate on the current codebase.

PR #604, opened 2026-02-27
fix: codegen: centralized source materialization in emit() boundary

### LINKED ISSUE #603: codegen: centralized source materialization in emit() boundary
## Problem

The `emit()` function in `lower.mach` converts rich `Operand` types into compact 8-byte `Inst` records. `operand_to_reg` returns 0 for non-REG operands (OPK_MEM, OPK_IMM), silently producing wrong code for any instruction that expects register operands.

This affects all register-only ops: CMP, FTOI, conversions (SEXT/ZEXT/TRUNC), and ALU (ADD/SUB/MUL/etc).

## Fix

Replace all scattered per-handler MEM/IMM patches with a single centralized materialization block in `emit()`:
- After the MOV/LOAD/STORE/ADDR handlers (which legitimately handle MEM operands)
- Before CMP, conversions, and ALU handlers
- Materializes OPK_MEM → LOAD to vreg, OPK_IMM → MOV to vreg

Also includes:
- `materialize_to_reg` helper for ABI-abstract ops (ARG, RETVAL, SCARG, CALL result, BRCOND)
- ISel fix: x86 SHL/SHR/SAR register form requires shift count in CL/RCX
- Aggregate handling in call.mach/stmt.mach: LOAD for small (≤ptr_size), ADDR for large

octalide:
Closes #603

### DIFF
--- a/src/compiler/codegen/lower.mach
+++ b/src/compiler/codegen/lower.mach
@@ -400,7 +400,7 @@ pub fun emit(ctx: *LowerContext, op: ir.Operator, dst: ir.Operand,
         var param_idx: u8 = src1.imm::u8;
         if (src1.imm == -1) { param_idx = ir.PARAM_SRET; }
         or (src1.imm == -2) { param_idx = ir.PARAM_VARARGS; }
-        val byte_size: u8 = dst.imm::u8;
+        val byte_size: u16 = dst.imm::u16;
         val align: u8 = dst.size;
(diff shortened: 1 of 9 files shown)


octalide:
Changes applied to fix/608 (PR #611) which includes these fixes rebased onto current dev (post-#606 DataPool/u16 changes).

PR #605, opened 2026-02-27
fix: sema: embedded record field offsets are zero when layout not y...

### LINKED ISSUE #602: sema: embedded record field offsets are zero when layout not yet computed
## Problem

When a record type contains another record by value (not by pointer), the embedded record's `size` may still be `0` at the time `compute_record_layout` runs. This causes all subsequent field offsets to be wrong — shifted down by the embedded record's actual size.

### Example

```mach
rec Allocator {
    ctx:       *void;
    alloc_fn:  *fun(*void, usize, usize) *void;
    resize_fn: *fun(*void, *void, usize, usize, usize) *void;
    free_fn:   *fun(*void, *void, usize, usize) void;
}

rec Arena {
    backing: Allocator;  # 32 bytes by value
    buf:     *u8;
    cap:     usize;
    off:     usize;
    align:   usize;
}
```

Arena's field offsets should be: `backing=0, buf=32, cap=40, off=48, align=56`. Instead, all fields after `backing` get offsets 32 bytes too small because `Allocator.size` is `0` when Arena's layout is computed.

### Root cause

`resolve_fields` in `resolve.mach` resolves each field's type via `resolve_type_expr`, which calls `ensure_type_resolved`. But `ensure_type_resolved` only creates the type shell (sets `resolved_type != nil`) — it does not compute the layout (fields, size, align). When `compute_record_layout` later reads `ft.size` for the embedded record, it sees 0.

### Fix

After resolving a field's type in `resolve_fields`, if the type is a value-type record or union with `size == 0`, recursively call `resolve_fields` on its declaration node before calling `compute_record_layout`.

octalide:
Closes #602

### DIFF
--- a/src/compiler/frontend/sema/resolve.mach
+++ b/src/compiler/frontend/sema/resolve.mach
@@ -342,7 +342,12 @@ fun ensure_layout(sema: *Sema, mod: *SemaModule, ft_node: *ast.Node, ft: *type.T
         if (ft_node != nil) {
             val ft_sym: *symbol.Symbol = ft_node.symbol::*symbol.Symbol;
             if (ft_sym != nil && ft_sym.decl_node != nil) {
-                resolve_fields(sema, mod, ft_sym.decl_node);
+                var owner: *SemaModule = mod;
+                if (ft_sym.module_path != nil) {
+                    val found: *SemaModule = find_module(sema, ft_sym.module_path);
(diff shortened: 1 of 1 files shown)


octalide:
Superseded by fix/603 (PR #604) which includes these commits plus additional fixes.

PR #851, opened 2026-03-27
fix: Codegen pipeline audit: ISel physical register model, regalloc...

### LINKED ISSUE #850: Codegen pipeline audit: ISel physical register model, regalloc architecture, asm IOC processing
## Context

The work on #846 (DWARF/MIR wiring) exposed fundamental codegen bugs that were patched incrementally over 35+ commits. The patches fixed symptoms but degraded code quality and introduced workarounds. PR #847 was closed without merge. This issue tracks the proper fix based on a full codebase audit.

## Root cause

The ISel emits instructions that use physical registers (RDI, RSI, RCX, RAX, RDX, R10, R11) in multi-instruction sequences that are invisible to the register allocator. The regalloc accumulated ad-hoc workaround functions to partially compensate, but they miss cases. Production compilers (LLVM, GCC, Cranelift) solve this by either using virtual registers exclusively in the ISel or by modeling all implicit register effects on every instruction.

### `zero_mi()` infinite recursion — isel.mach:27
The function calls itself on its own local declaration: `var mi: isa.Inst = zero_mi()`. This was introduced during a bulk replacement and crashes all ISel code with stack overflow. **This exists on the feat/846 branch only** — dev is unaffected since #847 was not merged.

### Division implicit clobbers undeclared — isel.mach:530-558
`sel_div` uses RAX and RDX as implicit operands for x86 DIV/IDIV. No clobber mask set. The regalloc assigns live values to RAX/RDX across division instructions. Same class: shift instructions hardcode RCX (isel.mach:514-515) without declaring it.

### ISel physical register temporaries — isel.mach: sel_copy, sel_div, sel_binop, sel_syscall, sel_call_indirect
Physical registers are written in instruction sequences spanning multiple ISel-emitted instructions. The regalloc cannot see these usages through the `Inst` operand fields. Functions affected:
- `sel_copy`: RDI, RSI, RCX for `rep movsb`
- `sel_div`: RAX, RDX for division
- `sel_binop`: RCX for shift count
- `sel_syscall`: RAX, RDI, RSI, RDX, R10, R8, R9 for syscall ABI
- `sel_call_indirect`: R11 (scratch_reg) for callee pointer

### Asm IOC specs not processed — stmt.mach:333-370, isel.mach:1125-1143
The parser supports `in`, `out`, `clb` on asm blocks. The MIR lowering only processes `in` specs. `out` and `clb` are silently ignored. The ISel ignores all specs. The standard library uses the old `{var}` curly-brace syntax instead of the proper IOC syntax. Tracked separately in octalide/mach-std#180.

### Hardcoded aggregate threshold — expr.mach:254, 506; stmt.mach:94
`lower_ident` and `lower_assign` use `agg_sz > 8` instead of `ctx.max_reg_ret`. 9-16 byte aggregates (which fit in two registers on x86_64) are incorrectly returned as pointers instead of loaded as values.

### `buf_append` silent data loss — isa.mach:167
If `reallocate` fails, the instruction is silently dropped. Code generation continues with missing instructions.

### MAX_ALLOCAS silent overflow — frame.mach:57, 80
Functions with >512 allocas silently skip extras. `slot_offset` returns 0, causing stack corruption.

### No `clobbers` field on Inst
The `Inst` struct has no way to declare implicit register effects. A `clobbers: u32` bitmask field is needed (each bit = one physical register). The regalloc must read this mask and create fixed intervals. Every ISel instruction that touches physical registers beyond `dst`/`src1`/`src2` must set it.

### Regalloc workaround functions
`track_param_liveness`, `track_phys_dst_liveness`, `record_phys_operand` in regalloc.mach are heuristic patches compensating for missing ISel declarations. They should be replaced by the uniform clobber mask mechanism once it works.

### Frame size / callee-saved split responsibility
`compute_frame_size` in compile.mach returns `local_size + spill_size`. The prologue adds callee-saved space separately. This implicit coupling is fragile — new ISA backends will get it wrong.

### No CFG awareness in regalloc
The allocator treats the instruction buffer as linear with no knowledge of basic blocks or branches. Range splitting was attempted and disabled because it broke on conditional branches. The current workaround (callee-saved-only for call-crossing ranges) increases register pressure.

### `emit_pcopy_mov` uninitialized fields — compile.mach:428
Creates an Inst without zero-initialization. `src2` sub-fields and any future fields contain stack garbage.

### SRET_FLAG packed into size — expr.mach:572
`0x8000` OR'd with return size in the same u16. No validation that sizes fit. Same field used for both encoding flags and aggregate sizes in frame.mach:77-92.

### Spill scratch register conflict — regalloc.mach:897-913
If an instruction has both a spilled memory base register and a spilled src1, both reload into `scratch_reg2`, corrupting the base address.

### O(n^2) interference checking — regalloc.mach:451-463
`is_preg_free` scans all ranges for every register candidate. LLVM uses O(log n) interval trees.

## Dead code to remove

From regalloc.mach: `split_range_at`, `split_at_calls`, `emit_split_stores`, `emit_split_loads`, `find_split_sibling`, `ensure_split_slot`, `resolve_vreg_at_any`, `find_any_range_for_vreg`. LiveRange fields: `split_parent`, `split_child`, `split_next`. Variable `at_call`. Import `use os`.

From expr.mach: `use allocator`, `use mem` (unused imports).

## Consteval gaps

`eval_expr` in consteval.mach does not handle: `NK_EXPR_IDENT` (cross-constant references), `NK_EXPR_CAST`, `NK_EXPR_FIELD`. These produce zeroinit globals silently.

## Implementation plan

1. Add `clobbers: u32` to `Inst`. Zero-init all Inst construction sites.
2. Set clobbers on every ISel instruction with implicit physical register effects (div, shift, rep movsb, syscall).
3. Regalloc reads clobbers mask in `build_ranges`, creates fixed intervals. Remove `track_param_liveness`, `track_phys_dst_liveness` workarounds.
4. Wire `out` and `clb` asm specs through MIR lowering and ISel.
5. Replace hardcoded `> 8` with `ctx.max_reg_ret`.
6. Clean dead code from regalloc.
7. Fix `buf_append` to propagate allocation errors.
8. Fix `emit_pcopy_mov` initialization.
9. Fix consteval gaps.
10. Re-apply DWARF/MIR wiring from #846 on top of the clean foundation.

octalide:
Closes #850

### DIFF
--- a/src/compiler/mir/compile.mach
+++ b/src/compiler/mir/compile.mach
@@ -109,7 +109,7 @@ fun compile_function(func: *ir.Function, module: *ir.Module, tgt: *target.Target
 
     val total_frame: usize = compute_frame_size(layout.frame_size, ra_result.spill_size);
 
-    var resolved_buf: isa.InstBuf = resolve_parallel_copies(alloc, ?ra_result.out);
+    var resolved_buf: isa.InstBuf = resolve_parallel_copies(alloc, ?ra_result.out, tgt);
     isa.buf_dnit(?ra_result.out);
 
     var final_buf: isa.InstBuf = resolved_buf;
@@ -302,14 +302,20 @@ fun compute_frame_size(local_size: usize, spill_size: usize) usize {
     ret total;
 }
 
-fun resolve_parallel_copies(alloc: *allocator.Allocator, insts: *isa.InstBuf) isa.InstBuf {
+fun resolve_parallel_copies(alloc: *allocator.Allocator, insts: *isa.InstBuf,
+                            tgt: *target.Target) isa.InstBuf {
     var out: isa.InstBuf = isa.buf_init(alloc);
     var i: i32 = 0;
 
     for (i < insts.count) {
         val inst: isa.Inst = insts.insts[i];
 
-        if (inst.opcode == isa.ISA_PCOPY || inst.opcode == isa.ISA_PCOPY_FENCE) {
+        if (inst.opcode == isa.ISA_PCOPY_FENCE) {
+            resolve_pcopy_group(insts, ?out, i, tgt);
+            i = i + 1;
+            cnt;
+        }
+        if (inst.opcode == isa.ISA_PCOPY) {
             i = i + 1;
             cnt;
         }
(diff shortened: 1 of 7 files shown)


octalide:
Closing — second audit found additional issues introduced by these commits (prologue mi.size=0, ISel physical register gaps). Rolling into next issue.

PR #912, opened 2026-04-04
fix: isel: ZEXT 4→8 bytes incorrectly uses MOVZX

### LINKED ISSUE #892: isel: ZEXT 4→8 bytes incorrectly uses MOVZX
## Summary

`sel_conv` handles OP_ZEXT by emitting MOVZX, but x86_64 MOVZX only supports zero-extension from 1 or 2 bytes to larger sizes. It cannot zero-extend 4 bytes to 8 bytes.

For 32→64 zero extension, the correct approach is a MOV to a 32-bit register, which implicitly zero-extends the upper 32 bits on x86_64.

## Location

`src/compiler/target/isa/x86_64/isel.mach`, `sel_conv` OP_ZEXT handler

## Impact

Any u32→u64 or i32→u64 zero extension emits an invalid or incorrect MOVZX encoding.

## Notes

octalide:
Closes #892


octalide:

PR #913, opened 2026-04-04
fix: abi: SRET return classification uses RAX instead of RDI

### LINKED ISSUE #891: abi: SRET return classification uses RAX instead of RDI
## Summary

In `sysv_classify_return()`, when a return value is classified as CLASS_SRET (aggregate > 16 bytes), the register is set to `defs.RAX` instead of `defs.RDI`.

Per SysV64 ABI: the caller passes a hidden pointer as the **first parameter** in RDI, the callee writes the result to that pointer, and returns the pointer in RAX. The classification should set `slot.reg = defs.RDI` for the incoming parameter side.

## Location

`src/compiler/target/abi/sysv64.mach:214`

## Impact

All functions returning aggregates > 16 bytes receive the return pointer in the wrong register, causing memory corruption.

## Notes

octalide:
Closes #891


octalide:

PR #914, opened 2026-04-04
fix: remove duplicate 0x66 prefix in MOV reg,imm16 encoding

### LINKED ISSUE #890: encoder: double 0x66 prefix on MOV reg, imm16
## Summary

The x86_64 encoder emits the 0x66 operand-size prefix twice when encoding `MOV reg, imm16`. Line 387 of `encode.mach` checks `FLAG_16BIT` and conditionally emits 0x66, then line 394 (in the `size == 2` branch) emits 0x66 unconditionally again.

This produces `66 66 B8+reg imm16` instead of the correct `66 B8+reg imm16`.

## Location

`src/compiler/target/isa/x86_64/encode.mach:387-394`

## Impact

Any 16-bit MOV reg,imm instruction encodes with a malformed byte sequence.

## Notes

octalide:
## Summary
- Removes duplicate `0x66` operand-size prefix emission in `encode_mov` for `MOV reg, imm16`
- The `FLAG_16BIT` check on line 387 already emits `0x66`, but the `size == 2` branch unconditionally emitted it again, producing two prefix bytes instead of one

Closes #890

### DIFF
--- a/src/compiler/target/isa/x86_64/encode.mach
+++ b/src/compiler/target/isa/x86_64/encode.mach
@@ -392,7 +392,6 @@ fun encode_mov(mi: *isa.Inst, sec: *of.Section, alloc: *allocator.Allocator) i32
             emit_byte(sec, alloc, 0xB0 + reg_bits(dst.reg));
             emit_byte(sec, alloc, imm::u8);
         } or (size == 2) {
-            emit_byte(sec, alloc, 0x66);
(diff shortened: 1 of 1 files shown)


octalide:
Fix applied differently in main working tree - FLAG_16BIT needs to be set from ISel emit helpers, not just removing the encoder's 0x66

PR #915, opened 2026-04-04
fix: skip zero-sized aggregate operations in MIR lowering

### LINKED ISSUE #898: lowering: zero-sized aggregate operations silently dropped
## Summary

Multiple places in the MIR lowering use the pattern:
```
if (is_aggregate && sz > 0 && sz <= 8) { agg_store(...); }
or (is_aggregate && sz > 8) { copy(...); }
or { store(...); }
```

When `sz == 0`, neither the `agg_store` nor `copy` branch is taken, and the scalar `store` branch is used for an aggregate type, which is incorrect. The operation is effectively dropped or miscompiled.

## Locations

- `src/compiler/mir/lower/expr.mach:521-528` (assignment)
- `src/compiler/mir/lower/stmt.mach:94-101` (variable declaration)
- `src/compiler/mir/lower/stmt.mach:249-253` (return)
- `src/compiler/mir/lower/lower.mach:264-273` (parameter initialization)

## Impact

Zero-sized aggregates silently lose their assignment/initialization/return semantics.

## Notes

octalide:
## Summary

Fixes #898.

- Added early `sz == 0` checks for aggregate types in four MIR lowering locations (`lower_assign`, `lower_var_decl`, `lower_ret`, `lower_param`) so zero-sized aggregates are treated as no-ops instead of falling through to the scalar `store` branch.

## Test plan

- [ ] Verify zero-sized aggregate types (e.g. empty structs) no longer emit spurious store/copy instructions
- [ ] Confirm non-zero aggregates still lower correctly (small agg_store, large copy, scalar store)

### DIFF
--- a/src/compiler/mir/lower/expr.mach
+++ b/src/compiler/mir/lower/expr.mach
@@ -518,6 +518,9 @@ fun lower_assign(ctx: *lctx.LowerCtx, lhs_id: ast.NodeId, rhs_id: ast.NodeId,
 
     val tid: type.TypeId = lctx.node_type(ctx, rhs_id);
     val sz: u32 = lctx.type_size(ctx, tid);
+    if (lctx.is_aggregate(ctx, tid) && sz == 0) {
+        ret value;
+    }
     if (lctx.is_aggregate(ctx, tid) && sz > 0 && sz <= 8) {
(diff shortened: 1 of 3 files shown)


octalide:

PR #1108, opened 2026-06-06
fix(global): emit aggregate global initializers as initialized data

### LINKED ISSUE #1106: codegen: module-level aggregate global initializers dropped to zeroed .bss (no init synthesis)
## Summary
A module-level `val`/`var` whose initializer is an **aggregate literal** (struct literal, array literal, or a `str`/fat-pointer constant) is emitted as a **zeroed `.bss` slot** — its initializer data is silently dropped. Scalar global initializers are unaffected (they comptime-fold to `const_int`). Local (in-function) aggregate initializers are unaffected (they lower to runtime stores).

This breaks **all of `std.print`**: `std.filesystem` exposes `pub val stdout: File = File{fd: 1};` / `stderr: File = File{fd: 2};`. The `fd` reads back as **0**, so every print is written to fd 0 and silently vanishes (or segfaults for `str` globals whose pointer field is zeroed). Blocks #1100.

## Minimal repro
```mach
use std.runtime;
$main.symbol = "main";
val SCALAR: i64 = 7;
rec Thing { a: i64; b: i64; }
val STRUCT: Thing = Thing{a: 3, b: 9};
fun main(argc: i64, argv: **u8) i64 {
    ret SCALAR + STRUCT.a + STRUCT.b;   # expect 19; actual 7 (struct fields read 0)
}
```
`str` globals are worse: `val GREET: str = "hi";` zeroes the fat pointer → `str_len(GREET)` segfaults.

## Root cause (traced)
1. `src/lang/fe/comptime.mach:245` — comptime eval **explicitly rejects** struct literals (`"struct literals are not comptime-evaluable"`). Array/str aggregate constants likewise don't fold to a scalar `CTValue`.
2. `src/lang/me/lower/decl.mach:382-413` (`lower_global`) — a non-comptime initializer is left as a **zero placeholder** (`value.const_int(0, gty)`). The comments say the real init logic is "materialised into the module init function by the driver's init-synthesis step."
3. **That init-synthesis step does not exist.** `ir.Module.init_fn` (declared `ir.mach:206`, set to `none()` at `ir.mach:245`) is **never populated and never called** anywhere in `src/` or `src/cli/`. `std.runtime`'s `_start` calls `main` directly with no global-ctor pass.
4. `src/lang/me/ir/value.mach` — `Value` has **no aggregate-constant kind** (only `CONST_INT/FLOAT/NULL/BYTES/GLOBAL/FN`), so even a folded aggregate has nowhere to live.

The bootstrap fixpoint is unaffected because the compiler's own code never relies on a module-level aggregate global initializer.

## Proper fix (design)
Global initializers should be **compile-time constants emitted as initialized data** (`.data`/`.rodata`), not runtime-synthesized. Concretely:
- Add a structured constant-aggregate representation (a `VAL_CONST_AGG`, or a pre-serialized byte blob + a relocation side-table) to `value.Value`.
- In lowering (`lower_global` / a const-aggregate folder, layout being backend-owned like the existing `$size_of` path), walk a struct/array/str literal of constant leaves and place each leaf at its field offset, recording relocations for pointer/`str` leaves (address-of-global, rodata string ptr).
- Extend the backend data emitter (`be/codegen` + `target/of`) to serialize the aggregate constant into `.data`/`.rodata` with those relocations. The `str`-local path already proves the reloc machinery exists; generalize it to global initializers.

This is deterministic (no startup ordering / ASLR), needs no `_start` ctor, and matches the "globals are constant-initialized" model. If genuinely runtime global initializers are ever required, that is a separate `init_fn`-synthesis feature.

## Acceptance
- The repro returns 19; `str`/struct/(array) globals carry their data.
- `std.print` produces output (unblocks #1100).
- `make clean && make` + `mach == mach2` fixpoint + `tools/test/differential.sh` all still pass.

octalide:
Module-level aggregate global initializers (struct literals, array literals, `str` constants) were dropped to a zeroed `.bss` slot instead of carrying their data. This breaks all of `std.print` (`fs.stdout = File{fd:1}` reads `fd==0`) and segfaults on `str` globals (null pointer).

## Design

Global initializers are compile-time constants emitted as **initialized data** in `.data`/`.rodata`, with relocations for pointer/`str` leaves — no runtime constructor.

- New `VAL_CONST_AGG` value kind: a pre-serialized blob + a relocation side-table (`AggReloc`: offset -> symbol + addend) for pointer leaves.
- A const-aggregate folder in lowering walks struct/array/str literals whose leaves are all comptime constants, placing each leaf at its IR-type field offset and recording a relocation for `str`/address-of-global leaves (reusing the existing anonymous-rodata-global mechanism).
- `pack_globals` serializes the blob into `.data`/`.rodata` and attaches one `of.Relocation` per side-table entry — generalizing the existing section-agnostic relocation infra the `.text` path already uses.

Closes #1106

### DIFF
--- a/src/lang/be/codegen.mach
+++ b/src/lang/be/codegen.mach
@@ -23,12 +23,16 @@ use std.types.result.is_err;
 use std.types.result.unwrap_ok;
 use std.types.result.unwrap_err;
 use std.types.size.usize;
+use std.types.option.Option;
+use std.types.option.is_none;
+use std.types.option.unwrap;
 use std.allocator.allocate;
 
 use writer: std.io.writer;
 
 use sess:   mach.lang.session;
 use intern: mach.lang.intern;
+use float:  mach.lang.float;
 use target: mach.lang.target;
 use of:     mach.lang.target.of;
 
@@ -423,6 +427,14 @@ fun pack_relocs(image: *of.ObjectImage, enc: *encode.EncoderOutput,
 # image: object image being built
 # irmod: IR module whose globals are packed
 # ret:   true on success, or an error string
+# true when IR type `id` is a 32-bit float.
+fun is_f32_type(types: *ir_type.IrTypeTable, id: ir_type.IrTypeId) bool {
+    val t_opt: Option[*ir_type.IrType] = ir_type.get(types, id);
+    if (is_none[*ir_type.IrType](t_opt)) { ret false; }
+    val t: *ir_type.IrType = unwrap[*ir_type.IrType](t_opt);
+    ret t.kind == ir_type.IRT_FLOAT && t.data.bits == 32;
+}
+
 fun pack_globals(s: *sess.Session, tgt: *target.Target, image: *of.ObjectImage,
                  irmod: *ir.Module) Result[bool, str] {
     var i: u32 = 0;
@@ -475,7 +487,13 @@ fun pack_globals(s: *sess.Session, tgt: *target.Target, image: *of.ObjectImage,
                 val br: Result[*u8, str] = allocate[u8](s.alloc, gsz::usize);
                 if (is_err[*u8, str](br)) { ret err[bool, str](unwrap_err[*u8, str](br)); }
                 bytes = unwrap_ok[*u8, str](br);
-                val bits: u64 = g.init.data.int_lit;
+                # a comptime float is always an f64 bit pattern; a 32-bit float
+                # global must carry the IEEE-754 single pattern of its own width,
+                # not the low 4 bytes of the f64 (which would corrupt the value).
+                var bits: u64 = g.init.data.int_lit;
+                if (g.init.kind == value.VAL_CONST_FLOAT && is_f32_type(?irmod.types, g.ty)) {
+                    bits = float.f64_to_f32_bits(bits)::u64;
+                }
                 var bi: u32 = 0;
                 for (bi < gsz) {
                     bytes[bi] = (bits >> (bi::u64 * 8::u64))::u8;
@@ -595,6 +613,24 @@ fun pack_agg_relocs(s: *sess.Session, tgt: *target.Target, image: *of.ObjectImag
 
         val r: Result[bool, str] = obj.add_relocation(image, rel);
         if (is_err[bool, str](r)) { ret err[bool, str](unwrap_err[bool, str](r)); }
+
+        # a function-pointer leaf targeting a symbol this object has not defined is
+        # a cross-module reference: declare a STT_FUNC extern stub now. every local
+        # function is already defined (pack_symbols ran before pack_globals), so an
+        # undefined fn-target is genuinely external. object (global) targets are
+        # left to pack_data_reloc_externs, which a later-packed global may still
+        # define.
+        if (ar.is_fn && ar.symbol != intern.STR_NIL && !image_has_symbol(image, ar.symbol)) {
+            var xsym: of.Symbol;
+            xsym.name    = ar.symbol;
+            xsym.section = of.SECTION_EXTERNAL;
+            xsym.offset  = 0;
+            xsym.size    = 0;
+            xsym.flags   = of.SYM_OBJ_FLAG_FUNCTION | of.SYM_OBJ_FLAG_EXTERN | of.SYM_OBJ_FLAG_GLOBAL;
+
+            val rx: Result[u32, str] = obj.add_symbol(image, xsym);
+            if (is_err[u32, str](rx)) { ret err[bool, str](unwrap_err[u32, str](rx)); }
+        }
         i = i + 1;
     }
     ret ok[bool, str](true);
--- a/src/lang/be/codegen.mach
+++ b/src/lang/be/codegen.mach
@@ -418,6 +418,14 @@ fun pack_relocs(image: *of.ObjectImage, enc: *encode.EncoderOutput,
     ret ok[bool, str](true);
 }
 
+# true when IR type `id` is a 32-bit float.
+fun is_f32_type(types: *ir_type.IrTypeTable, id: ir_type.IrTypeId) bool {
+    val t_opt: Option[*ir_type.IrType] = ir_type.get(types, id);
+    if (is_none[*ir_type.IrType](t_opt)) { ret false; }
+    val t: *ir_type.IrType = unwrap[*ir_type.IrType](t_opt);
+    ret t.kind == ir_type.IRT_FLOAT && t.data.bits == 32;
+}
+
 # materialise every module global into the image. a global with an initializer
 # becomes a `.rodata` (immutable) or `.data` (mutable) section; one with no
 # initializer becomes a zero-filled `.bss` entry sized from its type. each global
@@ -427,14 +435,6 @@ fun pack_relocs(image: *of.ObjectImage, enc: *encode.EncoderOutput,
 # image: object image being built
 # irmod: IR module whose globals are packed
 # ret:   true on success, or an error string
-# true when IR type `id` is a 32-bit float.
-fun is_f32_type(types: *ir_type.IrTypeTable, id: ir_type.IrTypeId) bool {
-    val t_opt: Option[*ir_type.IrType] = ir_type.get(types, id);
-    if (is_none[*ir_type.IrType](t_opt)) { ret false; }
-    val t: *ir_type.IrType = unwrap[*ir_type.IrType](t_opt);
-    ret t.kind == ir_type.IRT_FLOAT && t.data.bits == 32;
-}
-
 fun pack_globals(s: *sess.Session, tgt: *target.Target, image: *of.ObjectImage,
                  irmod: *ir.Module) Result[bool, str] {
     var i: u32 = 0;
--- a/src/lang/be/codegen.mach
+++ b/src/lang/be/codegen.mach
@@ -187,7 +187,45 @@ fun pack_image(s: *sess.Session, tgt: *target.Target, irmod: *ir.Module,
         ret err[of.ObjectImage, str](unwrap_err[bool, str](rxr));
     }
 
+    # an aggregate constant's `&G` / function-pointer leaf may target a symbol
+    # defined in another module (cross-object reference). every such data-section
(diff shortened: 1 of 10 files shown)


octalide:
## Verification

All checks pass:

- **struct repro** (`Thing{a:3,b:9}` + scalar) returns **19** (was 7)
- **str global repro** (`val GREET: str = "hithere"`) returns **7**, no segfault (was SIGSEGV)
- **std.print**: a `println("hi")` program prints `hi` and exits 0 — `fs.stdout = File{fd:1}` now carries `fd==1`
- **edge cases**: nested struct with a `str` field, `[N]i64`, and `[N]struct` array globals all read back correctly (combined → 181)
- **non-constant aggregate global** (`Thing{a: side(), b: 9}`) fails with a clean compile error rather than silently zeroing or crashing
- `make clean && make` exits 0
- **fixpoint**: `mach build . -o mach2` is **byte-identical** to `mach`
- `tools/test/differential.sh`: all 10 inputs consistent (-O0 vs -O2, smach vs mach)
- `mach test` corpus: unaffected (only the intentional `fails_on_purpose` fails)

## Design

`str` is a thin `*char`, so a `str` leaf is one pointer slot with an abs64 relocation to an anonymous `.rodata` blob — the exact machinery the local string-literal path already uses (`add_rodata_bytes` + a `VAL_GLOBAL` reference). Generalized that to global initializers:

- `VAL_CONST_AGG` carries a pre-serialized blob + an `AggReloc` side-table (offset → symbol + addend), chosen because it composes directly with the existing data emitter (`pack_globals` already takes a bytes/len pair) and keeps the backend free of IR-type walking.
- `lower/constagg` folds struct/array/str literals into that blob at IR field offsets / array strides, mirroring the `$size_of` fold path (record layout is backend-owned).
- `pack_globals` serializes the blob and attaches the side-table via the section-agnostic `of.Relocation` infra; `of.reloc_kind_name` exposes the abs64 kind name. The ELF writer already emits `.rela.<section>` for any relocated section, so no object-format change was needed.

octalide:
## Review fixes (pushed a247e34)

Governing change: `try_const_aggregate` is now **best-effort**. A genuinely non-constant leaf returns `ok(none)` so `lower_global` falls back to the zero-placeholder path (pre-PR behavior). `err` is reserved for malformed input / internal-invariant violations only. The `place*` helpers thread a `FoldResult` out-param (`FOLD_OK` / `FOLD_FALLBACK`) to distinguish a clean fold from a bail-out.

All MUST-FIX items addressed:

1. **f32 narrowing** — comptime floats are f64; a sub-8-byte float leaf is re-encoded to its IEEE-754 width with integer-only round-to-nearest-even (`f64_to_f32_bits`). The f32 narrowing *cast* is unencodable on the backend, so this avoids it entirely. Verified: 1.5→0x3FC00000, 0.0→0x0, 0.1→0x3DCCCCCD (round), -2.0→0xC0000000.
2. **Determinism** — blob now `zallocate`d; padding / trailing pad / union inactive bytes / unwritten slots are deterministically zero. Fixpoint byte-identical.
3. **Array OOB** — write bounded by `array_count` (declared length), not literal length. Over-long literal no longer corrupts the heap; extra elements dropped.
4. **Non-constant leaf falls back** — `ok(none)`, not `err`. Verified: struct-with-`side()`, array-with-runtime-element both compile.
5. **`$size_of`/`$align_of`/`$offset_of` leaves** — folded via `try_lower_comptime_intrinsic` (same path the scalar global uses). Verified `[2]usize{$size_of(A), $size_of(B)}`.
6. **IRT_UNION leaf** — routes through the struct-literal path (active member at offset 0). Verified a `Result`-shaped struct-with-union-field global.
7. **IRT_FN leaf** — function-pointer field emits a reloc to the function symbol. Verified a dispatch table.
8. **Address-of-global** — `?G` / `?G.field` emits a reloc to the global symbol with a field-offset addend. Verified both (read 99 and 22).
9. **Zero-size aggregate** — falls back instead of erroring. Verified `rec Empty{}; val e: Empty = Empty{};`.
10. **`module_source` hoisted** to one `lower.module_source`, called from expr/stmt/decl/constagg (was copied 4×); unused `src` imports removed.

Codegen: `pack_data_reloc_externs` declares undefined symbols for cross-module `?G` / fn-ptr data-reloc targets (data analog of `pack_reloc_externs`).

## Verification (all pass)

| case | result |
|---|---|
| struct repro | 19 |
| str global | 7, no segfault |
| `println("hi")` | prints `hi` |
| f32 field | 0x3FC00000 (and 0.0 / 0.1-round / -2.0 correct) |
| union-field (Result-shaped) | 43 (carries data) |
| fn-ptr table | 21 (correct relocs) |
| `?G` / `?G.field` | 99 / 22 (correct relocs + addend) |
| `$size_of` aggregate | 20 (folds) |
| over-long array literal | compiles, heap not corrupted |
| runtime-element aggregate | compiles (falls back) |
| empty-struct global | compiles |
| `make clean && make` | exit 0 |
| fixpoint (`mach build . -o mach2` + cmp) | **byte-identical** |
| `tools/test/differential.sh` | all 10 consistent |
| `mach test` corpus | unaffected |

octalide:
## Round-2 review fixes (pushed 0db2997)

All MUST-FIX + SHOULD-FIX + low-priority items addressed and verified.

**MUST-FIX**
1. **Orphan `.Lc` rollback** — `try_const_aggregate` snapshots `m.global_len` at the top and truncates it back on every fallback/err exit. Verified `Rec{s:"ZQXJWKV", n:rt()}`: x lands in `.bss` (runtime 1) and the unique string `ZQXJWKV` is **absent** from the binary. A *successful* str-in-struct fold still emits the string (verified separately).
2. **Cross-module `?mod.G`** — `place_addr_of` now checks `symbol_linkage_name(cur)` at the top of each iteration, so a node resolve bound to a symbol (bare ident or `mod.G`) is the base, never walked as a struct object. Verified `?data.G` across modules → reads 77, **no spurious "no such field" diagnostic**.
3. **Standalone f32 scalar global** — codegen narrows the comptime f64 to its IEEE-754 single pattern for a 32-bit-float global. Verified at the byte level: `3.14f32` → `c3 f5 48 40`, `1.0f32` → `00 00 80 3f`, and the old f64-truncation bytes (`1f 85 eb 51`) are **absent**. (Runtime read-back of an f32 global is unreliable due to a pre-existing f32 load/store backend gap, so verification is on emitted bytes.)
4. **Top-level pointer global** — `val PADDR: *Point = ?P` routes through `place_ptr`/`place_addr_of` and emits the reloc. Verified → reads 50 (42+8), no NULL deref.

**SHOULD-FIX**
5. **fn-ptr extern STT_FUNC** — `AggReloc` gains `is_fn`; `pack_agg_relocs` declares the cross-module stub as `SYM_OBJ_FLAG_FUNCTION`. Verified: in `main.o`, the undefined cross-module handler is typed `FUNC`, not `OBJECT`.
6. Dropped the unreachable `shift >= 64` guard in the f64→f32 subnormal path.
7. `lower_global` calls `try_comptime` only when the aggregate fold did not apply (no wasted `comptime.eval`).

**Low-priority**
8. `place_scalar` guards the comptime-intrinsic result kind (`VAL_CONST_INT`) so a future non-int intrinsic can't emit raw union bits.
9. `symbol_linkage_name` doc corrected to match what it enforces.

The f64→f32 narrowing (and `f64_bits`) moved to a shared `mach.lang.float` leaf so the middle end and back end share one deterministic implementation (no `be`→`me` dependency).

## Verification (all pass)

All round-1/2 cases (struct=19, str=7, edge=181, f32 bytes, union=43, fn-ptr=21, `?G`=99, `?G.field`=22, `$size_of`=20, over-long array safe, runtime-element falls back, empty struct, nil ptr=7) **plus** the new round-2 repros: orphan rollback clean, cross-module `?mod.G`=77, top-level `?P`=50, cross-module fn-ptr=105 with STT_FUNC.

`println("hi")` prints; `make clean && make` exits 0; **fixpoint byte-identical**; `tools/test/differential.sh` all 10 consistent; `mach test` corpus unaffected.

octalide:
## Final-round fixes (pushed a8e0682)

1. **`?fn` is_fn invariant completed** — `place_addr_of`'s symbol-bound base push now sets `is_fn` from `expr.references_function(cur)` instead of literal `false`. Verified: `val P = ?h.handler` (cross-module address-of-function) emits the extern `handler` as **FUNC** with a correct `R_X86_64_64` reloc — previously it would have been STT_OBJECT.
2. **Stale init-synthesis comment fixed** — `lower_global`'s fall-through comment no longer claims a non-existent driver init-synthesis step. It now states the truth: a non-foldable initializer stays a `.bss` placeholder (`ir.Module.init_fn` is never populated), which is exactly why const-address pointer inits are routed through the folder. One source of truth with the `constagg.mach` comments.
3. **`is_f32_type` doc placement** — moved above `pack_globals`' doc block so each function's doc sits with its definition.
4. **`is_const_addr_init` tightened** — only admits an IDENT/MEMBER pointer init that `references_function`; a plain data-global name no longer enters the folder to `zalloc` a throwaway blob and immediately fall back.

## Re-verify (all green)

- Item 1: cross-module `?fn` extern is **STT_FUNC** (verified in `main.o`).
- Item 2: comment matches reality (no init-synthesis step).
- Full case suite (18 repros) passes: struct=19, str=7, edge=181, f32 bytes, union=43, fn-ptr=21, `?G`=99, `?G.field`=22, `$size_of`=20, over-long-array safe, runtime-element falls back, empty struct, nil ptr=7, top-level `?P`=50, cross-module `?mod.G`=77, cross-module fn-ptr=105, orphan rollback clean + success folds.
- `make clean && make` exits 0; **fixpoint byte-identical**; `tools/test/differential.sh` all 10 consistent; `mach test` corpus unaffected.

Note (out of scope, pre-existing): a *call through* a function-typed top-level global (`P(5)` where `val P: fun(...) = f`) segfaults — the call site lowers `P` as a PC32-direct reference instead of loading the stored pointer. The data side is correct (the reloc to the function is emitted properly); this is a separate latent call-lowering bug, not introduced here.

PR #1128, opened 2026-06-06
fix(x64): float-to-float bitcast cast miscompile (2.5::f64)

### LINKED ISSUE #1126: codegen: explicit ::f64 cast on a float literal miscompiles (2.5::f64 wrong, bare 2.5 correct)
Found during #1089 verification. An explicit `::f64` cast on a float literal (e.g. `2.5::f64`) miscompiles to a wrong value, while the bare literal `2.5` is correct. std uses bare float literals, so it's not on the bootstrap path, but it's a real codegen bug in the float-literal cast path. Likely the cast re-interprets or re-widths the already-f64 literal incorrectly. Repro: compare `val x: f64 = 2.5;` vs `val x: f64 = 2.5::f64;` at runtime.

octalide:
## Summary

An explicit `::f64` cast on a float literal (e.g. `2.5::f64`) miscompiled to a wrong value at runtime, while the bare literal `2.5` was correct.

## Root cause

A same-width float-to-float cast lowers to an `OP_BITCAST` (`src/lang/me/lower/expr.mach`, `lower_cast`). In x64 instruction selection, `MIR_BITCAST` was unconditionally rewritten to `x64.MOV` — a **general-purpose register** move. Because the XMM and GP register id spaces overlap (XMM0 == RAX, XMM1 == RCX, ...), the GP MOV moved the aliased integer register's bytes instead of the float value, corrupting it. This is the same float-GP-alias hazard the `retag_float_moves` pass already guards for `MIR_MOV` / `MIR_LOAD` / `MIR_STORE`.

## Fix

A `MIR_BITCAST` whose operands are FP-class is a float register copy, so it joins `retag_float_moves` and becomes `MIR_FMOV` (MOVSS / MOVSD) like the ABI / phi float copies. A cross-domain reinterpret never reaches this point: lowering routes int<->float through the CVT family, so a `BITCAST`'s `[dst, src]` operands always share a register class.

## Verification

- `2.5::f64`, `0.0::f64`, `(-1.5)::f64`, `3.14::f64`, `1.0e9::f64`, `1234567.89::f64` all equal their bare literals at runtime.
- `x::f64` and `(x::f64)::f64` where `x` is an f64 variable are correct.
- `make clean && make` builds; FIXPOINT byte-identical; `mach test` and differential pass (see PR thread for full results).

## Out of scope (escalated separately)

While verifying I found two **pre-existing, unrelated** float bugs (present on `dev` before this change, not caused or fixed here):
- `i64::f64` (SI_TO_FP) and the reverse conversion produce wrong values.
- Runtime float negation (`-x` on a non-constant f64) is wrong.

Closes #1126

### DIFF
--- a/src/lang/target/isa/x64/isel.mach
+++ b/src/lang/target/isa/x64/isel.mach
@@ -155,6 +155,11 @@ pub fun select(alloc: *Allocator, tgt: *target.Target, fn: *mir.MirFunction) Res
 # selection like the float arith ops) makes regalloc route a spilled float operand
 # to its frame slot and the encoder emit MOVSS / MOVSD; the alias-free SSE form
 # also keeps the GP allocator from reusing the float's (GP-id-aliased) register.
+# a same-width float-to-float MIR_BITCAST (`f64::f64`, `f32::f32`) is likewise a
+# float register copy — its `[dst, src]` operands are both FP-class — so it joins
+# the retag rather than selecting to an aliased GP MOV that would corrupt the value
+# (a cross-domain reinterpret never reaches here: lowering routes int<->float
+# through the CVT family, so a BITCAST's operands always share a register class).
 # run before selection so isel's integer MIR_MOV / MIR_LOAD / MIR_STORE arms only
 # ever see integer/pointer operations.
 fun retag_float_moves(fn: *mir.MirFunction) {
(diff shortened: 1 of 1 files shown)


octalide:
## Verification (built from this branch's HEAD)

| check | result |
|---|---|
| original issue repro: `2.5` == `2.5::f64` at runtime | PASS |
| `2.5`, `0.0`, `(-1.5)`, `3.14`, `1.0e9`, `1234567.89` cast == bare | PASS |
| `x::f64` and `(x::f64)::f64` for f64 variable `x` | PASS |
| `make clean && make` | exit 0 |
| FIXPOINT (`mach` rebuilds itself byte-identical) | PASS (byte-identical) |
| `smach` == `mach` | byte-identical |
| `mach test --cwd .` | 194 PASS / 0 FAIL |
| `differential.sh` smach-vs-mach (-O0) arm | 10/10 PASS |
| `differential.sh` -O0-vs-O2 arm | -O2 BUILD-FAIL (known unrelated -O2 bug) |

The -O2 arm fails because the `--release` pipeline build itself errors with `mir.lower_ir: gep base is neither a memory nor a symbol reference`, which reproduces identically on `dev` before this change — a pre-existing -O2 issue, not a float-cast regression.

## Out-of-scope bugs found during verification (not touched here)

Both present on `dev` before this PR, both deferred:
- `i64::f64` (SI_TO_FP) and `f64::i64` (FP_TO_SI) conversions produce wrong runtime values.
- Runtime float negation (`-x` for a non-constant f64) is wrong (float `OP_NEG` lowers through the integer `MIR_NEG` GP path).

PR #1130, opened 2026-06-06
feat: comptime value-param monomorphization ($if/$or on a $param)

### LINKED ISSUE #1075: comptime: support $if/$or on a comptime function parameter (value-param monomorphization)
Split from #1066(c). `$if (mode == ...)` on a comptime `$mode` function parameter fails because the function body is lowered once and shared, and the param is an ordinary runtime value — making it usable as a comptime constant requires **monomorphizing the body per distinct comptime-argument value**, which the compiler doesn't do today. Currently documented as unsupported (doc/language/comptime-control.md). Implement comptime value-param monomorphization so comptime params can drive `$if`/`$or`.

octalide:
Implements comptime value-parameter monomorphization so `$if`/`$or` can be gated on a comptime function parameter (`$mode`).

Closes #1075

## Approach

Mirrors generic-TYPE monomorphization, keyed on comptime VALUES (CTValues) instead of TypeIds. A function with >=1 `$`-parameter is a template; each call site folds the comptime args to a CTValue sequence keying a value-instance whose mangled linkage name encodes the values in a namespace disjoint from the type `I...E` scheme. The drain emits one PUB|WEAK body per distinct key with `$param -> CTValue` bindings pushed into the comptime context, so the existing comptime evaluator selects each `$if` arm with no evaluator change.

## Front end

- resolve / sema bind and type-check ALL arms of a `$if` chain gated on a comptime parameter (value not yet known), deferring arm selection to lowering.
- comptime params are excluded from the interned TYPE_FUN; calls are arity-checked against the AST param list, and each `$`-argument must be compile-time foldable.

## Middle end

- ValueInstance worklist + `enqueue_value_instance` (linear dedup by mangled name, determinism preserved).
- injective CTValue mangle encoder under a `K...E` namespace (FLOAT encodes raw f64 bits).
- value instances stripped of comptime params from signature, ABI, and runtime call args.
- instance-count cap guards unbounded comptime recursion.

### DIFF
--- a/doc/language/comptime-control.md
+++ b/doc/language/comptime-control.md
@@ -69,14 +69,25 @@ fun apply($mode: u8, n: i64) i64 {
 Rules:
 
 - The argument bound to a `$`-parameter must be a compile-time constant at the
-  call site; a runtime value is rejected with `comptime argument is not a
-  compile-time constant`.
+  call site (a literal, a `pub val`, or another comptime parameter); a runtime
+  value is rejected with `comptime argument is not a compile-time constant`. A
+  cross-module constant works, whether imported by bare name or as a qualified
+  member (`alias.CONST`).
+- Each arm gate of a comptime-parameter `$if` must itself be comptime-foldable:
+  its identifiers must all be comptime (the comptime parameters or comptime
+  constants). A gate referencing a runtime local/parameter is rejected.
+- A comptime parameter has no storage, so its address cannot be taken
+  (`?$mode` is rejected with `cannot take the address of a comptime parameter`).
 - Comptime parameters carry no runtime cost: they are stripped from the lowered
   signature and ABI, so only the runtime parameters are passed.
 - A comptime parameter may be mixed freely with runtime parameters in any order.
 - A comptime parameter on a **generic** function (`fun f[T]($mode: u8, ...)`) is
   not yet supported — combining a type instance with a value instance is a
   pending extension and is reported with a clear diagnostic.
+- The function may live in any module: a value-parameter instance is emitted
+  against its declaring module and folds its `$if` gates against that module's
+  own comptime constants, so a library can export a comptime-parameter function
+  gated on its own `pub val`s.
 - A comptime parameter may gate per-target asm safely, since each instance only
   compiles its taken arm:
 
--- a/doc/language/comptime-control.md
+++ b/doc/language/comptime-control.md
@@ -23,6 +23,7 @@ The condition must be a comptime expression. Common shapes:
 
 - `$mach.*` reads for target / build conditions
 - Comparisons of comptime constants (`pub val` declarations)
+- Comparisons of a comptime function parameter (`$mode`) — see below
 
 ## Examples
 
@@ -42,13 +43,39 @@ $or {
 
 ### Dispatch on a comptime function parameter
 
-> **Not yet supported.** A comptime function parameter (`$order: Order`) is
-> accepted in a signature, but it is *not* yet usable as a comptime constant
-> inside the body — `$if (order == RELAXED)` reports `identifier is not a
-> comptime constant in scope`. Per-call-site dispatch on a comptime value
-> parameter requires monomorphizing the body per distinct argument value,
-> which the compiler does not yet do. The shape below describes the intended
-> behavior; until it lands, branch on such a parameter with a runtime `if`.
+A comptime function parameter (`$mode: u8`) is a compile-time-known argument
+fixed per call site. `$if` / `$or` may branch on it: the compiler
+**monomorphizes** the function body once per distinct comptime-argument value,
+and each instance compiles only the arm its value selects.
+
+```mach
+val MODE_DOUBLE: u8 = 0;
+val MODE_SQUARE: u8 = 1;
+
+fun apply($mode: u8, n: i64) i64 {
+    $if (mode == MODE_DOUBLE) {
+        ret n + n;
+    }
+    $or (mode == MODE_SQUARE) {
+        ret n * n;
+    }
+    ret 0;
+}
+
+# apply(MODE_DOUBLE, ..) and apply(MODE_SQUARE, ..) emit two distinct bodies,
+# each carrying only its selected arm.
+```
+
+Rules:
+
+- The argument bound to a `$`-parameter must be a compile-time constant at the
+  call site; a runtime value is rejected with `comptime argument is not a
+  compile-time constant`.
+- Comptime parameters carry no runtime cost: they are stripped from the lowered
+  signature and ABI, so only the runtime parameters are passed.
+- A comptime parameter may be mixed freely with runtime parameters in any order.
+- A comptime parameter may gate per-target asm safely, since each instance only
+  compiles its taken arm:
 
 ```mach
 pub fun load($order: Order, ptr: *i64) i64 {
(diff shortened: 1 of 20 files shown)


octalide:
Verification (all green):
- `make clean && make` exits 0; self-host fixpoint is **byte-identical** (`cmp -s out/bin/mach out/bin/mach2`)
- `out/bin/mach test --cwd .` — **194 pass / 0 fail**
- `tools/test/differential.sh` — **10 inputs, all consistent** across both the -O0/-O2 arm and the smach-vs-mach arm

Risk handling:
- int/bool/str param keys all verified (apply/pick/tag examples)
- non-constant $-arg → clear `comptime argument is not a compile-time constant` at the arg span
- mixed comptime+runtime params ($op at ordinal 1) passes runtime args in the right registers
- bounded comptime recursion (`sum_to(10)=55`) compiles; unbounded recursion hits the instantiation cap with a clear error (no OOM/hang)
- mangle injectivity: CTValue encoder in a disjoint `K...E` namespace (FLOAT encodes raw f64 bits); WEAK dedup by mangled name
- comptime+runtime views kept consistent: comptime ordinals stripped from the semantic TYPE_FUN, the lowered IR sig/params, and the runtime call args; arity checked against the AST param count

Known limitation (clean diagnostic, not a crash): a callee that is BOTH generic (`[T]`) and comptime-param (`$mode`) needs a combined type+value instance — reported as "comptime value parameters on a generic function are not yet supported" and documented.

octalide:
## Adversarial-review fixes (6 confirmed bugs)

Merged origin/dev first (float fixes #1124/#1129/#1128, #1122 expr.mach, -O2 GEP fix #1131) — clean auto-merge, no conflicts.

**1. FLOAT mangle non-injective** (`mangle.mach`) — `write_encoded_value` used `v.data.f::u64` (numeric truncation). Now uses `float.f64_bits` (raw IEEE-754 bits). Verified: `scale(1.5)`/`scale(1.9)` emit DISTINCT weak symbols (`scaledKf3ff8...E` vs `scaledKf3ffe...E`) and correct results.

**2. Cross-module `lc.ctx` not swapped** (`lower.mach`, `session.mach`, `build.mach`) — each module's ComptimeCtx is now registered with the session; `emit_one_value_instance` binds `lc.ctx` to the ORIGIN module's ctx (save/restore on every path). Verified: a library fn gated on its own `pub val`s, called cross-module, folds correctly.

**3. Cross-module comptime-ARG fold** (`expr.mach`, `check.mach`) — `fold_comptime_arg` resolves a module-qualified member arg (`alias.CONST`) to the defining module's const; sema defers such args to lowering. Verified: `disp.combine(disp.OP_SUB, ...)` works.

**4. Module-val shadows `$param`** (`comptime.mach`) — `lookup` now scans most-recent-first so an appended `$param` binding shadows an earlier same-named `val`. Verified: `val mode=1` no longer overrides the `$mode` param. Fixpoint + 194 corpus unaffected.

**5. Deferred-arm constness unenforced** (`sema.mach`) — `check_comptime_gate` requires each param-gated arm gate to reference only comptime things; a runtime-gated `$or` now gets a span'd sema error, not a span-less lower abort.

**6. `?comptime_param` fabricated a symbol** (`infer.mach`, `expr.mach`) — taking the address of a comptime param is rejected at sema (`infer_unary` UN_ADDR) with a clean message; the lower path is guarded too.

## Re-verification (all green)
- `make clean && make` exits 0; **FIXPOINT byte-identical**
- `mach test --cwd .` — **194 pass / 0 fail**
- `differential.sh` — **10 inputs, all consistent** (both -O0/-O2 and smach-vs-mach arms)
- float distinctness, cross-module (bare + aliased), module-val/$param shadow (param wins), runtime-gated arm (clean error), `?param` (clean error), plus the existing suite: int/bool/str keys, mixed comptime+runtime ABI, bounded + unbounded recursion (cap), generic+comptime rejection.

octalide:
## Round-2 review fixes (7 issues)

**1. lookup-reversal regression** (`comptime.mach`) — restored `lookup` to first-match and added a DEDICATED param scope (`push_param`/`param_mark`/`param_truncate`) consulted before `user_consts`. Verified BOTH: a block-local `val MODE=0` in fn A no longer shadows `pub val MODE=1` for fn B's gate (`b=50`), AND a `$mode` param still shadows a module `val mode=1` (`apply(0)=14`, `apply(1)=49`).

**2. check_comptime_gate incomplete** (`sema.mach`) — now tests actual foldability (folds each leaf via the evaluator, not by symbol kind) and recurses every expr kind. Verified: a runtime block-local val, and a runtime ident inside a CAST / CALL gate, all give span'd sema errors (not span-less lower aborts).

**3. bare comptime-param-fn reference** (`infer.mach`, `context.mach`) — rejected at sema via a transient `callee_eid` that permits only the call-callee position. Verified: `val fp = apply;` and `apply == nil` rejected; `apply(0,21)` still works.

**4. drain re-emits generics V+1 times** (`lower.mach`, `context.mach`) — persistent `inst_drained` cursor + a fixpoint loop. Verified with nm: a module with `box[i64]` called from 2 value-instances + `box[i32]` emits each generic body exactly ONCE.

**5. INT mangle non-injective on non-64-bit host** (`mangle.mach`) — encodes the full 64 bits as fixed hex (like FLOAT), not via usize. Verified: `0x100000000` vs `0x200000000` → distinct symbols (`pickKi0000000100000000E` vs `...0200000000E`), correct results.

**6. arg_names_module_constant too broad** (`check.mach`) — defers only SYM_IMPORTED; a local/runtime SYM_VAL arg is rejected eagerly. Verified: `apply(rt(), 21)` errors at sema, while module-level and local literal consts still fold.

**7. defensive skip** (`lower.mach`) — a missing origin registry is now a hard internal error, not a silent body skip.

## Re-verification (all green)
- lookup regression + $param shadow, gate constness (index/call/cast), bare-ref rejection, generic-emitted-once (nm), int injectivity, local-arg eager rejection
- existing suite: float distinctness (2 distinct symbols), cross-module bare+aliased, int/bool/str keys, mixed comptime+runtime ABI, bounded+unbounded recursion cap, generic+comptime rejection
- `make clean && make` exits 0; **FIXPOINT byte-identical**; `mach test --cwd .` **194 pass / 0 fail**; `differential.sh` **10 inputs, all consistent** (both arms)

octalide:
## Round-3 item 8: generic + comptime-value-param call rejection (Copilot review)

**Gap:** the generic+comptime rejection only covered the explicit-type-args path. A call relying on type inference — `f(0, 7)` where `f` is `fun f[T]($mode: u8, x: T)` — slipped past it into the comptime-call check against the still-generic stripped TYPE_FUN, producing a confusing `type mismatch: expected T, found i64` instead of the intended diagnostic. `lower_callee` also ignored `type_args_len` and would route a combined callee to `lower_value_callee` (wrong symbol).

**Fix:**
- `infer.mach` (`infer_call`) — detect the combined case (DeclFun `generics_len > 0` AND a comptime param) ONCE, before both the generic-call and comptime-call paths, regardless of `type_args_len`.
- `expr.mach` (`lower_callee`) — hard-error on the same combination so neither `lower_value_callee` nor `lower_generic_callee` can silently miscompile it.

**Verified — both call forms now give the clean span'd diagnostic:**
- `f[i32](0, 7)` → `comptime value parameters on a generic function are not yet supported` (already worked)
- `f(0, 7)` → same clean diagnostic (was the confusing `expected T, found i64`)

No arity error, no miscompile, span points at the call. No regression: core feature, generic-emitted-once (nm: `boxI3i64E`/`boxI3i32E` each ×1), bool/str/mixed, cross-module all intact.

**Suite:** `make clean && make` exits 0; **FIXPOINT byte-identical**; `mach test --cwd .` **194 / 0**; `differential.sh` **10 inputs, all consistent** (both arms).

octalide:
## Final round: 2 Copilot review items

1. **Nil-slice risk in `lower_value_callee`** (`expr.mach`) — replaced the `nil::str` fallback that flowed into `intern_span` with a hard error: if the comptime-param callee's origin module `SourceFile` is unavailable, the `vals`/`names` allocations are freed and the call returns `err("...origin module source unavailable")`, so a nil source is never sliced.
2. **`dnit_context` missed `inst_drained`** (`context.mach`) — now reset alongside the other instance-worklist fields, so a context reused without `init_context` cannot start a drain mid-worklist.

The other two Copilot comments (FLOAT `::u64` mangle, generic+comptime-call combo) were already addressed in round 2 (`f64_bits`) and round 3 item 8 — no action needed.

**Suite green:** `make clean && make` exits 0; **FIXPOINT byte-identical**; `mach test --cwd .` **194 / 0**; `differential.sh` **10 inputs, all consistent** (both arms). Feature sanity (same-module 5/5 example tests, cross-module bare+aliased) intact. Ready to merge.

PR #1149, opened 2026-06-06
refactor(comptime): scoped (mark-delimited) environment, fix block-local val leak

### LINKED ISSUE #1143: comptime: replace the flat per-module const table with a scoped (push/pop) environment
The comptime evaluator's constant environment is a single per-module, append-only `user_consts` list with first-match lookup. Block-local `val` comptime consts get registered into it during resolve and are **never popped**, so a block-local const leaks into the module scope permanently and can shadow a same-named module const in a *later* function. This was the root of several value-param-monomorphization (#1075) bugs — the lookup-order fix had to bolt on a separate param-binding stack to get shadowing right, and `$if` is eagerly arm-selected across three passes against this one flat context.

## Design
Give comptime evaluation a real lexical environment: a stack of scope frames with push/pop, so:
- a block-local comptime `val` is visible only within its block (popped on exit), shadowing outer consts within scope, never leaking;
- module-level consts, imported pub consts, `$mach.*` params, and value-instance `$param` bindings all live in clearly-ordered frames (innermost wins), unifying the ad-hoc param stack added in #1075;
- `$if`/`$or` selection composes naturally with the scoped environment.

This is a front-end refactor (resolve / sema / comptime eval). The only user-visible semantic change is in cases that are currently bugs (block-local-shadows-module-const in a later function); well-behaved code is unaffected.

Pre-1.0 (maintainer: critical). Best done on a stable base — fixpoint-delicate front-end area; design->implement->review arc with the standard review gate.

octalide:
Closes #1143.

Replaces the comptime evaluator's flat per-module constant table (plus the separate #1075 param stack) with a single scoped, mark-delimited environment. Fixes the block-local-comptime-`val` leak at the source.

## Step 1 gate (the cheap-design precondition)
Searched all of `src/` + `dep/mach-std/src/`: **no `$if`/`$or` gate inside a function block references a block-local comptime `val`.** Every real `$if`/`$or` condition reads only `$mach.*` target parameters. The leak-at-source removal therefore cannot change arm selection, so the cheap design is safe.

## What changed
- **`comptime.mach`**: `ComptimeCtx` now holds one `entries`/`entries_len`/`entries_cap` array (replacing the six `user_const*`/`param*` fields). A frame is implicit: the run `[mark, entries_len)`. `bind` appends; `frame_open`/`frame_close` (returning/consuming `FrameMark`) scope a frame; `lookup` is a single reverse scan where the innermost (latest-bound) name wins. `$mach.*` stays out of `entries` (resolved structurally).
- **module frame**: the bottom run `[0, module_mark)`, built once and never closed — what all three passes read. Populated by driver `register_val_for_load` + `merge_pub_consts` and resolve `collect_one_decl`, all via `bind`. Collect is now idempotent on an already-bound name, so the load+collect double-registration of module pub vals never duplicates.
- **leak fix** (`resolve.mach` `bind_local_decl`): dropped the block-local comptime-const registration. A block-local `val` is never put in the ctx, so it can't leak / shadow a later module const. Lexical scope is untouched, so runtime uses still resolve.
- **subsume #1075** (`lower.mach` `emit_one_value_instance`): `param_mark`/`push_param`/`param_truncate` → `frame_open`/`bind`/`frame_close` on the origin ctx, closing the frame on both the success and error paths.

## Verification
- `make clean && make`: exit 0 (full cmach → imach → smach → mach bootstrap).
- **Self-host fixpoint BYTE-IDENTICAL**: `mach build` → `mach2`, `cmp -s` identical. Fold-equivalence confirmed — no winning binding moved.
- Test suites + differential + value-param + leak regression: see PR comments.

### DIFF
--- a/examples/06_comptime/lib.mach
+++ b/examples/06_comptime/lib.mach
@@ -19,5 +19,5 @@ pub val LIB_BASE: i64 = 100;
 pub fun op($op: u8, a: i64, b: i64) i64 {
     $if (op == OP_ADD) { ret a + b; }
     $or (op == OP_SUB) { ret a - b; }
-    ret a * b;
+    $or { ret a * b; }
 }
--- a/examples/06_comptime/lib.mach
+++ b/examples/06_comptime/lib.mach
@@ -1,0 +1,23 @@
+# a sibling module for the comptime-scoping regression: it exports a comptime
+# value-parameter function gated on this module's OWN `pub val`s, so a caller in
+# another module folds the `$if` arms against THIS module's comptime constants
+# (the value-instance origin frame), not the caller's. it also exports a plain
+# comptime `pub val` to cover cross-module const reads.
+
+# distinct from main's MODE_* on purpose: a value-param instance must fold its
+# `$if` gate against these (lib's) consts, never against a same-named const that
+# might exist in the calling module.
+pub val OP_ADD: u8 = 0;
+pub val OP_SUB: u8 = 1;
+pub val OP_MUL: u8 = 2;
+
+# a cross-module comptime const, read both bare and via an alias.
+pub val LIB_BASE: i64 = 100;
+
+# a comptime value-parameter function defined in lib: each instance compiles only
+# the arm its `$op` selects, folded against lib's OP_* above.
+pub fun op($op: u8, a: i64, b: i64) i64 {
+    $if (op == OP_ADD) { ret a + b; }
+    $or (op == OP_SUB) { ret a - b; }
+    ret a * b;
+}
--- a/examples/06_comptime/main.mach
+++ b/examples/06_comptime/main.mach
@@ -47,7 +47,9 @@ fun apply($mode: u8, n: i64) i64 {
     $or (mode == MODE_SQUARE) {
         ret n * n;
     }
-    ret 0;
+    $or {
+        ret 0;
+    }
 }
 
 # comptime params also drive a mix of compile-time and runtime parameters: the
@@ -55,7 +57,7 @@ fun apply($mode: u8, n: i64) i64 {
 fun combine($op: u8, a: i64, b: i64) i64 {
     $if (op == 0) { ret a + b; }
     $or (op == 1) { ret a - b; }
-    ret a * b;
+    $or { ret a * b; }
 }
 
 # a comptime float parameter keys an instance by its exact value: floats that
@@ -63,7 +65,7 @@ fun combine($op: u8, a: i64, b: i64) i64 {
 fun scale($factor: f64, n: i64) i64 {
     $if (factor == 1.5) { ret n + n / 2; }
     $or (factor == 1.9) { ret n + n; }
-    ret n;
+    $or { ret n; }
 }
 
 # #1143 regression: a block-local comptime `val` must never leak into the module
--- a/examples/06_comptime/main.mach
+++ b/examples/06_comptime/main.mach
@@ -9,6 +9,12 @@ use std.runtime;
 use std.print.println;
 use std.print.printlnf;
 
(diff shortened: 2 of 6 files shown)


octalide:
## Verification results

**Step 1 gate (precondition for the cheap design):** searched all of `src/` + `dep/mach-std/src/`. **No `$if`/`$or` gate inside a function block references a block-local comptime `val`** — every real `$if`/`$or` condition reads only `$mach.*` target parameters. Removing the leak-at-source registration cannot change arm selection, so the cheap design is safe (the larger synchronized-block-frames design is not needed).

**Build:** `make clean && make` → exit 0 (full cmach → imach → smach → mach bootstrap).

**Self-host FIXPOINT — byte-identical (the key gate):**
`mach build` → `mach2`, `cmp -s out/bin/mach out/bin/mach2` → **identical**. Fold-equivalence confirmed: no winning binding moved. The module-const double-registration now folds the same init in the same env and is deduped at collect to a single entry, so reverse-scan lookup observes exactly the same CTValue the old first-match scan did.

**Tests:**
- `mach test --cwd .` → **200 PASS, 0 FAIL**.
- `examples/06_comptime` → **132 PASS, 0 FAIL** (incl. all 5 #1075 value-param cases + 5 new #1143 regressions).
- `examples/11_casts` → **128 PASS, 0 FAIL**.
- `tools/test/differential.sh` (both arms: -O0 vs -O2, and smach vs mach) → **11 inputs, all consistent**.

**#1075 value-param cases all pass:** same-module (`apply`/`combine`/`scale`), cross-module bare (`op`) + aliased (`lib.op`), `$param`-shadows-module-val (`shadow_param`), block-local-val-does-not-shadow-module-const-in-a-later-fn (`gated`).

**Regression discrimination (proves the new test is not vacuous):** built a leak-injected compiler (block-local val bound into the ctx without the dedup), ran the example suite against it → the `block-local comptime val does not leak` test **FAILED** (131 PASS / 1 FAIL). The same test passes against the fixed compiler. The fix is genuinely exercised.

octalide:
Merged latest `dev` (PR #1148 infer-cast followup) into the branch and re-verified the full gate set on the merged tree:

- `make clean && make` → exit 0
- self-host fixpoint → **byte-identical** (`cmp -s mach mach2`)
- `mach test --cwd .` → **200 PASS / 0 FAIL**
- `examples/06_comptime` → **132 / 0**, `examples/11_casts` → **128 / 0**
- `tools/test/differential.sh` (both arms) → **11 inputs, all consistent**

Copilot review: 4/4 files reviewed, no comments. Marking ready.

octalide:
Addressed all three Copilot review comments in ebabc66:

1. **`comptime.bind` error swallowed** — `try_register_comptime_const` now returns `Result[bool, str]` and propagates the bind result instead of discarding it; the `bind_local_decl` caller propagates in turn. An allocation failure can no longer leave a module const silently unregistered (which would surface as a confusing use-site "not a comptime constant" error). This matches the driver load-pass sibling, which already propagates. Best-effort skips (non-comptime initializer, already-bound, empty span) still return `ok`.
2. **Drifting line number** — removed the hardcoded `comptime.mach:702` reference from the driver.mach comment.
3. **Dead trailing arm** — closed each value-instance `$if`/`$or` chain (`op`, `combine`, `apply`, `scale`) with an unconditional `$or {}` else arm so every monomorphized instance compiles exactly one branch, matching the `gated`/`shadow_param` idiom already in the fixture.

Re-verified the full gate on the merged tree: build OK, byte-identical self-host fixpoint, `mach test --cwd .` 200/0, examples/06_comptime 132/0, examples/11_casts 128/0, differential consistent.

Issue #1272, opened 2026-06-10
comptime: u64-range integer constants compare as signed in $if / comptime arithmetic


octalide:
Follow-up to #1247 (epic #1259).

`CTValue` stores integers in an `i64` payload, so a literal in (2^63, 2^64) is negative inside the comptime evaluator. `$if` comparisons and comptime arithmetic over u64-range constants (e.g. `0xFFFFFFFFFFFFFFFF > 0`) therefore evaluate against the signed interpretation even though the runtime bit pattern is correct.

#1247 fixed the silent **wrap** (eval_lit_int now rejects literals past 2^64-1 and round-trips the full unsigned range through the payload bits), but left the signedness of comptime **comparison/arithmetic** unchanged: `literals.md` does not settle whether comptime integers are signed, unsigned, or carry per-value signedness, so the correct semantics is a spec decision rather than a localized bug fix.

Decide the comptime integer signedness model (likely: record signedness on `CTValue`, or widen it) and make comptime comparisons/arithmetic over u64-range constants correct. `src/lang/fe/comptime.mach`.

Issue #1259, opened 2026-06-10
epic: compiler cleanup — 2026-06 whole-codebase audit (125 findings, 45 confirmed bugs)


octalide:
Full-compiler audit (all 63k lines of `src/`, 12 subsystem territories, every bug claim adversarially verified; 3 refuted claims dropped). **125 findings: 45 confirmed/plausible bugs, 34 design problems, 46 smells/cleanups**, collected in the child issues below — each carries per-finding detail, evidence, and a suggested fix.

## Most severe (live miscompiles / corruptions, fix first)
- **regalloc FP-bank interference** — precolored-register tracking is GP-only; confirmed miscompile where swapped float args collapse to one value (#1254)
- **win64 callee-saved XMM6–15 never saved** — allocator hands them out, prologue ignores them; corrupts every caller's float state across calls (#1254) — *windows-release-relevant*
- **inline-asm clobbers ignored** — the language doc promises clobber inference that does not exist; two confirmed corruptions (#1254)
- **u32→float conversion converts as signed i32** + stale upper bits for sub-32-bit sources (#1255)
- **SysV float-bearing aggregates never classified to SSE registers** — silent ABI divergence for any C FFI passing `struct { double … }` by value (#1257)
- **mixed-signedness comparisons take signedness from the LHS only** (#1251)
- **comptime string in the wrong position panics the compiler**; cross-module generic `$if` folds against the wrong module (#1251)
- **`write_lock` heap overflow** (2 bytes per managed dep) (#1258)
- **weak/strong symbol resolution is order-dependent** (#1257)
- **constfold SIGFPEs the compiler on `INT_MIN / -1`** (#1252)

## Recurring structural themes (cross-cutting)
1. **Silent narrowing/truncation** is the house bug class: array lengths i64→u32, COFF/ELF section counts →u16, literal overflow wrap, 32-bit patch of 64-bit values — an audit-then-ban sweep of unchecked `::u32`/`::u16` casts on sizes/offsets/addresses would close a whole family. (#1248, #1256, #1247)
2. **Defensive fallbacks masking failure**: find_sym_index→0 (#1196, in progress), unknown reloc kinds→abs64, sym_remap OOM→corrupt object "success", verifier skipping dangling ids, branch-patch fallbacks. (#1256, #1250, #1255)
3. **Per-opcode hand-encoding** in encode.mach produced the same imm-width/REX bug three+ times — candidates: table-driven encoder core. (#1255)
[...]

Issue #1249, opened 2026-06-10
audit(comptime): name-text shadowing fold, divergent $ident semantics, query.mach staleness


octalide:
Findings from the 2026-06 whole-compiler audit, subsystem **comptime + query/editor**. Bug-class findings were adversarially verified (refuted claims removed). Parent epic: #1259.

**Subsystem health:** Split verdict. The comptime evaluator core (literal parsing, operator folding, the post-#1143 frame model with its append-only entries array and reverse-scan shadowing) is tight, well-documented, and bounded — frame_open/frame_close are used with care at the one call site that matters (the value-instance drain in me/lower.mach restores on every path). The weighty problems are at the seams: identifier evaluation is name-text-based rather than resolution-based (a runtime local shadowing a module const folds to the const — a real miscompile of `$if` gates), bare `$ident` means three different things to sema, eval, and lowering, and unknown os/arch tag names silently fold to UNKNOWN in direct contradiction of the file's own no-silent-placeholder rule. query.mach is the most concerning file: it presents as 'a complete memoization DB' (and #1164 plans to wire it in as-is), but the engine as written is not incrementally sound — validation is shallow (stale results through any two-deep derived chain), a failed recompute is laundered into a permanently-valid cached entry, and the documented automatic dep recording does not exist (no in-progress stack, no cycle detection, zero callers of record_dep). These should be fixed before #1164 integration, not discovered during it. editor.mach is a reasonable thin facade with honest tests; its defects are policy-level (fake 'host defaults' of os=0/arch=0/ptr=8 that make the LSP analyze target-gated code as freestanding, and an inconsistent diagnostics-reset lifecycle) rather than structural.

### [BUG/M] Comptime eval resolves identifiers by name text, so a runtime local shadowing a module comptime const silently folds to the const *(verified: confirmed)*
`src/lang/fe/comptime.mach:735-755 (eval_ident), src/lang/fe/sema.mach:588-594, src/lang/me/lower/stmt.mach:509`

[...]

Issue #1217, opened 2026-06-09
feat: expose project version via comptime `$project.version`


octalide:
Add a compiler-provided comptime project-metadata value to expose the `[project].version` from `mach.toml` so it can be consumed at compile time without manual duplication.

**Motivation:**
- Remove hardcoded compiler version string from `src/lang/driver.mach` (line ~952)
- Single source of truth: version lives only in `mach.toml`
- Reduces release burden (no manual sync required between manifest and code)

**Proposed addition:**
- Add `$project.version` (or similar) to the compiler's comptime namespace
- Consumer (driver.mach): replace hardcoded `"1.1.0"` with `$project.version`

**Scope:**
- Update compiler's comptime system to read and expose project version
- Update driver.mach to use the new value
- Document the new comptime capability

**See also:** #1215 (init scaffolding fix, exposed this duplication)

octalide:
Design linkage from the #1218 manifest-v2 spec (settled 2026-06-10): the comptime namespace becomes `$project.id/version` + `$target.os/isa/abi` (superseding the `$mach.os/arch` naming) + `$bin.name`, each fed from an explicit manifest stanza or the selected target tuple. Prerequisite before implementing: unify the three divergent `$ident` resolution paths flagged in #1249 so the new namespace lands on one semantics. Implementation target: v1.4.0.

octalide:
Unblocked: the #1249 `$ident` ruling (bare `$ident` rejected; closed shape set) means this lands as new path ROOTS beside `$mach`: `$project.id/version` (+ metadata), `$target.os/isa/abi` (superseding `$mach.os/arch` naming per the #1218 spec), `$bin.name` — all fed from the v2 manifest model (src/lang/manifest.mach, merged in PR #1305) and the selected build unit. Implementation queued behind manifest phase 2 (the defines wiring touches the same comptime-context seeding surface).

PR #1276, opened 2026-06-10
fix(comptime/query/editor): audit findings — gate shadowing, eval traps, query poison, editor host ctx (#1249)

octalide:
Addresses the fixable comptime + query.mach + editor.mach findings from the 2026-06 audit (#1249, epic #1259). Builds on #1268 (literal-overflow rejection already merged) and respects #1272 (comptime u64 signedness deferred as a spec decision).

## Fixed

- **[BUG/M] gate name-text shadowing** — a runtime local shadowing a same-named module comptime const no longer silently folds the `$if` gate to the module const. The resolver now binds non-param gate idents (so their resolution identity is recorded) and tags module-scope symbols `SYM_FLAG_MODULE`; sema validates every evaluated gate (both the param and non-param `$if` paths) and rejects a gate referencing a block-local runtime value. The name-based fold only ever runs on a gate proven to reference comptime bindings.
- **[BUG/M] divergent bare `$ident`** — a standalone `$ident` gate is fold-tested in sema and reported with a span instead of crashing lowering with a span-less `comptime path root must be $mach`. The full `$ident` semantics stay an open design question (see the issue comment).
- **[BUG/S] integer trap/wrap** — `apply_arith` traps `INT64_MIN / -1` and `% -1` as comptime errors (no more compiler SIGFPE). Literal overflow was already fixed by #1268.
- **[BUG/S] unknown os/arch tags** — `$mach.os.*` / `$mach.arch.*` reject unrecognized names; `freestanding` gets its own `OS_FREESTANDING` id distinct from `OS_UNKNOWN`.
- **[BUG/S] query failed-recompute poison** — a failed compute drops its entry instead of leaving a zero-dep (vacuously-valid) stale/nil cache.
- **[SMELL/S] frame_close** — surfaces a mark beyond the current depth as an error instead of silently ignoring it.
- **[SMELL/S] computed_at** — removed the dead, inconsistently-maintained field.
- **[DESIGN/S] editor host defaults** — buffers seed their comptime ctx from the real host target (`mach.lang.target.host_*`) instead of os=0/arch=0/ptr=8; the host helpers moved out of the driver into `target.mach`.
- **[SMELL/S] editor diagnostic lifecycle** — one reset-on-entry policy (`reset_diags`) shared by every public analysis entry; repeated `resolve`/`parse` no longer accumulate duplicate diagnostics.
- **[CLEANUP/S] `$`-stripping** — `comptime.comptime_ident_name` is the single owner; the four hand-rolled copies call it.
- **[CLEANUP/S] stale docs** — corrected the comptime `$mach.target.abi` claim and the editor `Q_FILE_TEXT` claim.

## Remaining (design items, kept on #1249)

- query.mach transitive (shallow) validation and the manual-dep-recording / in-progress-stack model are rewrites, not localized bug fixes — recommendation left on #1249 (gates #1164).
- the full bare-`$ident` resolution model (one shared resolution rule across sema/eval/lowering) — recommendation left on #1249 (gates the `$project.*`/`$target.*` namespace from #1217).

## Tests

Inline fail-before tests for the idiv trap, unknown/freestanding tags, frame_close overflow, query failed-recompute, editor host-ctx seeding, editor diagnostic reset, and an end-to-end editor+sema test that the shadowing gate is rejected (not folded to the const). `mach build .` clean, `mach test .` = 252 pass, byte-identical self-host fixpoint.

Refs #1249

### DIFF
--- a/src/lang/driver.mach
+++ b/src/lang/driver.mach
@@ -860,7 +860,7 @@ fun intern_required(i: *intern.Interner, text: str, missing: str) Result[intern.
 # find_target: resolve a target selector against the config's [targets.*]
 # entries. an empty selector defers to [project].target. the selectors
 # "native" and "host" resolve to the entry whose os/isa match the build
-# host (host_os_id()/host_arch_id()); any other selector matches an entry by
+# host (tgt.host_os_id()/host_arch_id()); any other selector matches an entry by
 # name. returns nil when nothing matches.
 fun find_target(c: *ProjectConfig, name: str, i: *intern.Interner) *TargetEntry {
     var sel: str = name;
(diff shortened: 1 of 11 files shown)


octalide:
Recreating: GitHub's merge-commit computation wedged on this PR (mergeable stuck UNKNOWN, pull_request workflows never scheduled). Continued as the next PR from the same branch.

PR #1336, opened 2026-06-12
feat(test): one executable per test, run as separate processes (#1235, #1274)

### LINKED ISSUE #1235: feat: restore `mach test .` individual binary behaviour
`mach test .` as originally designed created testing "artifacts" that were effectively a small, executable mach program with only the test function and required dependencies. This allowed for the creation of a system where tests could be compiled in bulk (incrementally even) and executed independently by the testing command or by the user directly, evaluating success and failure by program exit code.

This may not be the best design. Need to take a second look at it. Either way, the existing testing infra is ugly, slow, and does not produce the kind of results that a user would typically look for. We somehow lost a lot of features in the mix of the 1.0 release.

### LINKED ISSUE #1274: test: `mach test .` overwrites the compiler binary at out/linux/bin/mach with the corpus runner
Found during #1255 verification: `mach test .` writes its corpus-runner executable to the same path as the compiler (`out/linux/bin/mach` — the layout's binary slot), so running `mach test .` and then the integration suites against that path silently hands the suites the TEST RUNNER instead of the compiler. CI sidesteps it (separately seeded compiler), and local flows must remember to re-build or copy the compiler first — a footgun that will eventually eat someone's afternoon.

Proper fix likely falls out of the #1218 manifest v2 layout work: a test build is its own artifact and should resolve to its own path via the `out` template (e.g. a distinct artifact name), never the project binary's slot. Until then, worth at least a distinct filename.

Refs #1218, #1242.

octalide:
Reshapes `mach test` from a single monolithic in-process runner into the original per-test model on the new manifest foundation.

## What changed
- **One executable per test.** Every `test "..."` block compiles to its own standalone program — the project's transitive code plus a synthesized `main(){ret <test>();}` that returns the test's `i32` as the process exit status (0 = pass). Each is independently runnable by the user.
- **Separate process per test.** `mach test` spawns each executable, reads the exit status, and reports `name file:line PASS|FAIL(exit n|signal s)`. A crashing test (a signal) reports `FAIL(signal s)` and the run continues — one SIGSEGV no longer kills the suite. Ends with a `passed/failed/total` summary.
- **`tests` template key.** New `[project].tests = "out/{target}/{profile}/test/{name}"` joins the `out`/`obj`/`ir`/`asm` template family. Per-test executables land there — the runner never touches the project binary path again (**closes #1274**). `{name}` is the per-test name, expanded by `mach test`.
- **`--list` / `--filter`.** `--list` prints the collected tests without building/running; `--filter <substr>` selects by name substring.

## File-naming scheme (`{name}`)
`<index>-<sanitized label>`: the zero-based collection index keeps names unique (labels are not globally unique across modules) and stable; every label byte outside `[A-Za-z0-9._-]` becomes `_`, label portion truncated to 60 bytes.

## Architecture
The per-module objects (and external link inputs) are written/resolved **once** and shared by every test link; only a tiny per-test entry object is synthesized and linked per test. Sema + lowering + codegen of the module graph is shared across all tests. The per-test **link** is not yet shared (one link invocation per test) — noted as a follow-up optimisation on #1235.

## Execution platform
Linux-only execution for now (POSIX `fork`/`exec`). A win64 runner needs `CreateProcess` — tracked as the windows item on #1235, not implemented here.

## CI note
`ci.yml`'s Test Suite job runs `mach test .` with the **seed** (v1.3.1). The changes here are source-side, so the seed compiles them, but the seed binary that *executes* `mach test` is the old in-process runner reading the new manifest — it still produces a single runner binary at the project path. The new engine is exercised once a release seeds a compiler built from this branch. Local verification below uses a compiler built from this branch.

Closes #1235
Closes #1274


octalide:
## Verification (compiler built from this branch)

All green with a compiler built from this branch (`out/linux/bin/mach`):

- **`mach build .`** — clean.
- **`mach test .` full corpus** — `390 passed, 0 failed, 390 total` (~97s). Same tests, new engine.
- **Byte-identical fixpoint** — release-style `./a build .` → `./b build .`, `cmp` shows `s2 == s3` byte-identical (3,907,584 bytes). The compiler binary is unchanged by the test-infra work (it only changes the test build path), and even the seed-built stage matches.
- **Integration battery** — all 25 `.github/tests/*` suites pass against the branch binary, including `lower-testrunner` (the `--list`/label suite) and `manifest`.

### Crash-isolation demo (temporary tests, not committed)
A scratch project with a passing, a failing, a SIGFPE-crashing, and a following test:

```
arith adds /tmp/ttest/src/main.mach:10 PASS
always fails /tmp/ttest/src/main.mach:15 FAIL(exit 3)
divide by zero crash /tmp/ttest/src/main.mach:19 FAIL(signal 8)
runs after crash /tmp/ttest/src/main.mach:24 PASS

66 passed, 2 failed, 68 total
```

The crashing test reports `FAIL(signal 8)` and the run **continues** to `runs after crash` (PASS) — one crash no longer kills the suite. Exit code is 1 (failures present).

### Output sample (real corpus, `--filter manifest`, ~31s — only matching tests built)
```
manifest: a backslash in a path is rejected ./src/lang/manifest.mach:1865 PASS
...
manifest: an [os.<name>] overlay rejects an unknown key ./src/lang/manifest.mach:2203 PASS

23 passed, 0 failed, 23 total
```

### Naming scheme
`{name}` = `<index>-<sanitized label>`, e.g. `out/linux/debug/test/298-manifest__a_full_manifest_parses_every_stanza`. Index = zero-based collection order (labels are not globally unique, so the index disambiguates and keeps names stable); every label byte outside `[A-Za-z0-9._-]` → `_`, label portion truncated to 60 bytes. Each executable is independently runnable (verified: running one directly exits 0).

### Note on the repo's own `mach.toml`
The design called for the repo manifest to gain the `tests` key, but the bootstrap **seed (v1.3.1) rejects unknown manifest keys**, so adding it would break the seeded `mach build .`/`mach test .` (CI's Full Build and Test Suite jobs, and local bootstrap). The default template (`out/{target}/{profile}/test/{name}`) already keeps test artifacts off the project binary path, so the key is functionally optional. The repo manifest should adopt it once a release that understands the key becomes the seed. The `mach init` scaffold (self-consistent: same compiler writes and parses it) and the parser/default/docs all ship here.

Issue #1343, opened 2026-06-12
consumer member access through a fwd module re-export does not resolve


octalide:
Found while adding the `fwdreexport` integration suite (PR #1342).

`bind_fwd_module` (added in PR #1186) publishes a whole-module `fwd` re-export as a pub `SYM_USE` module alias, and the standalone build of such a library works. But a **consumer** cannot chain through that alias in expression position:

```mach
# dep "demo", src/lib.mach
fwd demo.alpha;            # module re-export

# consumer, with: use demo.lib;
ret lib.alpha.answer();    # error: no field 'answer' on the receiver type
```

Sema treats `lib.alpha` as a value/field access instead of resolving the module member `alpha` (itself a module alias) and continuing module-member resolution. Symbol re-exports (`fwd demo.alpha.answer;` then `lib.answer()`) work fine.

Two observations from the failing build:

1. `error: <src>: no field 'answer' on the receiver type` — the resolution gap itself.
2. A stray `error: mir.lower_ir: gep index into a non-aggregate type` printed in the same build, i.e. lowering appears to run (and emit an internal-looking diagnostic) despite sema errors — worth checking the error-gating between sema and lowering while fixing this.

Repro: take `.github/tests/fwdreexport/` from PR #1342 and change the consumer's body to `ret lib.alpha.answer() + lib.beta.value();`.

Practical impact today is low (mach-std consumers import real module paths like `use std.io;` directly), but the published re-export surface should be uniformly consumable: a module re-export that resolves nowhere is a silent API hole.

PR #1357, opened 2026-06-12
fix(sema): teaching diagnostics for never-value symbols in value position (silent-poison audit)

### LINKED ISSUE #1348: sema: audit decl_type_for's TYPE_NIL returns for never-typed decl kinds (silent-poison class)
Follow-up from #1343 / PR #1346.

The #1343 silent poison happened because a SYM_USE reaching value inference produced no diagnostic — sema completed clean and lowering surfaced internal errors (`gep index into a non-aggregate type`). PR #1346 fixed that instance by reporting module aliases in value position at both `infer_ident` and `infer_member`.

The class may be wider: `decl_type_for` returns TYPE_NIL for decl kinds that can never carry a value type. Any symbol kind that resolves cleanly but types as TYPE_NIL/TYPE_ERROR without a report lets a broken expression travel to lowering with zero sema diagnostics, defeating the sema/lower error gate.

Audit: enumerate every `decl_type_for` caller and every decl/symbol kind it can receive in value position; each never-typed kind needs either an explicit teaching diagnostic (like #1346's "a module alias is not a value") or proof it is unreachable there. Same bar as the silent-narrowing sweep from the #1259 audit.

octalide:
Closes #1348

## Audit first (house rule)

Full enumeration of `decl_type_for` callers, every symbol kind each can receive in value position, and a widened sweep of silent TYPE_NIL/TYPE_ERROR returns in sema inference is posted as [the audit table on #1348](https://github.com/octalide/mach/issues/1348#issuecomment-4691869198). Verdict: **5 reachable-silently classes**, 6 already-reported (each verified to report at or before the poison — no #1343-style never-fires traps), 8 unreachable/defensive.

## The (a) instances fixed

All previously passed sema with **zero** diagnostics and surfaced as link `undefined symbol: _M...` or span-less `lower:` errors:

| misuse | new diagnostic |
|---|---|
| `val p: Point = Point;` (local or imported rec, bare or `lib.Pt`) | `` `Point` is a record type, not a value; construct an instance with `Type{ ... }` `` |
| union name in value position | `` `U` is a union type, not a value; construct an instance with `Type{ ... }` `` |
| `def` alias in value position | `` `MyInt` is a type alias, not a value; use a value of the aliased type `` |
| generic type parameter in value position (`val t: T = T;`) | `` `T` is a generic type parameter, not a value; use it in type position or a comptime intrinsic like `$size_of(T)` `` |
| member access on a void expression (`nothing().x`) | `a void expression has no members; a member access requires a record or union value` |

## Implementation

One shared never-value classifier (`report_never_value`) consulted by `infer_ident` and `infer_member`'s qualified path — it folds in the existing #1346 module-alias report, so the never-value kind enumeration lives in exactly one place. An imported symbol is classified by its origin declaration's kind via `sess.module_ast` (the same route `callee_fun_decl` already uses). `SYM_PRIM` is included as a loud guard: unreachable today (primitives never bind into value scope) but it reports rather than poisons if that ever changes. The void-object arm in `infer_member` reports instead of silently early-outing on TYPE_NIL.

Not silently changed: `$size_of(T)` / `$align_of` / `$offset_of` route through `intrinsic_operand_type`, not value inference — verified unaffected. Function references, imported funs/vals/vars, and the #1346 module-alias diagnostic all verified unchanged. A type name in callee position (`Point(1, 2)`) now gets the teaching message instead of ``not callable: `Point` is not a function``.

## Tests

In-source `test` blocks are runtime-only (no negative-diagnostic support), so coverage follows the established `.github/tests/diaglocate` pattern: a new `valuepos` suite asserts each teaching message is emitted, located (`main.mach:line:col`), and that the adjacent positives (record literal, fn reference, `$size_of(T)`, qualified value access) still build.


octalide:
Verification (all local, plus CI):

- corpus: `mach test .` 409/409 with both the v1.4.1 seed and the stage-1 build
- integration: all 29 `.github/tests/*/run.sh` suites pass (including the new `valuepos`), windows target cross-compiled, wine suites included
- fixpoint: 3-stage self-build seeded from v1.4.1 — stage1 == stage2 == stage3, sha256 `c2642d8e...f4ec7b0663`, byte-identical
- CI: green on a4aa0010 (Full Build, Test Suite, Integration Tests), which contains all code and test changes. The final commit (26023031) only adds the CHANGELOG entry; GitHub stopped delivering `pull_request` events for this branch after 13:58 UTC (three pushes and a close/reopen produced no runs), so no check is attached to HEAD. If a run still hasn't appeared at merge time, re-trigger or use the admin-merge flow.

PR #1367, opened 2026-06-12
ci: native windows lane — build + test on windows-latest

### LINKED ISSUE #1351: ci: native windows lane — run mach.exe and the test suite on windows-latest
All windows verification today runs under wine driven by the linux mach binary. A native lane — `windows-latest` job that takes the windows artifact and runs `mach.exe build .` + `mach.exe test .` natively — tests the real OS instead of an emulation of it, and is the strongest answer to the foreign-target execution question (#1345): on the native OS no runner indirection is needed at all.

Blocked by octalide/mach-std#221 (native windows process spawning — without it mach.exe cannot exec git, test children, or run children on real windows). Until that lands, a partial lane (build-only, no test/dep) is possible but probably not worth the CI minutes.

Notes:
- use submodule checkout for deps in the lane; `mach dep pull` on native windows additionally needs the git-spawn path from mach-std#221.
- mach.exe has only ever been exercised under wine; expect the first native run to be a discovery exercise (filesystem/path semantics, console handles).
- the wine integration suites stay — they cover the cross-compilation path, which the native lane does not.

octalide:
Adds a native windows CI lane (`Windows (native)`, `windows-latest`) that builds mach and runs the in-source suite on the real OS, not the wine emulation the integration lane uses.

## What the lane covers

- `mach.exe build .` natively (the internal disk-based linker needs no child processes, so the build path works without a spawn backend).
- `out/windows/bin/mach.exe test .` natively — the freshly-built compiler spawns each test as a native child process via `CreateProcess`/`WaitForSingleObject`/`GetExitCodeProcess`.

This exercises the real windows PE loader and the windows exec/sync backends end to end. The wine integration suites stay — they cover the cross-compile path the native lane does not.

## Seeding design (reworked)

The lane no longer seeds from the released `mach-x86_64-windows.zip`. Instead:

- the existing **Integration Tests** linux job already cross-compiles `mach.exe` from *this PR's tree* (`out/linux/bin/mach build . --target windows`); it now uploads that binary as the `mach-windows-seed` artifact.
- the `Windows (native)` job (`needs: integration`) downloads that artifact, puts it on `PATH`, and uses it as the build seed.

So the lane tests each PR's **own** emitter on real windows and never depends on a release carrying a fixed emitter — breaking the released-seed chicken-and-egg permanently.

## Dependency

Bumps the vendored `dep/mach-std` to **v0.7.0** (`16020a1`), which carries the kernel32/ws2_32/advapi32 import-attribution fix (octalide/mach-std#235). `mach.toml`/`mach.lock` keep `ref = branch/main` — v0.7.0 is std main's tip, so the pin is consistent.

## Status: draft — blocked on a new mach-std import bug

The reworked lane is wired correctly (artifact handoff, native build/test). The native build step surfaced **another** wine-masked import mis-attribution, distinct from #235:

mach-std declares `WaitOnAddress` and `WakeByAddressSingle` (`src/system/os/windows/shared.mach:214-215`) without a `$<sym>.library` pin, so they bind to the first dependency, `kernel32.dll`. Real windows `kernel32.dll` does **not** export them — they live in `kernelbase.dll` / `api-ms-win-core-synch-l1-2-0.dll`. The native loader therefore rejects `mach.exe` with `STATUS_ENTRYPOINT_NOT_FOUND` (`0xC0000139`) before `main`. Wine's kernel32 exports them, which is why every cross-compile + wine lane stays green while the native lane cannot load the binary at all.

Proven on the `windows-2025` runner (temporary probe, since reverted): `kernel32.dll!WaitOnAddress`, `!WakeByAddressSingle`, `!WakeByAddressAll` all **ABSENT**; present in both `kernelbase.dll` and `api-ms-win-core-synch-l1-2-0.dll`.

Fix belongs in mach-std (pin the wait-on-address family to `api-ms-win-core-synch-l1-2-0.dll` via `$<sym>.library`, exactly like the ws2_32 fix), then cut a std release and move this pin to it. Holding this PR as draft until then rather than weakening the lane.

Closes #1351


octalide:
## Blocked: native Windows rejects mach's emitted PE (`ERROR_BAD_EXE_FORMAT`)

The lane is implemented and runs, and it immediately surfaced a more fundamental blocker than the one #1351 anticipated. The CI run fails at the very first native invocation:

```
mach.exe build .
Program 'mach.exe' failed to run: ... The specified executable is not a valid
application for this OS platform.   (ERROR_BAD_EXE_FORMAT / %1 is not a valid Win32 application)
```

The released `mach.exe` (v1.4.1 windows seed) **cannot be launched by the native Windows loader at all**, even though it runs under wine (the release smoke test) and Microsoft's own `dumpbin` parses its headers without complaint. Because the freshly-built compiler shares the same PE emitter, this blocks the *entire* native path — there is no green build-only subset available either.

### Verified on the runner (windows-2025, throwaway diagnostic, since reverted)
- A known-good native exe runs fine on the same runner (`cmd /c ver` → exit 0) — not an environment issue.
- `dumpbin /headers` on the seed: x64, PE32+, **relocations stripped**, DLL characteristics `0x100` (NX_COMPAT only — **no DYNAMIC_BASE**), fixed image base `0x13FFF0000`, section align `0x1000`, file align `0x200`, subsystem 3 (CUI), `SizeOfImage 0x4AF000`.
- System ASLR policy is **NOTSET** for ForceRelocateImages / RequireInfo / BottomUp / HighEntropy — so forced/mandatory ASLR is **not** the cause.

### Assessment
Static analysis shows the PE is structurally consistent (sections perfectly contiguous, `SizeOfImage` exact, data directories valid). The rejection is a Windows loader header-invariant violation that wine does not enforce. Leading suspects in the emitter (`src/lang/target/of/coff.mach`): the image is non-relocatable (`IMAGE_FILE_RELOCS_STRIPPED`, no `.reloc`, no `DYNAMIC_BASE`), and the 64 KiB `PE_IMAGE_BASE_RESERVE` places the first section at RVA `0x10000` while `SizeOfHeaders` is only `0x400` (a large header→first-section gap). Pinpointing the exact rejected field needs bisecting headers on a live Windows box — that is PE-emitter work, out of scope for this CI task.

### Note on the stated blocker
mach-std#221 (native windows `CreateProcess` exec backend) is listed as the blocker, but it is effectively **already implemented** in the current std submodule (`CreateProcessA` + `WaitForSingleObject` + `GetExitCodeProcess` in `system/os/windows/shared.mach`) — so once the PE bug is fixed, `mach test .` spawning should work natively.

### Recommendation
Keep this PR as a draft. File a compiler issue for windows PE-loader conformance (the `coff.mach` emitter), fix it, then this lane goes green and merges as-is. Holding rather than weakening the lane — making it pass while the bug exists would defeat its purpose.

octalide:
Holding as draft: blocked by #1374 (the native loader rejects the current PE output, so the build step cannot pass). One structural note for when #1374 lands — this lane seeds from the released windows zip, but that binary is itself unloadable and stays so until the NEXT release ships a fixed emitter. To break that chicken-and-egg (and to test the PR's own tree rather than the release), consider reworking the lane to cross-compile mach.exe in a linux job from the checked-out tree and hand it to the windows job as an artifact, then build + test natively from that seed. Revisit once #1374 merges.

octalide:
Status update: #1374 is merged — the PE layout is fixed and a mach-built program runs natively (proof: run 27431202210). The lane is now blocked on exactly one thing: octalide/mach-std#235 (std mis-declares ws2_32/advapi32 imports under kernel32 + a `SleepW` typo), which makes `mach.exe` itself die at import resolution (`STATUS_ENTRYPOINT_NOT_FOUND`). Once that lands std-side and this repo bumps its mach-std pin, rework the seeding here per the earlier comment (cross-compiled artifact, not the released zip) and the lane should go green.

octalide:
Lane rework is COMPLETE on this branch: seeds from a cross-built mach.exe artifact (linux job builds it from the PR's own tree — no released-zip dependency, tests each PR's own emitter), mach-std pinned to v0.7.0 (#235 fix included). The native run then surfaced wine-masked bug number three: octalide/mach-std#253 — `WaitOnAddress`/`WakeByAddressSingle` are unpinned and kernel32 does not export them on real Windows (only `api-ms-win-core-synch-l1-2-0.dll` / kernelbase do; wine's kernel32 does, hence the masking). `mach.exe` is rejected with STATUS_ENTRYPOINT_NOT_FOUND before main. Holding as draft until std#253 lands in a release and the pin bumps — that should be the last hop: the loader accepts the PE (#1374), the import directory is attributable (#1382), winsock/advapi32 are pinned (std#235), and this is the only remaining unresolvable import on a real loader.

octalide:
Native lane green: **468 passed / 0 failed** on windows-latest (run 27456524615). The 6 std exec tests that were the entire remaining red (exec.mach:218/283/293/303/318/340) now pass natively — confirming the #1395 stack-probe fix (merged via #1396, propagated here). Branch carries #1395 (dev merge) and std v0.10.0.

4th-descriptor (api-ms-win-core-synch) note: mach.exe links synch as its 4th PE import descriptor and loads/runs natively, but its code imports synch *dead* (no synch call site), and the std thread tests that would call it aren't in mach's corpus — so this lane doesn't exercise the synch *call* thunk. Verified separately with a thread spawn+join probe (same 4-DLL layout, synch trailing): its synch stubs target their own IAT slots (jmp→0x..2a4e0/0x..2a4e8) and it runs correctly under wine, empirically confirming #1388 covers the trailing descriptor.

PR #1383, opened 2026-06-12
fix(linker,driver): per-symbol DLL attribution + survive .symbol overrides

### LINKED ISSUE #1382: linker: no per-symbol DLL attribution — all windows imports forced onto kernel32.dll; ext-fun registration clobbers .symbol overrides
Root cause of octalide/mach-std#235 (mach.exe STATUS_ENTRYPOINT_NOT_FOUND on real windows). Two independent defects; both needed before mach-std can fix its bindings.

## 1. No per-symbol DLL attribution

The PE import directory is built entirely position-based:

- \`src/lang/be/linker.mach:1416\` — \`build_dynamic_info\` hardcodes **every** undefined \`ext\` import to \`lib_index = of.DYNLIB_ANY\`; no path sets a specific library.
- \`src/lang/target/of/coff.mach:2307-2313\` — \`import_lib_of\` maps \`DYNLIB_ANY\` → index 0, the first DLL in the manifest's \`[os.windows] libs\` — i.e. kernel32.dll.
- The parser (\`src/lang/fe/parser/decl.mach:134-142\`) treats \`ext\` as a bare flag; no library clause exists in the grammar, and the only per-symbol directive (\`\$name.symbol\`, \`src/lang/driver.mach:551-595\`) is a link-name override, not a library selector.

Net: every import lands on kernel32.dll. mach-std needs 14 winsock bindings on ws2_32.dll and \`SystemFunction036\` on advapi32.dll (full documentation-driven audit table in mach-std#235's thread). Adding the DLLs to \`libs\` alone just emits empty import descriptors.

Proposed: a per-symbol directive (e.g. \`\$foo.library = "ws2_32.dll"\`) parsed and stored on the session; \`build_dynamic_info\` sets \`lib_index\` by matching it against the \`DynLib\` list, falling back to \`DYNLIB_ANY\`. The COFF writer already supports N DLLs and pinned \`lib_index\` — the missing pieces are the directive and the routing.

## 2. \`register_ext_fun\` clobbers \`.symbol\` overrides

mach-std already carries \`\$ws2_socket.symbol = "socket"\`, \`\$SleepW.symbol = "Sleep"\`, etc. **They do nothing**: \`scan_export_directives\` registers the override, then \`register_ext_fun\` (\`src/lang/driver.mach:600-604\`) re-registers the bare identifier for the same decl, and \`register_export_name\` overwrites on double-registration (\`src/lang/session.mach:261\`). Since the ext-fun decl follows its directive in source order, the bare name always wins.

Empirical proof (llvm-readobj --coff-imports on a cross-built mach-std test exe): one import descriptor, kernel32.dll, with \`ws2_socket\`/\`ws2_send\`/\`SleepW\` (mach identifiers, not the .symbol values) hanging off it; wine aborts at call time with \`unimplemented function kernel32.dll.WSAStartup\`.

## Sequencing

This blocks octalide/mach-std#235, which in turn is the last blocker for native \`mach.exe\` (#1351 / native windows CI lane #1367). Once both defects land here and the std pin bumps, the mach-std side is a trivial follow-up: declare ws2_32.dll + advapi32.dll in \`[os.windows] libs\`, attribute 15 bindings, rename \`SleepW\`→\`Sleep\`.

octalide:
Closes #1382.

Two independent defects behind mis-attributed windows DLL imports (STATUS_ENTRYPOINT_NOT_FOUND on real windows; root cause of octalide/mach-std#235).

## 1. Per-symbol DLL attribution

`build_dynamic_info` (src/lang/be/linker.mach) hardcoded every undefined `ext` import to `DYNLIB_ANY`, so the COFF writer routed them all onto the first dependency — kernel32.dll. Adding extra DLLs to `[os.windows] libs` only emitted empty import descriptors.

New `$<sym>.library = "ws2_32.dll"` directive:
- parsed alongside the existing `$<sym>.symbol` directive in `register_one_export_directive` (src/lang/driver.mach), dispatched on the member name.
- stored per-decl on the session (`export_libraries`, keyed by `(ModuleId, DeclId)`), mirroring `export_names`.
- projected onto a link-name-keyed session map (`import_libraries`) by a post-scan pass (`register_import_libraries`) once each import's final link name is settled, so the key matches the object symbol the linker sees.
- consumed in `build_dynamic_info`: an import's library is matched against the dependency soname list and `lib_index` pinned; unattributed imports stay `DYNLIB_ANY`. A directive naming a library not among the link's dependencies is a hard link error, not a silent fallback.

The COFF writer already handled a pinned `lib_index` (`import_lib_of`, src/lang/target/of/coff.mach) — verified, no change needed.

## 2. `.symbol` clobber fix

`register_ext_fun` re-registered the bare `ext` identifier over a `.symbol` override that `scan_export_directives` had registered first (as a sibling), and `register_export_name` overwrites on double-registration — so the bare name always won and `.symbol` overrides did nothing. The bare name is now only a default: registration is skipped when an override already exists, while a `.symbol` directive scanned later still overwrites the default. Order-independent.

## End-to-end proof

Cross-compiled a windows program declaring:

```mach
$SleepW.symbol = "Sleep";                 # .symbol only -> kernel32 (fallback)
ext fun SleepW(ms: u32);
$WSAStartup.library = "ws2_32.dll";        # .library only -> ws2_32, bare name
ext fun WSAStartup(ver: u16, data: *u8) i32;
$ws2_socket.library = "ws2_32.dll";        # both -> ws2_32 as "socket"
$ws2_socket.symbol  = "socket";
ext fun ws2_socket(af: i32, kind: i32, proto: i32) i64;
```

with `[os.windows] libs = ["kernel32.dll", "ws2_32.dll"]`. `llvm-readobj --coff-imports`:

```
Import {
  Name: kernel32.dll
  Symbol: GetCommandLineA (0)
  Symbol: ExitProcess (0)
  Symbol: Sleep (0)
}
Import {
  Name: ws2_32.dll
  Symbol: WSAStartup (0)
  Symbol: socket (0)
}
```

- `socket` (the `.symbol` value, not the mach identifier `ws2_socket`) sits under a real ws2_32.dll import descriptor.
- kernel32 imports (the runtime's `GetCommandLineA`/`ExitProcess`) are untouched; `Sleep` falls back there as expected.

A directive naming an undeclared library fails the link cleanly:

```
error: import 'socket' pinned to library 'nope_not_declared.dll' not among the link's dependencies
```

Self-hosting byte-identical fixpoint holds.


octalide:
Closing — duplicate effort: the main mach session is already on #1382. This branch is being removed.

octalide:
For the record on the earlier "closing — duplicate effort" comment: the duplicate was resolved the other way around — the parallel session's work was discarded (nothing pushed) and THIS branch was restored from the PR head ref and finished (test suite, docs, changelog added on top of the reviewed implementation). Merging.

Issue #1402, opened 2026-06-13
target: drop the interner from register_all (convert entry_symbol/syscall_layer/libdir/dynamic_linker to str, intern at the linker consumer)


octalide:
Follow-up from #1377 (split out during that refactor). #1377 made the vtable `name`/`file_extension`/reg-class names immutable `str` constants, but `register_all` still takes the interner because the OS registrars per-session-intern `entry_symbol`, `syscall_layer`, `libdir`, `dynamic_linker` (and these are consumed as StrIds elsewhere).

## Ask

Achieve #1377's headline goal — `register_all` needs no interner — by converting the remaining per-session-interned vtable string-constant fields (`OsVTable.{entry_symbol, syscall_layer, libdir, dynamic_linker}`, and any ISA/ABI/OF equivalents) to plain `str`, and interning at the point of use in the consumers.

## Blast radius (why it's separate from #1377)

`entry_symbol` and `dynamic_linker` are read as StrIds in the **load-bearing linker entry path**:
- `linker.mach:702` — `entry_name: StrId = tgt.os.entry_symbol` (program-entry resolution, compared against symbol StrIds).
- `linker.mach:1365` — `dyn.info.interp = tgt.os.dynamic_linker` (PT_INTERP).

Converting them to `str` pushes interning into the linker (intern entry_symbol/interp at use — ~2-3 sites). Bounded, but it touches the entry/interp resolution, so it warrants its own review rather than riding the behavior-preserving #1377 name refactor. Verify with the byte-identical self-host fixpoint.

Refs #1377.

Issue #1377, opened 2026-06-12
target: make the isa/os/abi/of vtables truly immutable singletons (drop the per-session re-interned StrId fields)


octalide:
Follow-up deferred from #1319 (PR #1375), per that issue's "note on the vtable singletons" allowance.

The OF/ISA/OS/ABI vtables remain module-global statics whose `name` (and dead `file_extension`) `StrId` fields are re-interned per session by the registrars — harmless for the one-session-per-process CLI, but a latent cross-session hazard if two in-process sessions ever coexist, and it forces `register_all` to take an interner it otherwise wouldn't need.

Fix: make the vtable `name`/`file_extension` (and the reg-class names) plain `str` constants so the vtables are genuinely immutable and `register_all` needs no interner. Blast radius: vtable struct layout, the 10 registrars, `target.register_all`, and the three CLI entry points that currently pass `?s.interner`.

Behavior-preserving; verify with the byte-identical self-host fixpoint like #1319.

Parent: #1285. Epic: #1259.

octalide:
Closing: the immutable-vtable-names deliverable shipped in #1403 / v1.5.2 — the OF/ISA/OS/ABI vtable `name` (and reg-class names) are now immutable `str` constants set directly by the registrars, the dead `OfVTable.file_extension` is removed, and `mach info` reads the names directly. Behavior-preserving, verified by the byte-identical self-host fixpoint.

The remaining half of this issue — eliminating the interner from `register_all` entirely by also converting `entry_symbol`/`syscall_layer`/`libdir`/`dynamic_linker` to `str` and pushing interning into the linker entry/interp path — is tracked separately in #1402 (deferred: it touches the load-bearing linker entry/PT_INTERP resolution). #1402 is the single open tracker for it.

Issue #1319, opened 2026-06-11
of: make the isa/os/abi/of registries session-owned (piece 3 of #1285)


octalide:
Piece 3 of #1285, split out per that issue's allowance. Pieces 1 and 2 landed in #1317:

- **Piece 1** — relocation kinds are now a closed `of.RK_*` enum; the per-format interned-kind tables and the linker's `RelocKindIds` mirror are gone.
- **Piece 2** — the interner is threaded through `ObjectImage.interner` / `ParserFn` / `DynExecFn`; the `elf_interner`/`coff_interner` module globals are deleted. **This already removed the correctness hazard** the original issue describes (a process-global registry keyed off a module-global interner) — the registrars no longer capture any per-session state.

What remains (this issue): the `isa`/`os`/`abi`/`of` registries are still process-global arrays of vtable pointers (write-once-turned-rebinding per #1308). Moving them onto session/driver-owned state is now a pure ownership/cleanliness refactor, not a correctness fix.

## Design
- Per-dimension registry structs local to each module (avoids the target↔dimension import cycle): `isa.IsaRegistry`, `os.OsRegistry`, `abi.AbiRegistry`, `of.OfRegistry` — each just the `[8]*XVTable` slot array.
- `target.TargetRegistry` bundles the four; the `Session` (or driver) owns one.
- `register`/`lookup`/`registered_count`/`registered` in the four dimension modules take a `*XRegistry`; the process-global arrays + `ensure_*_registry` are deleted.
- `target.register_all(reg, i)` threads `?reg.isa` / `?reg.of` / … into the 10 registrars (`register_x64`, `register_coff`, `elf.register`, `linux.register_linux`, …), which push into the passed registry instead of a global.
- `target.select(reg, os, arch)` reads the passed registry; `driver` passes the session's.
- `mach info`'s enumeration (`registered_count`/`registered`, PR #1315) reads the owned registry.
- The CLI ordering invariant (register_all before select) becomes an ordinary init dependency on the owned registry.

## Note on the vtable singletons
[...]

Issue #1312, opened 2026-06-11
cli: `mach info` — at-a-glance compiler version + build information


octalide:
Maintainer request (2026-06-11): a quick way to identify the installed binary. `mach info` prints: compiler version, what the binary was built for (host os/isa — foldable at compile time via the new $target.* roots), and the registered capability surface (ISAs, OSes, object formats — enumerable from the registries; this doubles as 'which targets can this binary actually build'). Goals: one screen, no project required (must work outside a mach.toml directory), stable line-oriented format friendly to scripts. Also: surface the version in the bare `mach` usage banner, and accept `mach --version` as a one-line alias for tooling. Versions come from the interned compiler version string (the $project.version flip is staged as #1311 — info consumes the same single source once that lands).

octalide:
**Amendment (maintainer, 2026-06-11):** no global `mach --version` flag — the one-line version output lives at `mach info --version` instead. Bare `mach --version` is not recognized (normal unknown-flag handling). The usage-banner version line and the full `mach info` output are unchanged.

PR #1403, opened 2026-06-13
refactor(target): immutable vtable name/file_extension/reg-class str constants (#1377)

octalide:
Behavior-preserving refactor of #1377 (option A — the immutable-name half).

## Change
- `OsVTable.name`, `IsaVTable.name`, `AbiVTable.name`, `OfVTable.name`, and `RegClass.name` become plain `str` constants instead of per-session-interned `StrId` fields — the registrars set string literals directly, so the vtables are genuinely immutable and no longer re-intern their names each session.
- `OfVTable.file_extension` was **dead** (set by the 3 OF registrars, never read) → removed.
- `mach info` reads `vt.name` directly (a `str`), dropping `name_of`/`cap_fail` and the display interner.

## Verification
- **Byte-identical self-host fixpoint** (the #1319-style check): stage-1 = installed v1.5.1 building this source; stage-2 = stage-1 rebuilding this source; the two binaries are `cmp`-identical → the refactor changes no codegen.
- Suite green: **426 passed, 0 failed** (incl. the #1312 registry-enumeration test).
- Net `-133/+40` across 15 files (removed interning boilerplate).

## Scope note (the other half of #1377)
`register_all` still takes the interner: the OS vtables' `entry_symbol`/`syscall_layer`/`libdir`/`dynamic_linker` remain `StrId` because they're consumed as StrIds in the **load-bearing linker entry path** (`linker.mach:702` entry resolution, `:1365` PT_INTERP). Eliminating the interner entirely means converting those + interning at the linker consumers — split to **#1402** so that load-bearing change gets its own review rather than riding this behavior-preserving name refactor.

Refs #1377. (Leaving #1377 open for your call on closing-as-A vs keeping for #1402.)

### DIFF
--- a/src/cli/cmd/info.mach
+++ b/src/cli/cmd/info.mach
@@ -81,7 +81,7 @@ pub fun run(argc: usize, argv: **u8) i64 {
     print.print(host_isa());
     print.print("\n");
 
-    val code: i64 = print_capabilities(?reg, ?interner);
+    val code: i64 = print_capabilities(?reg);
     intern.dnit(?interner);
     ret code;
 }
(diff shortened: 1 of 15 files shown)


octalide:
Parking this pending octalide's call on the #1377 scope (per team-lead's defer decision). This branch carries the full **option A** (immutable vtable name/file_extension/reg-class str constants; self-host fixpoint byte-identical) — a valid head-start under either resolution: if octalide picks A, this is ready to mark ready + merge; if B (also convert entry_symbol/syscall_layer/libdir/dynamic_linker and push interning into the linker entry/interp path, per #1402), it builds on this. Dropped from the v1.5.2 bundle. Not for merge until octalide decides A vs B.

Issue #1431, opened 2026-06-14
epic: aarch64-linux primary triple


octalide:
## Scope

Bring up **aarch64-linux** as a new primary triple — the first v1.6 target, a lower-risk proving step before darwin (octalide's sequencing decision). It reuses the proven ELF object writer and the linux OS layer; only three orthogonal vtable compositions plus one structural linker change are missing:

- **ISA** (#1171) — `isa/arm64.mach` is a 130-line data-only stub (`select=nil`/`encode=nil`); becomes the x64 four-file shape (`register`/`isel`/`encode`/`printer` + leaf-defs) for fixed-width 32-bit A64.
- **ABI AAPCS64** (#1174) — `ABI_AAPCS64=3` is reserved and `canonical_abi` already routes `(linux,aarch64)→AAPCS64` (target.mach:104), but no `AbiVTable` is ever registered, so `select(linux,aarch64)` fails at `abi.lookup` today.
- **Reloc kinds** (#1163) — the abstract enum stops at `RK_ADDR32NB=5` (of.mach:68); ELF aarch64 map stops at `CALL26` (elf.mach:165).
- **Arch-aware linker** (new) — `apply_one` (be/linker.mach:1552) is x86-only and takes **no arch parameter**, so the internal static linker has **never** correctly applied an aarch64 relocation. Threading `arch_id` + A64 bit-packers is the central change behind "reuse ELF".

Sub-epic of #1180 (multi-target matrix). **Darwin is a separate later effort, not in this epic.**

## Critical correctness facts (from code-verified review)

The stub ships **four silent-miscompile-class** issues that **byte-tests cannot catch** — they MUST be fixed before the first runnable artifact (proof is qemu execution, not bytes):

1. **scratch_reg2/IP1 (X17) never reserved** — `seed_reserved` (regalloc.mach:706-712) reserves only `scratch_reg`; arm64 uses both IP0+IP1 as encoder scratch with a disjoint reload pool, so regalloc will color a live vreg onto X17 and immediate/address legalization clobbers it. Fix: extend `seed_reserved` to also reserve `scratch_reg2` (idempotent for x64 — R10 is already in the reserved reload pool).
2. **FP scratch V31 in the allocatable bank** — vector `count=32` while `fp_scratch_reg=V31`; an FP-scratch use corrupts a float vreg. Fix: hold V31 back (`count=31`), mirroring x64's xmm count=14.
[...]

PR #1461, opened 2026-06-14
fix(aarch64): f64/f32 memory load/store register-class dispatch + naked asm framing

### LINKED ISSUE #1459: aarch64: f64/f32 memory load/store ignore register class (self-host non-determinism)
Part of #1431.

The aarch64 backend's `encode_load` and `emit_store_mem` always used the GP register path, ignoring operand register class. A float loaded **from** memory was emitted as `ldr x` (bits land in a GP register, never the FP register); a float-register store **to** memory as `str x`. The value lands in the wrong bank and the FP register keeps stale garbage (a leftover, ASLR-randomized address).

Lowering creates a single `MIR_LOAD`/`MIR_STORE` with an XMM-class operand and expects the encoder to dispatch on class (as x86 does → MOVSS/MOVSD). aarch64 did not.

**Impact:** self-host fixpoint non-determinism. The compiler reads its own AST float literals via an f64 struct-field load; with the bug those reads returned the field *address*, leaking 24 distinct ASLR pointers into f64 immediate constants (109 sites) — different every run. Byte-identical cross-verification passed because both hosts emit the *same deterministically-wrong* bytes; only the self-host fixpoint (single-compiler-two-runs determinism) caught it.

Local float math is unaffected (locals live in XMM regs / spill through the correct FP path), which is why std's aarch64 suite stayed green.

**Fix:** add a vector-class branch to `encode_load` and `emit_store_mem` emitting `ldr/str Sd|Dd` via `emit_fp_mem`. Arch-isolated to arm64/encode.mach.

**Verified:** aarch64 self-host fixpoint converges byte-identical (s1==s2==s3); 0 residual pointer-shaped f64 constants.

### LINKED ISSUE #1460: aarch64: naked asm-only function framing corrupts _start argc
Part of #1431.

An inline-asm-only function with no frame (`_start`) was still getting the record-at-top prologue (`stp x29,x30,[sp,#-16]!; mov x29,sp`), which shifts SP before `_start` reads the kernel-supplied argc/argv off the entry stack — so a cross-built mach launched with arguments saw argc=0 and printed help.

Two parts, both arch-isolated to arm64/{encode,register}.mach:
- `asm_clobbers` (the AsmClobbersFn vtable slot) was `nil`, forcing the allocator's conservative all-register clobber → every asm function got framed. Implemented per-mnemonic GP clobber inference (mirrors x86-64's `asm_clobbers`).
- `function_is_naked`: an asm-only function with no frame / callee-saves / outgoing args emits NO prologue/epilogue — its inline asm owns the stack (mirrors x86-64).

**Verified:** cross-built mach under qemu reads argv correctly; self-host fixpoint converges.

octalide:
Bundles the two aarch64 backend fixes that unblock the **aarch64 self-host fixpoint** (v1.6.0 release gate). Arch-isolated to `arm64/{encode,register}.mach`; the x86 path is untouched.

## #1459 — f64/f32 memory load/store ignored register class (the self-host non-determinism)
`encode_load` / `emit_store_mem` always used the GP path. A float loaded **from** memory became `ldr x` (bits land in a GP reg, never the FP reg); a float-register store became `str x`. Lowering emits a class-tagged `MIR_LOAD`/`MIR_STORE` and expects class dispatch (x86 does → MOVSS/MOVSD); aarch64 didn't. Fix: vector-class branch in both → `ldr/str Sd|Dd` via `emit_fp_mem`.

The compiler reads its own AST float literals through an f64 field-load, so the bug leaked field **addresses** into f64 immediate constants — 24 distinct ASLR pointers across 109 sites, fresh each run. **Byte-identical cross-verification passed it** (both hosts emit the same deterministically-wrong bytes); only the self-host fixpoint's single-compiler-two-runs determinism check caught it. Local float math is unaffected (locals spill through the correct FP path) — which is why std's aarch64 suite stayed green.

## #1460 — naked asm-only function framing corrupted `_start` argc
`asm_clobbers` was `nil` → allocator's conservative all-register clobber framed every asm function; the frame-record `stp x29,x30,[sp,#-16]!` shifted SP before `_start` read kernel-supplied argc/argv. Fix: per-mnemonic GP clobber inference (`asm_clobbers`) + `function_is_naked` (no prologue/epilogue for a frameless asm-only function). Mirrors x86-64.

## Verification
- **aarch64 self-host fixpoint converges byte-identical**: s1 (x86-cross-built) == s2 == s2b == s3 (native, 2 gens), one sha `7ab82082c216`. Determinism (s2==s2b) and canonical fixpoint (s2==s3) both hold.
- **0** residual ASLR-pointer-shaped f64 constants in the final binary (was 109).
- Built against vendored mach-std dev e5086c10 (complete aarch64 runtime).
- x86 self-host regression: see PR comment (byte-identical).

Closes #1459
Closes #1460
Part of #1431

### DIFF
--- a/src/lang/target/isa/arm64/encode.mach
+++ b/src/lang/target/isa/arm64/encode.mach
@@ -999,6 +999,16 @@ fun emit_store_mem(st: *EncodeState, f: *mir.MirFunction, dst_mem: *mir.MirOpera
     val m: A64Mem = unwrap_ok[A64Mem, str](mr);
     if (m.has_index) { ret err[bool, str]("encode: indexed store not expected"); }
 
+    # a FLOAT store names a vector source: dispatch on the source class to the scalar-
+    # FP store (`str Sd|Dd, [base, #off]`). a symbol destination never reaches here for
+    # a float (lowering LEA-materializes it into a base vreg first).
+    if (reg_is_vec(val_op)) {
+        val rvr: Result[i32, str] = req_vec(val_op);
+        if (is_err[i32, str](rvr)) { ret err[bool, str](unwrap_err[i32, str](rvr)); }
+        emit_fp_mem(st, unwrap_ok[i32, str](rvr)::u32, m.base, m.off, width, false);
+        ret ok[bool, str](true);
+    }
+
     var rt: u32 = ZR;
     if (val_op.kind == mir.MIR_OP_IMM) {
         if (val_op.imm != 0) {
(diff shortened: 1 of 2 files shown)


octalide:
## Verification

**aarch64 self-host fixpoint — byte-identical convergence** (built against vendored mach-std dev e5086c10, the complete aarch64 runtime):
- s1 (x86-cross-built) == s2 == s3 (native under qemu, 2 generations) — one sha `bcd5c05ac25d…`
- Determinism (single compiler, two runs of the same build): byte-identical — the non-determinism that this fixed is gone.
- 0 residual ASLR-pointer-shaped f64 constants in the final binary (was 24 distinct pointers across 109 sites pre-fix).

**x86 self-host regression — shared path unperturbed:** x86 mach built from this branch reproduces itself byte-identically (sha `aec54708…`). The changes are arm64-only (the arm64 backend code only executes when targeting aarch64), so x86 codegen is unchanged and deterministic.

**Diff scope:** arch-isolated to `arm64/encode.mach` + `arm64/register.mach` only.

**/code-review (max):** the f64 load/store fix + naked-asm framing verified clean (callers correctly gated, MIR class-dispatch contract holds, load/store widths unclamped/correct, no SYM reaches the float branch). Review also surfaced a latent under-approximation in the new `asm_clobbers`: pre/post-index writeback `ldp`/`stp` forms (which the inline-asm encoder accepts) write back the base register, which wasn't reported — fixed in commit 2 (`asm_mark_ldstp_wb`). Latent (no current inline-asm uses writeback ldp/stp), so the fixpoint is byte-identical with or without it; re-verified s1==s2==s3 after adding it.

Ready for review.

Issue #1476, opened 2026-06-15
decorators: backtick per-decl codegen directives


octalide:
Per-declaration **codegen decorators** — leading, one per line, call syntax — on the (currently reserved) backtick.

- Lexer: make backtick a real token (today it is a lex error).
- Parser: leading `` `name(args)` `` / `` `flag` `` clauses on a decl; args are comptime exprs.
- Directives:
  - `symbol(str)` — emitted/imported symbol name (fns & globals)
  - `section(str)` — section placement (fns & globals)
  - `inline` — inlining hint (fns)
  - `align(usize)` — alignment (data & types)
  - `library(str)` — per-extern import routing: which of the `mach.toml [os.*] libs` a symbol imports from. The 1:1 replacement for std's `$<sym>.library` setter; pairs with `symbol(...)`.
- **Scope: codegen only** — NOT visibility (that's `pub`). `library(...)` ROUTES an extern to one of the libs declared in `mach.toml [os.*] libs`; it does NOT declare the lib (the link set stays in the manifest, so no second linking surface).

**Acceptance:** decorators parse, attach, and affect emission; `` `align($align_of(u64))` `` (comptime-expr value) works; a windows extern with `` `library("ws2_32.dll")` `` + `` `symbol("socket")` `` imports correctly.

Part of the v1.7 comptime epic.

Issue #1488, opened 2026-06-16
ci: self-host fixpoint coverage gaps — no PR fixpoint; release cmp is stage2==stage3 (blind to uniform codegen changes); aarch64 never byte-diffed


octalide:
Surfaced during the #1484 pre-merge review. The self-host fixpoint — the project's load-bearing correctness invariant — is **not actually measured where it matters**, so a whole class of codegen change can land without the fixpoint guarding it.

### Gaps
1. **No fixpoint on PRs.** Every `ci.yml` lane is single-stage: `mach build .` once (seed → stage1) + tests, and only checks the binary exists. No lane rebuilds the compiler with the freshly-built binary and `cmp`/sha-compares. So a PR's *self-host emission under its own new codegen* is never exercised against a reference — it only runs against the integration test apps.
2. **The release fixpoint is the wrong comparison for codegen-uniformity changes.** `release.yml` (`Build to the fixpoint`, tag-push only) does `./a build . -> b; ./b build . -> out; cmp -s b out` — i.e. **stage2 vs stage3**, both emitted by the *new* compiler. A *uniform* codegen change (e.g. a global-section-alignment policy change applied to every global) is applied identically to `b` and `out`, so `cmp` passes and the change is invisible. The **stage1(seed)-vs-stage2** diff — the one that would catch it, because stage1 is emitted by the *old* seed — runs nowhere.
3. **aarch64 is never byte-diffed.** The aarch64 lane cross-compiles + qemu-smokes a tiny program; it never self-hosts to a fixpoint. So "byte-identical on both backends" is unverified even at release. Structural codegen coverage (e.g. the aligndec section-align assertions) is x86_64/ELF-only.

### Why it matters
#1476's global-section-alignment change (`pack_globals`: hardcoded `align=8` → `max(8, byte_align(g.ty), g.align)`) is safe **only** because of the language invariant "no type has natural alignment > 8" — not because any job proved the emission unchanged. The moment a >64-bit scalar / SIMD / over-aligned type is added, an un-decorated global of that type would silently diverge from the seed, and **nothing in CI would catch it** (build + tests still pass; the blind stage2==stage3 cmp still passes).

### Recommend
- Add a **stage1(seed)-vs-stage2 byte-diff** on PRs (the comparison that actually catches a uniform codegen change). Cheap: one extra rebuild + `cmp`.
[...]

PR #1493, opened 2026-06-16
fix(codegen): emit each sectioned function once

### LINKED ISSUE #1489: fast-follow #1476: function-section — emit each sectioned function once (drop dead-bytes duplication)
**Fast-follow to #1476** (octalide chose merge-now + fast-follow). The shipped function-section impl is correct but leaves each `section(...)` function duplicated: a live copy in its named section AND a dead, un-GC-able copy in `.text`. Every object using the feature carries the function twice; the duplication is inherited into the final binary. The #1484 review flagged it as "a structural compromise that should not remain the final design."

### Fix (the "B" plan, already reasoned out by the implementer)
Build `.text` from the **non-sectioned** functions only — excise each sectioned `[start,end)` range and re-home all `.text` marks/relocs by the cumulative excised offset; sectioned functions go to their named sections as today. **Additive:** only triggers when a `section(...)` decorator is present, so the un-decorated path stays byte-identical. ~60–80 lines, all in `src/lang/be/codegen.mach` (`pack_text` / `make_func_section` / `build_section_placements`); encoders untouched (section assignment is object-assembly, not instruction-encoding).

### Watch
- **`enc.text` ownership:** in the repartition branch `.text` becomes a new image-owned buffer, so `enc.text` is no longer a section's bytes and must be explicitly `deallocate`d (with `s.alloc`) — leak/double-free if the branch/free split is wrong. Localized to the sectioned branch; does NOT touch the fixpoint (compiler hits the default branch).
- Edge: a module where *every* function is sectioned → empty `.text` (len 0).

### Verify
- True 3-stage fixpoint **incl. seed(stage1)-vs-stage2** byte-identical on both backends (per #1488 — don't rely on the blind stage2==stage3 cmp).
- `sectiondec` e2e still green, incl. `main`/`g_def` now in the **rebuilt** `.text`/`.data` at correct offsets, running on ELF (x64 + aarch64) and COFF/wine.

Part of v1.7 epic #1477.

octalide:
Fast-follow to #1476. The shipped function-section impl left each `section(...)` function duplicated: a live copy in its named section AND a dead, un-GC-able copy in `.text`. This makes each sectioned function emit once.

## Change (the "B" plan, all in `src/lang/be/codegen.mach`)
- `.text` is built from the **non-sectioned** functions only. Each sectioned `[start, end)` range — already copied into its named section — is excised, and the remaining bytes are packed contiguously into a fresh image-owned buffer.
- Every `.text`-resident symbol mark / relocation is re-homed by the cumulative excised length below it (`rebased_text_offset`); sites inside a sectioned range are re-homed to the named section as before.
- **Additive:** with no `section(...)` decorator (`count == 0`) the encoder blob is adopted whole and emission is **byte-for-byte unchanged**.
- Intra-function branch fixups need no rebasing (they are position-independent within a copied contiguous range); inter-function calls are relocations the linker resolves by symbol.

## Ownership
`pack_text` is the last reader of `enc.text`, so it owns the blob's disposition: adopt it (default path) or free the now-orphaned blob (repartition path) with `s.alloc`. A module whose every function is sectioned yields an empty `.text` (nil bytes, len 0).

## Verify
- True 3-stage seed-inclusive self-host fixpoint (stage1==stage2==stage3) on x86_64 and aarch64 (qemu).
- `sectiondec` e2e green (`main`/`g_def` in the rebuilt `.text`/`.data`, `f_sec`/`g_sec` in their named sections).
- Un-decorated fixpoint unchanged from the dev baseline.

Closes #1489

### DIFF
--- a/src/lang/be/codegen.mach
+++ b/src/lang/be/codegen.mach
@@ -147,10 +147,11 @@ pub fun codegen_module(s: *sess.Session, tgt: *target.Target,
 
 # build the object image from the encoder output and the module's globals.
 #
-# creates the `.text` section from the encoded bytes, registers every symbol
-# mark against it, attaches the encoder's relocations, then materialises each
-# module global into a data / rodata / bss section with its defining symbol. on
-# any builder failure the partially built image is torn down before returning.
+# moves each `section("...")` function into its named section, builds `.text`
+# from the remaining non-sectioned bytes, registers every symbol mark against its
+# section, attaches the encoder's relocations, then materialises each module
+# global into a data / rodata / bss section with its defining symbol. on any
+# builder failure the partially built image is torn down before returning.
 # ---
 # s:     shared session; supplies the allocator and interner
 # tgt:   resolved target; supplies type sizes for bss global layout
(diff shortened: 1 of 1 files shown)


octalide:
## Verification (local)

**Self-host fixpoint — true 3-stage, seed-inclusive (#1488):**
- x86_64: seed(1.6.0)→stage1→stage2→stage3, `cmp` all three **byte-identical** (stage1==stage2==stage3).
- aarch64 (qemu-cross): x86 stage3 cross-builds A1; A1→A2→A3 under `qemu-aarch64`, **byte-identical** (and cross-built A1 == native-built A2).
- aarch64 emitter byte-verify (`.github/tests/aarch64enc`): all PASS.

The seed (pre-#1476) agreeing byte-for-byte with the new compiler confirms the un-decorated (`count == 0`) path is unchanged.

**`sectiondec` e2e — all three object formats, exit 123:**
| target | run | `.text` | `f_sec` | dead copy |
|---|---|---|---|---|
| ELF x86_64 | exit 123 | 0x42 = just `main` | `.hottext` | **gone** |
| ELF aarch64 (qemu) | exit 123 | 0x44 = just `main` | `.hottext` | **gone** |
| COFF/PE (wine) | exit 123 | 0x47 = just `main` | `.hottext` | **gone** |

Object inspection (x86_64): `.text` is exactly `main`'s body — no second copy of `f_sec`; `f_sec` appears once in `.hottext`; `main`→`f_sec` is an `R_X86_64_PC32` reloc at the correct rebased offset; `g_def` in `.data`, `g_sec` in `.machsec`. Each sectioned function now emits **once**.

Issue #1483, opened 2026-06-15
comptime: correctly-rounded float-literal folding (large exponents lose precision)


octalide:
The compiler's float-LITERAL folding is imprecise for large exponents: `1.0e100` folds to `1.0000000000000006e100` (~6 ULP off true `1e100`). Confirmed by mach-std — feeding `write_f64` the EXACT `1e100` bits yields `"1e100"`, but the source literal yields the wrong value.

Cause: the comptime float-literal folder accumulates error (naive `*= 10` per exponent step) instead of a correctly-rounded decimal→f64 conversion.

Fix: correctly-rounded decimal→binary float-literal folding (same algorithm family as mach-std's dragon4-inverse bignum). Part of the v1.7 "correct floats" story (owner-confirmed) alongside std's `parse_f64` rewrite (mach-std#287).

Repo: mach (comptime float-literal folding). v1.7.

octalide:
**Confirmed repro + root cause** (verified by mach-std against exact dragon4 goldens while building write_f64):

- The literal `1.0e100` lexes to `1.0000000000000006e100` — ~6 ULP off the true nearest double to 1e100.
- Proof it's PARSING not formatting: a correct formatter prints the literal `1.0e100` as "1.0000000000000006e100" but prints the EXACT 1e100 bit pattern `0x54B249AD2594C37D` as "1e100". So the wrong value enters at literal folding.
- Small/moderate literals lex fine (pi @ 16 sig digits, 1e20, 1e21); only large-magnitude exponents drift — consistent with `eval_lit_float`'s naive `value *= 10` per exponent digit instead of correctly-rounded decimal→binary.

**Fix approach (per the build-order ruling):** rewrite `eval_lit_float` (src/lang/fe/comptime.mach) to correctly-rounded decimal→f64, REUSING mach-std's `std.math.bignum` from #287 (same algorithm family as their parse_f64) rather than duplicating — call their correctly-rounded path or build on the bignum. Seed-safe (pure old syntax). Gated: after #1476; after #287 lands; the compiler's vendored-std pin must advance to the (seed-safe) #287 bignum commit so `eval_lit_float` can import it while staying pre-#282.

**Test golden:** `1.0e100` must fold to `0x54B249AD2594C37D`. Add the full round-trip set (large +/- exponents, subnormals, ties-to-even boundaries) in .github/tests.

PR #1502, opened 2026-06-17
fix: reconcile vendored mach-std pin; harden the DECOUPLE freeze (#1486)

### LINKED ISSUE #1486: fix: reconcile vendored mach-std pin (gitlink #281 vs mach.lock #265); harden the DECOUPLE freeze
**Corrected (octalide, authoritative): `mach build` resolves deps by PATH, not by commit. `mach dep` is the only version-resolution mechanism. The freeze must be done the mach-native way — not by wrangling a git submodule.**

### How mach deps actually resolve (verified: `mach help dep` + `src/cli/cmd/dep.mach`)
- A dep has one source form: a git url + **ref**, or a path. Ref formats: `branch/<n>` (advanceable — `mach dep update` moves it to the remote tip), `tag/<n>` and `commit/<n>`/bare-sha (**immutable** — update is a no-op). Acquired into `dir_dep/<name>/` (here `dep/mach-std/`).
- `mach dep pull` realizes the manifest (clone/resolve/repair) and records each git dep's resolved commit in `mach.lock`.
- **`mach build` resolves by PATH** — it uses the realized `dep/mach-std/` directory and **never shells out to git**. So `mach.lock` is `mach dep`'s lockfile *output*, not what `build` consults.

### The actual problem
`dep/mach-std` is **also a git submodule** (`.gitmodules` + a `160000` gitlink at `e8ce7dc`) — a second, non-`mach dep` mechanism bolted onto the same path. It tracks `branch = dev`, so `git submodule update --remote` would advance the gitlink; once #282 merges to mach-std `dev`, that HEAD carries `va:`/`$each` syntax the v1.6.0 seed cannot parse → self-host breaks. Meanwhile `mach.toml` pins `ref = "branch/main"` (advanceable, not a freeze). The submodule is the source of both the gitlink/lock divergence and the auto-advance foot-gun; it should not be a resolver at all.

### Fix (the mach-native freeze)
1. **Pin the commit in `mach.toml`:** `[deps.mach-std] ref = "commit/<frozen-pre-#282-sha>"` (immutable — `mach dep update` is a no-op on `commit/`).
2. **`mach dep pull`** to realize it into `dep/mach-std/` + record in `mach.lock`.
3. **Remove the conflicting git submodule** for `dep/mach-std` (drop the `.gitmodules` entry + gitlink); let `mach dep` own the path. Update CI to `mach dep pull` before `mach build` instead of `checkout --recurse-submodules`.
4. **When stable (post v1.7 re-seed):** revert `mach.toml` to `ref = "branch/main"` + `mach dep pull`. Commit-pinning is a temporary freeze only; `branch/main` is the steady state.

### Acceptance
- `mach.toml` is the single source of truth (`commit/<sha>` during the freeze); no git submodule for `dep/mach-std`; `mach dep pull` populates the path; `mach build` uses it offline.
- `git submodule update --remote` can no longer advance the vendored std (the mechanism is gone).
- Self-host fixpoint green; the revert-to-`branch/main` step documented for the post-release sweep.

### Gate
Land before #282 merges to mach-std `dev` (gated behind compiler features #1474/#1475). Part of v1.7 DECOUPLE integrity (epic #1477).

octalide:
Closes #1486

Reconciles the divergent vendored-`mach-std` pin and replaces the auto-advance-prone git submodule with the mach-native immutable freeze.

## Investigation: how the build resolves `mach-std` (verified in source)

- **`mach build` resolves deps purely by filesystem PATH.** `driver.resolve_dep` (`src/lang/driver.mach:1460`) builds `<root>/<dir_dep>/<alias>/` (`dep/mach-std/`), reads that dir's own `mach.toml`, and records the directory as `vendor_root`. It never reads `mach.lock`, never reads `mach.toml`'s `git`/`ref` fields, and never shells out to git.
- **`mach dep` is the only version-resolution mechanism.** `mach dep pull` (`src/cli/cmd/dep.mach`) realizes the manifest into `dep/mach-std/` and records the resolved commit in `mach.lock`; `mach dep update` is the only lock-advancer, and `ref_advanceable` (`dep.mach:631`) returns **false** for `tag/`, `commit/`, and bare-sha refs — so `commit/<sha>` is an immutable no-op on update.

**Consequence:** the build content is whatever sits in `dep/mach-std/` on disk. CI/release populate it with `actions/checkout … submodules: true` → the **submodule gitlink `e8ce7dc`**; only local `mach dep pull` runs populated it from `mach.lock` (`c0f1d2c8`). That is the divergence.

## Topology (mach-std)

- `main` HEAD = `e8ce7dc` — **the submodule gitlink; what CI and `release.yml` actually build against today**.
- `mach.lock` = `c0f1d2c8` — a strict **ancestor** of `e8ce7dc` (main advanced past the lock); only local builds used it.
- `main` and `dev` (`c100ea4`) have diverged; **both are pre-variadic-rewrite**. `dev` HEAD is the merge-base of the variadic branch, i.e. the rewrite sits directly on top of `dev`.

## Freeze target: `commit/e8ce7dc1921e8d02d92584ee371de4d1dfe9d2be`

The newest pre-variadic commit on the **release line** that the compiler is **already proven** to self-host byte-identically against (current dev CI green; v1.6.0 shipped against it). Reconciling **up** to the gitlink (not down to the older lock) is a true *freeze*: it preserves the exact std CI/release already compile, changing nothing about the seed inputs — only making the pin immutable and mach-native.

## Changes

- `chore(deps)`: remove the `dep/mach-std` git submodule (`.gitmodules`, gitlink, local config); ignore `/dep/`.
- `build(deps)`: pin `mach.toml` to `ref = "commit/e8ce7dc…"`; regenerate `mach.lock`.
- `ci`: realize `mach-std` via `mach dep pull` instead of `submodules: true`; add a content-based seed-safety guard.

## Revert path (post v1.7 re-seed)

Restore `ref = "branch/main"` + `mach dep pull`. Commit-pinning is a temporary freeze; `branch/main` is the steady state.

## Verification

- [x] `mach build .` against the reconciled pin
- [x] self-host fixpoint byte-identical: `mach dep pull` (fresh clone of `dep/mach-std` @ `e8ce7dc`) → seed→stage1, stage1→stage2, `cmp stage1 stage2` identical — `sha256 caec8c4adc12e6c02ca568e2cdf1e465ea252096d79d0e0172fc47f633620d22` (freeze is inert vs the released seed)
- [x] seed-safety guard passes on the frozen std and the regex catches `va:`/`$each`/`$mach.build.*` (verified live, not a no-op)

### DIFF
--- a/.github/tests/aarch64run/run.sh
+++ b/.github/tests/aarch64run/run.sh
@@ -18,7 +18,7 @@ REPO_ROOT="$(cd "$HERE/../../.." && pwd)"
 STD="$REPO_ROOT/dep/mach-std"
 
 if [ ! -d "$STD/src" ]; then
-    echo "error: vendored std not found at $STD (run: git submodule update --init)" >&2
+    echo "error: vendored std not found at $STD (run: mach dep pull)" >&2
     exit 2
 fi
 
--- a/.github/tests/absout/run.sh
+++ b/.github/tests/absout/run.sh
@@ -17,7 +17,7 @@ REPO_ROOT="$(cd "$HERE/../../.." && pwd)"
 STD="$REPO_ROOT/dep/mach-std"
 
 if [ ! -d "$STD/src" ]; then
-    echo "error: vendored std not found at $STD (run: git submodule update --init)" >&2
+    echo "error: vendored std not found at $STD (run: mach dep pull)" >&2
     exit 2
 fi
 
--- a/.github/tests/aligndec/run.sh
+++ b/.github/tests/aligndec/run.sh
@@ -23,7 +23,7 @@ REPO_ROOT="$(cd "$HERE/../../.." && pwd)"
 STD="$REPO_ROOT/dep/mach-std"
 
 if [ ! -d "$STD/src" ]; then
-    echo "error: vendored std not found at $STD (run: git submodule update --init)" >&2
+    echo "error: vendored std not found at $STD (run: mach dep pull)" >&2
     exit 2
 fi
 
--- a/.github/tests/asmcarry/run.sh
+++ b/.github/tests/asmcarry/run.sh
@@ -18,7 +18,7 @@ REPO_ROOT="$(cd "$HERE/../../.." && pwd)"
 STD="$REPO_ROOT/dep/mach-std"
 
 if [ ! -d "$STD/src" ]; then
-    echo "error: vendored std not found at $STD (run: git submodule update --init)" >&2
+    echo "error: vendored std not found at $STD (run: mach dep pull)" >&2
     exit 2
 fi
 
--- a/.github/tests/asmclobber/run.sh
+++ b/.github/tests/asmclobber/run.sh
@@ -19,7 +19,7 @@ REPO_ROOT="$(cd "$HERE/../../.." && pwd)"
 STD="$REPO_ROOT/dep/mach-std"
 
 if [ ! -d "$STD/src" ]; then
-    echo "error: vendored std not found at $STD (run: git submodule update --init)" >&2
+    echo "error: vendored std not found at $STD (run: mach dep pull)" >&2
     exit 2
 fi
 
--- a/.github/tests/asmcomment/run.sh
+++ b/.github/tests/asmcomment/run.sh
@@ -19,7 +19,7 @@ REPO_ROOT="$(cd "$HERE/../../.." && pwd)"
 STD="$REPO_ROOT/dep/mach-std"
 
 if [ ! -d "$STD/src" ]; then
-    echo "error: vendored std not found at $STD (run: git submodule update --init)" >&2
+    echo "error: vendored std not found at $STD (run: mach dep pull)" >&2
     exit 2
 fi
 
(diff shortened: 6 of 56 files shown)


octalide:
Admin-merged over a red **Self-host Fixpoint** lane. The red is the *inherited* intentional float-fold divergence from #1483: CI builds this PR merged with dev, which now carries #1483's precise float folding, so the v1.6.0 seed-built stage1 differs from the self-built stage2. **This PR is itself inert** — the pin freeze was verified byte-identical (same caec8c4a fixpoint sha as dev); it changes only dep vendoring, not codegen. Every other lane is green. The v1.7.0 release re-seeds the fixpoint gate and clears this.

octalide:
Admin-merged over a red **Self-host Fixpoint** lane. The red is the *inherited* intentional float-fold divergence from #1483: CI builds this PR merged with dev, which now carries #1483's precise float folding, so the v1.6.0 seed-built stage1 differs from the self-built stage2. **This PR is itself inert** — the pin freeze was verified byte-identical (same caec8c4a fixpoint sha as dev); it changes only dep vendoring, not codegen. Every other lane is green. The v1.7.0 release re-seeds the fixpoint gate and clears this.

Issue #288, opened 2026-02-17
overhaul `mach test` CLI output and per-test execution


octalide:
## Summary

Rewrite the test runner to execute each test in its own process with captured output, replacing the current module-level pass/fail with per-test granularity.

### CLI
- Remove `-m`/`--modules` flag (redundant with `-v`)
- Replace `--filter` with `--test <pattern>` and `--module <pattern>` for targeted selection
- Add `-q`/`--quiet` to suppress stdout (stderr always shown)

### Architecture
- One binary per module, dispatched once per test via argv index
- Each test runs in its own forked process with stdout/stderr captured via pipe
- Test binary remains target-agnostic (pipe/dup2 only in the compiler's test runner)
- Two-pass approach: count tests first, then compile and run

### Output Format
```
[module.path] [ 1/496] [pass] test name
[module.path] [ 2/496] [fail] test name
<captured output>

[results] [pass] 494
[results] [fail] 2
[results] [fail] [compile] 0
[results] [fail] [runtime] 2
- module.path: test name
```

- Default: only show failing tests and their output
- `-v`: show all tests in the stream
- `-q`: suppress captured stdout (stderr always displayed)
- Progress counter `[n/total]` with left-padded n

### Test Names
Test names from `test "name" { ... }` are the primary identifier in all output. Modules are organizational context, not the unit of testing.

PR #1503, opened 2026-06-17
fix(comptime): correctly-rounded float-literal folding

### LINKED ISSUE #1483: comptime: correctly-rounded float-literal folding (large exponents lose precision)

octalide:
Closes #1483.

Replaces the lossy `*= 10` per-exponent-step float-literal folding in `eval_lit_float` with correctly-rounded decimal→f64 (round-to-nearest-even) via exact bignum arithmetic — the same algorithm family as std's `parse_f64`. A literal now folds to the SAME bits `std.parse_f64` produces for that string.

### Approach
- `mach.lang.bignum`: leaf port of std's proven fixed-capacity bignum (std.math.bignum landed on mach-std dev via #288). Ported rather than imported because that path is not yet on the branch the compiler pins (`branch/main`); advancing the pin would be a broad cross-repo integration, out of scope for this fix.
- `mach.lang.float.bits_f64`: u64→f64 reinterpret (inverse of `f64_bits`).
- `eval_lit_float`: accumulates the exact significand into a `Big` + decimal exponent (768-digit cap with a sticky-truncation bit), then `decimal_to_f64` rounds correctly.

### Verification
- Inline goldens assert exact bits for hard cases: large exponents (1e100→0x54B249AD2594C37D, 1e308, 1e-308), subnormals, smallest/largest normals, ties-to-even (down/up) and the sticky tie, overflow→inf.

### DIFF
--- a/src/lang/bignum.mach
+++ b/src/lang/bignum.mach
@@ -1,0 +1,266 @@
+# mach.lang.bignum: fixed-capacity big integer for exact decimal<->binary float
+# conversion.
+#
+# A leaf module (std-only deps) so comptime float-literal folding can fold a
+# literal to the exact correctly-rounded f64 bits without a lossy shortcut.
+# Ported from std.math.bignum (the proven path behind std.parse_f64); kept in
+# parity so the compiler folds a literal to the same bits std.parse_f64 yields.
+#
+# limbs are little-endian 32-bit; n counts active limbs (n == 0 is zero, no
+# leading-zero limbs). 128 limbs (4096 bits) covers every f64 conversion
+# (<= ~1100 bits). all multiplies are by small constants, so no 64x64->128
+# is needed.
+
+use std.types.bool.bool;
+use std.types.size.usize;
+
+# Big: a fixed-capacity unsigned big integer.
+pub rec Big {
+    n: usize;
+    d: [128]u32;
+}
+
+# zero: set b to zero.
+pub fun zero(b: *Big) {
+    b.n = 0;
+}
+
+# norm: drop leading-zero limbs so n is exact.
+pub fun norm(b: *Big) {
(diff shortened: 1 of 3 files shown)


octalide:
**Admin-merge over a red Self-host Fixpoint lane — intentional, verified.**

This PR makes the compiler's float-literal folding correctly-rounded, which changes the bits of float literals in the compiler's *own* source. So the v1.6.0 seed (old folding) and this compiler (new folding) emit different constants → `stage1 != stage2` → the #1488 fixpoint lane is RED by design (its "intentional codegen change" case).

Verified locally that this is intentional and self-consistent, not a bug:
- **stage2 == stage3 byte-identical** (sha256 `6738135e622437acbce09c59d4080c2a2e669e6edef494d722b68d371f99e882`) — the new compiler reproduces itself.
- stage1 (seed-built) differs, as expected.
- All 25 exact-bits float goldens pass (Test Suite green); the algorithm is a faithful port of std's vetted `decimal_to_f64`.

The **v1.7.0 release re-seeds** at this compiler, after which the fixpoint lane returns to green. Admin-merging per the gate's designed override.

Issue #1391, opened 2026-06-13
driver: build all manifest targets into one deduplicated Project (multi-target graphs for tooling)


octalide:
## Problem

`build_project` builds a single selected target's import closure. Long-lived tooling sessions (mach-lsp) need the union of all manifest targets: files reachable only from a non-default target are invisible to cross-file features today, so e.g. a workspace-wide rename silently skips them (octalide/mach-lsp#58). Multi-artifact projects are in the pipeline, so this is now required, not theoretical.

## Ask

A driver API that builds every target of a manifest into one Project (or an equivalent queryable union), with module identity deduplicated across targets — a module shared by two targets must appear once, so consumers don't emit duplicate locations/edits for it. Dedup belongs in the driver, where module identity (FQN/path ↔ ModuleId) is defined.

Design notes from the consumer side:
- Per-target entry FQNs already exist (`build_project_sel`); the union is conceptually "dfs_load from every target entry into one module table", which the by-FQN dedup map may already make cheap within a single Project.
- Conditional compilation: `$if` branches taken can differ per target — the API needs a stance (per-target resolve results vs a declared-target-set resolve), even if v1 just documents "default target's branches win".

Related: #1389 (incremental rebuild / source reclamation) touches the same driver lifecycle — worth designing together.

Tracking on the consumer side: octalide/mach-lsp#58.

Issue #1218, opened 2026-06-09
chore: review `mach.toml` schema


octalide:
Conduct a comprehensive review and generalization of the `mach.toml` manifest schema.

**Context:**
The current schema works but should be evaluated for consistency, extensibility, and future-proofing as the project evolves.

**Areas to review:**
- Field naming and structure consistency
- Default value handling
- Optional vs required fields
- Support for future cross-compilation scenarios
- Backward compatibility concerns
- Documentation clarity

**Deliverable:**
Updated `mach.toml` with refined schema (to be determined in manual review).

**Related:**
- #1215 (init scaffolding now generates multi-target manifests)
- #1217 (version duplication issue)

octalide:
## Design spec: mach.toml v2 (settled with the maintainer, 2026-06-10 — implement in v1.4.0)

**Principle:** nothing automatic, nothing derived, nothing hidden. The manifest is the complete, readable statement of what a project builds. Cleanliness comes from eliminating repetition *between axes* — the cartesian product lives in the build engine, never in the file. Rules the file states (the layout template) are acceptable; silent built-in behavior is not.

**Root defect being fixed:** `[targets.<name>]` fuses three orthogonal axes — artifact (what), platform (where it runs), layout (where outputs go). Multi-artifact projects (e.g. a coreutils with dozens of binaries) explode M×N, and even mach's own manifest repeats `entrypoint` per platform and hand-declares convention paths.

### Schema

```toml
[project]
id      = "coreutils"            # required
version = "0.1.0"                # required
name    = "Core Utilities"       # optional metadata: name, description, license, authors
src     = "src"                  # explicit project dirs (dir_* prefixes dropped)
out     = "out"
dep     = "dep"
target  = "linux"                # default target (a [target.*] key), explicit
layout  = "{out}/{target}/{profile}/bin/{name}{ext}"   # the ONE layout rule, visible

[target.linux]                   # platforms: fully-spelled tuples; nothing inferred
isa = "x86_64"
os  = "linux"
abi = "sysv64"

[target.windows]
isa  = "x86_64"
os   = "windows"
abi  = "win64"
ext  = ".exe"                    # explicit, lives where it belongs; default ""
libs = ["kernel32.dll"]          # platform-level link overlay
defines = []                     # per-target comptime defines (#1191)

[bin.ls]                         # artifacts: every one declared, named by table key
entry = "src/bin/ls.mach"

[bin.cat]
entry = "src/bin/cat.mach"

[...]

Issue #1485, opened 2026-06-15
doc: document all modules (any target) + mirror module path in doc/ tree (drop doc/api)


octalide:
`mach doc` needs two changes:

**1. Document ALL modules regardless of release target.**
It currently FAILS on a multi-artifact project, requiring `--bin`/`--lib` selection. `mach doc` should document the whole project's module set across every declared artifact/target — not force a single-artifact selection. This is the same multi-artifact union need as the LSP (#1481): document the union of every artifact's import closure (cf. `driver.build_project_union`), so a `bin`+`lib` project documents both.

**2. Output structure mirrors the module path — drop the flat `doc/api`.**
Stop emitting into a generic `doc/api`. Instead write one file per module at `doc/<module-path-subdirs>/<module>.md` — e.g. module `mach.cli.cmd.build` → `doc/mach/cli/cmd/build.md`, so `doc/mach/cli/*.md` etc. mirror the module namespace as a directory tree.

Repo: mach (`cli/cmd/doc.mach` + doc output). Related: #1481 (shared multi-artifact-union need in the driver).

octalide:
Dropped. Output already mirrors the FQN path (just under `doc/api/`) and #1481 closed the artifact-union half. Not worth tracking as a standalone issue; re-file narrowly if the `doc/api` prefix-drop is still wanted.

Issue #1481, opened 2026-06-15
driver: union over artifacts so the LSP loads multi-artifact projects (bin+lib)


octalide:
mach-lsp fails to load any project whose `mach.toml` declares MULTIPLE artifacts (e.g. both `[bin.blit]` and `[lib.blit]`): cross-module features are silently disabled, with

> `mach-lsp: failed to load project at .../blit: mach.toml: multiple artifacts declared; select one with --bin/--lib (bin blit, lib blit) (cross-module features disabled for this project)`

**Root cause:** mach-lsp loads the project graph via `driver.build_project_union` (`mach-lsp/src/project.mach:672`), but that path still resolves a SINGLE artifact and errors when several are declared (`resolve_artifact` → `mach/src/lang/manifest.mach:1672`). The LSP can't pass `--bin`/`--lib` and wants the whole module graph anyway.

**Fix (proposed):** extend the union to ARTIFACTS, mirroring the multi-TARGET union (#1391). The union path (`build_project_union` or a sibling the LSP calls) should union the import closures of EVERY declared artifact × every target into one deduplicated Project — never require a single-artifact selection on the union/tooling path. Then a multi-artifact project's bin + lib modules all load and cross-module features (refs / rename / diagnostics across the whole graph) work.

**Acceptance:**
- A project with `[bin.X]` + `[lib.X]` loads in mach-lsp with cross-module features ENABLED.
- The union includes modules rooted by every artifact (not just one).
- Single-artifact projects unchanged; `mach build` (which legitimately requires one artifact) is NOT affected — only the union/tooling path.

Fix lives primarily in **mach** (driver/manifest union path); **mach-lsp** is the consumer (likely no change, or trivial). Surfaced from real LSP use on project `blit`.

octalide:
Related: #1485 (mach doc) needs the same multi-artifact union — both want build_project_union to document/load all artifacts without --bin/--lib selection.

octalide:
Duplicate of #1505 — same fix (union the import closure over every declared artifact on the driver/tooling path). The mach-lsp multi-artifact load failure reported here is exactly that symptom and needs no LSP-side change. Implemented by PR #1507. Closing in favor of #1505.

PR #1507, opened 2026-06-19
feat(driver): union the import closure over every artifact (#1505)

### LINKED ISSUE #1505: driver: union the import closure over every artifact, not just one (multi-artifact tooling graphs)
## Problem

`driver.build_project_union` (the long-lived multi-target tooling entry, #1391) unions the import closure over every declared **target**, but still resolves a **single artifact**. Its chokepoint, `load_config`, calls `manifest.resolve_build_unit` which resolves ONE artifact via `manifest.resolve_artifact`. When a `mach.toml` declares multiple artifacts (any mix of `[bin.*]`/`[lib.*]`) and no `--bin/--lib` selector is given, `resolve_artifact` returns nil and `resolve_build_unit` errors with:

```
mach.toml: multiple artifacts declared; select one with --bin/--lib (...)
```

That error aborts the whole union load, so tooling (the mach-lsp language server) over a multi-binary project gets **nothing** — no symbols, no go-to-def, no rename across any module.

The multi-artifact matrix was explicitly deferred — see `src/lang/manifest.mach` `resolve_artifact` ("the multi-artifact matrix lands in a later phase of #1218").

A real example: a project declaring `[lib.co]`, `[bin.co]`, `[bin.tmp]` loads nothing in the LSP.

## Goal

Extend `build_project_union` so it loads the import closure over the full (target × artifact) matrix — every declared artifact's entry module is loaded into the single deduplicated Project — mirroring how the target axis is already unioned (#1391). Workspace-wide tooling then sees every module reachable from any artifact entry.

## Design (mirror the existing target-axis union)

1. **Primary selection unchanged.** `build_project_sel` and every single-artifact selected build stay byte-for-byte identical. Under `for_union` only, the primary picks the first declared artifact as the ctx (mirroring `resolve_native` falling back to the first declared target) instead of erroring on ambiguity.
2. **Union the rest.** Record every declared artifact's entry FQN on the Project (a `union_entries`/`union_entry_count` array, parallel to `union_tuples`), populated in `load_config` under `for_union`, freed in `dnit_project`, initialized in `init_project`. After the primary `dfs_load` in `build_project_impl`, `dfs_load` each additional entry into the same Project (dedup by FQN is automatic).
3. **The ambiguous-artifact error still fires for `build_project_sel`** with no selector (CLI `mach build`/`mach run` ambiguity is intentional).

## Tests

Add a driver union regression test: a multi-artifact fixture where one module is reachable ONLY from a non-default artifact's entry; assert the union Project loads it and the single-selected build does not.

Refs #1218, #1391.

octalide:
## Summary

`driver.build_project_union` (the long-lived multi-target tooling entry, #1391) unioned the import closure over every manifest **target** but still resolved a **single artifact**. When a `mach.toml` declares multiple artifacts (any mix of `[bin.*]`/`[lib.*]`) and no `--bin/--lib` selector, `manifest.resolve_build_unit` returned the `multiple artifacts declared; select one with --bin/--lib (...)` error, aborting the whole union load — so tooling (mach-lsp) over a multi-binary project saw nothing.

This mirrors the existing target-axis union on the artifact axis so the union loads the closure over the full (target × artifact) matrix.

## Changes (`src/lang/driver.mach`)

- **Primary selection under `for_union`:** when several artifacts are declared with no selector, the union takes the **first declared artifact** as the primary ctx (mirroring `resolve_native`'s first-declared-target fallback) instead of erroring. `build_project_sel` and every selected/single-artifact build are unchanged — the ambiguity error still fires for them.
- **`union_entries`/`union_entry_count` on `Project`** (parallel to `union_tuples`): every declared artifact's entry FQN (each artifact's base `entry` plus any per-cell `[*.target.*]` entry override), deduped by FQN. Populated in `load_config` under `for_union`, initialized in `init_project`, freed in `dnit_project`.
- **`build_project_impl`** loads each `union_entries` FQN after the primary `dfs_load` (dedup by FQN is automatic, so the primary and shared modules re-load as no-ops).

A module that does not fully resolve under the default ctx still records per-module diagnostics and leaves the Project usable — the union is not aborted (existing #1391 behavior).

## Tests

Two new driver-union tests:
- every artifact's entry closure is loaded, not just the primary's — a `beta_only` module reachable ONLY from the non-default artifact's entry is present in the union.
- an ambiguous multi-artifact default errors for a single build but loads for the union — the selector-less `build_project_sel` still errors, a `--bin alpha` build loads only alpha's closure (never `beta_only`), and the union loads every artifact.

## Verification

- `mach build .` succeeds; byte-identical self-hosting fixpoint holds (seed-built == self-built).
- `mach test .` — **526 passed, 0 failed** (9 driver-union tests including the 2 new ones).
- `manifest` and `module` integration suites pass, including `ambiguous 'mach run .' requires --bin` (the CLI ambiguity guard is preserved).

Refs #1218, #1391. Closes #1505.

### DIFF
--- a/src/lang/driver.mach
+++ b/src/lang/driver.mach
@@ -291,6 +291,14 @@ pub rec Project {
     union_tuples:       *TargetTuple;
     union_tuple_count:  u32;
 
+    # union build (#1505): the entry-module FQN of every declared artifact, so the
+    # union loads the import closure over the whole (target × artifact) matrix — not
+    # just the primary artifact's. Recorded in `load_config` under `for_union`,
+    # walked after the primary `dfs_load` in `build_project_impl` (dedup by FQN is
+    # automatic). nil/0 on a single-artifact or single-target build.
+    union_entries:      *intern.StrId;
+    union_entry_count:  u32;
+
     compiler_name:  intern.StrId;
     compiler_ver:   intern.StrId;
 
@@ -477,6 +485,20 @@ fun build_project_impl(s: *sess.Session, project_root: str, sel: *manifest.Selec
         ret err[Project, str](unwrap_err[sess.ModuleId, str](load_r));
     }
 
+    # union build (#1505): load every other declared artifact's entry into the same
+    # Project so tooling sees the closure over the whole (target × artifact) matrix,
+    # not just the primary artifact's. dfs_load dedups by FQN, so the primary entry
+    # (also present in union_entries) re-loads as a no-op, as do shared modules.
+    var ei: u32 = 0;
+    for (ei < p.union_entry_count) {
+        val ear: Result[sess.ModuleId, str] = dfs_load(?p, p.union_entries[ei]);
+        if (is_err[sess.ModuleId, str](ear)) {
+            dnit_project(?p);
+            ret err[Project, str](unwrap_err[sess.ModuleId, str](ear));
+        }
+        ei = ei + 1;
+    }
+
     # the module table is stable past the load pass; publish every module's
     # AST to the session registry so later phases can reach cross-module
     # declarations (imported record fields, generics from transitive deps).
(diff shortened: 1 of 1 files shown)


octalide:
Note on the **Self-host Fixpoint** check (red): this is a pre-existing, dev-wide divergence from the released **v1.6.0** seed, not introduced by this PR.

- This change is entirely in `driver.mach`'s `for_union` paths and is a strict **no-op for `mach build .` of the single-artifact `mach` project itself** (`for_union` is false there), so it cannot alter how the compiler emits itself.
- Locally, the exact CI fixpoint (seed → stage1 → stage2) is **byte-identical** with the real v1.6.0 release tarball seed, on both this branch and the pristine base `0f49f626`.
- The latest merge to `dev`, **#1508**, shows the **same** `Self-host Fixpoint: fail` and was merged anyway — the codegen has drifted from the v1.6.0 seed (e.g. the post-release comptime/bignum work), which the next release will re-seed. The check's own error text calls this out ("if the codegen change is intentional … the next release re-seeds").
- All other checks pass: Full Build, Test Suite (526/526), Integration Tests, AArch64 (cross), Windows (native), scrub.

`src/lang/driver.mach` is unchanged between this PR's base and current `dev`, so the change merges cleanly.

octalide:
Rebased onto current `dev` (2.1.0 line, post-2.0.0 collapse). The only conflict was the diverged `register_export_names` doc comment (the `$sym.attr` setter → backtick-decorator rewrite); resolved by keeping dev's comment and inserting the two new union functions before it. The change is still confined to `src/lang/driver.mach` (+272/−15).

**Fixpoint now green.** Against the released **2.1.0-line seed (mach 2.0.1)**, the self-host fixpoint is byte-identical: seed → stage1 → stage2 → stage3 all `726ab287…`. The union code runs only under `for_union`, so `mach build .` of the single-artifact compiler (`for_union` false) is unaffected — its self-emission is unchanged, and single-artifact/selected builds resolve through the same `resolve_build_unit` path as before (the `eff_sel` copy is field-identical to the selection for every non-union build).

`mach test .` — **540 passed, 0 failed**, including the two #1505 driver-union tests (the `beta_only` module reachable only from the non-default artifact's entry, and the ambiguous-default single-build-errors / union-loads case).

Supersedes #1481 (already closed as a duplicate of #1505). Note: #1485 (mach `doc` over a multi-artifact project) wants the same union on the doc path — out of scope here, left as a follow-up.

PR #1508, opened 2026-06-19
fix(dep): find git.exe on windows (#1506)

### LINKED ISSUE #1506: dep: mach dep pull can't find git.exe on Windows (searches for 'git', misses the .exe)
`mach dep pull` fails on Windows with `error: git not found on PATH (needed to fetch git dependencies)` **even though git is clearly on PATH**. CI diagnostic on the windows-native runner:

```
where.exe git
  C:\Program Files\Git\bin\git.exe
  C:\Program Files\Git\cmd\git.exe
  C:\Program Files\Git\mingw64\bin\git.exe
git --version  ->  git version 2.54.0.windows.1
mach.exe dep pull  ->  error: git not found on PATH
```

So mach's own git lookup in the dep resolver searches PATH for a file named literally `git` and never finds `git.exe` on Windows (the classic missing-`.exe` bug). Fix: when locating the git executable, try `git.exe` (or honor `PATHEXT`) on Windows.

Surfaced by #1486 (submodule → `mach dep pull` migration). The **v1.6.0 seed carries this bug and can't be fixed retroactively**, so the #1486 CI works around it by realizing the std on the windows lane with a direct `git clone` (reading the pinned commit from mach.lock). Once a seed with this fix ships, that windows workaround should be removed.

**Target v1.7.0** (it's the seed we re-seed to for the whole v1.7 line; getting the fix in means the workaround can be dropped in v1.7.1). Windows-native target bug.

octalide:
The dep resolver looked up the bare name `git` on PATH, which never resolves on Windows (it's `git.exe`) — `mach dep pull` failed there even with git installed and on PATH (`where git` finds it in three places). Now looks for `git.exe` on Windows via the seed-safe `$if ($mach.target.os == $mach.os.windows)` form.

Self-host verified byte-identical (stage1==stage2==stage3); inert/additive change.

Closes #1506

### DIFF
--- a/src/cli/cmd/dep.mach
+++ b/src/cli/cmd/dep.mach
@@ -925,7 +925,11 @@ fun read_dep_manifest(alloc: *Allocator, dir: str) Result[toml.Table, str] {
 # error naming the operation that needed git.
 fun ensure_git(alloc: *Allocator, gt: *GitTool) Result[bool, str] {
     if (gt.ready) { ret ok[bool, str](true); }
-    val git_r: Result[str, str] = find_on_path(alloc, "git");
+    # windows executables carry a `.exe` extension; the bare `git` name does not
+    # resolve on PATH there (it's `git.exe`), so look for the right name (#1506).
+    var gname: *u8 = "git";
+    $if ($mach.target.os == $mach.os.windows) { gname = "git.exe"; }
+    val git_r: Result[str, str] = find_on_path(alloc, gname);
     if (is_err[str, str](git_r)) { ret err[bool, str]("git not found on PATH (needed to fetch git dependencies)"); }
     val env_r: Result[**u8, str] = build_env(alloc);
     if (is_err[**u8, str](env_r)) { ret err[bool, str]("failed to build child environment for git"); }


octalide:
Admin-merged over a red **Self-host Fixpoint** lane. The red is the *inherited* intentional float-fold divergence from #1483: CI builds this PR merged with dev, which now carries #1483's precise float folding, so the v1.6.0 seed-built stage1 differs from the self-built stage2. **This PR is itself inert** — self-host verified S1==S2==S3 on a local pre-#1483 build; the git.exe lookup doesn't affect emission. Every other lane is green. The v1.7.0 release re-seeds the fixpoint gate and clears this.

octalide:
Admin-merged over a red **Self-host Fixpoint** lane. The red is the *inherited* intentional float-fold divergence from #1483: CI builds this PR merged with dev, which now carries #1483's precise float folding, so the v1.6.0 seed-built stage1 differs from the self-built stage2. **This PR is itself inert** — self-host verified S1==S2==S3 on a local pre-#1483 build; the git.exe lookup doesn't affect emission. Every other lane is green. The v1.7.0 release re-seeds the fixpoint gate and clears this.

PR #1597, opened 2026-06-24
fix(lower): widen gep index to the machine word before emit_gep

### LINKED ISSUE #1596: Compiler miscompiles a sub-word (u8) array/pointer index: GEP index not widened to the machine word
## Summary
`lower_index_lvalue` (`src/lang/me/lower/expr.mach`) passes a runtime GEP index to `emit_gep` at its **source integer width**. The MIR back end scales the **full index register** (`be/codegen/mir/context.mach`: "GEP index scaling is pointer arithmetic over machine-word index registers"), so a sub-word index (e.g. a `u8`) reaches codegen with **undefined high register bits** and `lea (base, idx, scale)` produces a wild address.

The leading aggregate index (`const_int(0, index_type)`) is correctly 64-bit, but the actual index operand isn't widened. The comparison path zero-extends (`movzbl`); the index path does not — that inconsistency is the bug.

## Why it's been latent
Every in-tree index site casts the index (`x::u32` / `::usize`), which forces a clean register. The bug surfaces the moment an index is used at sub-word width without a cast (e.g. `ti.prims[kind]` with `kind: u8`): the `kind` register's garbage high bits index far out of bounds → SIGSEGV.

## Diagnosis
gdb on a self-host crash: fault in `allocate_raw` dispatching a null allocator, traced to `map.grow` on a zeroed map, traced to a wild GEP in the interner; the index register carried a valid low byte with garbage high bits. Restoring the `::u32` cast or widening the index in the lowering both resolve it.

## Fix
Widen the runtime index to the 64-bit machine-word index type (zext unsigned / sext signed) in `lower_index_lvalue` before `emit_gep`.

## Validation
Self-host byte-identical fixpoint (`-O2`, stage2 == stage3) and full test corpus (590/590) pass with the fix.

octalide:
Closes #1596.

`lower_index_lvalue` passed the runtime GEP index to `emit_gep` at its source integer width. The MIR back end scales the full index register, so a sub-word index (e.g. a `u8`) reached codegen with undefined high bits → wild address. The leading aggregate index was 64-bit but the real index operand wasn't widened; the comparison path zero-extends, the index path didn't.

### Change
Added `widen_index` and call it in `lower_index_lvalue`: a runtime index narrower than 64 bits is extended (zext unsigned / sext signed) to the machine-word index type before `emit_gep`; 64-bit indices are unchanged.

### Why latent
Every in-tree index site casts (`x::u32`/`::usize`), forcing a clean register. Surfaces on an uncast sub-word index (e.g. `ti.prims[kind]`, `kind: u8`).

### Validation
- Self-host **byte-identical fixpoint** at `-O2` (stage2 == stage3).
- Test corpus: **590/590** pass.
- `widen_index` runs on every runtime index (incl. the many `u32` indices in the compiler), so the byte-identical self-host directly exercises it.

### DIFF
--- a/src/lang/me/lower/expr.mach
+++ b/src/lang/me/lower/expr.mach
@@ -887,7 +887,14 @@ fun lower_index_lvalue(ctx: *lower.LowerContext, eid: aid.ExprId, e: *aexpr.Expr
 
     val ir2: Result[value.Value, str] = lower_rvalue(ctx, e.data.index.index);
     if (is_err[value.Value, str](ir2)) { ret ir2; }
-    val ix: value.Value = unwrap_ok[value.Value, str](ir2);
+    var ix: value.Value = unwrap_ok[value.Value, str](ir2);
+
+    # a gep index is machine-word arithmetic: the back end scales the full index
+    # register, so a sub-word index (e.g. a u8) must be widened to 64 bits first —
+    # otherwise its undefined high register bits corrupt the scaled offset.
+    val ixw: Result[value.Value, str] = widen_index(ctx, ix, lower.expr_type_of(ctx, e.data.index.index));
+    if (is_err[value.Value, str](ixw)) { ret ixw; }
+    ix = unwrap_ok[value.Value, str](ixw);
 
     if (is_ptr) {
         # `*T[ix]` strides over the pointee element type T. a raw untyped `ptr` has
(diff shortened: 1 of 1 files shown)


octalide:
The `Self-host Fixpoint` check is red **by design**: this PR intentionally changes codegen (the index zext), so the 2.5.8 seed-built `stage1` differs from the self-built `stage2`. The new compiler is self-consistent — local `stage2 == stage3` is byte-identical — and all other lanes (build, test corpus 590/590, integration, windows, aarch64) pass. The next release re-seeds. Follow-up for target-aware index normalization: #1598.

Issue #1178, opened 2026-06-07
target OS: macOS / Darwin (runtime, syscalls, entry)


octalide:
The std darwin runtime/syscall layer exists: dep/mach-std/src/runtime/darwin/x86_64.mach, system/os/darwin/shared.mach (~1109 lines over BSD syscalls). Needs: the Mach-O object format (issue #1176), dynamic libSystem linkage via #1159, and an entry-symbol fix (os/darwin.mach advertises _main but the runtime defines start — a one-line latent link failure). Apple Silicon (aarch64-darwin) additionally requires the aarch64 ISA (#1171) + AAPCS64 (#1174); x86_64-darwin (Intel) does not.

Issue #1159, opened 2026-06-07
epic: linker dynamic-import/bind infrastructure (+ LINK_SHARED)


octalide:
The linker is static-only — it hard-errors on any undefined `ext` symbol and has no import/bind tables, and `LINK_SHARED` is stubbed. This is the linchpin gating Windows (PE imports against kernel32), macOS (libSystem dynamic linkage), and shared-library linking. Work: route undefined/`ext` symbols into per-object-format import/bind tables, accept DLL/dylib lib-specs, implement `LINK_SHARED`. Touches `src/lang/be/linker.mach` and the object-format vtables under `src/lang/target/of/`. Prerequisite for the Windows and macOS target epics.

PR #1652, opened 2026-06-27
feat(target/of): Mach-O object format (writer, reader, executable)

### LINKED ISSUE #1176: target object format: Mach-O (writer, reader, executable)
src/lang/target/of/macho.mach is scaffolding — emit_object (Mach header, LC_SEGMENT_64 + sections, LC_SYMTAB nlist_64, LC_DYSYMTAB, X86_64_RELOC_*) and parse_object are stubs; Mach-O emit_exec + libSystem dynamic linkage (LC_LOAD_DYLINKER + LC_LOAD_DYLIB + dyld bind/rebase opcode streams — modern macOS rejects static syscall binaries) is unimplemented and depends on #1159. arm64 also needs LC_CODE_SIGNATURE. Needs the page-relative reloc kind from #1163. Required for any Darwin triple.

octalide:
Implements the Mach-O 64-bit object-format substrate (#1176), the last object
format needed to unlock the Darwin triples once the Darwin OS runtime (#1178)
lands. Pure substrate authoring: the only file changed is
`src/lang/target/of/macho.mach` (plus byte-verify fixtures and a CI lane) - no
shared-backend edits.

## What it does

- `emit_object`: a relocatable `MH_OBJECT` with one `LC_SEGMENT_64` holding every
  section, an `LC_SYMTAB`/`LC_DYSYMTAB` pair (symbols ordered local, then
  external-defined, then undefined), and per-section `relocation_info` arrays.
- `parse_object`: the mirror reader, round-tripping sections, symbols, and
  relocations back into an `of.ObjectImage`.
- `emit_exec`: a static `MH_EXECUTE` (leading `__PAGEZERO`, one `LC_SEGMENT_64`
  per laid-out load segment, `LC_UNIXTHREAD` carrying the entry point).
- Relocation mapping for both `x86_64` (`X86_64_RELOC_*`) and `arm64`
  (`ARM64_RELOC_*`), keyed on the ISA id forward and the header `cputype` in
  reverse - the same `tgt_isa.id` approach the COFF substrate uses, so no
  per-format hook is added to the shared `IsaVTable`.

## Notes

- Mach-O has no RELA addend field: addends live in-place in the patched bytes
  (x86_64, and the 8-byte `UNSIGNED` form on both arches), with pc-relative kinds
  folding the next-instruction `P + 4` bias exactly as COFF does. arm64's
  page/page-offset/branch kinds carry a non-zero addend in a preceding
  `ARM64_RELOC_ADDEND` entry.
- arm64's `PAGEOFF12` is scale-agnostic on the wire (one type for ADD and every
  LDR/STR width). The parser decodes the patched instruction to recover the exact
  `RK_ADD_LO12` / `RK_LDST*_LO12` kind the linker's `apply_reloc` needs, so the
  round-trip is lossless without touching the shared linker.
- `emit_dyn_exec` is left nil: dyld bind/rebase + libSystem dynamic linkage (and
  the runnable-image details: header inside `__TEXT`, `LC_MAIN`, code signature)
  are the Darwin-runtime follow-up (#1178), gated on #1159. This PR is the object
  format only.

## Verification

- Self-host fixpoint (`mach build` x3, `cmp`) and `mach test .` green.
- In-source byte tests round-trip both arches and assert the on-wire header, load
  commands, sections, nlist symbols, and relocations.
- A `test/macho/` cross-compile fixture byte-verifies the emitted Mach-O object
  and executable with the llvm tools (Darwin binaries cannot run on the Linux CI
  host, so structural byte-verification is the bar, exactly as the riscv64 lane).

Closes #1176

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -176,6 +176,45 @@ jobs:
       - name: cross-compile and byte-verify the rv64 fixture
         run: bash test/riscv64/verify.sh "$PWD/m"
 
+  # the Darwin triples are cross-compile-only here: the Darwin runtime (#1178) is
+  # a later slice, so the compiler cannot be built or self-tested for darwin yet.
+  # this lane builds a Mach-O-capable compiler from the PR source, cross-compiles
+  # the freestanding fixture for x86_64-darwin and aarch64-darwin, and byte-verifies
+  # the emitted Mach-O object and executable with the llvm tools. there is no
+  # execution step: Darwin binaries cannot run on the Linux host, and a Darwin exit
+  # needs the libSystem runtime the #1178 slice adds.
+  cross-darwin:
+    name: cross darwin (macho)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: install llvm tools
+        run: |
+          set -euo pipefail
+          sudo apt-get update
+          sudo apt-get install -y llvm
+
+      - name: fetch released mach
+        env:
+          GH_TOKEN: ${{ github.token }}
+          SEED: mach-*-x86_64-linux.tar.gz
+        run: |
+          set -euo pipefail
+          tmp="$(mktemp -d)"
+          gh release download -R "$GITHUB_REPOSITORY" -p "$SEED" -D "$tmp"
+          tar -xzf "$tmp"/$SEED -C /usr/local/bin mach
+          chmod +x /usr/local/bin/mach
+
+      - name: build macho-capable compiler from source
+        run: |
+          set -euo pipefail
+          mach dep pull
+          mach build . -o m
+
+      - name: cross-compile and byte-verify the macho fixture
+        run: bash test/macho/verify.sh "$PWD/m"
+
   build-windows:
     name: build ${{ matrix.target }}
     runs-on: ${{ matrix.runs-on }}
--- a/src/lang/target/of/macho.mach
+++ b/src/lang/target/of/macho.mach
@@ -1999,36 +1999,25 @@ fun ts_bad_input() u8 {
     ret 0;
 }
 
-# the Mach-O byte tests run on every host except windows. each `test` block is
-# built into its own executable by the `mach test .` driver, and the windows
-# driver is at its per-run memory ceiling - one more test executable fails the
-# link with "coff: unwind table alloc failed" (a pre-existing test-driver memory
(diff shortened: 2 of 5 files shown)


octalide:
### CI note: `test x86_64-windows` red is pre-existing (#1653), not this PR

Every lane is green except `test x86_64-windows`, which fails in the COFF exec
writer with `coff: unwind table alloc failed`. This is a pre-existing
test-driver memory cliff, filed as #1653, and is independent of this change:

- The in-source Mach-O byte tests are comptime-gated off the windows host
  (`$if ($mach.build.os != $mach.os.windows)`); I verified the gate genuinely
  excludes `test` blocks, so the windows test suite here is identical to the base.
- The impl-only commit (no macho tests) passed `test x86_64-windows`; the gated
  build (same windows suite) then failed twice — so the base windows suite is
  already at the runner's memory cliff, flaky/failing regardless of this PR.
- `build x86_64-windows` (one COFF exec link) passes; only the per-test-link
  driver run exhausts memory.

Green here: build x86_64-linux / aarch64-linux / x86_64-windows, test
x86_64-linux, test aarch64-linux (both run the full suite incl. the Mach-O byte
tests), release x86_64-linux, cross riscv64-linux, and the new
**cross darwin (macho)** lane which byte-verifies the emitted Mach-O object and
executable for both Darwin triples with the llvm tools.

The windows gate is temporary; remove it once #1653 lands.

octalide:
Holding this for merge pending #1653. The `test x86_64-windows` red is deterministic, not flaky: I re-ran the job on the same commit and it failed identically at `coff: unwind table alloc failed`. The suite grew past the windows test-driver memory ceiling via today's merges (#1650, #1651), and this PR's module tips it over.

CI runs only on pull_request against the live base, so merging this would red the windows lane for every subsequent PR, not leave a pre-existing condition. The comptime gate is a workaround for a real structural bug (the per-test memory leak in #1653), so the right order is: fix #1653, then merge this with the macho byte tests un-gated. The Mach-O substrate itself is verified green by the `cross darwin (macho)` lane and is not in question. Dispatching the #1653 fix now.

octalide:
Update: #1654 (the test-driver memory fix) is merged to dev. I rebased this branch onto current dev and removed the temporary windows gate, so the two Mach-O in-source byte tests now run on every host. All CI lanes are green, including `test x86_64-windows` (pass). The earlier "holding for #1653" notes are resolved.

Issue #1600, opened 2026-06-25
Epic: retargeting architecture - make targets arbitrary, drop-in plugins


octalide:
> ## Status: foundation delivered (v2.6.0, 2026-06-26)
>
> The retargeting foundation landed across #1601-#1615 and ships in v2.6.0. The north
> star - research a platform, build a target, drop it in, no core edits - is met for the
> register-machine class. A new register-machine target now plugs in as substrate modules
> plus vtable hooks with no shared-backend edits (see #1172 riscv64, the first such
> target).
>
> What landed: prim_desc (#1601), index/address width off the pointer width (#1603), dead
> syscall_layer removed (#1604), x64 abi layering (#1605), the MachineModel record
> (#1606), regalloc and mir read the model (#1607, #1608), name-based composition with
> name-keyed registries (#1609), the isa declares its ELF e_machine (#1610), setup_registry
> bring-up (#1611), the isa owns its relocation packers (#1612), raw flat-image output plus
> a freestanding os (#1613), and object-format selection by name with a freestanding e2e
> (#1615).
>
> Achieved differently than the locked design said: the shared stages became model-driven
> rather than swapped through a per-stage `StageVTable`. For every register-machine target
> the model-driven stages suffice, so the StageVTable seam is deferred until a target that
> cannot run them exists (see Deferred below).
>
> ### Handed off
> - Concrete targets (the actual platforms): #1180 multi-target matrix. Next is #1172
>   riscv64.
> - Parallel and incremental compilation (themes 4 and 5, a separate axis, never target
>   authorship): #1164 incremental epic. The back-half binding is #1618; content-addressed
>   identity (#1579) already landed.
> - Raw flat-image build-pipeline gap (raw cannot round-trip through `.o`): #1616.
>
> ### Deferred (gated on a non-register-machine target, e.g. 6502 / embedded)
> - Per-stage `StageVTable` seam: only buys stage replacement, which a register machine
>   never needs. Premature until a target cannot run the model-driven stages.
> - Model-parametric MIR: MIR still assumes a register file and a downward stack frame.
>   This is the deep rework, and it pays off only for a target the current MIR cannot
>   express. Gate it on actually committing to such a target.
> - Composable type spelling (`u<N>`/`f<N>` end to end): declined. prim_desc (#1601) makes
>   adding a concrete primitive cheap, and arbitrary-width spelling carries real
>   legalization cost (masking, soft-float) for little demand.
>
[...]

Issue #1179, opened 2026-06-07
target OS: freestanding / no-OS (embedded / bare-metal)


octalide:
A target with NO operating system — no syscalls, no OS runtime. The $mach.os.freestanding value already exists. Needs: a freestanding runtime (custom entry/_start, no syscall layer, no libc), a suitable output (raw flat binary and/or ELF without a dynamic loader), and a memory-layout/linker-script surface for placing code/data at fixed addresses. Decouples the compiler from any OS so Mach can target microcontrollers / bare-metal. Composes with any ISA (x86_64 ✓, aarch64, riscv64).

octalide:
Closing — every capability this asked for has shipped, verified against the current tree rather than assumed:

- **Freestanding runtime** (custom entry/`_start`, no syscall layer, no libc): shipped with the `freestanding` OS and the `#[naked]` entry model; exercised end to end by the BareMetal work and its int fixtures.
- **Raw flat-binary output**: `of = "raw"` shipped in v2.6.0; #1616 (the object-centric path for it) is CLOSED, and #2402 fixed the flat image's zero-fill, #2409 its entry placement.
- **Memory layout / fixed-address placement**: the `base` load-address override on `[target.*]` shipped in 4.4.0 (#2426/#2427), which is how a bare-metal image is placed without a linker script.
- **Composes with any ISA**: x86_64 and aarch64 are exercised; aarch64 gained system-register access (#2480) and both gained raw-encoding directives (#2479). riscv64's remaining gap is inline-asm CSR access, tracked separately as #2481 — an asm-surface gap, not a missing freestanding capability.

The bare-metal direction now proceeds through the platform-tag model (#2426) and the milestone rather than this epic.

PR #1661, opened 2026-06-27
fix(target): raw flat-image build path via produces_image predicate

### LINKED ISSUE #1616: target: raw object format has no path through the object-centric build pipeline
The `raw` object format (`of/raw`, #1613) implements only `emit_exec`. It deliberately
stubs `emit_object` and `parse_object`, because a flat image has no relocatable
container. But `mach build` is object-file-centric: it runs codegen, calls
`tgt.of.emit_object` per module, then `tgt.of.parse_object` on every input at link time,
and only then `emit_exec`. So a `raw` target fails before the linker with
`raw output supports only executables` and never reaches `emit_exec`.

Confirmed by #1615's freestanding e2e. The identical minimal program built as
`os=freestanding, of=elf` builds end to end and produces a working image whose entry
resolves to the freestanding base (`0x0`), which satisfies raw's entry-at-base contract.
The blocker is purely raw's missing relocatable round-trip, not the entry contract and
not freestanding.

## Fix (a build/link pipeline change, not a raw-writer patch)

Two candidate designs:

- Link `produces_image` formats straight from the in-memory codegen output, bypassing
  the per-module `.o` emit/parse round-trip entirely.
- Give raw a minimal in-memory object representation so it round-trips like elf.

The first is the cleaner fit for the substrate model: a format declares whether it needs
a relocatable round-trip, and the driver dispatches on that instead of assuming every
target emits `.o` files. This is the `needs_link` / `produces_image` predicate the
retargeting design (#1600) called for but did not land.

The of-by-name selection and the raw writer itself are done (#1613, #1615). This is the
last gap to a real `os=freestanding, of=raw` flat-image build (#1179).

### LINKED ISSUE #1617: diag: unknown-os diagnostic omits freestanding
`select_target` reports an unknown target os with `(expected: linux, darwin, windows)`
(`src/lang/driver.mach:1708`), but `os_id_for` has accepted `freestanding` since #1613.
The diagnostic should list it: `(expected: linux, darwin, windows, freestanding)`.

Cosmetic, one line. Surfaced by #1615's review.

octalide:
Closes #1616
Closes #1617

The `raw` flat-image object format (`of/raw`, #1613) implements only `emit_exec`
and stubs `emit_object`/`parse_object` (a flat image has no relocatable
container). But `mach build` was object-file-centric: it codegens each module,
writes a `.o` per module (`emit_object`), then the linker re-parses every input
(`parse_object`) before laying out the executable. So a raw target failed before
the linker with `raw output supports only executables` and never reached
`emit_exec`. This was the last gap to a real `os=freestanding, of=raw`
flat-image build (#1179).

## Fix: a build/link pipeline change, dispatched on a vtable predicate

A format now declares whether it round-trips through relocatable `.o` files. The
driver dispatches on that instead of assuming every target emits and re-parses
objects.

- **`of.OfVTable.produces_image`** (new): `true` for a container-free flat image
  (raw), `false` for a relocatable-object format that round-trips through `.o`
  files (elf, coff, macho). This is the predicate the retargeting design (#1600)
  called for but did not land.
- **`linker.link_images`** (new): links an already-built `of.ObjectImage` array
  straight from memory. `link` (disk inputs) and `link_images` (in-memory codegen
  images) now share a `link_modules` core that never owns the modules - the
  caller frees them. The merge, symbol resolution, relocation patching, and
  segment layout are identical, so the existing elf/coff/macho output is
  byte-unchanged.
- **`cli/build` branch**: when `produces_image`, the build links the in-memory
  codegen images directly (`link_image_artifact` -> `linker.link_images`),
  bypassing the per-module `.o` round-trip entirely. The relocatable formats keep
  writing and re-parsing per-module objects unchanged. A flat image has only an
  executable form, so library mode and `--emit obj` are rejected for it.

No defensive fallback for the raw-can't-round-trip case: the driver dispatches on
the predicate so the flat-image path never reaches `emit_object`/`parse_object`.

## #1617 (same area, trivial)

`os_id_for` has accepted `freestanding` since #1613, but the unknown-os
diagnostic still read `(expected: linux, darwin, windows)`. Now lists
`freestanding`.

## Verification

A new `test/freestanding` fixture (a minimal `os=freestanding, of=raw` x86_64
program: a counted loop summing 0..9, then a raw `exit` syscall) builds end to
end to a flat image. `verify.sh` (modeled on `test/macho` / `test/riscv64`):
asserts no `.o` was written (round-trip bypassed), confirms the output is a
container-free flat binary (no ELF header), then mmaps the position-independent
image into an executable page and runs it, asserting exit code 45. A CI lane
builds a compiler from PR source and runs it.

Built with the modified compiler:
- `mach build .` on the fixture prints `link 1 image(s)` and produces the flat
  image with zero `.o` files.
- The 176-byte flat image disassembles to the expected prologue/loop/`exit`
  syscall, entered at offset 0 (the image base), and runs to exit code 45.
- The unknown-os diagnostic now reads
  `(expected: linux, darwin, windows, freestanding)`.

Self-host fixpoint (`a->b->c`, `cmp b c`) and `mach test .` results are added as
a comment once the rebuild completes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -215,6 +215,38 @@ jobs:
       - name: cross-compile and byte-verify the macho fixture
         run: bash test/macho/verify.sh "$PWD/m"
 
+  # the freestanding/raw flat-image build path (#1616): builds a compiler from the
+  # PR source, builds the minimal os=freestanding, of=raw x86_64 fixture (which
+  # links straight from the in-memory codegen images, no per-module .o), confirms
+  # the output is a container-free flat binary, then mmaps the position-independent
+  # image and runs it to assert its computed exit code. needs only a C compiler for
+  # the loader harness - no llvm, no qemu.
+  cross-freestanding:
+    name: cross freestanding (raw)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: fetch released mach
+        env:
+          GH_TOKEN: ${{ github.token }}
+          SEED: mach-*-x86_64-linux.tar.gz
+        run: |
+          set -euo pipefail
+          tmp="$(mktemp -d)"
+          gh release download -R "$GITHUB_REPOSITORY" -p "$SEED" -D "$tmp"
+          tar -xzf "$tmp"/$SEED -C /usr/local/bin mach
+          chmod +x /usr/local/bin/mach
+
+      - name: build a compiler from source
+        run: |
+          set -euo pipefail
+          mach dep pull
+          mach build . -o m
+
+      - name: build and run the freestanding raw flat image
+        run: bash test/freestanding/verify.sh "$PWD/m"
+
   build-windows:
     name: build ${{ matrix.target }}
     runs-on: ${{ matrix.runs-on }}
--- a/src/cli/cmd/build.mach
+++ b/src/cli/cmd/build.mach
@@ -1281,6 +1281,14 @@ fun link_artifact(p: *driver.Project, emit_obj: bool, verbose: bool, root: *u8,
     val te: *driver.TargetEntry = p.target_entry;
     if (te == nil) { print.eprint("error: no target selected\n"); ret 2; }
 
+    # a flat-image object format (`produces_image`) has no relocatable `.o`
+    # container: it cannot emit per-module objects or re-parse them at link time.
+    # link the in-memory codegen images straight into the flat executable instead
+    # of round-tripping through `.o` files.
+    if (p.target.of != nil && p.target.of.produces_image) {
+        ret link_image_artifact(p, emit_obj, verbose, root, out_override, images, out_path);
+    }
+
     # the per-module object tree root: `<root>/<resolved obj template>`.
     var obj_base: [4096]u8;
     if (!resolve_obj_base(p, root, ?obj_base[0], 4096)) {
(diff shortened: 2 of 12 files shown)


octalide:
Verification (built with the released v2.7.0 seed against this PR source):

- **Self-host fixpoint**: `a = mach build .` -> `b = ./a build .` -> `c = ./b build .`, `cmp b c` -> identical. Fixpoint converges.
- **Object-centric path byte-unchanged**: `cmp a b` -> identical. The base compiler (released seed) and the PR compiler emit byte-identical ELF for the same source, so the elf/coff/macho round-trip path is unchanged.
- **`mach test .`** (run with the fixpoint compiler `c`): 657 passed, 0 failed. Includes the raw vtable / emit_exec tests and the #1615 `select_of:freestanding_raw_emits_flat_image` test.
- **Freestanding raw e2e** (`bash test/freestanding/verify.sh`):
  - `mach build .` prints `link 1 image(s)` and produces the flat image with zero `.o` files written (the per-module round-trip is bypassed).
  - The 176-byte output carries no ELF header (a container-free flat binary), is entered at offset 0 (the image base), and its disassembly is the expected prologue + counted loop + `mov rax,0x3c` (`exit`) + `syscall`.
  - mmapped into an executable page and run, it exits with code 45 (sum 0..9), proving the emitted machine code is correct end to end.
- **#1617**: the unknown-os diagnostic now reads `(expected: linux, darwin, windows, freestanding)`.

PR #1676, opened 2026-06-28
chore: repoint relocated repos to the briar-systems org

octalide:
Repoints all references to the repos that moved from `octalide` to `briar-systems` (mach, mach-std, mach-boot) across docs, badges, install scripts, the dependency manifest/lock, and source. Stay-behind repos (mach-sieve) and the octalide funding/copyright identity are left untouched. `mach dep pull`/`dep update --all` confirmed the lock resolves to the same commit on the relocated mach-std.

Authorized org-relocation hotfix; ref-only diff, merging to both main and dev.


octalide:
Automated org-relocation hotfix. Diff is ref-only (octalide -> briar-systems for mach/mach-std/mach-boot; stay-behind repos and the octalide funding/copyright identity untouched) and all CI checks pass.

Could not auto-merge with `--admin`: the **main** ruleset requires status-check contexts named exactly `Test Suite` and `Full Build`, which no CI job reports (actual jobs are `build <target>` / `test <target>`), so both sit permanently as 'expected'; plus 1 approving review is required and admin bypass is not honored for this actor. Needs a reviewer with ruleset bypass (or a fix to the required-check context names) to merge. Companion dev PR: #1677.

PR #1677, opened 2026-06-28
chore: repoint relocated repos to the briar-systems org (dev)

octalide:
Same ref-only repoint as #1676, targeted at `dev` so the org relocation isn't lost on the next release sweep. Moves all references to mach, mach-std, and mach-boot from `octalide` to `briar-systems`; leaves stay-behind repos (mach-sieve) and the octalide funding/copyright identity untouched.

Companion to #1676 (-> main).


octalide:
Companion to #1676, targeting `dev` so the relocation survives the next release sweep. Ref-only diff. Could not auto-merge with `--admin`: the `dev` ruleset requires 1 approving review and admin bypass is not honored for this actor (author cannot self-approve). Needs a reviewer to merge.

Issue #1722, opened 2026-06-29
codegen/link(macho): PIE Mach-O executable writer + rebase stream (Apple Silicon arm64 exec)


octalide:
Part of #1718. Blocks the aarch64 leg of #1680.

## What

macOS arm64 (Apple Silicon) refuses to exec a non-PIE main executable — proven on
macos-14 in #1718 (the linker even forces PIE: `-no_pie ignored for arm64`). mach
emits a non-PIE fixed-absolute image (`__TEXT` at `0xfffff000`, empty rebase stream),
so its arm64 darwin binaries are `Killed: 9` at exec. This issue makes mach emit a
genuine **PIE** Mach-O executable so the darwin self-host runs on Apple Silicon.

The investigation (#1718) already prototyped the cheap parts and ruled out the
expensive alternatives — keep raw `svc` (no libSystem syscall routing), classic
`LC_DYLD_INFO` rebase (no chained fixups), zero dylib dependency. The prototypes live
on branch `invel/1718-scope` (env-gated `LC_MAIN`/`LC_BUILD_VERSION`/`LC_UUID`/`MH_PIE`
toggles in `emit_dyn_exec`); use them as the starting point.

Scope:
- **Writer (`src/lang/target/of/macho.mach`):** emit a position-independent image —
  `MH_PIE` header flag, a relocatable layout (PIE base, not the current fixed
  `0xfffff000`), an `LC_MAIN` entry (replacing `LC_UNIXTHREAD` for the darwin main
  executable), `LC_BUILD_VERSION`, `LC_UUID`, and an `LC_DYLD_INFO(_ONLY)` carrying a
  **rebase stream** — classic rebase opcodes (`REBASE_OPCODE_*`) covering every
  in-image absolute pointer, so dyld slides them by the load bias. No imports ⇒ no
  dylibs / stubs / GOT / bind stream. Keep the ad-hoc `LC_CODE_SIGNATURE` (#1679) and
  the header-in-`__TEXT` layout (#1717).
- **Linker (`src/lang/be/linker.mach`):** collect the in-image absolute relocations
  (global pointers, vtable function pointers, absolute data relocs) into rebase
  records for the writer. arm64 TEXT is already PC-relative; the rebase work is in
  absolute DATA pointers. This is the dominant cost of the issue.
- Route the compiler's darwin self-host through this PIE writer (it has no dynamic
  imports, so the dynamic path collapses to "PIE + rebase, zero dylibs").
[...]

Issue #1717, opened 2026-06-28
codegen/link(macho): static executables map the mach header outside __TEXT → SIGKILL on Apple Silicon


octalide:
Blocks #1680 (aarch64 leg). Completes the real, on-Mac acceptance of #1679.

The static Mach-O executable writer (`emit_exec` in `src/lang/target/of/macho.mach`)
places the mach header + load commands at file offset `[0, hdr_span)` **outside any
segment** — only `__PAGEZERO` (filesize 0) precedes it, and `__TEXT` starts at fileoff
4096. Standard macOS executables — and mach's own `emit_dyn_exec` (macho.mach:~2126:
`__TEXT` fileoff 0, filesize `hdr_span + segs[0].filesz`, `text_va = base - hdr_span`)
— map `__TEXT` from file offset 0 so it contains the header.

The compiler's darwin binary is a **static** `LC_UNIXTHREAD` executable (mach-std's
darwin runtime issues raw `svc 0x80` syscalls; it does not link libSystem), so it goes
through `emit_exec`, not `emit_dyn_exec`.

## Repro

Cross-build the compiler for `aarch64-darwin`, run it on Apple Silicon:
`./mach-darwin-seed dep pull` → `Killed: 9` (SIGKILL, exit 137) at first exec. The
Apple Silicon kernel rejects the image even though the ad-hoc signature is
structurally valid off-Mac (all 1513 SHA-256 page hashes match, `codeLimit` == sig
offset, `execSegBase`/`execSegLimit` cover `__TEXT`, `CS_EXECSEG_MAIN_BINARY` set).
Reproduced on a `macos-14` runner via #1680's `workflow_dispatch` — the FIRST time any
mach-built arm64 darwin binary was actually executed on a Mac. `x86_64-darwin` runs
fine (x86_64 does not enforce signatures, so it tolerates the non-standard layout).

This is #1679's acceptance criterion ("arm64-darwin executables exec on Apple Silicon,
no `Killed: 9`") failing for real — it was only ever structurally validated before,
because the `cross-darwin` CI lane never executes anything on a Mac.

## Expected

A mach-built static arm64 darwin executable execs on Apple Silicon (no `Killed: 9`)
and self-hosts; the native `a→b→c` fixpoint converges on a `macos-14` runner.

## Fix

Make `emit_exec` map the mach header into `__TEXT` at file offset 0, mirroring
[...]

Issue #1679, opened 2026-06-28
codegen/link: emit ad-hoc Mach-O code signature so Darwin executables run on Apple Silicon


octalide:
Blocks #1680.

## What

The Mach-O executable writer in `src/lang/target/of/macho.mach` emits structurally
valid static (`LC_UNIXTHREAD`) and dynamic (`LC_LOAD_DYLINKER` / `LC_LOAD_DYLIB` +
`LC_DYLD_INFO_ONLY`) Darwin executables, but emits **no** `LC_CODE_SIGNATURE`. The
writer documents this itself (`macho.mach:1666-1668`):

> Apple Silicon additionally requires a code signature the kernel verifies before
> exec; this writer emits no `LC_CODE_SIGNATURE`, so its output is byte-verifiable
> but not runnable on arm64 without an external ad-hoc signature.

On `x86_64-darwin` an unsigned binary runs. On `aarch64-darwin` the kernel `SIGKILL`s
any unsigned executable (`Killed: 9`). So a mach-built `aarch64-darwin` binary cannot
execute — which means it cannot rebuild itself, so the native self-host fixpoint
(`a → b → c`, `b == c`) is impossible on Apple Silicon. That blocks an
`aarch64-darwin` release entirely.

The fix is to emit an **ad-hoc** code signature for the executable writers:

- Append an `LC_CODE_SIGNATURE` load command whose `dataoff`/`datasize` point into
  `__LINKEDIT` at an embedded signature **SuperBlob** (`CSMAGIC_EMBEDDED_SIGNATURE`
  `0xFADE0CC0`) containing a single **CodeDirectory** blob
  (`CSMAGIC_CODEDIRECTORY` `0xFADE0C02`).
- CodeDirectory fields: a supported `version`, the `adhoc` flag
  (`0x0002`) in `flags`, an `identifier` derived from the output name,
  `hashType = 2` (SHA-256), `hashSize = 32`, `pageSize = 12` (log2 of 4096),
  `codeLimit` = the file offset of the signature, and the per-page SHA-256 hashes of
  the file bytes `[0, codeLimit)`. No special slots, no CMS blob (that is what makes
  it ad-hoc).
- Hashing uses the existing `std.crypto.hash.sha256` already vendored in `mach-std`
  (`init`/`update`/`final`, or the one-shot `hash`, `DIGEST_SIZE = 32`) — no new
  crypto primitive is needed.

Signing **must be deterministic**: no timestamps, no nonces. Identical input pages
[...]

Issue #325, opened 2026-02-17
Replace physical register references in lowering with ABI-provided values


octalide:
## Parent
Child of #316

Depends on: dead frame code removal (Sub-Issue D)

## Problem
Lowering references `ir.RAX`, `ir.RBP`, `ir.RSP`, `ir.RDI`, etc. directly. These should come from the ABI spec.

### Extend ABI record (`abi/spec.mach`)
Add fields:
- `stack_reg: i32` — stack pointer register (RSP=4 for sysv64)
- `frame_reg: i32` — frame pointer register (RBP=5 for sysv64)
- `syscall_regs: *i32` — syscall argument registers
- `syscall_reg_count: i32` — number of syscall arg registers
- `syscall_num_reg: i32` — register for syscall number (RAX=0 for sysv64)

### Populate in sysv64.mach
`stack_reg=RSP(4)`, `frame_reg=RBP(5)`, `syscall_regs=[RDI,RSI,RDX,R10,R8,R9]`, `syscall_num_reg=RAX(0)`

### Rewrite call.mach
- `ir.RSP` → `ctx.abi.stack_reg`
- `ir.RAX` return value → `ctx.abi.ret_reg` (already on ABI!)
- Syscall regs → `ctx.abi.syscall_regs[i]`
- Syscall num RAX → `ctx.abi.syscall_num_reg`

### Rewrite lower.mach
- `ir.RBP` in `frame_mem()` → `ctx.abi.frame_reg`

### Move physical register constants out of ir.mach
Delete `RAX` through `R15` (L152-168). They remain in `isa/x86_64/` and `abi/sysv64.mach`.
Keep `VREG_START = 1024`.

## Verification
`make` — all 4 stages pass. Critical: wrong register number silently generates broken code. 4-stage bootstrap catches this.

PR #1723, opened 2026-06-29
feat(macho): PIE arm64 darwin executables (MH_PIE + LC_MAIN + rebase stream)

### LINKED ISSUE #1718: epic: dynamic libSystem-linked darwin executables (Apple Silicon rejects static images)
> ## Status (2026-06-28)
> Scoped (see the investigation findings comment). The exec gate on Apple Silicon is
> **PIE**, proven on macos-14 — not signing, load commands, or libSystem. Awaiting a
> direction call: commit to the PIE work / x86_64-interim / shelve. #1680 aarch64 leg
> and #1716 are blocked on this. Not started.

## What

macOS on arm64 (Apple Silicon) refuses to exec a **non-PIE** main executable. Proven
on a macos-14 runner (#1680 / scope investigation): EVERY mach variant — static,
dynamic+libSystem+LC_UNIXTHREAD+raw-svc, +LC_MAIN, +LC_BUILD_VERSION, +LC_UUID,
+MH_PIE flag, +libSystem call — is `Killed: 9`, both mach-signed and after Apple
`codesign -s -`. System clang on the same runner: `clang -nostdlib` raw-`svc` with
zero dylibs runs (→42); `clang -no_pie` emits `ld: warning: -no_pie ignored for
arm64` — the linker forces PIE. mach emits a non-PIE fixed-absolute image (`__TEXT` at
`0xfffff000`, empty rebase stream), so the kernel kills it.

What this rules OUT (shrinks the epic): raw `svc` works → no libSystem syscall
routing; classic `LC_DYLD_INFO` works → no chained-fixups; a dyld-linked image with
zero dylibs runs → no libSystem dependency. The **only** missing hard requirement is
genuine PIE.

Goal: emit a position-independent darwin executable so arm64 binaries exec on Apple
Silicon and the compiler self-hosts there.

## Sized scope (from the investigation)

- **mach (codegen/link) — LARGE.** Emit a real PIE Mach-O: `MH_PIE` + a relocatable
  base/layout, `LC_MAIN` / `LC_BUILD_VERSION` / `LC_UUID` (all prototyped on the scope
  branch), and the dominant cost — a **rebase stream** (classic `LC_DYLD_INFO` rebase
  opcodes) for every in-image absolute pointer (globals, vtable fn-pointers, absolute
  data relocs). arm64 TEXT is already PC-relative; the work is in absolute DATA
  pointers and the linker collecting them as rebase records. Route the darwin
  self-host through this writer (no imports ⇒ zero dylibs/stubs/got/bind).
- **mach-std (os:darwin) — SMALL.** Keep raw `svc`; only change the darwin arm64 entry
  from the `LC_UNIXTHREAD` stack convention (`[sp]=argc`) to the `LC_MAIN` register
  convention (`x0=argc, x1=argv, x2=envp`).

## Children

- [ ] #1722 — mach: PIE Mach-O writer + rebase stream — MH_PIE, relocatable layout,
  LC_MAIN/LC_BUILD_VERSION/LC_UUID, and the rebase-stream emission for absolute
  pointers (linker collects absolute relocs → rebase records). Route the darwin
  self-host through it. (LARGE)
- [ ] briar-systems/mach-std#324 — mach-std: darwin arm64 entry LC_UNIXTHREAD →
  LC_MAIN register convention (keep raw svc). (SMALL)
- [ ] #1680 (aarch64 leg) — re-enable + validate native fixpoint + exec on macos-14
  once #1722 + mach-std#324 land.

## Sequencing

mach PIE writer + mach-std LC_MAIN entry, validated together on a macos-14 runner.
Then #1680 aarch64. The #1717 static-writer header fix already landed (correct for
static x86_64-darwin + freestanding); x86_64-darwin tolerates non-PIE/static and is
unaffected by this epic.

## Future follow-ons (enabled, NOT in scope here)

The `BaseReloc` set #1722 adds to the linker is format-neutral, so once it's proven on
macOS, two OPTIONAL ASLR-hardening follow-ons become small encoder issues consuming the
same set (to be filed against the proven interface, parallelizable with each other):
- **ELF PIE (linux ASLR):** `ET_DYN` + `R_*_RELATIVE` in `.rela.dyn`. Caveat — mach's
  no-libc static linux binaries also need a self-relocation startup (apply RELATIVE
  relocs before `main`, since there's no `ld.so`), so this is more than the encoder.
- **PE/COFF ASLR (windows):** `.reloc` base-relocation blocks + `DYNAMIC_BASE`. Closer
  to pure-encoder.

Neither is forced — linux/windows run fixed-address fine — so both are deferred until
ASLR hardening is wanted. Captured here so the generalization isn't lost.

octalide:
Closes #1718's investigation child via #1722 — makes mach emit position-independent arm64 Darwin executables so they exec on Apple Silicon. Pairs with mach-std #325 (the arm64 `LC_MAIN` entry).

## Why
#1718 proved on `macos-14` that the kernel refuses a non-PIE arm64 main executable (`-no_pie ignored for arm64`); mach emitted a non-PIE fixed-address image, so its arm64 binaries were `Killed: 9`. The investigation ruled out the expensive alternatives: raw `svc` works, classic `LC_DYLD_INFO` rebase is accepted (no chained fixups), and no libSystem dependency is needed. So the fix is "emit a genuine PIE image".

## What
- **Linker (format-neutral):** a new `of.BaseReloc` set on `of.DynamicInfo` — the in-image absolute pointers a position-independent image must slide (the same concept ELF `.rela.dyn` / PE `.reloc` encode). The linker collects each in-image `RK_ABS64` pointer relocation as a base relocation when the target requires PIE. The per-arch PIE policy lives on the OS vtable (`darwin`: arm64 yes, x86_64 no), so the linker never branches on ids; a PIE target routes through the dynamic writer even with zero dylibs and reserves a one-page header gap so `__TEXT` maps at the OS base. The `of.WriterFn` boundary is unchanged — each format serializes the set its own way (only Mach-O here; ELF/PE are future ASLR work).
- **Mach-O writer:** when `dyn.pie`, emit `MH_PIE`, the mach header at `0x100000000` (`__PAGEZERO = [0, base)`), an `LC_MAIN` file-offset entry, `LC_BUILD_VERSION`(macOS) + a deterministic name-based `LC_UUID`, and the `LC_DYLD_INFO` rebase stream (classic `REBASE_OPCODE_*`, sorted for a byte-identical fixpoint). The ad-hoc signature (#1679) and header-in-`__TEXT` (#1717) are preserved. Non-PIE (x86_64 darwin) output is byte-for-byte unchanged.
- **mach-std #325:** the darwin arm64 `_start` reads `argc/argv/envp` from `x0/x1/x2` (the `LC_MAIN` register ABI); raw `svc` retained.

## Validation
- **Off-Mac (this PR's CI):** `test/macho/verify.sh` byte-verifies the new arm64 PIE shape (`MH_PIE`, `LC_MAIN`, `LC_BUILD_VERSION`, `LC_UUID`, no `LC_UNIXTHREAD`) for the import-less and dynamic fixtures, and a new rebase fixture asserts `llvm-objdump --rebase` resolves the `__DATA` pointer slots. x86_64 unchanged. Linux self-host fixpoint stays byte-identical (`b == c`); all 686 in-source tests pass.
- **On-Mac (`macos-14`):** the exec proof (no `Killed: 9`, native `a→b→c` fixpoint, `mach test .`) is run via a dispatch and reported on the issue — see #1718.

## Merge coupling
Depends on mach-std #325 (the `LC_MAIN` entry). `mach.toml`/`mach.lock` stay on `branch/main` here (clean; CI's freestanding darwin fixtures don't need the runtime); **before this lands, bump `mach.lock` to the merged mach-std commit** carrying #324.

### DIFF
--- a/mach.lock
+++ b/mach.lock
@@ -3,4 +3,4 @@
 [deps.mach-std]
 url = "https://github.com/briar-systems/mach-std"
 ref = "branch/main"
-commit = "2a765f79a94282274d02111e4ad5d41930b36e2b"
+commit = "0262511fd60e21efb6aa95a66200b534940ea2f7"
--- a/src/lang/be/linker.mach
+++ b/src/lang/be/linker.mach
@@ -2379,12 +2379,14 @@ fun target_requires_pie(tgt: *target.Target) bool {
     ret tgt.os.pie_exec(tgt.arch.id);
 }
 
-# the page granularity used to align segment virtual addresses,
-# read from the OS vtable (`tgt.os.page_size`).
+# the page granularity used to align segment virtual addresses, read from the OS
+# vtable: the per-architecture `page_size_for` policy when present (Darwin maps 16
+# KiB pages on arm64, 4 KiB on x86_64), otherwise the fixed `page_size`.
 # ---
-# tgt: active target - supplies the OS page size
+# tgt: active target - supplies the OS page size and architecture
 # ret: the page granularity for segment alignment
 fun os_page_size(tgt: *target.Target) u64 {
+    if (tgt.os.page_size_for != nil) { ret tgt.os.page_size_for(tgt.arch.id); }
     ret tgt.os.page_size;
 }
 
--- a/src/lang/be/linker.mach
+++ b/src/lang/be/linker.mach
@@ -1575,9 +1575,22 @@ fun build_dynamic_info(s: *session.Session, tgt: *target.Target, dyn: *DynState,
         dyn.info.interp = R.unwrap_ok[intern.StrId, str](dr);
     }
 
-    # one DynLib per soname, in input order.
-    if (soname_count > 0) {
-        val lr: R.Result[*of.DynLib, str] = A.zallocate[of.DynLib](alloc, soname_count::usize);
+    # a PIE image implicitly links the OS platform library (darwin libSystem): a
+    # dyld-loaded executable that names no LC_LOAD_DYLIB is rejected, so it is
+    # recorded as a trailing dependency after the manifest sonames.
+    var implicit: u32 = 0;
+    var implicit_soname: intern.StrId = intern.STR_NIL;
+    if (dyn.info.pie && !str_empty(tgt.os.platform_lib)) {
+        val pr: R.Result[intern.StrId, str] = intern.intern(?s.interner, tgt.os.platform_lib);
+        if (R.is_err[intern.StrId, str](pr)) { ret R.err[bool, str](R.unwrap_err[intern.StrId, str](pr)); }
+        implicit_soname = R.unwrap_ok[intern.StrId, str](pr);
+        implicit = 1;
+    }
+
+    # one DynLib per soname, in input order, then the implicit platform library.
+    val total_libs: u32 = soname_count + implicit;
+    if (total_libs > 0) {
+        val lr: R.Result[*of.DynLib, str] = A.zallocate[of.DynLib](alloc, total_libs::usize);
         if (R.is_err[*of.DynLib, str](lr)) { ret R.err[bool, str](R.unwrap_err[*of.DynLib, str](lr)); }
         dyn.info.libs = R.unwrap_ok[*of.DynLib, str](lr);
         var i: u32 = 0;
@@ -1585,7 +1598,8 @@ fun build_dynamic_info(s: *session.Session, tgt: *target.Target, dyn: *DynState,
             dyn.info.libs[i].soname = sonames[i];
             i = i + 1;
         }
-        dyn.info.lib_count = soname_count;
+        if (implicit == 1) { dyn.info.libs[soname_count].soname = implicit_soname; }
+        dyn.info.lib_count = total_libs;
     }
 
     dyn.import_map = map.init[intern.StrId, u32](alloc, map.hash_u32, map.eq_u32);
--- a/src/lang/be/linker.mach
+++ b/src/lang/be/linker.mach
@@ -114,18 +114,20 @@ rec SymbolLoc {
 # plan's import array. empty for a static link; the writer consumes `info` + the
 # fixups. owns the `info.libs` / `info.imports` arrays and the fixup array.
 # ---
-# info:        the dynamic-link plan handed to `emit_dyn_exec`
-# import_map:  imported symbol name -> index into `info.imports`, for the patch pass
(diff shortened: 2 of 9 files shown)


octalide:
## macos-14 exec proof — PASSED

Validated end-to-end on a `macos-14` (Apple Silicon) runner via a `workflow_dispatch` dispatch (throwaway branch `inv/1722-proof`, with the paired mach-std runtime pinned):

- **PROOF 1 — exec:** the minimal, rebase, and full-compiler arm64 PIE seeds all **EXEC** (no `Killed: 9`).
- **PROOF 2 — fixpoint:** the native `a→b→c` darwin self-host **converges, `b == c`** (deterministic).
- **PROOF 3 — `mach test .` on Apple Silicon:** **425 passed, 1 failed** (`mach.lang.driver:union_ctx_restored_to_default_tuple`).

### What it took (each found via the on-Mac oracle)
Beyond the PIE writer (MH_PIE, LC_MAIN, rebase stream, LC_BUILD_VERSION, LC_UUID, `__DATA_CONST` SG_READ_ONLY), three more were required, each a real Apple-Silicon contract:
1. **16 KiB segment alignment** — Apple Silicon maps 16 KiB pages and SIGKILLs a main executable whose segments aren't 16 KiB aligned (mach used 4 KiB). This was the dominant blocker.
2. **implicit libSystem** — a dyld classic-`LC_DYLD_INFO` image with no `LC_LOAD_DYLIB` is not a valid darwin executable (the system linker won't even produce one).
3. **`CS_LINKER_SIGNED`** — CoreTrust admits a linker-signed ad-hoc main executable but rejects a hand-rolled ad-hoc one; mach is the linker, so it stamps the flag.

Plus paired mach-std darwin syscall fixes (mkdirat/unlinkat/faccessat numbers) on briar-systems/mach-std#325, which surfaced once the runtime actually ran.

### The one failing test
`union_ctx_restored_to_default_tuple` (driver.mach) is a **pre-existing, darwin-host-specific** failure unrelated to this PR: its union-walk source only handles linux/windows hosts and it exercises the comptime ctx-restoration path, which this PR does not touch. It passes on the linux PR CI. It only became observable now that the suite can run on a darwin host. Recommend a separate follow-up issue.

### Merge coupling
Still depends on mach-std#325; bump `mach.lock` to its merged commit before this lands.

Issue #1680, opened 2026-06-28
infra: CD-authored darwin release build (x86_64 + aarch64, native fixpoint)


octalide:
Depends on #1679.

## What

The CD pipeline (`.github/workflows/cd.yml`) builds and publishes release archives
for `x86_64-linux`, `aarch64-linux`, and `x86_64-windows`, each built natively on its
own host to a self-host fixpoint (`a → b → c`, `b == c`) and seeded from the previous
release's binary. There is **no darwin job**, so no release ships a darwin binary —
v2.9.0's assets are the three above plus `SHA256SUMS`. This is the last gap in the
darwin milestone: mach can *target* darwin (#1176, #1178) but the project never
*ships* one.

Add a `build-darwin` job, matching the established native-fixpoint release model
(cross-compiling-and-shipping is explicitly rejected — every target is built natively
on its own host, per the `cd.yml` header):

- **Runners:** `macos-13` → `x86_64-darwin`, `macos-14` (Apple Silicon) →
  `aarch64-darwin`. A matrix mirroring `build-linux`.
- **Bootstrap seed (chicken-and-egg):** no prior darwin release exists to seed from,
  exactly like the riscv64 backend predating its seed. For the first darwin release,
  cross-build a darwin `mach` on the Linux runner (the `cross-darwin` lane already
  proves Mach-O cross-compilation works), publish/pass it to the macOS runner as the
  stage-0 seed, then run the native `a → b → c` fixpoint there. Once the first darwin
  archive is released, subsequent CD runs self-seed from it like every other target —
  collapse the bootstrap path at that point.
- **Fixpoint + package:** same shape as `build-linux` — `mach dep pull`, three
  `--profile release` builds, `cmp -s b c`, then tar `mach` + `LICENSE` into
  `dist/mach-$ver-$TARGET.tar.gz`, uploaded as `release-$TARGET`. The existing
  `release` job already globs `dist/mach-*.tar.gz`, so darwin archives flow into
  `SHA256SUMS` and the published release with no change there.

CI scope (decided): the native darwin fixpoint lives **only** in `cd.yml`, which is
tag-only (releases cut from `main`), so it runs a handful of times per release rather
than on every PR. `ci.yml` runs on every PR to `dev`/`main`, where native macOS
runners would cost ~10x ubuntu minutes for no commensurate benefit — so `ci.yml`
**keeps the existing ubuntu `cross-darwin` byte-verify lane unchanged** and gets no
[...]

PR #1726, opened 2026-06-29
fix(driver): cover darwin hosts in union ctx-restoration test fixture

### LINKED ISSUE #1724: driver/sema: union_ctx_restored_to_default_tuple fails on a darwin host (pre-existing, host-specific)
Surfaced by (not caused by) #1722 — the first time `mach test .` ran on a darwin host.

The test `mach.lang.driver:union_ctx_restored_to_default_tuple` (`src/lang/driver.mach:4298`)
FAILs when the test suite runs on an **arm64 darwin host**, but PASSes on x86_64-linux,
aarch64-linux, and x86_64-windows CI. Because it correlates with **host OS = darwin**
(not arch — aarch64-linux passes) and lives in the driver/comptime path that #1722's
codegen/link changes never touch, it is a pre-existing, host-data-dependent bug, not a
regression from the PIE work.

## Repro

Run `mach test .` with a self-hosted compiler on an arm64 macOS host (e.g. #1680's
macos-14 `workflow_dispatch`): `mach.lang.driver:union_ctx_restored_to_default_tuple
... FAIL(exit 1)` (425 passed, 1 failed). Passes on every linux/windows CI lane.

## Likely cause

The fixture exercises comptime union-context restoration to the host's "default tuple",
and its expectations only enumerate linux/windows hosts — on a darwin host it hits an
unhandled/default branch. Either the test fixture needs a darwin host case, or it
exposes a real but narrow ctx-restoration leak that linux/windows hosts can't observe
(their host default coincides with a branch tuple). Confirm which during the fix.

## Acceptance

`mach test .` is 426/426 on an arm64 darwin host; the union-walk ctx-restoration is
correct (or the fixture covers darwin hosts), with the linux/windows lanes still green.

octalide:
## Summary

`mach.lang.driver:union_ctx_restored_to_default_tuple` (`src/lang/driver.mach`) FAILed only on an arm64 **darwin** host (425/1), while passing on x86_64-linux, aarch64-linux, and x86_64-windows. This blocked the darwin (and all) releases: #1680's `build-darwin` self-test gates the `release` job.

## Root cause — (a) test-fixture gap, not a restoration leak

The test asserts the restored comptime ctx equals the host's native-resolved default tuple:

```
or (cm.target_os_id != target.host_os_id())     { bad = 1; }
or (cm.target_arch_id != target.host_arch_id()) { bad = 1; }
```

`resolve_native` (`manifest.mach`) scans the manifest's declared targets for one matching the host `(os, arch)`; on **no match it falls back to `m.targets[0]`**. The fixture manifest `ut_manifest()` declared only `linux` (x86_64), `windows` (x86_64), and `linux-arm64` (aarch64-linux) — **no darwin target**. On an arm64 darwin host nothing matches, so the selected default tuple is x86_64-linux while `host_os_id()`/`host_arch_id()` return darwin/aarch64, and the host-tie assertions fail.

The union ctx-restoration itself is correct: modules `a`/`b` are seeded from `p.target` and `main` restores to the same selected default, so `cm`/`ca`/`cb` stay internally consistent on darwin too — only the host-tie comparison was wrong, because the fixture had no host-matching target for darwin.

## Fix

Declare a `darwin-arm64` (`aarch64`-`darwin`-`aapcs64`) target in `ut_manifest()` so an arm64 macOS host resolves a matching native target, restoring the `selected default == host ids` equivalence the assertion relies on. No masking, no test skip. Every declared target has a distinct `(isa, os)` so `native` is never ambiguous, and each CI host keeps matching its own declared target.

## Validation

- x86_64-linux: PR-source compiler runs `mach test .` -> **426 passed, 0 failed**; all `union_*` tests pass.
- arm64 darwin (the proof): validated via a macos-14 `workflow_dispatch` (see PR comments) — the self-test goes **426/0**.

Closes #1724

### DIFF
--- a/src/lang/driver.mach
+++ b/src/lang/driver.mach
@@ -4042,10 +4042,15 @@ fun ut_cc3(alloc: *A.Allocator, a: str, b: str, c: str) str {
     ret buf::str;
 }
 
-# the shared three-target manifest the union tests scaffold. `native`
-# resolves the default to the host-matching target (linux, windows, or linux-arm64 in CI).
+# the shared five-target manifest the union tests scaffold. `native` resolves the
+# default to the host-matching target: linux (x86_64-linux), windows (x86_64-windows),
+# linux-arm64 (aarch64-linux), darwin-x86_64 (x86_64-darwin), or darwin-arm64
+# (aarch64-darwin) in CI. both darwin targets are declared so either a macos-13
+# (x86_64) or a macos-14 (arm64) host has a matching native target - otherwise
+# `resolve_native` falls back to the first target (linux) and the host-tie assertions in
+# union_ctx_restored_to_default_tuple read the wrong default for a darwin host (#1724).
 fun ut_manifest() str {
-    ret "[project]\nid = \"uniontest\"\nname = \"u\"\nversion = \"0.0.0\"\nsrc = \"src\"\ntarget = \"native\"\nout = \"out/{target}/bin/{name}{ext}\"\n\n[target.linux]\nisa = \"x86_64\"\nos = \"linux\"\nabi = \"sysv64\"\n\n[target.windows]\nisa = \"x86_64\"\nos = \"windows\"\nabi = \"win64\"\next = \".exe\"\n\n[target.linux-arm64]\nisa = \"aarch64\"\nos = \"linux\"\nabi = \"aapcs64\"\n\n[bin.uniontest]\nentry = \"main.mach\"\n";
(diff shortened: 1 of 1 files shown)


octalide:
Root cause confirmed as **(a) a test-fixture gap** (not a comptime restoration leak). `resolve_native` falls back to the first declared target on no host match; the fixture declared no darwin target, so on an arm64 darwin host the selected default tuple was x86_64-linux while `host_os_id()`/`host_arch_id()` are darwin/aarch64, failing the host-tie assertions only there.

**Off-Mac:** PR-source compiler runs `mach test .` on x86_64-linux -> 426 passed, 0 failed; all `union_*` tests pass.

**On-Mac proof:** dispatched CD (`workflow_dispatch`) on proof branch `ci/1724-darwin-proof` (this fix layered on the #1680 darwin lane), headSha `7816fc76840c910dc69738b63723f64d340b279e`. Watching the macos-14 (aarch64) `build-darwin` self test for 426/0. Run: https://github.com/briar-systems/mach/actions/runs/28346292929

octalide:
Scope correction: `build-darwin` runs **both** macos-13 (x86_64-darwin) and macos-14 (aarch64-darwin), and the host-tie failure hits both. Expanded the fix to declare both darwin targets (`darwin-x86_64` sysv64 + `darwin-arm64` aapcs64), mirroring the project's own mach.toml. Linux stays 426/0.

Re-dispatched CD on `ci/1724-darwin-proof` headSha `e0f462b7695118afdacc4efb5784ee4f520b329c` (the old run was cancelled). Watching both macos lanes for 426/0. Run: https://github.com/briar-systems/mach/actions/runs/28346429709

Issue #1627, opened 2026-06-26
riscv64: instruction selection + lp64d call lowering (slice 3)


octalide:
Slice 3 of the riscv64 backend (#1172). Instruction selection over the slice-2 encoder,
plus completion of the lp64d calling convention.

## Scope
- NEW `src/lang/target/isa/riscv64/isel.mach` (SelectFn): generic MIR -> concrete RV64
  opcodes; flags-less compare lowering (slt/sltu + branch, RISC-V has no condition-flags
  register); large immediates via lui+addi; address materialization via auipc+addi;
  three-address ALU forms.
- Complete the deferred lp64d rules in `src/lang/target/abi/lp64.mach`: hard-float struct
  flattening, and the 9-16 byte register/stack split.
- Wire `select` in `register.mach`.

## Dependency
Blocked by the float-bank fix #1624 - `fp_class_index` must find the float bank by
`RC_FLOAT` kind before RV64D floats can lower. Land #1624 first or as the opening commit.

## Size
isel ~600-1200 lines; lp64 delta + the fp fix small. Rough.

## Width boundary
- Width-agnostic: the selection patterns (MIR-op -> RISC-V form), the flags-less
  compare-and-branch idiom, three-address forms, the lp64/ilp32 register-role assignment.
- RV64-specific: *W-vs-full-width op choice for 32-bit ops in 64-bit registers,
  LD/SD-vs-LW/SW by access width, 64-bit const materialization, lp64d (vs ilp32d) slot
  width and 2*XLEN thresholds. All width-parameterized, not arch-flagged.

Part of #1172. Blocks relocations + printer (slice 4).

PR #1732, opened 2026-06-29
fix(abi): shared register/stack-split + mixed-class aggregate slot primitive

### LINKED ISSUE #1637: abi/codegen: aggregate arg-passing needs a shared register-plus-stack split / mixed-class slot-emit primitive
riscv64 lp64d (slice 3, #1627) completed homogeneous-float-aggregate passing (CLASS_HFA),
but three psABI aggregate rules could not be expressed with the shared backend's current
argument-emission machinery:

- a mixed float + integer struct (members split across the FP and GP register files),
- a struct that flattens to two floats of different fundamental width (e.g. {f32, f64}),
- the 9-16 byte integer case that splits one register + one stack slot when the argument
  registers are nearly exhausted.

The shared classes today are CLASS_HFA (uniform-width, FP-only) and a GP-only register pair
at fixed 0/8 offsets. There is no register-plus-stack split class and no mixed FP+GP slot
emit, so these aggregates currently take the conservative integer / by-value-stack path.

Impact: the conservative path is SELF-CONSISTENT across a mach caller and callee, so mach
programs are correct. It diverges from the C psABI only at an FFI boundary for these
specific struct shapes (passing such a struct to or from C would mis-place members).

This is a SHARED-backend extension (a general `emit_arg_slot`-style primitive supporting
register-plus-stack split and mixed-class aggregates), not riscv64-specific. x86_64 sysv and
aarch64 aapcs64 have analogous split/mixed rules - investigate whether they handle these
cases or also take a conservative path, so the primitive serves all three.

Surfaced and escalated (not worked around) by riscv64 slice 3 (#1627). Not on the path to a
working riscv64 self-host (mach<->mach is correct); needed for full C psABI FFI conformance
with mixed/split aggregates.

octalide:
Closes #1637.

The shared backend could not express three RISC-V lp64d aggregate placements (escalated, not worked around, by riscv64 slice 3 #1627): a mixed float+integer struct (members split across the FP and GP files), a struct that flattens to two floats of different fundamental width, and the 9–16 byte integer aggregate that splits across the last integer register and the stack. The `reg`/`reg2`/`reg_count`/`reg_width` slot could only describe a single register, a GP pair at fixed 0/8, or a uniform-width FP-only `CLASS_HFA` run.

**Investigation outcome:** the gap is lp64d-specific. x86-64 SysV already places mixed FP+GP aggregates via the `reg`/`reg2` pair (the FP half rides a class-dispatched MOVSD) and never splits a register placement across the stack; AArch64 AAPCS64 is all-or-memory (HFA, GP pair, byref, or whole-stack) and never mixes or splits; Win64 passes anything >8 bytes by reference. So only lp64d needs the new expressiveness — but the primitive is shared.

## Plan
- [x] **Part 1 — the primitive (this commit).** `ParamPiece` (GP/FP/stack chunk with a source offset, stack offset, and width) + a `pieces[]`/`piece_count` list on `ParamSlot`. Every constructor fills it; the IR→MIR consumer materializes all register/stack aggregates through one uniform per-piece path. Behavior-preserving — all 686 tests and the riscv64/macho/freestanding byte-verify fixtures pass unchanged.
- [ ] **Part 2 — the register-plus-stack split (case C).** lp64d emits a split slot (GP reg + stack chunk) for a 9–16 byte integer aggregate when exactly one integer register remains.
- [ ] **Part 3 — mixed + different-width float (cases A, B).** A flatten descriptor (the per-member is-float/offset/width layout the per-ABI classifier needs but the type oracle owns) feeds lp64d's mixed float+integer and different-width float-pair rules.

## Verification
`mach build .` self-host, `mach test .` (686/686), and `test/{riscv64,macho,freestanding}/verify.sh` all green at each commit.

### DIFF
--- a/src/lang/be/codegen/mir/abi.mach
+++ b/src/lang/be/codegen/mir/abi.mach
@@ -701,8 +701,11 @@ pub fun lower_call(ctx: *context.LowerCtx, mb: *mir.MirBlock, inst: *instruction
             slot = classify_arg(pidx::i32, asz, aal, afl, 0, aag, gp_used, fp_used, 0, 0);
             # round a stack-passed tail argument up to its slot alignment before
             # placing it (identity on x86; honors a 16-aligned AAPCS64 by-value arg).
+            # a register-plus-stack split's tail stack chunk is eight-byte-granular.
             if (slot.class == abi.CLASS_STACK || slot.class == abi.CLASS_STACK_BYREF) {
                 stack_off = round_up_i64(stack_off, stack_arg_align(slot.class, aal));
+            } or (slot_has_stack_piece(?slot)) {
+                stack_off = round_up_i64(stack_off, 8);
             }
             soff = stack_off;
         }
@@ -751,6 +754,10 @@ pub fun lower_call(ctx: *context.LowerCtx, mb: *mir.MirBlock, inst: *instruction
                 stack_off = stack_off + context.stack_slot_bytes(slot.size);
             } or {
                 advance_arg_cursors(?slot, ?gp_used, ?fp_used);
+                # a split tail also consumes its stack chunk's outgoing slot.
+                if (slot_has_stack_piece(?slot)) {
+                    stack_off = stack_off + split_stack_bytes(?slot);
+                }
             }
         }
 
--- a/src/lang/be/codegen/mir/abi.mach
+++ b/src/lang/be/codegen/mir/abi.mach
@@ -327,7 +327,9 @@ fun agg_flatten(ctx: *context.LowerCtx, ty: type.IrTypeId, base_off: u64,
                 out: *abi.AggLayout, count: u32) u32 {
     if (count > AGG_MAX_LEAVES) { ret count; }
     val it: O.Option[*type.IrType] = type.get(ctx.types, ty);
-    if (!O.is_some[*type.IrType](it)) { ret count + 1; }
+    # an unresolvable type has no known leaf to record; poison the walk past the cap so
+    # the caller rejects the whole aggregate (it falls to the integer convention).
+    if (!O.is_some[*type.IrType](it)) { ret AGG_MAX_LEAVES + 1; }
     val t: *type.IrType = O.unwrap[*type.IrType](it);
 
     if (t.kind == type.IRT_STRUCT) {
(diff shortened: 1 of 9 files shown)


octalide:
Part 1 (the ParamPiece primitive + unified consumer) and Part 2 (the 9–16 byte register-plus-stack split, case C) are landed and verified end-to-end (riscv64 qemu fixture runs to exit 105). Part 3 (cases A mixed float+int and B different-width float pair, via a per-member flatten descriptor) is in progress — the consumer already materializes arbitrary mixed-bank/different-width pieces, so this is classifier-side only.

octalide:
Part 3 landed scoped (commit f66a7392): lp64d cases **A** (mixed float+integer → fa+a) and **B** (different-width float pair → fa0:fa1 per-member width) in `classify_arg`/`classify_return`, via a new `AggLayout` per-member flatten descriptor; bank-exhaustion falls through to the integer convention. +2 unit tests (689/689).

All three issue cases (A/B/C) are now implemented on the ABI side. **A/B's end-to-end float validation is gated on #1733** — a pre-existing RISC-V FP load/store mis-encoding (integer `ld`/`sd` instead of `fld`/`fsd`) surfaced by this work, kept separate per one-issue-per-PR. The float e2e fixture (HFA + A + B) lands with #1733's encoder fix.

Verified independently: `mach build` clean, `mach test` 689/689, riscv64 (105) / macho / freestanding (45) fixtures green; CI green across all targets.

octalide:
Max-effort code review run before merge surfaced one **confirmed miscompile** — the lp64d mixed float+int rule emitted the integer leaf's GP chunk at a hardcoded 8-byte width, so a `struct { f32; i32 }` read/wrote 4 bytes out of bounds (masked only by #1733). Fixed in 43033793: the mixed GP piece now uses the leaf width and the four consumer piece loops honor `p.width` (byte-identical for every existing width-8 piece — riscv fixture holds at 105), with a unit test for the sub-word shape. Also hardened: `agg_flatten` now poisons on an unresolvable type instead of failing open, loop early-outs past the leaf cap, `agg_layout` reuses the caller's `is_aggregate`, `make_slot_pieces` is bounded by `PARAM_MAX_PIECES`, and the piece constructors are routed through `make_piece`.

Deferred (design-debt cleanups the review flagged) → **#1734**. Pre-existing riscv variadic-float ABI bug → **#1735**. A/B end-to-end float validation remains gated on **#1733**.

Final state: `mach test` 690/690, all three byte-verify fixtures green, CI green across all targets. Merging.

Issue #1734, opened 2026-06-29
refactor(abi): consolidate the ParamPiece slot model — drop vestigial parallel state, unify the leaf-walkers and piece-dispatch


octalide:
Follow-up to #1637 (the ParamPiece slot-model refactor); surfaced by that PR's max-effort code review. These are behavior-preserving consolidations that finish the "one uniform per-piece mechanism" the refactor set out to build — none is a correctness bug (the confirmed bug from that review is fixed in #1637), but together they remove the duplicated/vestigial state the design rules flag.

## What

After #1637, the backend materializes every register/stack aggregate through the `ParamSlot.pieces` list, but several pre-piece structures linger as duplicated or dead parallel state:

1. **Vestigial `reg2` / `reg_count` / `reg_width` / `CLASS_HFA`.** Post-refactor no non-test, non-constructor code reads these — `pieces` is the sole representation the lowering consumer (`emit_arg_slot`/`emit_param_slot`/`emit_ret_slot_to_vreg`/`lower_ret`/`advance_arg_cursors`) consumes. Every classifier still fills them and the per-ABI unit tests assert on them, so the slot carries duplicated state hand-kept in sync with `pieces`, and the tests verify a representation codegen ignores. Remove the shorthand fields (and the `CLASS_HFA` discriminant where it is now redundant), and rewrite the per-ABI placement tests to assert on `pieces`.

2. **Three independent leaf-walkers.** `mark_eightbyte_classes` (SysV SSE mask), `hfa_walk` (AAPCS64 HFA shape), and `agg_flatten` (lp64d AggLayout) each re-implement the same IRT_STRUCT/IRT_ARRAY/IRT_UNION/scalar recursive leaf traversal in `be/codegen/mir/abi.mach`. Factor a single shared per-leaf iterator the three consumers build on, so a change to leaf flattening (a new IR type kind, an offset/alignment fix) lands once.

3. **Four hand-rolled piece-dispatch loops.** `emit_arg_slot`, `emit_param_slot`, `emit_ret_slot_to_vreg`, and `lower_ret` each hand-roll the same PIECE_GP/PIECE_FP/PIECE_STACK walk (the register arms are identical in load/store pairs). Factor a shared per-piece materializer parameterized by direction/destination, so a move-width or new-PieceKind change lands once.

[...]

PR #1736, opened 2026-06-29
fix(codegen/riscv64): encode FP loads/stores as fld/fsd, not integer ld/sd

### LINKED ISSUE #1733: codegen(riscv64): FP loads/stores from memory are mis-encoded as integer ld/sd
Surfaced by #1637 (the lp64d hard-float aggregate ABI work); it gates the end-to-end validation of #1637 cases A/B.

On the riscv64 backend, floating-point loads and stores to/from memory are encoded as **integer `ld`/`sd`** of the GP register that aliases the FP register's index, instead of `fld`/`fsd`. Any f32/f64 value that round-trips through memory — a stack-local float, a struct float member spilled/reloaded, a global float — is read from or written to the wrong register file and produces garbage. It is a silent miscompilation of all riscv64 hard-float that touches memory.

## Repro

```mach
var x: f64 = 2.0;
var g: f64 = x + 3.0;
ret g::i64;
```
Cross-compiled `--target linux-riscv64` and run under `qemu-riscv64`.

## Expected
Exit code `5` (the reload of `x` is an `fld`, the `fadd` then sees `2.0`).

## Actual
Exit code `0`. The reload of `x` emits an integer `ld` into a GP register, so the `fadd` reads an unloaded FP register.

## Root cause

In `src/lang/target/isa/riscv64/encode.mach`:
- `encode_load` (~`:804`) resolves the destination with `req_gpr(?mi.operands[0])` **unconditionally** and emits the integer `emit_mem_access(...)`, with no `reg_is_fp` check.
- `emit_store_mem` (~`:706`) resolves the stored value with `req_gpr(val_op)` unconditionally → integer store.
- `emit_global_load` (~`:760`) and `emit_global_store` (~`:776`) — the folded-global `%pcrel` paths — likewise use the integer LOAD/STORE encoding + `req_gpr`.
- The correct FP-memory primitives **already exist** but are wired only into the regalloc FP-spill path, never into the general `MIR_LOAD`/`MIR_STORE` encoding: `req_fpr` (~`:1084`), `reg_is_fp` (~`:1097`), `fp_mem_funct3` (~`:1103`), and `emit_fp_mem` (~`:1113`, which packs `LOAD_FP`/`STORE_FP`).
- The isel (`src/lang/target/isa/riscv64/isel.mach`) FP-special-cases only `MIR_MOV` (→ `FMV`, register-to-register); `MIR_LOAD`/`MIR_STORE` fall through to the integer encoder.

It was latent because the riscv64 backend is cross-compile-only and the test corpus had **zero riscv float code** until the #1637 hard-float aggregate work added a float fixture, which exposed it.

## Impact
All riscv64 hard-float that touches memory is broken end to end: scalar floats, the existing lp64d HFA flattening, and the lp64d mixed / different-width float aggregate passing (#1637 cases A and B). The ABI-classifier side of #1637 (A/B) is correct and unit-tested but cannot be validated end to end until this lands. Not a current release blocker — riscv64 is cross-only with no float users today — but a silent miscompilation that must be fixed before any riscv float code is trusted.

## Proposed fix
Small, reuses the existing infra:
- `encode_load`: when `operands[0]` is `reg_is_fp`, resolve via `req_fpr` and emit `emit_fp_mem(..., is_load=true)` at `mi.src_width`.
- `emit_store_mem`: when the value operand is `reg_is_fp`, use `req_fpr` + `emit_fp_mem(..., false)` at `mi.width`.
- Handle the folded-global FP load/store sub-case in `emit_global_load`/`emit_global_store` (the `%pcrel` paths need an `fld`/`fsd` low half), or make that sub-case a loud error if deferred.

## Acceptance
- The repro returns `5` under qemu-riscv64.
- A float-bearing probe in `test/riscv64` — a scalar float round-trip plus the #1637 mixed and different-width aggregate probes — runs to its asserted exit code under qemu, proving lp64d hard-float (scalars, HFA, and cases A/B) end to end.
- The disassembly shows `fld`/`fsd` (not `ld`/`sd`) for FP memory access; existing integer load/store byte output is unchanged.

## Environment
- Target: `linux-riscv64` (cross-compile; `isa=riscv64, os=linux, abi=lp64`).
- Toolchain: self-hosted `mach` (dev), `qemu-riscv64`.
- Host: linux.

octalide:
Closes #1733.

The riscv64 `encode_load` / `emit_store_mem` resolved every operand via `req_gpr` and emitted the integer load/store form, so an f32/f64 round-tripping through memory was read/written from the GP register aliasing the FP index — a silent miscompile of **all** riscv64 hard-float that touches memory (scalar floats, the lp64d HFA flattening, and the #1637 mixed / different-width float aggregates). Latent because the backend is cross-only and the corpus had no riscv float code until the #1637 hard-float fixture exposed it.

## Fix
Both encoders dispatch a float-bank operand (`reg_is_fp`) to `req_fpr` + `emit_fp_mem` / `emit_fp_slot_store` (flw/fld, fsw/fsd), reusing the helpers already wired into the regalloc spill path. The folded-global FP symbol load/store sub-case has no float `la`-form yet and errors loudly rather than mis-encoding (it appears unreachable — riscv lowering materializes a global's address and loads via the pointer path).

## Verification
`test/riscv64` gains float-bearing probes — scalar round-trip plus #1637 case A (mixed f64+i64), **case A sub-word (f32+i32, the 4-byte GP chunk the #1637 review flagged)**, and case B (f32+f64 pair) — so the qemu e2e proves lp64d hard-float end to end. `mach build` clean, `mach test` 690/690, riscv64 fixture **exit 133**, macho + freestanding green. The fixture's many existing integer load/store probes still fold correctly, so the GP path is unregressed.

This unblocks the end-to-end validation of #1637 cases A/B (the ABI classifier side merged in #1732).

### DIFF
--- a/src/lang/target/isa/riscv64/encode.mach
+++ b/src/lang/target/isa/riscv64/encode.mach
@@ -704,6 +704,18 @@ fun encode_mov(st: *encode.EncodeState, f: *mir.MirFunction, mi: *mir.MirInstr) 
 # folded-global symbol destination emits the auipc-based `la` store sequence; otherwise
 # a non-register value (an immediate) is materialized into the scratch register first.
 fun emit_store_mem(st: *encode.EncodeState, f: *mir.MirFunction, dst_mem: *mir.MirOperand, val_op: *mir.MirOperand, width: u8) R.Result[bool, str] {
+    # a float-bank value stores through the F/D store form (fsw / fsd), recovering its
+    # f-register index from the composite id; the folded-global symbol destination has
+    # no float la-store form yet, so it errors loudly rather than mis-encoding as an
+    # integer store of the aliasing GP index.
+    if (reg_is_fp(val_op)) {
+        if (dst_mem.kind == mir.MIR_OP_SYM) {
+            ret R.err[bool, str]("encode: float store to a folded-global symbol is unsupported");
+        }
+        val rr: R.Result[i32, str] = req_fpr(val_op);
+        if (R.is_err[i32, str](rr)) { ret R.err[bool, str](R.unwrap_err[i32, str](rr)); }
+        ret emit_fp_slot_store(st, f, dst_mem, R.unwrap_ok[i32, str](rr)::u32, width);
+    }
     if (dst_mem.kind == mir.MIR_OP_SYM) {
         ret emit_global_store(st, dst_mem, val_op, width);
     }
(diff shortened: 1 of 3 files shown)


octalide:
**Review resolution** (high-effort code review):

- **[confirmed] spilled float load/store operand errors at encode time** — a float `MIR_LOAD` dest / `MIR_STORE` value that regalloc spills to a frame-slot MEM operand isn't caught by `reg_is_fp` (PREG-only) and falls through to `req_gpr`. This is a deeper gap: unlike float `MIR_MOV` (isel tags it `FMV` pre-spill so `encode_fmov` handles the slot), load/store carry no float marker for the encoder post-spill. **It's a loud compile error (fail-safe, not a miscompile), pre-existing (errored before this PR too), and rare (needs FP register pressure).** This PR is a strict improvement — it turns the common register-operand case from a *silent integer miscompile* into correct `fld`/`fsd`. Tracked as **#1737** rather than blocking on the isel/regalloc design change.
- **[cleanups] resolve_mem + FP/GP dispatch duplication** — folded into the broader piece/encoder consolidation **#1734**.
- **[acceptable] the folded-global FP symbol loud-errors** — guards an invariant the lowering already enforces; kept as the loud (not silent) form.

Net: register-operand FP load/store now correct and e2e-proven (riscv fixture exit 133, exercising scalar floats + #1637 cases A/A-subword/B). Merging.

PR #1738, opened 2026-06-29
refactor(abi): consolidate the ParamPiece slot model (#1734)

### LINKED ISSUE #1734: refactor(abi): consolidate the ParamPiece slot model — drop vestigial parallel state, unify the leaf-walkers and piece-dispatch

octalide:
Closes #1734.

Behavior-preserving consolidation finishing the #1637 piece-model refactor. One cleanup per commit, each verified: mach test 690/0, the riscv64 / macho / freestanding byte-verify fixtures unchanged (133 / ok / 45), the riscv64 object byte-identical to the pre-refactor compiler, and a self-host fixpoint (the refactored compiler rebuilds itself byte-identically). The leaf-walker change additionally checked byte-identical on x86 via a float-struct SSE-mask probe object.

Cleanups (issue order):
1. ✅ Removed vestigial `reg2` / `reg_count` / `reg_width` and the `CLASS_HFA` discriminant from `ParamSlot`; `make_slot_hfa` labels its run `CLASS_FP`; per-ABI tests rewritten to assert on `pieces`.
2. ✅ Unified the two offset-based leaf-walkers (`mark_eightbyte_classes` + `agg_flatten`) behind one `collect_leaves` enumerator. `hfa_walk` is kept separate **by design** — it is a homogeneity reduction (union = widest member's count, not a leaf sum), so folding it in would change union-HFA placement on an untested path; a load-bearing comment guards it. (Acceptance note added to #1734.)
3. ✅ Factored the four piece-dispatch loops into one `materialize_pieces` (PiecePlan-parameterized); the GP/FP register arms collapse.
4. ✅ Deduped the lp64d two-leaf rule across `classify_arg` / `classify_return` (`two_leaf_float_count` + `make_two_leaf_slot`).
5. ✅ Gated the `agg_layout` flatten on `AbiVTable.consumes_agg_layout`.

Left as **draft** for maintainer review of the full diff before merge.

### DIFF
--- a/src/lang/be/codegen/mir/abi.mach
+++ b/src/lang/be/codegen/mir/abi.mach
@@ -560,7 +560,12 @@ pub fun classify_signature(ctx: *context.LowerCtx, sig: type.IrTypeId, ret_type:
     var ret_hm:    u8   = 0;
     var ret_he:    u8   = 0;
     hfa_info(ctx, ret_type, ?ret_hm, ?ret_he);
-    out.ret_slot = classify_ret(ret_size, 0, ret_float, ret_eb, ret_aggr, ret_hm, ret_he, agg_layout(ctx, ret_type, ret_aggr));
+    # only the lp64d classifier reads the flattened-aggregate descriptor; for every
+    # other convention skip the per-value flatten walk and pass the no-information one.
+    val wants_agg: bool = ctx.tgt.abi.consumes_agg_layout;
+    var ret_agg: abi.AggLayout = abi.agg_none();
+    if (wants_agg) { ret_agg = agg_layout(ctx, ret_type, ret_aggr); }
+    out.ret_slot = classify_ret(ret_size, 0, ret_float, ret_eb, ret_aggr, ret_hm, ret_he, ret_agg);
 
     var gp_used:   i32 = 0;
     var fp_used:   i32 = 0;
@@ -614,7 +619,9 @@ pub fun classify_signature(ctx: *context.LowerCtx, sig: type.IrTypeId, ret_type:
         var phm:  u8   = 0;
         var phe:  u8   = 0;
         hfa_info(ctx, pty, ?phm, ?phe);
-        var slot: abi.ParamSlot = classify_arg(i::i32, psz, pal, pfl, peb, pag, gp_used, fp_used, phm, phe, agg_layout(ctx, pty, pag));
+        var pagg: abi.AggLayout = abi.agg_none();
+        if (wants_agg) { pagg = agg_layout(ctx, pty, pag); }
+        var slot: abi.ParamSlot = classify_arg(i::i32, psz, pal, pfl, peb, pag, gp_used, fp_used, phm, phe, pagg);
 
         # record the real outgoing stack offset on the slot (the ABI classifier
         # returns a sizing-only slot with offset 0); the param / call lowering
--- a/src/lang/be/codegen/mir/abi.mach
+++ b/src/lang/be/codegen/mir/abi.mach
@@ -949,6 +949,85 @@ fun lower_value_addr(ctx: *context.LowerCtx, mb: *mir.MirBlock, v: value.Value) 
     ret R.ok[mir.MirOperand, str](mir.op_vreg(addr));
 }
 
+# how `materialize_pieces` transfers a placement's register / stack chunks: the
+# direction (register load vs store), the per-chunk register-side memory operand, and
+# the stack-side base for a register-plus-stack split's PIECE_STACK chunk.
+# ---
+# store:      true to store each register chunk to its memory operand (a callee spill /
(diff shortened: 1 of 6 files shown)


octalide:
**Reviewed (coordinator pass) — sound, behavior-preserving, ready for your merge.**

Design:
- `collect_leaves` cleanly unifies the two true leaf *enumerations* (SSE-eightbyte mask + lp64d AggLayout flatten) behind one iterator, with a `strict` flag for their real divergence (the flatten rejects unions / unresolvable types; the mask enumerates overlapping union members) and a `cap` for their different limits — net -14 lines, not an abstraction-for-its-own-sake.
- `hfa_walk` is correctly kept **separate** with a load-bearing comment: it's a union-*max* homogeneity reduction, not an offset enumeration, so routing it through `collect_leaves` would silently change union-HFA placement (an untested path). #1734's acceptance was amended to document this exception.
- `materialize_pieces` + `PiecePlan` collapse the four hand-rolled caller/callee/return piece loops into one, honoring `p.width` (preserves the #1637 sub-word fix). Vestigial `reg2`/`reg_count`/`reg_width`/`CLASS_HFA` dropped; per-ABI tests now assert on `pieces`. `agg_layout` gated per-vtable.

Verification (independently re-run, plus the worker's): `mach test` 690/690; riscv64 133 / macho / freestanding 45 unchanged; **self-host fixpoint** (refactored compiler rebuilds itself byte-identically) and per-path .o byte-identity vs the pre-refactor compiler. CI green (windows test finishing).

Leaving the merge to you — it's a refactor of the ABI core, so you may want to eyeball the approach, and it's not urgent. Ready when you are.

PR #1746, opened 2026-06-29
fix(abi/sysv): pass >16B aggregate args by value on the stack (C-ABI conformance) (#1739)

### LINKED ISSUE #1739: abi/codegen(sysv): >16B aggregate argument passed by hidden reference instead of by value on the stack (C-ABI divergence, x86_64)
Found by the foundation-stability audit (#1180) and verified directly against the tree. A **C-ABI conformance divergence on the primary target** (x86_64-linux, and x86_64-darwin — both use `sysv.mach`).

## What
`sysv.classify_arg` passes a **>16-byte (MEMORY-class) aggregate argument by hidden reference** — `CLASS_BYREF`, a pointer in the next GP argument register — whenever a GP register is free (`src/lang/target/abi/sysv.mach:158-160`), and only falls back to by-value-on-stack (`CLASS_STACK`) when the GP bank is exhausted (`:163`). The locking test asserts this (`sysv.mach:~520-534`).

The SysV AMD64 psABI requires a **MEMORY-class argument to be passed by value on the stack** — it consumes **no** integer register and is **not** a pointer. (The hidden-pointer form is for a MEMORY-class *return*, in RDI/RAX — and that path is correct here.) So mach puts a pointer in a GP register where gcc/clang put the struct's bytes on the stack, and also wrongly consumes a GP slot, shifting every subsequent argument.

## Why it's latent (not a regression for any shipping path)
mach↔mach is self-consistent — caller and callee both use `CLASS_BYREF`, so pure-mach programs are correct. It diverges **only at the C-FFI boundary**, and linux is no-libc / no C-FFI fixture, so nothing exercises a >16B struct across the boundary today. Same shape as the riscv lp64d conservative paths #1637 addressed — but here it's presented as the intended design, not flagged as a deferred divergence.

## Fix
Drop the `CLASS_BYREF` branch for a >16B aggregate **argument**: classify it `CLASS_STACK` (by value at full size) unconditionally — the existing no-GP-register fallback is already the psABI-correct placement, so the by-value copy / outgoing-cursor / callee-read machinery is in place. Rewrite the locking test and the doc comment. (Confirm the full MEMORY-class conditions — size and field alignment — match the psABI, not just `size > 16`.)

Note: this **changes mach↔mach codegen** for any call passing a >16B struct by value (including the compiler's own code), so it is **not byte-identical** — re-establish the self-host fixpoint at the new placement and confirm the corpus + fixtures stay green. A C-FFI conformance probe (a mach↔C call with a large struct) would lock it down if the harness allows.

## Acceptance
- A >16B aggregate argument is passed by value on the stack, consuming no GP register; placement matches gcc/clang for the same struct.
- Self-host fixpoint re-established; `mach test` + all fixtures green.

octalide:
## What

`sysv.classify_arg` passed a **>16-byte (MEMORY-class) aggregate argument by hidden reference** (`CLASS_BYREF` — a pointer in the next GP argument register) whenever a GP register was free, only falling back to by-value-on-stack (`CLASS_STACK`) once the GP bank was exhausted. The locking test asserted the BYREF behavior.

The SysV AMD64 psABI requires a **MEMORY-class argument to be passed by value on the stack**, consuming **no** integer register. The hidden-pointer form is for a MEMORY-class **return** (the sret path) — `classify_return` is unchanged.

## Fix

Drop the `CLASS_BYREF` branch in `classify_arg` for the >16-byte aggregate argument path so it unconditionally returns `CLASS_STACK` at full `size`. The existing no-GP-register fallback was already the psABI-correct placement, so the by-value-copy / outgoing-cursor / callee-read machinery (in `be/codegen/mir/abi.mach`) is already in place. Rewrote the locking unit test and updated the doc comments. **Only `sysv.mach` is touched** — win64/aapcs64/lp64 keep their own large-aggregate rules.

Caller and callee both classify through the same whole-signature oracle, so mach↔mach stays self-consistent at the new placement.

## Validation

- **Bug confirmed real**: the old test locked `s3.class == CLASS_BYREF` for a >16B arg with a free GP register; the return/sret path (`classify_return` → `CLASS_SRET`) is a separate function, unaffected.
- **Self-host fixpoint re-established at the new placement**: `mach build . -o a` → `./a build . -o b` → `./b build . -o c`; `b` and `c` are **byte-identical**. A clean-dev baseline fixpoint differs from this one (`c0 != c`), confirming the change does alter the compiler's own codegen — i.e. the compiler source itself passes some >16B struct by value, now emitted by value on the stack. **Output changed; the fixpoint holds.**
- **`mach build .` clean**; **`mach test .` 430/430** (project) and **690/690 with `--include-deps`**, run with the new fixpoint compiler.
- **All three fixtures green** with the new compiler: `freestanding` (exit 45), `macho` (both Darwin triples, static + dynamic + rebase + signatures), `riscv64` (RV64 ELF runs to exit code 133 under qemu — unchanged, the sysv change doesn't affect riscv/lp64d).

Closes #1739

### DIFF
--- a/src/lang/target/abi/sysv.mach
+++ b/src/lang/target/abi/sysv.mach
@@ -3,11 +3,12 @@
 # Implements `abi.AbiVTable` and self-registers under `abi.ABI_SYSV` at startup.
 # Owns the SysV AMD64 classification of parameters and return values -- the
 # <=8-byte one-register, 9-16-byte two-register (RDI:RSI for args, RAX:RDX for
-# returns), and >16-byte by-reference / sret rules -- plus the GP/FP/callee-saved
-# register files, the 16-byte stack alignment, the 128-byte red zone, and the
-# variadic save-area hooks. The three classify / arg_passing / ret_passing vtable
-# function pointers are type-erased (`*u8`); a consumer casts each back to the
-# concrete signature declared here (`ClassifyFn`, `ArgPassingFn`, `RetPassingFn`).
+# returns), >16-byte by-value-on-stack argument, and >16-byte sret return rules --
+# plus the GP/FP/callee-saved register files, the 16-byte stack alignment, the
+# 128-byte red zone, and the variadic save-area hooks. The three classify /
+# arg_passing / ret_passing vtable function pointers are type-erased (`*u8`); a
+# consumer casts each back to the concrete signature declared here (`ClassifyFn`,
+# `ArgPassingFn`, `RetPassingFn`).
 
 use std.types.bool.bool;
 use std.types.bool.false;
@@ -121,10 +122,10 @@ pub fun classify(size: u64, align: u64, is_float: bool, fp_eightbytes: u8,
 
 # assign registers or a stack slot for one argument (SysV)
 #
-# Aggregates larger than 16 bytes are passed by hidden reference (CLASS_BYREF)
-# when a GP argument register remains -- the pointer occupies that register; once
-# the GP registers are exhausted such an aggregate is MEMORY-class, passed BY
-# VALUE on the stack (its full byte size, CLASS_STACK), not by reference.
+# Aggregates larger than 16 bytes are MEMORY-class: passed BY VALUE on the stack at
+# their full byte size (CLASS_STACK), consuming NO GP register and never a hidden
+# pointer -- the System V psABI reserves the hidden-pointer form for a MEMORY-class
+# RETURN (the sret path), not an argument.
 # Aggregates of 9-16 bytes are classified PER EIGHTBYTE: each eightbyte takes a GP
 # argument register, or an SSE eightbyte (all its fields floating-point, per
 # `fp_eightbytes`) takes an XMM register, so `{f64; f64}` rides XMM0:XMM1 and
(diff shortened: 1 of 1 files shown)


octalide:
**Reviewed (coordinator pass) — verified, merging.**

- Fix is psABI-correct: >16B aggregate **argument** now unconditionally `CLASS_STACK` (by value on the stack, no GP register, no hidden pointer); the hidden-pointer form stays reserved for the MEMORY-class **return** (sret), which is a separate function and untouched.
- **Non-byte-identical primary-target ABI change** (expected): the compiler's own source passes at least one >16B struct by value, so those sites now emit by-value-on-stack instead of a register pointer. Independently re-verified the **self-host fixpoint holds at the new placement** (built seed→a→b→c, `stage2==stage3` byte-identical), so the new compiler reproduces itself; mach↔mach stays consistent (caller + callee classify through the same oracle).
- Independent re-verify: `mach test` 430/430 (690/690 with deps per the worker), riscv64 (133) / macho / freestanding (45) fixtures green, CI green across all lanes.

Note: the `reg2`/`reg_count` doc staleness the worker spotted is pre-existing (from #1734's field removal) and spans the ABI doc comments — folding it into a separate quick doc sweep rather than this PR.

PR #1748, opened 2026-06-29
fix(codegen): encode spilled-float load/store via the FP scratch on riscv64 + arm64 (#1737)

### LINKED ISSUE #1737: codegen(arm64+riscv64): spilled float load/store operand errors at encode time (cross-ISA)
Surfaced by the #1733 review. #1733 fixed FP load/store for register operands (the common case, e2e-proven); this is the remaining spilled-operand gap. It is reachable *because* #1733 enables riscv float code at all: a float-heavy lp64d function under enough FP register pressure to spill the value being stored / the result of a load will now fail to compile.

A float `MIR_LOAD` whose destination vreg is spilled, or a float `MIR_STORE` whose value vreg is spilled, reaches the encoder with that operand rewritten by regalloc into a **frame-slot MEM operand** (the documented spilled-float convention; `regalloc.mach:2457`/`:2474`). `encode_load` / `emit_store_mem` detect float operands via `reg_is_fp`, which matches only a float **PREG** — a spilled MEM operand falls through to `req_gpr` and hard-errors (`encode: expected a register operand`).

## Repro
A riscv64/lp64d function with enough simultaneously-live f64 values to exhaust the FP registers, storing/loading a spilled float (e.g. a large unrolled FP reduction). Currently aborts at encode time.

## Expected
The spilled float load/store encodes via the FP scratch + the slot (flw/fld into / fsw/fsd from the frame slot), exactly as `encode_fmov` already does for a spilled float **MOV**.

## Actual
`encode: expected a register operand` — the build aborts.

## Root cause
`encode_fmov` (`encode.mach:1482`) handles spilled-float MEM operands, but it knows the operand is float because isel routes a float `MIR_MOV` to `FMV` **before** regalloc spills (an opcode tag). `MIR_LOAD`/`MIR_STORE` carry no analogous float marker, so post-spill the encoder cannot tell a spilled load-dest / store-value is float from the operand alone (`mi.src_width` 4/8 doesn't distinguish f32 from i32).

## Fix candidates
1. Have isel tag float loads/stores (a float-load/store opcode, or a flag on the MIR instr) so the encoder routes them like `FMV` does, surviving the spill.
2. Have the encoder consult the function's vreg class for the load-dest / store-value even when the operand was spilled to a slot.
3. Extract an `emit_fp_slot_load` (the mirror of the existing `emit_fp_slot_store`) and dispatch a spilled FP load/store through it, reusing `emit_fp_mem` / `emit_fp_slot_store` / `FP_SCRATCH`.

## Acceptance
- A float-heavy riscv64 function that spills FP compiles, and a `test/riscv64` probe that forces an FP spill runs to its asserted exit code under qemu.
- Existing integer and register-operand FP load/store byte output is unchanged.

## Environment
- Target `linux-riscv64` (cross-only); self-hosted `mach` (dev); `qemu-riscv64`. Loud compile error, not a miscompile — fail-safe, but blocks float-heavy code under register pressure.

octalide:
Closes #1737.

## The bug
When register allocation spills a float, it leaves the spilled load destination / store value in place as a frame-slot MEM operand (`op_mem_slot(v)` carrying the float vreg id) for the encoder to fulfill — the contract `encode_fmov` already honored for spilled float MOVs. But the riscv64 and arm64 **load/store** encoders dispatched the scalar-FP form only on a float PREG (`reg_is_fp` / `req_vec`), so a spilled float (now a MEM with no PREG) fell through to `req_gpr` → `encode: expected a register operand` abort. Latent on arm64 because the integer-heavy self-host never spills floats; a float-heavy program under FP pressure aborts on **both** backends.

## The fix (approach (a), encoder-local — signed off)
Detect a spilled-float slot operand by its slot-key vreg's class (`is_fp_slot` via `vreg_is_fp`) — no isel/regalloc change, matching the existing operand-class FP seam and regalloc's own "identify the float side by vreg class" rule — and route it through the reserved FP scratch (never live across the single instruction's expansion, mirroring `encode_fmov`):
- spilled-dest **LOAD**: src → fp scratch → dest slot
- spilled-value **STORE**: value slot → fp scratch → dest mem

via the new `emit_fp_slot_load` mirroring `emit_fp_slot_store`, in both `riscv64/encode.mach` and `arm64/encode.mach`.

## Validation (before → after)
- Both ISAs **abort on pre-fix dev** (`encode: expected a register operand`) and **run correctly after** the fix.
- **riscv64 fixture**: a `fp_spill_probe` (forty simultaneously-live f64s, loaded from + stored back to a global buffer) folds into the exit code; fixture 133 → **153** under qemu-riscv64.
- **arm64 fixture (new `test/arm64/`)**: same probe, run under qemu-aarch64 to **exit 20**, asserting the spilled float encodes as scalar-FP `ldr/str Dd` (not an integer ld/st or an abort).
- `mach test` 690/0; macho + freestanding fixtures unchanged.

Draft for review — not marked ready.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -176,6 +176,44 @@ jobs:
       - name: cross-compile, byte-verify, and run the rv64 fixture
         run: bash test/riscv64/verify.sh "$PWD/m"
 
+  # aarch64-linux is cross-compile-only on this x86_64 host: the integer-heavy self-host
+  # never spills floats, so the spilled-float load/store encoding (#1737) has no other
+  # e2e guard. this lane builds an aarch64-capable compiler from the PR source,
+  # cross-compiles the freestanding fixture (forty simultaneously-live f64s forcing an FP
+  # spill), asserts the spilled float encodes as scalar-FP ldr/str Dd, and runs it under
+  # qemu-aarch64 to assert its inline-asm `exit` exit code.
+  cross-arm64:
+    name: cross aarch64-linux
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: install llvm tools and qemu
+        run: |
+          set -euo pipefail
+          sudo apt-get update
+          sudo apt-get install -y llvm qemu-user
+
+      - name: fetch released mach
+        env:
+          GH_TOKEN: ${{ github.token }}
+          SEED: mach-*-x86_64-linux.tar.gz
+        run: |
+          set -euo pipefail
+          tmp="$(mktemp -d)"
+          gh release download -R "$GITHUB_REPOSITORY" -p "$SEED" -D "$tmp"
+          tar -xzf "$tmp"/$SEED -C /usr/local/bin mach
+          chmod +x /usr/local/bin/mach
+
+      - name: build aarch64-capable compiler from source
+        run: |
+          set -euo pipefail
+          mach dep pull
+          mach build . -o m
+
+      - name: cross-compile, byte-verify, and run the arm64 fixture
+        run: bash test/arm64/verify.sh "$PWD/m"
+
   # the Darwin triples are cross-compile-only here: the compiler cannot run on the
   # Linux host, so it is not built or self-tested for darwin. this lane builds a
   # Mach-O-capable compiler from the PR source, cross-compiles the freestanding
--- a/src/lang/target/isa/arm64/encode.mach
+++ b/src/lang/target/isa/arm64/encode.mach
@@ -964,6 +964,15 @@ fun emit_store_mem(st: *encode.EncodeState, f: *mir.MirFunction, dst_mem: *mir.M
         ret R.ok[bool, str](true);
     }
 
+    # a spilled float VALUE: regalloc left it as its frame slot. load it into the reserved
+    # FP scratch, then store the scratch to the destination at the float store width.
+    if (is_fp_slot(f, val_op)) {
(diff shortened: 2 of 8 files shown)


octalide:
**Reviewed (coordinator pass) — verified both ISAs, CI-guarded, merging.**

- Fix is encoder-local approach (a): `is_fp_slot` (a spilled float slot = MIR_OP_MEM with a float-classed slot-key vreg) + a new `emit_fp_slot_load` mirroring `emit_fp_slot_store`, routing a spilled-MEM float load/store through the reserved FP scratch in both riscv64 and arm64 encoders — fulfilling the spilled-FP-slot contract regalloc already documents. Error propagation + the folded-global loud-error are correct; FP scratch is never live across the expansion.
- **Before→after, both ISAs**: a 40-live-f64 probe aborted with `encode: expected a register operand` on dev (both targets), and after the fix compiles + runs correctly under qemu — disasm shows scalar-FP `fld/fsd` (riscv) / `ldr/str Dd` (arm64), independently re-verified locally (riscv 153, arm64 20).
- **Both fixes are CI-guarded**: the riscv `fp_spill_probe` rides the existing `cross-riscv64` lane (133→153), and the new `cross aarch64-linux` lane runs the new `test/arm64` fixture under qemu-aarch64 — green on this PR. Bonus: that's the first linux-arm64 *execution* coverage in CI (the integer-heavy self-host never spills floats, so this path had none).

All 14 CI checks green. Clears the last cross-ISA correctness bug.

PR #1751, opened 2026-06-29
feat(coff): PE base relocations + DYNAMIC_BASE for windows ASLR (#1729)

### LINKED ISSUE #1729: link(coff): emit ASLR (.reloc base relocations) via the shared BaseReloc set (windows ASLR)
Deferred follow-on of #1718, enabled by #1722's format-neutral `BaseReloc` collection.

Consume the linker's `BaseReloc` set (the in-image absolute pointers) to emit PE
`.reloc` base-relocation blocks + `IMAGE_DLLCHARACTERISTICS_DYNAMIC_BASE` for **windows
ASLR**. Closer to a pure encoder than ELF-PIE: the Windows loader applies `.reloc`
itself, so no self-relocation startup is needed.

Optional hardening — windows runs fixed-base fine, no forcing function. Deferred until
ASLR is wanted.

octalide:
Closes #1729.

## The gap
Windows PE executables loaded at a fixed address: the COFF writer set `IMAGE_FILE_RELOCS_STRIPPED` + `NX_COMPAT` only — no `DYNAMIC_BASE`, no `.reloc`. So no ASLR.

## The fix
Make Windows PE images position-independent (ASLR), consuming the **same** shared format-neutral `BaseReloc` set the Mach-O ASLR writer (#1722) does.

1. **`windows pie_exec → true`** (x86_64): Windows is position-independent, so the link routes through the **existing** shared PIE base-relocation collection (in-image `RK_ABS64` pointers) rather than a windows-special branch. Windows already takes the dynamic path (mach-std imports `ExitProcess` from kernel32), so this adds only the base-reloc collection + the one-page PIE header gap (compatible with PE's ImageBase-one-header-span-below layout). **linux/darwin pie policy untouched.**
2. **COFF writer**: emit a `.reloc` section of `IMAGE_BASE_RELOCATION` blocks (DIR64 fixups, sorted + grouped per 4 KiB page), wire the BASERELOC data directory, set `DllCharacteristics` DYNAMIC_BASE (keeping NX_COMPAT), clear RELOCS_STRIPPED — all **gated on the image being position-independent**, so the static/fixed-base COFF path stays byte-identical. Mirrors the `.idata` synthesized-section path.

The Windows loader applies the `.reloc` fixups at load — no startup self-relocation (unlike ELF PIE).

## Validation
Cross-built the compiler for windows and byte-verified with `llvm-readobj`:
- `.reloc` section present with 45 DIR64 base relocations (`--coff-basereloc`)
- DllCharacteristics `0x140` = DYNAMIC_BASE + NX_COMPAT
- file Characteristics `0x2E` — RELOCS_STRIPPED **cleared**
- image parses cleanly
- `mach test` 690/0 (static COFF path unchanged)

The load-bearing run-proof is the **windows CI lanes**: with DYNAMIC_BASE set the loader ASLRs the image to a randomized base and applies `.reloc`, so a wrong fixup or pie-layout fails to load and build/test/release-windows go red. I'll report the lane result.

Draft for review — not marked ready.

### DIFF
--- a/src/lang/target/of/coff.mach
+++ b/src/lang/target/of/coff.mach
@@ -33,6 +33,7 @@
 
 use std.allocator.page;
 use std.collections.map;
+use sort: std.collections.sort;
 use std.filesystem;
 use std.memory;
 use std.types.bool.bool;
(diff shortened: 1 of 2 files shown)


octalide:
**Reviewed (coordinator pass) — verified, merging.**

- Design is the clean one: `windows pie_exec -> true` feeds the EXISTING shared PIE base-reloc collection (linux/darwin untouched), and the COFF writer consumes it to emit `.reloc` (per-4KiB-page IMAGE_BASE_RELOCATION blocks, DIR64 fixups), wires the BASERELOC data directory, sets DYNAMIC_BASE, clears RELOCS_STRIPPED — all gated on `pie`, so the non-pie COFF path is byte-identical. Mirrors the Mach-O ASLR pattern.
- The 64 KiB-alignment fix is the right resolution and a good catch: the PIE one-page header gap + windows's manual +page reserve double-counted, leaving ImageBase 4 KiB- but not 64 KiB-aligned. Dropping the manual reserve lets the PIE gap provide it; ImageBase stays `0x140000000`.
- **Load-bearing validation is the CI**: all three windows lanes (build/test/release) self-host the compiler and RUN it, so with DYNAMIC_BASE the real Windows loader ASLRs the image to a randomized base and applies `.reloc` — green means the fixups + layout are correct. This is exactly why the lane matters: llvm-readobj/wine accepted the misaligned image; only the real loader rejected it (and now accepts it). All 14 checks green.

Net: windows PE binaries now ship with ASLR (DYNAMIC_BASE + .reloc), no startup self-relocation needed (the loader applies it).

Issue #1760, opened 2026-06-29
int: structural (field) + freestanding (flat-loader) producers


octalide:
Part of #1758. Depends on #1758's first child (the framework).

## What

Add the two non-execution producers, **additively** — new producer functions in `int/lib/produce.sh` plus `targets.conf`/case entries, with no change to `int/run.sh`'s golden-diff core. This is the additivity the epic's design claims; this issue is the proof.

- **`field` producer** — `coreutils` (`od`/`dd`) reads of known header offsets, for format facts execution can't observe (e.g. the PE ASLR `DllCharacteristics` bit). Per decision: escalate a single check to an LLVM tool only if the `od` math gets genuinely hairy — a deliberate, flagged, CI-only call — and **never** a mach-native reader.
- **`flat-loader` producer** — build the `os=freestanding`, `of=raw` flat image and run it via a tiny C loader (`mmap` + jump); capture its output as the observable.

Seed the matching pins from the removed fixtures:
- `int/regression/` (or a freestanding surface case): the freestanding flat-image probe.
- format-fact pins: macho PIE/rebase shape and PE/ASLR (`DYNAMIC_BASE`), via `field`.

## Acceptance
- Both producers plug into the existing core with no structural change to `run.sh`.
- The freestanding and format-fact pins are green in CI on their declared targets.
- Dependency surface unchanged: mach + OS + coreutils; no LLVM.

Issue #1758, opened 2026-06-29
epic: integration-test system under int/


octalide:
> ## Status (2026-06-29)
> Design converged. v1 implementation in the first child; the two non-execution producers deferred to the second.

## What

A deliberate integration-test system under `int/`, replacing the ad-hoc `test/<target>/verify.sh` fixtures removed in #1756/#1757. The goal is a set-and-forget framework where adding coverage is *data*, never a new CI lane.

Model:
- **Execution-first on the native GitHub runner matrix.** Each target runs on its matching runner (real silicon/OS); `qemu-user` only for `linux-riscv64`, the one target with no native runner. darwin/windows execute for real — no structural byte-poking required for the common case.
- **One golden-diff core, swappable producers.** Every case: build → produce a normalized text observable → diff a golden. Producers: `exec` (native / qemu), `field` (coreutils header reads — deferred), `flat-loader` (freestanding — deferred).
- **`int/targets.conf` is the single source of truth** for both the harness run-mode and the CI matrix (generated by `int/lib/ci-matrix.sh`), so adding a target is one row and touches no workflow.
- **A case is data:** `mach.toml` + `src/main.mach` + `expect.txt`, plus an optional line-based `case.conf` to deviate. Two buckets: `int/surface/` (the surface-area × target matrix) and `int/regression/` (one pin per historical finding).
- **Dependencies:** mach + the OS everywhere; coreutils for the structural residue; qemu CI-only, confined to riscv64. **No LLVM.**

## Children
- [ ] #1759 v1: harness + `targets.conf` + CI wiring + starter surface set + an exemplar regression pin
- [ ] #1760 structural (`field`) + freestanding (`flat-loader`) producers + their pins

## Sequencing

The first child stands up the framework and the execution matrix. The second adds the two non-exec producers purely additively — new producer functions plus `targets.conf` rows, no change to the harness core. That additivity is the design's main correctness claim: a new verification mode never forces a redesign.

octalide:
Complete — both children merged.

- **#1759** (PR #1761): the framework — golden-diff harness, `targets.conf` as the single source of truth, the data-driven per-PR CI matrix, the `exec` producer, the starter surface set + the #1750 regression pin.
[...]

PR #1761, opened 2026-06-29
int: v1 framework + CI wiring + starter surface set

### LINKED ISSUE #1759: int: v1 framework + CI wiring + starter surface set
Part of #1758.

## What

Stand up the `int/` framework and the execution-first starter set.

**Framework**
- `int/run.sh` — harness entrypoint: `--target <name>`, `--bless`, `--filter <glob>`, `<compiler>`. For each case in `int/{surface,regression}/`: read defaults (overridable by an optional line-based `case.conf`), and for each applicable target × profile, build the case with the from-source compiler, produce a normalized text observable, and `diff` it against `expect.txt` (or `--bless` to regenerate). Nonzero exit if any case fails.
- `int/targets.conf` — the central table `name  runner  run-mode  cadence`:
  - `linux  ubuntu-latest  native  pr`
  - `linux-arm64  ubuntu-24.04-arm  native  pr`
  - `linux-riscv64  ubuntu-latest  qemu  pr`
  - `windows  windows-latest  native  pr`
  - `darwin-x86_64  macos-13  native  nightly`
  - `darwin-arm64  macos-14  native  nightly`
- `int/lib/produce.sh` — the `exec` producer (`native` → run directly; `qemu` → `qemu-<arch>`). `field` / `flat-loader` are stubs that error "deferred to the next child".
- `int/lib/ci-matrix.sh <cadence>` — emit the GH Actions `include:` matrix JSON from `targets.conf`, filtered by cadence.
- `ci.yml` — one data-driven `integration` job: an `int-matrix` setup step runs `ci-matrix.sh pr` → the job matrixes over it, builds the compiler from source per leg, runs `bash int/run.sh --target <t> ./m`. Plus a nightly scheduled workflow (also `workflow_dispatch`) that runs `ci-matrix.sh nightly` for the darwin legs.

**Cases** (observable = **stdout golden**: the fixture computes and prints deterministically, exits 0; a miscompile diverges the printout or crashes)
- `int/surface/`: `sign-extension`, `aggregate-abi`, `float`, `control-flow`, `syscall-inline-asm` — each runs on every executable target, both `-O0` and `-O2`.
- `int/regression/`: one exemplar pin, `1750-i32-signext`, to establish the bucket + the pattern (a minimal negative-i32-at-the-call-boundary repro). Reference the removed fixtures (git history of `test/`) for assertion content.

## Acceptance
- `bash int/run.sh --target linux ./m` and `--target linux-riscv64 ./m` (under qemu) pass locally against a compiler built from current-dev source (NOT the installed seed).
- `bash int/lib/ci-matrix.sh pr` emits valid matrix JSON; `nightly` emits the darwin legs.
- The CI `integration` job is green on the per-PR legs (linux, linux-arm64, linux-riscv64, windows); the darwin legs pass via a manual `workflow_dispatch` of the nightly workflow on the branch.
- Adding a case is a new directory with no CI edit; adding a target is one `targets.conf` row.
- Dependency surface is mach + the OS + (qemu for riscv64 only). **No LLVM.**

octalide:
Part of #1758. Stands up the `int/` integration-test framework and the execution-first starter set.

## What

**Framework**
- `int/run.sh` — harness: `--target <name> [--bless] [--filter <glob>] <compiler>`. For each case under `int/{surface,regression}/` with a `mach.toml`: load defaults (overridable by an optional line-based `case.conf`), and for each applicable target × profile build the case, run the producer to get a normalized stdout observable, and `diff` it against `expect.txt` (`--bless` regenerates). Nonzero exit on any failure.
- `int/targets.conf` — the single source of truth (`name runner run-mode cadence`); both the harness run-mode and the CI matrix read it. Adding a target is one row.
- `int/lib/produce.sh` — the `exec` producer (`native` → run directly; `qemu` → `qemu-<arch>`). `field` / `flat-loader` are stubs deferred to #1760.
- `int/lib/ci-matrix.sh <cadence>` — emits the GH Actions matrix JSON from `targets.conf`, filtered by cadence.
- CI: one data-driven `integration` job in `ci.yml` (per-PR legs), plus `int-nightly.yml` (scheduled + `workflow_dispatch`) for the darwin legs.

**Cases** (observable = stdout golden; the fixture computes and prints deterministically, exits 0)
- `int/surface/`: `sign-extension`, `aggregate-abi`, `float`, `control-flow`, `syscall-inline-asm`.
- `int/regression/`: `1750-i32-signext` (the exemplar pin).

## Invariants held
- One golden-diff core; the producer is the only thing that varies.
- `targets.conf` drives both harness and CI; adding a target is one row, zero workflow edits.
- A normal case is 3 files (mach.toml + src/main.mach + expect.txt); `case.conf` only to deviate.
- Shared `expect.txt` across targets for exec (output is target-independent).
- No LLVM. Dependency surface = mach + OS + (qemu for riscv64 only).
- Portable bash (no bash-4-isms).

## Status
Draft — implementation per #1759. See PR thread for the local run output, ci-matrix JSON, and judgment calls.

Closes #1759
Refs #1758


octalide:
## Verification (draft — for review)

Built a from-source current-dev compiler (`mach build . -o m`, **not** the seed) and verified:

- `mach test .` → **430 passed, 0 failed** (src/ untouched).
- `bash int/run.sh --target linux ./m` → all 6 cases × {debug,release} **PASS**.
- `bash int/run.sh --target linux-riscv64 ./m` (qemu-riscv64) → all **PASS** against the *same* linux-blessed goldens — the cross-target proof (shared `expect.txt`).
- ci-matrix JSON, both cadences valid:
  - `pr`: `[{"target":"linux",...},{"target":"linux-arm64",...},{"target":"linux-riscv64",...,"runmode":"qemu"},{"target":"windows",...}]`
  - `nightly`: `[{"target":"darwin-x86_64",...},{"target":"darwin-aarch64",...}]`
- **CI is green on every PR int leg** (int matrix, linux, linux-arm64, linux-riscv64, windows) and all existing build/test/release lanes.
- Every case also compile-checks for all 6 targets (linux, linux-arm64, linux-riscv64, windows, darwin-x86_64, darwin-aarch64).

## Judgment calls / deviations from spec

1. **Print mechanism = integers/strings only; floats folded to ints.** Found a **riscv64 (lp64) backend bug**: a freshly-computed/returned `f64` passed through `printf`'s variadic pack miscompiles (debug: inline `x+y` → `0`; release: all float varargs → `0`). Integer varargs and direct float compute/pass/return are all correct, and x86_64 is unaffected. So the float case computes/passes/returns floats normally but folds each result to an exact integer before printing — the harness output path stays reliable on every target while still catching float miscompiles. **This is a separate, unfiled rv64 bug** (out of #1759 scope) — repro: `printlnf("{}", addf(2.5,1.25))` built `--target linux-riscv64`, run under qemu, diverges from x86_64.
2. **arm64 darwin target named `darwin-aarch64`, not `darwin-arm64`.** The compiler's own `mach.toml` (and `cd.yml`) call it `darwin-aarch64`; the nightly seed job cross-builds the *compiler*, which fails with `darwin-arm64` (`no target named`). Standardized int/ on `darwin-aarch64` so compiler-build, fixture-build, and run.sh all agree. (`darwin-x86_64` is consistent everywhere.)
3. **`syscall-inline-asm` exempts windows** (`case.conf: exempt: windows`) — windows has no syscall ABI (console write is a kernel32 import). It runs on linux ×3 + darwin ×2; the asm arms mirror `std.system.panic`.
4. **windows handling.** run.sh keys the exe suffix off the *host* (uname), resolves a compiler passed as `./m` to `./m.exe`, and builds to a case-relative `out/` path (the native windows compiler can't resolve an MSYS `/tmp` path). `int/.gitattributes` pins `eol=lf` (autocrlf was turning LF goldens into CRLF). CI invokes `./m` uniformly.

## Not yet validated: the nightly darwin legs

`gh workflow run int-nightly.yml --ref feat/1759-int` returns 404 — `workflow_dispatch` requires the workflow on the **default branch**, so the darwin legs can only be triggered once this lands on `dev`. Fixtures + compiler **cross-build** for both darwin targets locally (compile-only), but real macOS execution is unverified. Please dispatch `int-nightly.yml` after merge (or it runs on the daily schedule).

Issue #1778, opened 2026-06-30
link(elf): RELRO hardening for relocated pointers under --pie (PT_GNU_RELRO + runtime mprotect)


octalide:
Follow-on to #1727. The initial `--pie` ships a deliberate hardening tradeoff: the linker forces any segment holding a relocated absolute pointer (RK_ABS64) WRITABLE (`pie_seg_flags`), because static-PIE has no RELRO and the self-relocation startup must write those pointers at runtime. So relocated *constant* pointers — including vtables — lose read-only hardening under `--pie`. This is the proper fix.

## What
Restore RO hardening of relocated pointers under `--pie` via the standard RELRO mechanism (writable during relocation, read-only after):
- **linker (mach, src/lang/target/of/elf.mach):** route reloc-bearing read-only data into a `.data.rel.ro`-style section and emit a `PT_GNU_RELRO` program header covering it (rather than forcing the whole segment RW). Keep the data writable at load so the self-reloc can apply the RELATIVE relocs.
- **runtime (mach-std, src/runtime/linux):** after `_rt_relocate` applies the relocations, `mprotect` the `PT_GNU_RELRO` region read-only, before `main`. Gated on `$mach.build.pie` like the rest of the self-reloc runtime.

Cross-repo (mach + mach-std), same flow as #1727: mach-std change to dev -> dev->main release -> mach.lock bump in the mach PR.

## Acceptance
- Under `--pie`, relocated constant pointers (vtables, pointer-init `val` globals) are read-only after startup: a write to one faults (verified, e.g. via the int/ harness or a targeted probe).
- ASLR + correctness preserved: the relocated-pointer int guard from #1727 still passes; non-PIE byte-identical.
- `PT_GNU_RELRO` present and correctly sized under `--pie`; absent without it.

octalide:
Implementation split across the two repos (same flow as #1727):

- **Linker (mach):** briar-systems/mach#1811 — emits `PT_GNU_RELRO` over any read-only-natured segment that carries a base relocation (`.rodata` holding relocated constants). The segment is still mapped writable at load for the self-reloc; the header is the runtime's cue to re-protect it. Adds an `elf-relro` int structural guard + a writer unit test. Independent and CI-green against the current runtime.
[...]

Issue #1727, opened 2026-06-29
link(elf): emit PIE (ET_DYN) executables via the shared BaseReloc set (linux ASLR)


octalide:
Deferred follow-on of #1718, enabled by #1722's format-neutral `BaseReloc` collection.

The linker now collects a format-neutral set of in-image absolute pointers (the
`BaseReloc` set) for position-independent images; #1722 ships the Mach-O encoder. ELF
can consume the *same* set to emit PIE (`ET_DYN`) executables for **linux ASLR**:
encode each `BaseReloc` as an `R_*_RELATIVE` entry in `.rela.dyn`, set `ET_DYN`, add
`PT_DYNAMIC`/`DT_RELA*`.

## Caveat (more than an encoder)

mach's linux binaries are no-libc static (direct syscalls, no `ld.so`), so a static-PIE
needs a **self-relocation startup** — apply its own `R_*_RELATIVE` relocs before `main`.
That's a runtime/codegen addition beyond the writer.

Optional hardening — linux runs fixed-address (`ET_EXEC`) fine, so there is no forcing
function. Deferred until ASLR is wanted.

PR #1811, opened 2026-07-02
feat(link/elf): RELRO hardening for relocated read-only pointers under --pie

octalide:
## What

Restores read-only hardening of relocated constant pointers (vtables, `val` pointer globals) under `--pie` via the standard RELRO mechanism: writable during self-relocation, read-only after.

Under `--pie` the linker forces any segment holding a relocated absolute pointer (`RK_ABS64`) writable, because a no-libc static-PIE self-relocates its own `.rela.dyn` RELATIVE slots at startup and must write them. That left relocated *constants* — which live in read-only data (`.rodata`) — permanently writable, losing their hardening.

This PR (the **linker half**) emits a `PT_GNU_RELRO` program header spanning any read-only-natured segment that carries a base relocation. The segment is still mapped writable at load (so the self-reloc can apply its RELATIVE relocs); the `PT_GNU_RELRO` is the runtime's cue to `mprotect` it read-only before `main`.

- **Coverage:** a segment is a RELRO region iff its source permissions are read-only (no write, no execute) *and* it carries a base relocation. In mach's canonical layout that is exactly `.rodata` when it holds relocated constants. Executable code and genuinely-writable `.data`/`.bss` are excluded and stay writable.
- **Selection & sizing:** `pie_relro_seg` returns the single read-only reloc-bearing segment mach's layout produces (`.rodata`) — no multi-segment span. `PT_GNU_RELRO` starts at that segment's page-aligned `vaddr` with `p_memsz` **page-rounded** (DATA_SEGMENT_RELRO_END style), so the runtime re-protects whole pages with no arithmetic. This is safe because every merged kind starts on a fresh page (`assign_section_vaddrs`), so the rounded end reaches exactly to the next segment and never into it. Forced-writability (`pie_seg_flags`) and RELRO selection share one predicate (`seg_has_base_reloc`) so they can't drift.
- **Non-PIE and PIE-without-RO-relocs are unaffected:** the header is emitted only by the static-PIE writer, only when a read-only reloc-bearing segment exists.

## Cross-repo (part of #1778)

The **runtime half** lands separately in **mach-std** (`_rt_relocate` walks `PT_GNU_RELRO` and `mprotect`s it `PROT_READ` after applying the RELATIVE relocs, gated on `$mach.build.pie`). Same flow as #1727: mach-std change → dev → main release → `mach.lock` bump + exec-fault guard in a follow-up mach PR. This PR is self-contained and CI-green against the current runtime (the header is emitted; with the current runtime the region simply stays writable, so behavior is unchanged and correct).

mach-std PR: briar-systems/mach-std#335.

## Verification

`readelf -l` on a `--pie` probe carrying a `val` pointer, a constant pointer table, and a constant function-pointer table (all relocated constants in `.rodata`), plus a `var` pointer (writable `.data`):

```
  Type           Offset   VirtAddr           FileSiz  MemSiz   Flg Align
  LOAD           0x03d000 0x000000000043d000 0x000040 0x000040 RW  0x1000   # .data (var m_ptr)
  LOAD           0x03e000 0x000000000043e000 0x001e59 0x001e59 RW  0x1000   # .rodata (relocated constants, writable at load)
  LOAD           0x040000 0x0000000000440000 0x0001d0 0x0001d0 R   0x1000   # synthesized .rela.dyn/.dynamic
  DYNAMIC        0x040180 ...
  GNU_RELRO      0x03e000 0x000000000043e000 0x002000 0x002000 R   0x1       # page-rounded, covers the .rodata segment's pages
```

`p_memsz` is page-rounded (`0x1e59` → `0x2000`), ending exactly at the next segment's page (`0x440000`). Dynamic relocations confirm the split — the RELATIVE entries land in the RELRO'd `.rodata` range (`0x43e000`–`0x440000`), and the single `var` pointer's entry lands in the writable `.data` segment (`0x43d038`), correctly **not** covered.

- Program runs correctly under `--pie` (debug and release): `s=9012`.
- Self-host build reaches the **byte-identical fixpoint**.
- `mach test .`: 439 ok. The elf module's `emit_pie_exec:relro_covers_ro_reloc_segment` asserts exactly one page-rounded RELRO header over `.rodata`, a sibling **writable** reloc-bearing segment excluded, and the covered PT_LOAD emitted R+W.
- int: `elf-relro` (dedicated `relro` producer) passes on the x86_64/arm64/riscv64 ELF legs; `elf-pie` (e_type only) and `pie-reloc` unchanged.

## Acceptance mapping

- ✅ `PT_GNU_RELRO` present and correctly sized under `--pie`; absent for non-PIE and for `--pie` images whose only relocs are in writable `.data`.
- ⏳ *Relocated constants read-only after startup (write faults)* — delivered by the mach-std runtime half + lock bump follow-up.
- ✅ ASLR + correctness preserved; non-PIE byte-identical.


octalide:
Runtime half (the `_rt_relocate` mprotect): briar-systems/mach-std#335. This linker PR is independent and CI-green against the current runtime; the mach-std change lands + releases, then a `mach.lock` bump activates the runtime re-protection and adds the write-fault exec guard.

octalide:
Review changes addressed (pushed as `refactor(link/elf): tighten RELRO selection, sizing, and its int guard`, CI green on all 14 checks incl. int on x86_64/arm64/riscv64):

**7. Open-coded scan ×3 → shared predicate.** Extracted `seg_has_base_reloc`; both `pie_seg_flags` (forced writability) and `pie_relro_seg` (RELRO selection) key on it, so they can't drift. (The "reloc emission" loop maps each reloc to its segment rather than scanning per-segment, so it's a different operation and left as is.)

**8. min/max span → single segment.** Replaced `pie_relro_range` with `pie_relro_seg`, returning the one read-only reloc-bearing segment mach's layout produces (`.rodata`). No bounding box, so nothing can engulf an interposed writable segment.

**9. Layout contract.** Fixed to TEXT, DATA, RODATA, BSS (the `of.SK_*` order) in both `elf.mach` and `linker.mach:65`. Stated the "every merged kind starts on a fresh page → each owns its trailing page" invariant at `assign_section_vaddrs` as the contract RELRO depends on, and cross-referenced `EXEC_PAGE_SIZE` with the linker's OS page size (`LINUX_PAGE_SIZE` via `os_page_size`).

**10. Comment altitude.** The writer doc now frames the header as the runtime's *cue*; re-protection ships in mach-std (a freestanding/non-mach-std `--pie` carries the header but is never re-protected). Softened the `.data.rel.ro` claim: whole-`.rodata` RELRO means all constants are writable from load until re-protection — a deliberate simplification, a wider window than a real split.

**11. elf-pie relro assertion.** Dropped — `field_elf` reverts to `e_type` only, and the incidental `relro=1` claim (mach-std internals, not the fixture) is gone from elf-pie's golden and comment. RELRO is now owned by elf-relro via a dedicated `relro` producer.

**12. Test gaps.** The writer unit test now has three segments incl. a **writable** reloc-bearing one; it asserts exactly one PT_GNU_RELRO over `.rodata` (page-rounded size), the writable segment excluded, and the covered PT_LOAD emitted **R+W** (the forced-writable precondition). elf-relro now runs the `relro` producer on the **x86_64, arm64, and riscv64** ELF legs (mirrors the pie-reloc sibling).

Also: PT_GNU_RELRO `p_memsz` is now emitted **page-rounded** (DATA_SEGMENT_RELRO_END style) so the runtime does no arithmetic — see the mach-std reply for the matching gate.

Re-verified end-to-end (write to a relocated `val` faults; normal reads + `.data` writes succeed) on all three arches; self-host fixpoint holds; `mach test .` 439 ok.

Issue #1784, opened 2026-06-30
chore: remove the orphaned --color / ColorMode / config.color_enabled surface


octalide:
Follow-up to #1777.

## What
The ASCII-only framed diagnostic renderer (#1777) dropped the `color` argument from the CLI diagnostic flush, leaving `--color`, the `ColorMode` type, and `config.color_enabled` with **no readers**. `--color` now parses but no-ops. The renderer is permanently black-and-white ASCII by design (see #1773), so color is not returning.

Remove the dead surface cleanly:
- the `--color` flag + its help text,
- the `ColorMode` type,
- `config.color_enabled`,
- any now-unused plumbing that threaded color through.

## Acceptance
- No `--color` / `ColorMode` / `color_enabled` remains in the tree.
- `mach <cmd> --color` reports an unknown flag (the flag is gone from help), rather than silently no-opping.
- `mach test .` green; self-host fixpoint unaffected.

octalide:
Merged in #1808. One acceptance nuance, found in review: the "reports an unknown flag" criterion is structurally unmeetable today — mach has no unknown-flag rejection at all, so a leftover `--color <mode>` invocation now has its value token claimed as the project-path positional (e.g. `mach build --color always .` resolves the path to `always`). The flag surface itself is fully gone (help, docs, parser, is_value_flag). The missing general rejector is tracked as #1810 so the divergence is owned rather than lost.

PR #1817, opened 2026-07-02
feat(driver): `mach info targets` — derive the supported target-tuple matrix

### LINKED ISSUE #1806: driver: `mach info targets` — derive and print the supported target-tuple matrix from vtable capability declarations
## What

`mach info` prints the capability surface as four independent dimension lists
(`isa:`, `os:`, `abi:`, `of:`), which reads as "any combination composes." The
joint support facts say otherwise — the COFF writer's relocation mapping covers
x86_64 only, the linker's `apply_reloc` hooks are bound per-ISA, `emit_dyn_exec` is
nil for some formats — and none of that is visible until a build fails, one missing
piece per attempt, deep in codegen or link.

Scope decision for the support claim: **"supported" means this compiler binary can
compose and emit the tuple end-to-end** — classify, select, encode, relocate,
write, link. Standard-library and runtime availability is explicitly not part of
the claim: std coverage is a property of `mach-std`, not of the compiler, and
targets may substitute their own runtime or stdlib entirely (freestanding). Under
that scoping every fact in the claim lives inside the compiler and can be declared
or introspected.

Design: joint capabilities are self-described on the vtables that own them, and the
matrix is derived — never curated.

- `OfVTable` declares the ISAs its relocation mapping and writers cover (today:
  elf declares x86_64/aarch64/riscv64; coff declares x86_64; macho declares
  x86_64/aarch64; raw is isa-agnostic).
- `AbiVTable` declares its ISA (a calling convention is per-ISA by construction:
  sysv64/win64 → x86_64, aapcs64 → aarch64, lp64d → riscv64).
- Existing nil-checks stay authoritative for what they already express
  (`emit_dyn_exec` presence, per-ISA `apply_reloc` binding).

`mach info targets` enumerates registered combinations, filters by the joint
declarations, and prints the supported tuples. `select_of` reuses the same checks,
so a tuple whose dimensions all exist but whose joint cell is unfilled fails at
selection with the missing capability named (e.g. "coff does not cover aarch64
relocations"), instead of surfacing as an encoder or linker error.

## Acceptance

- `mach info targets` prints the derived tuple matrix; windows-aarch64 is absent
  today because coff declares x86_64 only.
- Freestanding/raw tuples appear for every ISA with an encoder, independent of any
  stdlib or runtime consideration.
- Selecting an uncovered tuple fails at selection and names the missing joint
  capability; per-dimension "not registered" errors are unchanged.
- Adding a capability declaration is the only step needed for a new joint cell to
  appear in the matrix (no curated list to update).

octalide:
Closes #1806.

## What

Adds `mach info targets`, which prints the **supported target-tuple matrix** —
one `<os>-<isa>` per line — for exactly the tuples this binary can compose and
emit end-to-end. The matrix is **derived** from the vtables, never curated.

```
linux-x86_64
linux-aarch64
linux-riscv64
darwin-x86_64
darwin-aarch64
windows-x86_64
freestanding-x86_64
freestanding-aarch64
freestanding-riscv64
```

`windows-aarch64` and `darwin-riscv64` are correctly absent (COFF covers x86_64
only; Mach-O covers x86_64/aarch64). Bare `mach info` is unchanged.

## Where the capability data comes from

The joint cells that the four independent capability lists hide are now
**self-declared on the vtables that own them**:

- `of.OfVTable` gains `covers_isas` / `covers_isa_count` / `covers_isa_any` and
  an `of.covers_isa` predicate — the ISAs a format's relocation mapping and
  writers cover. elf → x86_64/aarch64/riscv64, coff → x86_64, macho →
  x86_64/aarch64, raw → isa-agnostic.
- `abi.AbiVTable` gains `arch_id` and an `abi.covers_isa` predicate — a calling
  convention is per-ISA (sysv64/win64 → x86_64, aapcs64 → aarch64, lp64d →
  riscv64).

`target.select_of` enforces the same two cells at selection, so an uncovered
tuple fails at composition **naming the missing capability** instead of surfacing
deep in codegen or link. The matrix in `mach info targets` and the selection
check read the identical predicates, and adding a capability declaration to a new
vtable is the only step needed for a tuple to appear — nothing is curated.

## Sample selection error

```
$ mach build <a windows-aarch64 project>
error: object format 'coff' does not cover aarch64 relocations
```

## Verification

- `mach build .` reaches the byte-identical self-host fixpoint.
- `mach test .`: **442 passed, 0 failed**, including new tests for
  `of.covers_isa`, `abi.covers_isa`, and `select_of` rejecting an
  object-format-uncovered ISA and an abi/isa mismatch (each asserting the error
  names the capability).
- `mach info targets` output verified against the acceptance criteria;
  end-to-end selection diagnostic verified with a throwaway windows-aarch64
  project.

## Notes / out of scope

- Display is a line-oriented `<os>-<isa>` list (matching commit `46231078`'s root
  target naming and `mach info`'s script-stable contract), not a 2D grid. Happy
  to switch to a grid if preferred.
- Bare `mach info` (currently the full capability screen, not the installed
  build's own target) is left untouched per the issue comment.
- Pre-existing, unrelated: `help.mach`'s `global_flags()` still lists
  `--color <mode>`, orphaned after the color-surface removal (#1784/#1808) — left
  alone here; worth a separate cleanup.


octalide:
## Review round 1 addressed

Pushed 4 commits (`34a3779f`..`63a16fa8`). All 8 items:

1. **One shared predicate.** `target.tuple_capability(os, isa, abi, of) -> TUPLE_*` is the single joint check. `select_of` turns a non-OK code into a named diagnostic; `mach info targets` keeps a `<os>-<isa>` tuple only when some registered (abi, of) yields `TUPLE_OK`. The ISA-codegen check now lives **inside** the shared predicate, so a description-only ISA is rejected at selection too (no longer hidden-from-matrix-but-composes).

2. **(os, of) axis added.** `OsVTable` declares `supported_ofs` (linux→elf, darwin→macho, windows→coff, freestanding→raw+elf), read via `os.supports_of`. `windows + elf` now fails at selection (`operating system 'windows' does not support object format 'elf'`) instead of dying at link on `mainCRTStartup`. This also **resolves the inconsistency you flagged**: the matrix and the emit-check now agree — `windows-aarch64` is absent both ways. Note this makes `windows + raw` a *rejected* pair (COFF-only OS contract); bare-metal-with-win64-ABI is `os = freestanding`, which is where a flat image belongs. Flag if you'd rather windows advertise raw.

3. **Union tolerates uncoverable targets.** `populate_union_tuples` skips a non-composable declared target with a `warning:` on the session sink and keeps the rest; `mach build <tuple>` stays strict. New test `union_skips_uncoverable_target` (coverable + windows-aarch64 → count 1, no abort).

4. **Bounded coverage.** Replaced hand-set counts with `of.cover_isa` / `os.support_of` append helpers that `panic` past capacity — no more silent write/read past the fixed arrays, and the count can't drift from the entries.

5. **Loud lookup.** `print_targets` iterates the object-format registry through the shared predicate; it no longer resolves an OS's `default_of` and so can't silently skip an OS on a missed lookup.

6. **Verbatim naming.** The abi-mismatch diagnostic dropped `arch_name_for(abi.arch_id)`; every dimension is spelled with the vtable's own `.name` (`calling convention 'aapcs64' does not target instruction set 'x86_64'`).

7. **Doc drift.** Dropped the hand-written tuple snapshot from `doc/cli.md`; it now describes the derivation and points at the command, and the prose matches the completed joint model.

8. **freestanding-aarch64/riscv64 coverage.** Not adding an int case — the int runner always builds **and runs** a producer against a golden, and there is no x86_64-host way to run an aarch64/riscv64 image (the `freestanding` case is `run: flat-loader`). Instead the claim is covered by composition: `freestanding-<isa>` = the `<isa>` isel+encode (run end-to-end by the `int linux-arm64` / `int linux-riscv64` legs) + the raw flat-image writer (run end-to-end by `int freestanding-x86_64`) + freestanding entry-at-base (isa-independent). Every component is exercised by an existing int **run** leg; only the specific freestanding+aarch64/riscv64 *combination* isn't, because it needs cross-execution (qemu) the freestanding case lacks. The compose+emit path itself is unit-tested via `select_of`. Happy to add a build-only int producer if you'd prefer explicit coverage.

**Verification (settled tree):** byte-identical self-host fixpoint holds; `mach test .` = **450 passed / 0 failed** (added `has_codegen`/`supports_of` paths, the os-of and union rejection tests); `mach info targets` output unchanged (9 tuples); both new selection diagnostics confirmed end-to-end. CI running.

PR #1850, opened 2026-07-03
incremental: bind the back half (Q_LOWER / Q_CODEGEN / Q_LINK) into the query engine

### LINKED ISSUE #1618: incremental: bind the back half (Q_LOWER / Q_CODEGEN / Q_LINK) into the query engine
The incremental front end is done. Parse (#1580), resolve (#1582), and sema (#1583) all
run as memoized queries. The back half is still outside the engine: `Q_LOWER`,
`Q_CODEGEN`, and `Q_LINK` are reserved kinds in `query.mach` but unbound, so lowering,
codegen, and link re-run unconditionally on every build and cannot be cached or
parallelized across modules.

Bind them so a module whose typed surface and lowered IR are unchanged reuses its prior
`.o` instead of re-lowering and re-emitting. First slice: per-module hash-and-skip on the
codegen output (reuse an on-disk `.o` when a module's lowered inputs are unchanged), which
is also the prerequisite for parallel codegen.

This is the back-half / execution-scale axis the retargeting epic #1600 surfaced and
explicitly handed here. It is not target authorship. Builds on the now-landed stable
cross-build identity (#1579), which keeps a cached `.o` valid across a graph reshape.

Part of #1164.

octalide:
Binds the back half of the pipeline into the incremental query engine so lowering, codegen, and link cache and parallelize per module instead of re-running every build (#1618). Keyed on `StableModuleId`; the CLI's cold-cache path is behaviour-preserving (always recomputes/links), and the reuse is proven on a long-lived session by driver tests.

## What landed
- **Catalog doc-drift fix** — `Q_LOWER`/`Q_CODEGEN`/`Q_LINK` key on `StableModuleId`.
- **`Q_LOWER`** — owned-handle optimized `ir.Module`. Deps: `Q_SEMA(module)`, `Q_TARGET`, and `Q_SEMA` of the module's **transitive import closure** (a generic/comptime instance from another module is monomorphized into this module's IR; no per-hop lowering firewall, so the full transitive closure is required for soundness). Driver `run_lower_pass`; lower+optimize collapse into one cached unit.
- **`Q_CODEGEN`** — owned-handle `of.ObjectImage`. Deps: `Q_LOWER(module)`, `Q_TARGET`. `--emit-asm` renders via a separate deterministic codegen pass (assembly is a streaming side-channel not in the cached image), so the common path stays single-pass.
- **`Q_LINK`** — derived link fingerprint keyed on unit. Folds the revision of every link input: `Q_TARGET` (target/profile/pie), the new **`Q_LINK_CONFIG`** input (link-only config — output path, product, output kind, external libs/objects/sonames), and `Q_CODEGEN` of every module. `run_link_pass` peeks the prior revision, demands the entry, and returns "re-link needed?"; the CLI (both the normal and flat-image link paths) reuses the prior artifact when unchanged. New `pub query.peek_revision`.
- The CLI routes lower/codegen/link through the driver passes; IR and images are query-owned, borrowed via shallow view arrays; `free_modules` retired.

- Byte-identical self-host fixpoint.
- Full unit suite: **480 passed / 0 failed** (475 baseline + 5 new incremental tests).
- int/ golden-diff harness: all cases pass on the linux leg.
- Cross-target: `linux-arm64` and `windows-x86_64` builds succeed and are byte-deterministic.
- Incremental behavior (driver tests, `last_changed`-probe style): no-change rebuild reuses lower/codegen/link; a single-module edit re-lowers/re-codegens/re-links only what it must; a link-only config change (output path) with every object reused **still** re-links (the `Q_LINK` fingerprint-completeness pin).

## Notes
- `--pie` already lives in `Q_TARGET` (it forks compilation via `$mach.build.pie`), so a `--pie` toggle re-codegens every module and re-links through the `Q_CODEGEN` deps; it is also folded into `Q_LINK_CONFIG` for completeness. The genuine link-only axis the completeness test exercises is the output path.

## Follow-ups (surfaced, being filed by the lead — not in this PR)
- Test-mode `main`-neutralization mutates the query-owned IR in place — safe on the CLI's single-shot session, latent only for a long-lived test-driving session.
- No lowered-surface fingerprint firewall yet: an impl-only edit to a dependency re-lowers importers even when their inlined slice is unchanged — the future twin of `Q_TYPED_EXPORTS`.

Closes #1618

### DIFF
--- a/src/cli/cmd/build.mach
+++ b/src/cli/cmd/build.mach
@@ -1503,6 +1503,42 @@ fun link_artifact(p: *driver.Project, emit_obj: bool, root: *u8,
         ret 2;
     }
 
+    # record the link inputs into Q_LINK_CONFIG and consult Q_LINK: when no link
+    # input changed and the prior artifact is present, reuse it instead of
+    # re-linking. the external link inputs are paths[topo_len..len) (paths[0..topo_len)
+    # are the per-module objects Q_CODEGEN already tracks). a cold-cache CLI build
+    # always re-links (Q_LINK is a fresh entry), so this is behaviour-preserving for
+    # a one-shot build and only skips on a warm session.
+    val ext_from:  u32  = p.topo_len;
+    val ext_count: u32  = len - p.topo_len;
+    val cfg_r: R.Result[bool, str] = driver.set_link_config_input(p, ?artifact[0], artifact_product_name(p), linker.LINK_EXE::u32,
+                                                                  ?paths[ext_from], ext_count, sonames, soname_len);
+    if (R.is_err[bool, str](cfg_r)) {
+        free_paths(p.s.alloc, paths, len, len);
+        free_sonames(p.s.alloc, sonames, soname_len);
+        print.eprint("error: ");
+        print.eprint(R.unwrap_err[bool, str](cfg_r));
+        print.eprint("\n");
+        ret 2;
+    }
+    val relink_r: R.Result[bool, str] = driver.run_link_pass(p);
+    if (R.is_err[bool, str](relink_r)) {
+        free_paths(p.s.alloc, paths, len, len);
+        free_sonames(p.s.alloc, sonames, soname_len);
+        print.eprint("error: ");
+        print.eprint(R.unwrap_err[bool, str](relink_r));
+        print.eprint("\n");
+        ret 2;
+    }
+    if (!R.unwrap_ok[bool, str](relink_r) && filesystem.is_file((?artifact[0])::str)) {
+        free_paths(p.s.alloc, paths, len, len);
+        free_sonames(p.s.alloc, sonames, soname_len);
+        @out_path = dup_cstr(p.s.alloc, ?artifact[0]);
+        readout.item(p.s.readout, readout.PH_LINK, (?artifact[0])::str, t_link);
+        readout.phase_end(p.s.readout, readout.PH_LINK, 1, "binary", "binaries");
+        ret 0;
+    }
+
     val res_link: R.Result[bool, str] =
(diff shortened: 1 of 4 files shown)


octalide:
Review round 1 addressed (commit `0cbeac3c`), rebased onto current dev, full CI green (all build/int/release/test lanes incl. linux-riscv64).

**1 (MAJOR) — external inputs fingerprinted by path only:** fixed. Static external link inputs (`.o`/`.a`, baked into the output) are now folded by path **and** by a sha256 of their content (`fp_file_content`), so a same-path rebuild of a vendored archive invalidates `Q_LINK_CONFIG` and re-links. Mirrors the front half's source-file fingerprint (which stores full text; the link hashes since an archive can be large). Sonames stay identity-only — they are `DT_NEEDED` runtime deps, never baked into the binary, so their content can't change this output.
**2 (MAJOR) — external axis untested:** fixed. New test `incremental_link_external_content_fingerprint` varies an external input's content at the same path (pins fix 1), its path, and adds a soname — each must re-link; a same-config rebuild reuses.
**3 (minor) — wrong tracker in comment:** fixed → #1858.
**4 (minor) — 2× shard scans in the link fold:** fixed. `record_dep_rev` records a dep and returns the revision it captured, so `q_link_compute` no longer peeks each dep a second time.
**5 (minor) — verify_ir not a dep:** fixed. New `Q_LOWER_FLAGS` input carries the IR-verifier toggle; `Q_LOWER` depends on it, so a warm session re-lowers (and re-runs the verifier) when it's toggled on. Scoped to lowering — a toggle does not re-resolve/re-sema.
**6 (minor) — build-4 probes one module:** fixed. Now probes `a`, `b`, and `main` codegen stability across the link-only change.

Verification on the rebased head: byte-identical self-host fixpoint; unit suite 487/0; int all legs; arm64+windows cross-builds deterministic; CI green.

PR #1855, opened 2026-07-03
feat(link): split reloc-bearing read-only data into a .data.rel.ro segment kind

### LINKED ISSUE #1844: link(elf): split reloc-bearing read-only data into a .data.rel.ro segment kind
Follow-up to #1778, targeting the post-2.13.0 cycle. First of the RELRO end-state trio
(with the max-page-size layout issue and the mach-std fatal-reprotect issue).

## What

The shipped RELRO maps the entire merged `.rodata` writable during self-relocation because
mach's 5-kind section model has no home for "read-only data that carries base relocations."
Pure constants (string literals, lookup tables) are therefore writable from process entry
until `_rt_relocate` re-protects — a startup-only window, but wider than the conventional
layout, and hardening audit tools expect the split.

Add the sixth merged kind: `.data.rel.ro`. Classification happens where globals are assigned
sections: a read-only global whose initializer carries a base relocation (an address) goes to
the rel.ro kind; everything else stays pure `.rodata`. The writer then maps pure `.rodata`
read-only from load (no forced `PF_W`), maps only the rel.ro segment writable, and emits
`PT_GNU_RELRO` over exactly it. The runtime half is unchanged — the header remains the cue
and the region is page-rounded by the writer as today.

## Acceptance

- `readelf -l` on a `--pie` binary: pure `.rodata`'s `PT_LOAD` is `R` (never `RW`); the
  rel.ro segment is `RW` at load with `PT_GNU_RELRO` covering exactly it.
- A write to a string literal faults from process START (extend the int write-fault guard
  with a pure-constant case), and the existing `pie-relro-fault` case still passes.
- Non-PIE and freestanding layouts unaffected; self-host fixpoint byte-identical modulo the
  new section placement; full suite + int matrix green.

octalide:
Closes #1844.

## What

Adds the sixth merged section kind, `SK_RELRO` (`.data.rel.ro`): read-only data that carries base relocations (relocated constant pointers - vtables, `val` pointer globals, `str` literals).

- **codegen** (`be/codegen.mach`): a read-only global whose aggregate initializer carries a relocation side-table (`VAL_CONST_AGG` with `reloc_count > 0`) is routed to `.data.rel.ro`; every other read-only constant (scalar, byte blob, null) stays pure `.rodata`.
- **linker** (`be/linker.mach`): a sixth merged kind, ordered `TEXT, DATA, RODATA, RELRO, BSS, DEBUG` (rel.ro before bss so the freestanding flat image keeps its zero-fill trailing). It becomes its own page-aligned, read-only-natured load segment.
- **ELF PIE writer** (`of/elf.mach`): no logic change needed — `seg_has_base_reloc` / `pie_relro_seg` / `pie_seg_flags` already key on "the read-only-natured segment carrying a base relocation," which is now exactly `.data.rel.ro`. Pure `.rodata` carries no base relocation, so it stays mapped read-only from load; only `.data.rel.ro` is mapped writable during self-relocation and covered by `PT_GNU_RELRO`. The ELF parser re-derives the kind from the `.data.rel.ro` section name.
- **other formats** (`of/macho.mach`, `of/coff.mach`, raw): Mach-O and COFF have no RELRO; their loaders handle reloc-bearing read-only data natively, so `SK_RELRO` maps to read-only const data equivalent to `.rodata` (Mach-O coalesces it into `__const`; COFF emits read-only `.rdata`-style characteristics). The raw flat writer is permission-driven and needs no kind change.

## Acceptance status

- [ ] `readelf -l` on `--pie`: pure `.rodata` PT_LOAD is `R`; `.data.rel.ro` is `RW` at load with `PT_GNU_RELRO` over exactly it
- [ ] write to a string literal faults from process start (int write-fault guard extended); existing `pie-relro-fault` still passes
- [ ] non-PIE / freestanding unaffected; self-host fixpoint byte-identical modulo new section placement; full suite + int matrix green


octalide:

**Self-host fixpoint** — byte-identical: stage2 == stage3 (7,397,376 bytes). stage1 (built by the 2.13.0 seed, no split) is +4096 = exactly one page smaller, consistent with the new page-aligned `.data.rel.ro` segment.

**`readelf -l` on a `--pie` compiler binary** — the end state the issue asks for:
```
LOAD  0x00ad5000 0x00ad5000 R    filesz 0x38cd9   # pure .rodata: R only (was RW before)
LOAD  0x00b0e000 0x00b0e000 RW   filesz 0x170     # .data.rel.ro: RW at load
LOAD  0x00b0f000 0x00b0f000 RW   memsz  0x2c98    # .bss
GNU_RELRO 0x00b0e000 0x00b0e000 R  memsz 0x1000   # covers exactly .data.rel.ro
```
The 232 KB pure `.rodata` is now mapped `R` from load; only the 368-byte rel.ro region is transiently writable and covered by `PT_GNU_RELRO`. The `--pie` compiler runs and rebuilds a binary byte-identical to the non-pie stage2 (relocations apply, RELRO region re-protects cleanly).

**Unit suite** — 475 passed, 0 failed (object-format tests: elf 20, coff 20, macho 3, raw 3).

**int matrix** — `int: all cases passed on linux` (incl. macho-pie x86_64/aarch64, freestanding x3, elf-pie, elf-relro, pie-reloc, pie-relro-fault, pie-rodata-fault). riscv64 under qemu: all RELRO cases pass. arm64 RELRO verified via direct `qemu-aarch64` (`read=1001` then SIGSEGV); the arm64 leg runs native in CI (`targets.conf`), not qemu.

**Regression caught & fixed in-branch**: the macho writer folds `.data.rel.ro` into the read-only `__const` bucket, but the relocation-write loop initially missed the `coalesce_kind` normalization — counted the relocs but didn't write them, truncating the array. Fixed in `00b70fee`; macho-pie now builds on both arches.

Issue #1742, opened 2026-06-29
test/infra: riscv64 self-host investigation (corpus qemu lane now provided by int/)


octalide:
Surfaced by the #1180 foundation audit — the structural fix for the riscv64 latent-bug cluster (#1733/#1735/#1737 all hid behind thin test coverage).

## What
riscv64 is exercised only by one ~246-line qemu exit-code fixture — no self-host, no corpus, no opt2 — which is exactly why float/edge bugs stayed latent until run. **mach-std now ships a riscv64 runtime** (`dep/mach-std/src/runtime/linux/riscv64.mach`, `system/os/linux/riscv64.mach`), so the `mach.toml:23` / ci comments asserting "no riscv64 linux runtime yet" are **stale** and a qemu self-host/corpus lane is feasible.

## Fix
Attempt a riscv64 cross-build of the compiler + run the test corpus under qemu-riscv64 (and ideally a self-host fixpoint). Update the stale mach.toml/ci comments either way. This graduates riscv64 from cross-only and closes the whole cluster's root cause.

Part of #1180.

octalide:
Re-scoping. The **corpus / qemu-lane** half is now provided by the **int/ system** — the riscv64 per-PR int leg cross-builds + runs the suite under qemu-riscv64, and the one-off corpus proof landed **57→0** after #1750/#1753/#1762. The stale `mach.toml`/ci "no riscv64 runtime yet" comments were also corrected during the test/→int/ migration.

What **remains** here is narrower: now that mach-std ships a riscv64 runtime, does the compiler riscv64 **self-host** (build `--target linux-riscv64` + run the compiler under qemu to compile something)? That's the open question this issue now tracks — an investigation, not the lane (the lane exists).

octalide:
## riscv64 self-host investigation — findings

### TL;DR
Self-host is **not viable today.** The compiler cross-builds for `linux-riscv64` and the resulting binary runs under `qemu-riscv64` for trivial inputs, but the riscv64-hosted compiler **aborts in register allocation** — `regalloc: GP vreg assigned a non-GP register` (SIGTRAP, exit 133) — as soon as it compiles any function with control flow that keeps a value live across a basic-block boundary. This is a **riscv64 backend miscompilation of the compiler's own regalloc/liveness code**: host-dependent, target-independent, and opt-level-independent.

### Setup
- From-source current-dev compiler (2.13.0), native x86_64: `mach build . -o m-native`.
[...]

Issue #1762, opened 2026-06-29
codegen(riscv64): f64 through a native variadic pack (va.../$each) miscompiles to 0 (lp64)


octalide:
On riscv64 (lp64), a freshly-computed or returned `f64` passed through a **native** comptime variadic pack (`va: ...` / `$each`, e.g. `std.print.printlnf`) miscompiles to `0`. This is the language's own variadic mechanism, not C variadic FFI — so it is a real backend bug, not the known C-variadic-ABI limitation.

## Repro

```mach
use p: std.print;
fun addf(x: f64, y: f64) f64 { ret x + y; }

# in some entrypoint:
p.printlnf("{}", addf(2.5, 1.25));   # passes an f64 through the native variadic pack
```

Build `--target linux-riscv64`, run under `qemu-riscv64`:
- **debug:** an inline freshly-computed arg (`addf(...)` / `x + y`) prints `0`; binding it to a temp `var` first prints correctly.
- **release (-O2):** all float varargs print `0`.

Integer varargs and direct (non-variadic) float compute / pass / return are all correct. x86_64 and aarch64 are unaffected.

## Expected

The `f64` flows through the native variadic pack and prints its value (`3.75`), identically to x86_64/aarch64.

## Actual

Prints `0`.

## Note on cause

RISC-V lp64 passes variadic (unnamed) arguments — including floats — in the **integer** registers / on the stack, never the FP registers (the lp64 variadic rule). The pattern here (an FP-register-resident inline value fails; a temp-var'd value that gets stored/reloaded works in debug; everything fails once -O2 keeps values in FP registers) points to the variadic-argument lowering leaving / reading a freshly-computed `f64` on the FP side instead of moving it to the integer side for the pack. Same lp64-ABI-mishandling family as #1750 (32-bit sign-extension) and #1733/#1737 (FP load/store + spill).

## Environment

- Triple: `linux-riscv64` (lp64d), cross-built then run under `qemu-riscv64`.
- Toolchain: current `dev` (post-#1754/#1755).
- Surfaced by: building the `int/` integration harness (#1759). The `int/surface/float` fixture currently folds each result to an exact integer before printing specifically to avoid this path; once fixed, a pin that passes a freshly-computed `f64` through the variadic pack should guard it.

## Acceptance

- `printlnf("{}", addf(2.5, 1.25))` cross-built for `linux-riscv64` prints the correct value under qemu in **both** debug and release.
[...]

PR #1857, opened 2026-07-03
fix(codegen): materialize phi immediate merge moves at logical width

### LINKED ISSUE #1851: codegen: phi immediate merge moves materialize at full register width, breaking riscv64's sign-extension invariant
Blocks #1742.

The riscv64 backend miscompiles the compiler's own cross-block liveness code: a riscv64-hosted `mach` (cross-built from current dev, run under qemu-riscv64) aborts with `regalloc: GP vreg assigned a non-GP register` (SIGTRAP, exit 133) whenever it compiles a function in which a GP value is live across a basic-block boundary. Straight-line functions compile fine, including with heavy spill pressure (40 live locals); any `if`/`for` carrying a value across a block edge panics. Full investigation record: https://github.com/briar-systems/mach/issues/1742#issuecomment-4871619460

## Repro

On an x86_64 linux host, dev @ 96358fe8, from-source compiler:

1. Cross-build the compiler: `mach build . --target linux-riscv64` (call the artifact `mach-rv64`).
2. Minimal input, as a freestanding-riscv64 build:

   ```mach
   fun start(x: i64) i64 { if (x > 0) { ret x; } ret 0; }
   ```

3. Compile it with the hosted compiler under emulation: `qemu-riscv64 mach-rv64 build ...`

## Expected

Exit 0 and a working artifact — the native-hosted compiler compiles the identical input successfully (exit 0).

## Actual

`regalloc: GP vreg assigned a non-GP register`, SIGTRAP, exit 133.

## Characterization

- **Host-dependent**: same source, same flags; only the executing host differs → `mach-rv64` itself is miscompiled. The input's target is irrelevant.
- **Target-independent**: the hosted compiler panics emitting x86_64, aarch64, and riscv64 alike → the corrupted code lies in shared regalloc/liveness (`extend_range_for_block` / `block_spans` / the live-in/out dataflow in `src/lang/be/codegen/regalloc.mach`), not in riscv64-target codegen paths.
- **Opt-independent**: repros at -O0 and --release.
- Trigger is cross-block liveness, not register pressure.
- Same shape as #1762 (riscv64 isel dropped a same-bank float bitcast): a riscv64 isel/encode defect that only manifests once the produced binary is the compiler itself. The fix belongs in the riscv64 backend — restructuring regalloc source to dodge the construct is a workaround and out of the question.
- Coverage gap (tracked separately): the unit suite cross-compiled to riscv64 passes 475/475 under qemu — it exercises regalloc helpers, never the full allocator over a multi-block function; the int/ riscv64 leg runs produced binaries only, so the compiler never executes on riscv64 in CI.

## Environment

- dev @ 96358fe8, from-source compiler
- qemu-riscv64 (user-mode) on an x86_64 linux host

## Acceptance

- The minimal repro compiles cleanly under qemu-riscv64 and its artifact matches the native-hosted compiler's.
- The #1742 self-host chain proceeds past regalloc (report how far it gets; further distinct failures are new issues, not this one).
- A regression unit test drives the full allocator over a multi-block function, runnable under the existing `--target linux-riscv64 --runner qemu-riscv64` unit lane.

octalide:
Fixes the riscv64 backend miscompiling the compiler's own cross-block liveness code, which blocked riscv64 self-host (#1742): a riscv64-hosted `mach` aborted in its own register allocator with `regalloc: GP vreg assigned a non-GP register` whenever it compiled a function carrying a value across a basic-block boundary.

## Root cause
Out-of-SSA phi destruction (`src/lang/be/codegen/mir/lower.mach`) spliced every phi merge copy at full register width (`gpr_width`). That is correct for a register/symbol source — a bit copy of a value already in its target-canonical register form is width-agnostic — but wrong for an **immediate** source, which is a *materialization* of a constant. The encoder canonicalizes an immediate by its operand width (riscv64 sign-extends a negative i32 to the lp64 "i32 held sign-extended in a 64-bit register" convention via `imm_sext_for_width`, which only re-signs width 4). A full-width (8) move left a negative i32 phi constant un-sign-extended, and RISC-V's 64-bit `slt` then read it as a large positive.

Concretely, `regalloc.mach`'s allocation loop does `var preg: i32 = -1; if (hint) { preg = pick_hint(...); } if (preg < 0) { preg = pick_gp(...); }`. The `-1` initializer flows through a phi; on riscv64 `preg < 0` read false, so `pick_gp` was skipped and the `-1` no-register sentinel reached `assign`, which rejected it as a bad register class.

IR / riscv64 asm of the minimal repro:
```
%23 = phi i32 [bb1 %11, bb2 4294967295]
cmp_lt_s i8 %23, 0
```
```
mv a1, 4294967295      # -1 materialized as 0x00000000FFFFFFFF (not sign-extended)
cmp.lt_s a0, a1, 0     # 64-bit slt reads +4294967295 -> false
```
The riscv64 encoder is **correct**; the shared phi-lowering fed it the wrong width. x86-64 was masked because its i32 signed compare selects a width-aware 32-bit instruction. Full investigation: #1851 (issuecomment-4871748567).

## Fix
Carry each phi merge copy's logical width (`op_width_of(phi.ty)`) and apply it to **immediate sources only**; register/symbol copies stay full-width. The immediate/register split is semantic: a register copy transfers an already-canonical value, an immediate source materializes a constant that needs its logical width to be canonicalized.

## Sibling-site sweep
Swept every MIR site that materializes an immediate into a register. All other width-sensitive sites (call-argument setup, scalar returns, spilled register pieces) already carry the value's logical width via `emit_mov_w` / `scalar_reg_move_width`; address materializations correctly stay full-width; immediate stores are safe (truncate + width-aware reload). The phi merge move was the sole defect of this class. Details in the sweep comment below.

## Tests
Two produced-code regression tests in `regalloc.mach`, runnable on every lane including the qemu riscv64 unit lane:
- `cross_block_liveness:1851` — a multi-block function carrying GP values across a loop back edge and an if/else, with a signed compare of a loop-carried negative i32.
- `phi_neg_i32_signed_compare:1851` — the minimal shape; a loop-derived `probe` keeps the branch/phi from folding to -O2.

Both **fail** when built for riscv64 by a pre-fix compiler (run under qemu) and **pass** post-fix; both pass natively — i.e. validated to actually catch the defect.

## Verification
- x86_64 self-host fixpoint: stage2 == stage3 byte-identical (no host regression).
- Full unit suite: 477 passed native, 477 passed riscv64-under-qemu.
- int matrix: linux and linux-riscv64 legs all pass (arm64/windows/darwin run on their CI runners); arm64 codegen smoke-tested locally under qemu-aarch64.
- #1742 self-host chain attempt under qemu: reported in a comment.

Closes #1851

### DIFF
--- a/int/regression/1851-rv64-phi-signext/expect.txt
+++ b/int/regression/1851-rv64-phi-signext/expect.txt
@@ -1,0 +1,2 @@
+v_lt_0=true
+acc=48
--- a/int/regression/1851-rv64-phi-signext/mach.toml
+++ b/int/regression/1851-rv64-phi-signext/mach.toml
@@ -1,0 +1,52 @@
+[project]
+id      = "case"
+name    = "case"
+version = "0.0.0"
+src     = "src"
+dep     = "dep"
+target  = "linux"
+out     = "out/{target}/{profile}/bin/{name}{ext}"
+
+[target.linux]
+isa = "x86_64"
+os  = "linux"
+abi = "sysv64"
+
+[target.linux-arm64]
+isa = "aarch64"
+os  = "linux"
+abi = "aapcs64"
+
+[target.linux-riscv64]
+isa = "riscv64"
+os  = "linux"
+abi = "lp64"
+
+[target.windows]
+isa = "x86_64"
+os  = "windows"
+abi = "win64"
+ext = ".exe"
+
+[target.darwin-x86_64]
+isa = "x86_64"
+os  = "darwin"
+abi = "sysv64"
+
+[target.darwin-aarch64]
+isa = "aarch64"
+os  = "darwin"
+abi = "aapcs64"
+
+[profile.debug]
+opt = 0
+
+[profile.release]
+opt = 2
+
+[bin.case]
+entry = "main.mach"
+
+[deps.mach-std]
+git = "https://github.com/briar-systems/mach-std"
+ref = "branch/main"
--- a/int/regression/1851-rv64-phi-signext/src/main.mach
+++ b/int/regression/1851-rv64-phi-signext/src/main.mach
@@ -1,0 +1,36 @@
+# regression pin for #1851: a negative i32 constant reaching a signed compare
+# through a phi (a branch merge). the phi's constant edge must materialize at the
+# value's logical width so riscv64 sign-extends it to the lp64 register convention;
+# a full-register merge move left it un-sign-extended and the 64-bit `slt` read -1 as
+# a large positive, so `v < 0` was false and a loop-carried negative i32 took the
+# wrong arm. x86_64 / aarch64 stay correct (width-aware 32-bit compare). the loop-
+# derived `probe` keeps the branch and its phi from folding at -O2.
+
+use std.runtime;
+use std.types.size.usize;
+use p: std.print;
+
+#[symbol("main")]
+fun main(argc: usize, argv: **u8) i64 {
+    var probe: i32 = 0;
+    var i:     i32 = 0;
+    for (i < 5) { probe = probe + i; i = i + 1; }
+
+    var v: i32 = -1;
+    if (probe > 100) { v = probe; }
+    if (v < 0) { p.println("v_lt_0=true"); } or { p.println("v_lt_0=false"); }
+
+    # a loop-carried negative i32 across a back edge and an if/else.
+    var a:   i32 = 3;
+    var b:   i32 = -7;
+    var acc: i32 = 0;
+    var j:   i32 = 0;
+    for (j < 6) {
+        if (b < 0) { acc = acc + a; } or { acc = acc - a; }
+        b   = b + 1;
+        a   = a + 2;
+        j   = j + 1;
+    }
+    p.printlnf("acc={}", acc);
+    ret 0;
+}
--- a/src/lang/be/codegen/mir/lower.mach
+++ b/src/lang/be/codegen/mir/lower.mach
@@ -318,6 +318,15 @@ fun lower_block_phis(ctx: *lctx.LowerCtx, fn: *ir.Function, blk: *ir.Block) R.Re
         ret R.err[bool, str](R.unwrap_err[*mir.MirOperand, str](rs));
     }
     val srcs: *mir.MirOperand = R.unwrap_ok[*mir.MirOperand, str](rs);
+    # per-copy logical width, parallel to dsts / srcs: the phi result's byte width,
+    # so an immediate incoming value materializes at the constant's true width.
+    val rw: R.Result[*u8, str] = A.allocate[u8](ctx.alloc, np::usize);
+    if (R.is_err[*u8, str](rw)) {
+        A.deallocate[u32](ctx.alloc, dsts, np::usize);
+        A.deallocate[mir.MirOperand](ctx.alloc, srcs, np::usize);
+        ret R.err[bool, str](R.unwrap_err[*u8, str](rw));
+    }
+    val wids: *u8 = R.unwrap_ok[*u8, str](rw);
 
     var err_msg: str = nil;
     # iterate the distinct predecessor block ids of `S`. preds were copied from the
@@ -369,6 +378,7 @@ fun lower_block_phis(ctx: *lctx.LowerCtx, fn: *ir.Function, blk: *ir.Block) R.Re
                 if (phi.operands[k].data.int_lit::u32 == pred_id) {
                     dsts[nc] = dst_vreg;
                     srcs[nc] = lctx.lower_value(ctx, phi.operands[k + 1]);
+                    wids[nc] = lctx.op_width_of(ctx, phi.ty);
                     nc = nc + 1;
                     found = true;
                 }
@@ -378,7 +388,7 @@ fun lower_block_phis(ctx: *lctx.LowerCtx, fn: *ir.Function, blk: *ir.Block) R.Re
         }
 
         if (err_msg == nil && nc > 0) {
-            val r: R.Result[bool, str] = emit_edge_copies(ctx, pred_mb, dsts, srcs, nc);
+            val r: R.Result[bool, str] = emit_edge_copies(ctx, pred_mb, dsts, srcs, wids, nc);
             if (R.is_err[bool, str](r)) { err_msg = R.unwrap_err[bool, str](r); }
         }
         pi = pi + 1;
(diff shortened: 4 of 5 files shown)


octalide:
## Sibling-site sweep — swept, all other sites width-correct

Audited every MIR site that materializes an immediate into a register (the "immediate materialized at gpr_width instead of logical width" defect class), across the shared lowering and the riscv64 encoder.

**Already width-correct (carry the value's logical width via `emit_mov_w` / `scalar_reg_move_width`):**
- Call-argument scalar setup — `abi.mach:1111` (`emit_arg_slot`): `emit_mov_w(..., scalar_reg_move_width(size))`, 4 for i32 else gpr_width. Its comment already documents the RV64 lp64 sign-extension convention.
- Scalar return-value materialization — `abi.mach:1474`: same `scalar_reg_move_width` mechanism.
- Spilled register-piece moves — `abi.mach:1015`: `emit_mov_w(..., p.width)`.

**Safe by construction (no width-based re-sign needed):**
- Immediate stores — `context.mach` `emit_store_imm` / `emit_store_to*`, encoder `emit_store_mem`: the store truncates to the store width (`sb/sh/sw/sd`) and the reload sign/zero-extends (`lb/lh/lw`), so the low bytes are always correct.
- Address materializations — GEP/ALLOCA bases (`context.mach` GEP emitters; `lower.mach:2197` `fold_gep_mem`): a compile-time address is a full 64-bit pointer and correctly stays gpr_width; narrowing would truncate it.
- Boolean / count immediates — the constant `cbr` condition (`lower.mach:888`, a 0/1 bool) and the SysV vector-count `MOV AL, fp_used` (`abi.mach:876`, a small non-negative count) are width-agnostic.
- ALU immediate folds — `encode.mach` `encode_alu` / `encode_shift`: fold only small immediates into the width-aware `addiw`/`slliw` (etc.) form, else route through `resolve_source` → `imm_sext_for_width`.

**The one defect (fixed):** the out-of-SSA phi merge move (`lower.mach` `insert_mov_before_terminator`) was the sole site hardcoding `gpr_width` for an immediate source. Now carries the phi's logical width for immediate sources; register/symbol copies stay full-width.

Note: `imm_sext_for_width` re-signs only width 4; sub-32-bit (i8/i16) integers are held zero-extended in registers on this backend (loads use `lbu`/`lhu`), so their signed handling is a separate concern and unchanged by this fix (the phi move now passes the true logical width, which for i8/i16 leaves behavior identical to before). Not in scope for #1851; flagged, not touched.

octalide:
## Verification — full results

- **x86_64 self-host fixpoint**: stage2 == stage3 byte-identical. No host regression.
- **Unit suite**: 477 passed / 0 failed native; 477 passed / 0 failed cross-compiled to riscv64 and run under qemu.
- **int matrix**: `linux` and `linux-riscv64` legs all cases pass (incl. the existing `surface/sign-extension` and `regression/1750-i32-signext`, `regression/1762-rv64-float-vararg`). arm64/windows/darwin run on their own CI runners; arm64 codegen additionally smoke-tested locally under qemu-aarch64 (correct).
- **Regression tests**: both new tests fail when built for riscv64 by a pre-fix compiler (under qemu) and pass post-fix; both pass natively.

## #1742 self-host chain — RESOLVED, chain completes

The fixed riscv64 compiler (cross-built on x86_64) was run under `qemu-riscv64` to compile the **entire compiler source**:

```
qemu-riscv64 ./mach-rv64  build .  -o rv_selfhost_out      # exit 0
```

- The riscv64-hosted compiler compiled the whole compiler → **exit 0** (previously aborted in regalloc at exit 133).
- `rv_selfhost_out` (riscv64→riscv64) is **byte-identical** to the x86_64→riscv64 build of the same source: the compiler running on riscv64 produces the same bytes as the compiler running on x86_64 → **riscv64 self-host fixpoint**.
- The self-hosted riscv64 compiler then correctly compiled the original #1851 repro under qemu (exit 0).

No new distinct failure appeared past regalloc; the self-host chain now runs to completion. This unblocks #1742.

octalide:
## Review response — both findings addressed

### Finding 1 (CI-lane gap) — fixed
Added `int/regression/1851-rv64-phi-signext`, a produced-code golden-diff case the existing qemu riscv64 int leg runs under emulation (the shape: a negative i32 constant reaching a signed compare through a phi, plus a loop-carried negative i32 across block edges). The in-source unit tests are kept for native coverage.

Falsifiability validated via the harness (not just locally):
- `int/run.sh --target linux-riscv64` with a **pre-fix** compiler (released seed) → **FAIL**, golden diff `v_lt_0=false` / `acc=42` vs expected `v_lt_0=true` / `acc=48`.
- Same, with the branch compiler → **PASS**; `--target linux` (x86_64) → **PASS**.

### Finding 2 (sub-word partial-register hazard) — clamp applied
`imm_merge_width` now clamps the phi-immediate materialization width: i32 stays width 4 (the fix), every other width goes to `gpr_width`, mirroring `mir.abi.scalar_reg_move_width` (cited in the comment). Confirmed at the asm level: an i8 phi's constant edges now emit `mov rax, 255` / `mov rax, 5` (full register defined) instead of the prior `mov al, ...`.

On the requested narrow-phi regression test: **I could not construct a falsifiable one**, and I think that's a substantive finding worth surfacing rather than papering over with a test that passes both ways. I swept, on x86-64, the un-clamped branch vs the released seed across: widening casts (`i8/i16/u8 -> i64`), signed/unsigned sub-word compares, sub-word arithmetic, a cross-function i8 argument, phi-of-phi, ~20-live-i8 register pressure (forced spills), array store/reload, and variadic `printlnf` formatting (the #1762 path). Every case matched — mach re-narrows every sub-word read (movsx/movzx, byte-ops, and the already width-aware `cbr`), so the stale upper bits from `mov al` are never observed. Corroborating: the **entire un-clamped compiler** self-compiled **byte-identically** and passed the full unit suite + int + the riscv64 self-host chain, so the hazard isn't triggered in the compiler itself either.

So the clamp is the correct **canonical-form** fix (restores the fully-defined-register invariant the pre-PR width-8 move gave, and that `scalar_reg_move_width` maintains for args/returns), not a fix for an active observable miscompile. I did not add a non-falsifiable narrow-phi test since it would give false regression confidence. Happy to add one purely as path-coverage if you'd prefer — say the word.

### Re-verification (clamped compiler)
- x86_64 self-host fixpoint: stage2 == stage3 byte-identical (re-checked; the clamp changes emitted bytes but still converges).
- Unit suite: 477 passed native, 477 passed riscv64-under-qemu.
- int: `linux` and `linux-riscv64` legs all pass, including the new `1851-rv64-phi-signext`.

Issue #1691, opened 2026-06-28
debuginfo: MirInstr loc + encoder PC<->line row table


octalide:
## What
Add a loc field to MirInstr (be/codegen/mir.mach:343-350), forward it through mir lowering, and have the encoder emit a NEW retained (text_offset, file, line) row array in EncoderOutput (mirroring SymbolMark/PendingReloc in be/codegen/encode.mach). Wire it into insert_text relaxation and codegen pack_image re-homing (site_placed/rebased_text_offset) so rows track the final bytes.

## Why
Byte offsets exist only transiently inside encode_function; block offsets are freed by scratch_release and never reach output. A line table (and CodeView line info) needs a retained PC->source map that stays correct under branch relaxation and #[section(...)] re-homing. No PC-tracking infra is missing — only per-instruction granularity and retention. Offset->line/col reuses the existing source.position machinery.

## Acceptance
- MirInstr carries loc; EncoderOutput exposes ordered (text_offset, file, line) rows per function.
- Rows remain correct after a forced insert_text branch relaxation and a #[section(...)] function; offsets match final .text (unit test).

Depends-on: #1689. Foundation; reused by all three primes.

Issue #1690, opened 2026-06-28
debuginfo: central IR mutation API (replace/erase/clone) + pass migration


octalide:
## What
Introduce replace_value / erase_instruction / clone_instruction in me/ir/builder.mach and migrate the five passes (cse, mem2reg, dce, constfold, algebraic) onto them. Loc (and later OP_DBG_VALUE) maintenance becomes an invariant of the primitive instead of five hand-rolled copies. Behavior-preserving.

## Why
No central mutation API exists; each pass rolls its own substitution — cse's Subst/apply_substitutions, mem2reg's subst/rename (pass/mem2reg.mach:994), dce's dead[] worklist. Debug-value survival across optimization is structurally unsupportable until mutation is centralized. The maintenance data already exists: cse's subst[old->new] table IS the RAUW map; mem2reg's per-alloca reaching-def rename IS the variable-location source.

## Acceptance
- The three primitives exist and preserve/merge loc on rewrite and clone; nothing is dropped silently.
- All five passes use them; no pass keeps a private substitution/free path.
- Loc survives a mem2reg->constfold->dce run on a sample function (verified by IR dump); corpus/regression output byte-unchanged.

Depends-on: #1689. Foundation; reused by all three primes.

octalide:
## Implementation notes — corrected model + erase reframing

Two premises in the issue body needed adjustment against the live tree (implemented in PR #1889, plan approved):

**1. "All five passes use them" → each pass routes its mutations through the primitives, using the subset that applies.** cse/algebraic/constfold/mem2reg do value replacement (the shared `Rewrite` RAUW map); inline does clone + a single immediate replacement (`clone_instruction` + `replace_uses`); dce and mem2reg-strip do erasure (`erase_instruction`). dce has no substitution of its own — that's expected, not a gap. Per the epic's C2 note, inline's `replace_value_uses` and `clone_instructions` were absorbed too, though inline isn't in the body's five-pass list.

[...]

Issue #1688, opened 2026-06-28
debuginfo: shared source-location + debug-model foundation


octalide:
## What
The format- and ISA-agnostic substrate every debug-info emitter (DWARF for ELF/Mach-O, CodeView for COFF) consumes. Repairs the source-location chain severed at the AST->IR seam and builds the shared backend debug model: a loc field on IR, central mutation primitives, per-instruction PC<->(file,line) rows, the OP_DBG_VALUE variable-location records the optimizer maintains, the DebugInfo vtable seam, the three missing abstract reloc kinds (RK_ABS32, RK_SECREL32, RK_SECTION) and their per-format/per-ISA maps, the per-ISA DWARF/CodeView register-number tables, linker carry of non-allocated debug sections, the -g flag, and a CI verification lane.

## Why
Spans die at lowering today: IR Instruction/Value carry no location and the builder emit_* API takes no span. No central RAUW/erase primitive exists, so loc and dbg-values cannot survive the optimizer. The DebugInfo vtable #1594 calls 'preserved unchanged' does NOT exist (only SK_DEBUG=4 does), and the abstract reloc set has no unsigned RK_ABS32 and no RK_SECREL32/RK_SECTION at all — so DWARF cross-section refs and CodeView code references are physically unencodable. DWARF register numbers are ISA-specific and unowned, silently breaking arm64/riscv64 locals. All three format emitters block on this layer, and there is no -O0 escape hatch (default `mach build` runs mem2reg->constfold->dce), so value tracking is mandatory.

## Acceptance
- IR carries a (FileId, offset) loc that survives mem2reg/dce/cse/constfold/algebraic.
- The encoder exposes a retained PC<->(file,line) row table.
- OP_DBG_VALUE + the mutation primitives land; DWARF and CodeView reuse them unchanged.
- The DebugInfo vtable seam + RK_ABS32/RK_SECREL32/RK_SECTION + per-ISA register tables + SK_DEBUG-by-name linker carry exist.
- `mach build -g` emits valid relocatable .o debug sections behind the seam on every shipped ISA; the CI verification lane (incl. the Windows runner) is green on a fixture.

Sub-issues: #1689, #1690, #1691, #1692, #1693, #1694, #1695, #1696, #1697, #1698.

Out of scope (deferred-with-note): bin/raw (flat sectionless formats — debug info is structurally impossible), ar (debug travels transparently inside member .o files), and the freestanding OS (no debugger).

Depends-on: nothing (this is the foundation layer).

[...]

Issue #1694, opened 2026-06-28
debuginfo: linker carries/merges non-allocated debug sections into executables


octalide:
## What
In be/linker.mach: stop excluding SK_DEBUG from executables — key the debug merge by section NAME (not the single kind-collapsed .debug), so .debug_info/.debug_line/.debug_abbrev/.debug_str survive distinctly; apply debug relocations to final vaddrs (the compiler is its own linker); and extend the ExecFn/DynExecFn seam in of.mach (and emit_exec/emit_dyn_exec) to carry the non-loaded, relocation-patched debug sections to the writers. SK_DEBUG stays non-SHF_ALLOC/non-loaded.

## Why
Today is_loadable_kind excludes SK_DEBUG and build_load_segments collapses every debug input into one kind-keyed section, so debug bytes never reach an executable — the common debugging case. The exec writers receive only LoadSegment[] + ExecFunction[], so there is no channel for debug bytes. This is the only seam-signature change the milestone forces.

## Acceptance
- A linked executable contains distinct, relocation-resolved .debug_* / __debug_* sections (non-alloc).
- Loadable layout unchanged; the relocatable .o path needs no of/ writer edits; Mach-O ad-hoc code-signature still covers the file and `codesign --verify` passes.

Depends-on: #1692. Foundation; reused by the DWARF and CodeView exec paths.

---

## Status (foundation split)

- **ELF** (seam + name-keyed linker merge + static/PIE/dynamic emission): landed in #1882 (merged).
- **Mach-O + COFF** exec emission: tracked in #1887, implemented in #1894. Mach-O carries the debug sections in a non-mapped `__DWARF` segment below the ad-hoc code signature (so `code_limit` covers them); COFF carries them as discardable PE sections, with long names via a COFF string table. Native `codesign --verify` / dyld-load (macOS) and PE-loader (Windows) verification is gated on the #1698 runners (checklist posted on #1698); structure round-trips in per-format unit tests.

octalide:
Complete across all formats: linker carry + ELF emission in PR #1882, Mach-O (__DWARF under code_limit) + COFF (discardable sections, /offset string table) in PR #1894 (which closed #1887). Runtime verification checklist for the native runners is on #1698; the darwin main-merge lane exercises Mach-O natively at the next main integration.

PR #1883, opened 2026-07-03
feat(debuginfo): source location on IR instructions + builder loc cursor

### LINKED ISSUE #1689: debuginfo: source location on IR instructions + builder loc cursor
## What
Add a (FileId, offset) loc field to the IR Instruction record (me/ir/instruction.mach:175-183). Give Builder an LLVM-IRBuilder-style current-debug-location cursor alongside its existing fn/block cursor (builder.mach:67-87), stamped by the terminal constructor pool_append (builder.mach:1032). Set the cursor once per statement/expr in me/lower/{expr,stmt,decl}.mach from each AST node's .span + the already-available ctx.a.file_id — NO span argument added to the ~40 emit_* functions.

## Why
The location chain is severed here: lowering reads .span for diagnostics then discards it on every emit. A module spans multiple files, so a bare token.Span is insufficient and len is irrelevant for line tables — (FileId, offset) is the minimum. This single additive seam fix unlocks line tables, variable locations, and CodeView.

## Acceptance
- Instruction has a loc field; pool_append stamps it from the builder cursor.
- me/lower/* set the cursor; emitted instructions carry correct file+offset, and source.position (source.mach:294) round-trips it to line/col.
- IR printer/verify updated; output byte-identical without -g.

Depends-on: nothing. Foundation; reused by all three primes.

octalide:
Closes #1689. First landable slice of the debug-info foundation epic #1688.

## What

- **`source.SrcLoc`** — a shared `(FileId, offset)` location primitive with a `FILE_NIL` sentinel and `loc` / `loc_nil` / `loc_is_valid` helpers. Format- and ISA-agnostic; the later MIR loc + PC↔line tables (#1691) reuse it.
- **`Instruction.loc`** — an additive `SrcLoc` field on the IR instruction record. Pure debug metadata: never read by codegen, `FILE_NIL` for compiler-synthesized instructions.
- **Builder debug-location cursor** — a `loc` cursor on `Builder`, an LLVM-IRBuilder-style `set_loc` / `current_loc`, reset to `FILE_NIL` on `set_function`, and stamped onto every instruction in `pool_append`. No span argument is added to any `emit_*`.
- **Lowering sets the cursor** — `me/lower/{stmt,expr,decl}` point the cursor at each statement / expression / function from the AST node's `.span` and `ctx.a.file_id`. `stmt`/`expr` save and `fin`-restore the cursor so a subexpression never leaks its location to the enclosing operation (correct per-operation attribution).
- **Pass hygiene** — inline's cloned instructions faithfully carry their source loc; genuinely synthesized nodes (inline continuation br / result phi, mem2reg SSA-join phis) are `FILE_NIL`. (Loc *survival* through the optimizer is #1690/#1696; this only avoids leaving the new field at a misleading default.)

## Invariant: byte-identical without `-g`

There is no `-g` surface yet, so the entire change must be output-preserving. The IR printer/verifier need no change — `loc` is inert metadata with no textual surface, so IR dumps stay byte-identical.

Verification (dev-baseline compiler vs this branch, both from source via the release seed, compiling the identical pristine dev tree):

- **Byte-identical emitted output on all six shipped targets** — linux-x86_64, linux-arm64, linux-riscv64, windows-x86_64, darwin-x86_64, darwin-aarch64 all `cmp`-identical.
- **Self-host fixpoint converges** (b == c) on the branch.
- **Unit suite: 503 passed, 0 failed** (dev baseline 501; +2 are the two tests added here: `me.ir.builder` cursor-stamping and `source.srcloc` round-trip).
- **int harness green** on the linux leg (all surface + regression cases, debug and release).

### DIFF
--- a/src/lang/me/ir/builder.mach
+++ b/src/lang/me/ir/builder.mach
@@ -21,6 +21,7 @@
 # second terminator emission is rejected with an error rather than silently
 # overwriting the first.
 
+use std.allocator.page;
 use std.types.bool.bool;
 use std.types.bool.true;
 use std.types.size.usize;
@@ -29,21 +30,27 @@ use std.types.string.str;
 use A: std.allocator;
 use R: std.types.result;
 
+use mach.lang.intern;
 use mach.lang.me.ir;
 use mach.lang.me.ir.id;
 use mach.lang.me.ir.instruction;
 use mach.lang.me.ir.type;
 use mach.lang.me.ir.value;
+use mach.lang.source;
 
 # an insertion cursor over an IR module
 # ---
 # m:     the module being built (owns instruction-pool growth via its functions)
 # fn:    the function currently being emitted into (nil until set_function)
 # block: the block new instructions land in (BLOCK_NIL until set_block)
+# loc:   current debug-location cursor stamped onto every emitted instruction;
+#        FILE_NIL until set_loc, and reset to FILE_NIL on each set_function so a
+#        new function's prologue never inherits a stale location
 pub rec Builder {
     m:     *ir.Module;
     fn:    *ir.Function;
     block: id.BlockId;
+    loc:   source.SrcLoc;
 }
 
 # construct a builder positioned on a module with no current function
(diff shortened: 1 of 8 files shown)


octalide:
Ready for review. Rebased onto current dev (`1a98d686`); no file overlap with the sibling epic work that landed (`#1693` register tables live under `target/isa/*`).

**CI note — the one red check (`int linux-arm64`) is pre-existing and not caused by this PR.** It fails only on `surface/pie-reloc` and `surface/pie-relro-fault` with `std.runtime.linux.reloc: mprotect of the PT_GNU_RELRO region failed`, an environment-level RELRO-hardening fault on the arm64 runner. The **same four cases fail identically on `dev` itself** (run 28662033489) and on every sibling branch (`#1692`, `#1693`, `#1694`, `chore/bump-std-pins`). This PR's emitted output is **byte-identical to the dev-baseline compiler on all six shipped targets including linux-arm64**, so the produced binary is the same one dev produces — the failure is independent of this change. Every check that this change can affect is green (all builds, all `test` legs, `int` linux/riscv64/windows, all `release` legs).

PR #1884, opened 2026-07-03
fix(link): require #[library] attribution for dynamic imports on two-level-namespace formats

### LINKED ISSUE #1800: link(coff): unattributed dynamic imports silently bind to dependency 0 — require attribution on two-level-namespace formats
Depends on briar-systems/mach-std#334 (mach-std must pin its kernel32 imports before the
default can flip, or every windows link of a mach-std program breaks). Related: #1799
(pin auto-registration, the natural companion), #1382 (introduced `#[library]`), #1159.

## What

The format-neutral `DynamicInfo` model is ELF-first. An `ext` import with no `#[library]`
attribution is recorded as `DYNLIB_ANY` — "the binding is left to the loader's global
search" (`src/lang/target/of.mach`). On ELF that is a real concept: the dynamic symbol
table records no provider and the runtime loader resolves through the flat global
namespace; the ELF writer never reads `lib_index` at all.

PE has no such concept. Its namespace is two-level — every import physically names a
providing DLL in the import directory. So the COFF writer needed *some* answer for the
sentinel, and it binds an unattributed import to **dependency index 0** (`import_lib_of`,
`src/lang/target/of/coff.mach`: "the first dependency, which on Windows is the canonical
system import library"). That comment encodes an assumption, not a design: it holds only
while `kernel32.dll` happens to be the first collected DLL, which the manifest cascade
("topological dep then declaration order") guarantees for nobody.

#1382 added `#[library]` as an *override* for the exceptions (ws2_32, the synch apiset —
mach-std#253), leaving the fallback as the default. The result is an emergent invariant:
mach-std's ~56 unpinned kernel32 imports require kernel32.dll at slot 0, and any other
library's unpinned imports require *their* DLL at slot 0. The first mixed-DLL program
(mach-std + octalide/mach-glfw's `glfw3.dll`) makes both claims at once: whichever set
loses the ordering mis-binds, the link succeeds silently, and the loader rejects the exe
with `STATUS_ENTRYPOINT_NOT_FOUND` on a machine that may not be the build host. Confirmed
by inspection of `build_dynamic_info` (`src/lang/be/linker.mach`), `import_lib_of`, and
both libraries' declarations; wine reproduces the load failure.

The same cliff is waiting for Mach-O: dyld's namespace is two-level by default, so when
Mach-O dynamic imports land (#1176), `DYNLIB_ANY` will be equally meaningless there.

## Fix

Preferred design: on a two-level-namespace format, an unattributed dynamic import is a
**hard link error**. The diagnostic names the symbol, lists the link's DLL dependencies,
and points at `#[library]` (comptime-conditional declarations cover the per-OS split, as
mach-std already does with its per-OS source tree). ELF keeps `DYNLIB_ANY` unchanged,
because there it is true. The check belongs at the same seam that builds the import plan
— gate it on whether the selected object format consumes attribution, so Mach-O inherits
the right behavior for free when #1176 lands.

Candidates considered and rejected:

- Keep a fallback only when the link has exactly one DLL dependency (unambiguous), error
  on two or more. Still an implicit contract — adding a dependency to a working project
  flips its meaning — and the single-DLL case barely exists in practice since mach-std
  contributes DLLs to every windows link.
- Resolve attribution at link time by probing each DLL's export table. Rejected outright:
  it requires Windows artifacts on the build host, destroying the verbatim-record/no-probe
  property that makes cross-compiling from linux work with zero Windows SDK material.
  That property is correct and must be preserved.

Rollout: land briar-systems/mach-std#334 first (pin the kernel32 set), then flip the
default here. #1799 (auto-registration) is independent but complements this — together
they make the declaration the single source of truth.

## Acceptance

- Linking a PE image with an unattributed dynamic import fails with a diagnostic that
  names the symbol and the candidate dependencies and suggests `#[library]`.
- After mach-std#334, mach-std windows programs (including the self-hosted compiler and
  the native test suite) link and run unchanged.
- A regression fixture linking two DLLs with imports attributed to each loads and runs
  under the wine/CI lane regardless of dependency order.
- ELF links are byte-identical before and after the change.

octalide:
Closes #1800. Depends on mach-std#334 (merged; std pin bumped on dev in #1879).

## What

An `ext` import with no `#[library]` attribution reached the import plan as `DYNLIB_ANY`. On ELF that is a real concept — the loader's flat global namespace resolves it. On PE it is not: the two-level import directory names a providing DLL per import, so the COFF writer had to invent an answer for the sentinel and bound it to **dependency index 0** (`import_lib_of`). That is correct only while the providing DLL happens to sort first; the first mixed-DLL link (mach-std + a windowing dep's `glfw3.dll`) silently mis-binds one set and the loader rejects the exe with `STATUS_ENTRYPOINT_NOT_FOUND` — off the build host.

## Fix

Attribution becomes mandatory at the import-plan seam (`build_dynamic_info`) when the object format's namespace is two-level. An unattributed dynamic import is a hard link error naming the symbol, listing the link's dependency libraries, and pointing at `#[library]`:

```
error: dynamic import 'some_unattributed_import' has no #[library] attribution; a two-level-namespace target requires every import to name its providing library — annotate the declaration with #[library("...")]; candidate libraries: kernel32.dll
```

The gate is a self-reported `OfVTable.two_level_namespace` capability (COFF + Mach-O true; ELF + raw false) rather than a branch on format id, so Mach-O inherits the rule for free when its dynamic imports land (#1176). ELF keeps `DYNLIB_ANY` unchanged.

## Verification

- **Self-host fixpoint**: gen2 == gen3 byte-identical.
- **Unit suite**: 501 passed, 0 failed — count-neutral vs pristine dev (501).
- **ELF byte-identical**: a fixed linux program built by the pristine-dev compiler and by this branch produces `cmp`-identical output (the ELF path is untouched).
- **int matrix (linux leg)**: all cases pass, including every ELF case (elf-pie, elf-relro, pie-reloc, pie-relro-fault), the macho-pie cross-builds (flag inert today, no regression), pe-aslr, and the new negative case. The windows/arm64/riscv64/darwin legs run in CI.
- **Negative test (falsifiable)**: `int/surface/pe-import-attribution` cross-builds a win64 PE with one unattributed import on the linux leg and diffs the link diagnostic. The pre-change linker binds it silently, so the case fails on baseline and passes here.
- **Positive**: pinning the same import links, and it lands under the named DLL even with that DLL placed last in `libs`.

## Harness

Adds a `build-fails` run-mode to the int harness: it asserts the compile is rejected and takes the compiler's `error:` line as the golden observable (deterministic — symbol + candidate libraries, no paths). Documented in `run.sh` and `produce.sh`.

### DIFF
--- a/int/lib/produce.sh
+++ b/int/lib/produce.sh
@@ -20,6 +20,10 @@
 #                 for a cross-built target with no host runner (a freestanding
 #                 aarch64/riscv64 image on the x86_64 leg). the observable is a
 #                 constant, so the golden is the fact "it emitted".
+#
+# build-fails is a run-mode but not a producer: it asserts the compile is REJECTED
+# and takes the compiler's 'error:' diagnostic as the observable. it is handled in
+# run.sh (there is no artifact to run), noted here for discoverability.
 
 # the directory this file lives in (int/lib), used to find flat_loader.c. resolved
 # from the sourced path so it does not depend on run.sh's variables.
--- a/int/run.sh
+++ b/int/run.sh
@@ -196,25 +196,52 @@ for dir in "$here"/surface/$filter "$here"/regression/$filter; do
             fi
         fi
 
-        if ! (cd "$dir" && "$compiler" dep pull && $buildcc build . --target "$build_target" --profile "$profile" $case_build_flags -o "$relbin") >"$tmp/build.log" 2>&1; then
-            echo "FAIL $label (build)"
-            sed 's/^/    /' "$tmp/build.log" >&2
-            fails=$((fails + 1))
-            rm -rf "$tmp" "$dir/out/int"
-            continue
+        if (cd "$dir" && "$compiler" dep pull && $buildcc build . --target "$build_target" --profile "$profile" $case_build_flags -o "$relbin") >"$tmp/build.log" 2>&1; then
+            build_ok=1
+        else
+            build_ok=0
         fi
 
-        if produce "$case_run" "$runmode" "$target" "$bin" >"$tmp/out.txt" 2>"$tmp/err.txt"; then
-            prc=0
+        # a build-fails case asserts the compile is rejected and takes the compiler's
+        # 'error:' diagnostic as its observable (deterministic: no paths, just the
+        # message text). every other mode requires a clean build and runs a producer
+        # on the artifact.
+        if [ "$case_run" = build-fails ]; then
+            if [ "$build_ok" -eq 1 ]; then
+                echo "FAIL $label (build succeeded; expected a link error)"
+                fails=$((fails + 1))
+                rm -rf "$tmp" "$dir/out/int"
+                continue
+            fi
+            grep '^error:' "$tmp/build.log" >"$tmp/out.txt" || true
+            if [ ! -s "$tmp/out.txt" ]; then
+                echo "FAIL $label (build failed without an 'error:' diagnostic)"
+                sed 's/^/    /' "$tmp/build.log" >&2
+                fails=$((fails + 1))
+                rm -rf "$tmp" "$dir/out/int"
+                continue
+            fi
         else
-            prc=$?
-        fi
-        if [ "$prc" -ne 0 ]; then
-            echo "FAIL $label (producer exit $prc)"
-            sed 's/^/    /' "$tmp/err.txt" >&2
-            fails=$((fails + 1))
-            rm -rf "$tmp" "$dir/out/int"
-            continue
+            if [ "$build_ok" -eq 0 ]; then
+                echo "FAIL $label (build)"
+                sed 's/^/    /' "$tmp/build.log" >&2
+                fails=$((fails + 1))
+                rm -rf "$tmp" "$dir/out/int"
+                continue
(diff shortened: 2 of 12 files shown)


octalide:
CI: every check passes except **`int linux-arm64`**, which is a **pre-existing failure on `dev`, not introduced by this branch**.

The failing cases are `surface/pie-reloc` and `surface/pie-relro-fault` (`std.runtime.linux.reloc: mprotect of the PT_GNU_RELRO region failed`, `relro_write=exit133`) — an aarch64 RELRO/mprotect runtime issue. The identical four failures appear on a pristine `dev` PR run at `873450ee` ([run 28662033489](https://github.com/briar-systems/mach/actions/runs/28662033489)), before this work existed.

This change is format-gated on `two_level_namespace` (false for ELF), so it cannot affect RELRO/PIE on any ELF target. Relevant to this PR: `int windows` passes (the fully-pinned std links and runs on a native windows runner with the new hard-error compiler), and `int linux` / `int linux-riscv64` pass (ELF paths unchanged). The arm64 RELRO breakage should be tracked and fixed separately.

Issue #1845, opened 2026-07-02
link(elf): target-parameterized max page size so aarch64 images load and harden on 16K/64K-page kernels


octalide:
Follow-up to #1778, targeting the post-2.13.0 cycle. Second of the RELRO end-state trio;
the mach-std fatal-reprotect issue depends on this one.

## What

The ELF writer lays segments out on a hardcoded 4 KiB page (`EXEC_PAGE_SIZE`, cross-referenced
with `LINUX_PAGE_SIZE`). aarch64 kernels can run 16 KiB or 64 KiB pages; there a 4 KiB-aligned
image either fails the kernel's `p_offset ≡ p_vaddr (mod pagesize)` exec check (binary is
unloadable) or, if tolerated, forces the RELRO runtime to skip re-protection (its
page-congruence gate — the documented large-page degradation).

Make the layout page a per-target parameter — the conventional max-page-size: 65536 for
aarch64-linux (lld's default), 4096 for x86_64 and riscv64. One home: the target supplies it,
`assign_section_vaddrs` aligns every merged kind with it, the writer's `p_align` and the
header-reservation math consume the same value (unifying the two hardcoded constants). With
segments aligned to max-page ≥ any supported kernel page, binaries load everywhere and the
runtime's congruence gate always passes — the skip path becomes dead, unblocking the fatal-
reprotect follow-up.

## Acceptance

- aarch64 `--pie` images carry 64 KiB `p_align` with congruent offsets; x86_64/riscv64 output
  byte-identical to today.
- The writer has one target-supplied page-size home; no independent 4096 constants remain.
- Documented size cost (inter-segment padding) measured on the compiler corpus in the PR.
- Unit tests pin alignment + congruence per ISA; suite/int/fixpoint green.

octalide:
Blocks the runtime half of the end-state: briar-systems/mach-std's fatal-reprotect issue (filed alongside this) adopts the glibc stance once this layout guarantee exists.

PR #1888, opened 2026-07-03
fix(link): emit honest PT_GNU_RELRO extent — arm64 PIE RELRO mprotect (#1885)

### LINKED ISSUE #1885: runtime/link: arm64 PIE binaries panic at startup — RELRO mprotect fails on native arm64 hosts (64K images × fatal re-protect)
arm64 PIE binaries built by the current toolchain (mach with 64 KiB max-page images, #1845, + mach-std with fatal RELRO re-protection, mach-std#347) die at startup on native arm64 linux hosts: `std.runtime.linux.reloc: mprotect of the PT_GNU_RELRO region failed (cannot restore read-only hardening)`, SIGTRAP, exit 133. **Both changes shipped in v2.14.0 / mach-std main, so this affects the released toolchain.**

## Repro

CI reproduces it on every current branch including dev: `int linux-arm64` (native arm64 runner), `surface/pie-reloc` and `surface/pie-relro-fault`, debug and release — e.g. dev run 28662033489 @ 873450ee. The identical binaries pass under qemu-aarch64 (user-mode emulation does not reproduce native mprotect semantics).

## Expected

The PIE image self-relocates, re-protects PT_GNU_RELRO read-only, and runs (`relro_write=faulted` for the fault probe; exit 0 for pie-reloc).

## Actual

`mprotect` fails, the (now-fatal, mach-std#347) re-protection path panics, producer exits 133. The fault probe golden diff shows `relro_write=exit133`.

## Timeline / masking

- PR #1871 (64K images) had a GREEN native arm64 int leg — but at that pin, re-protection failure was still best-effort: mprotect may have already been failing SILENTLY (unhardened binaries reported as passing).
- The failure became fatal-and-visible once the mach-std pin picked up #347. Intermediate CI runs (release train PRs #1876–#1878) were green, so either the failure is host-kernel-dependent (GitHub's arm64 runner fleet mixes page sizes) or layout-dependent — the investigation must determine which.

## Suspected mechanism (to verify, not assume)

On a 4K/16K-page kernel, a 64 KiB-aligned image is loaded on runtime-page boundaries, but if the runtime's re-protect (or the writer's PT_GNU_RELRO extent) rounds the region using the IMAGE's max-page rather than AT_PAGESZ, `mprotect` can span address space the loader never mapped → ENOMEM. Alternatively the #347 congruence gate itself computes against the wrong page. Whichever it is, the fix belongs at the correct layer (writer extent vs runtime rounding), not as a tolerated failure — the fatal semantics of #347 stay.

## Acceptance

- Root cause identified with errno-level evidence from a native arm64 host (a scratch GitHub Actions run on the arm64 runner is an acceptable lab).
- arm64 PIE binaries harden and run on native 4K/16K/64K-page kernels; `int linux-arm64` green on dev.
- The RELRO fault probe (`pie-relro-fault`) and the writer-level unit guards updated if the extent model changes; qemu-vs-native gap noted in the harness docs if qemu cannot exercise the fixed path.
- Assessment of the released v2.14.0: statement of exactly which host configurations are affected, for octalide's hotfix decision.

octalide:
The mach (writer) half of the critical arm64 startup fix #1885. The mach-std (runtime) half is briar-systems/mach-std#360 and is the actual crash fix; this makes the header honest so the two agree.

## What

The PIE writer rounded `PT_GNU_RELRO` p_memsz up to the IMAGE max-page (64 KiB on aarch64, #1845). The kernel maps the tiny `.data.rel.ro` segment rounded to the RUNTIME page, so on a runtime page smaller than the max-page (a 4K/16K-page aarch64 kernel) the declared region overshot the mapping and the runtime's re-protect faulted `ENOMEM`. This emits `PT_GNU_RELRO` p_memsz as the segment's **actual extent**; the runtime (mach-std#360) rounds it up to `AT_PAGESZ` — the pages the kernel actually mapped. The rounding moves from the build-time max-page to the runtime page, where it belongs.

Byte-changes the RELRO p_memsz field on all targets (it was over-rounded before; the runtime no longer relies on it). Updates the #1845 writer guard (`emit_pie_exec:relro_splits`) to assert the actual extent.

## Sequencing

mach-std#360 lands + releases + pin bumps first (it repairs already-released binaries on rebuild, no new mach release needed). This PR then makes the writer's declared extent match what the runtime protects. `int linux-arm64` (native) is the end-to-end acceptance and goes green once the mach-std pin is in.

## Verification status

- [x] self-host fixpoint byte-identical; unit suite 511/511 (updated RELRO guard)
- [ ] native arm64: probe built with this writer + fixed runtime runs hardened, honest PT_GNU_RELRO p_memsz (scratch CI, in progress)
- [ ] int linux-arm64 green (after mach-std pin)

### DIFF
--- a/src/lang/target/of/elf.mach
+++ b/src/lang/target/of/elf.mach
@@ -1693,14 +1693,15 @@ fun write_pie_phdrs(buf: *u8, phdr_offset: usize, phdr_count: u32,
     # with relocated constants), when present. the region is mapped writable above for the
     # self-reloc; this header is the runtime's cue to re-protect it read-only (see the
     # `pie_seg_flags` contract). pure `.rodata` carries no base relocation, so it is never
-    # covered here and stays read-only from load. p_memsz is page-rounded so the runtime
-    # re-protects whole pages with no arithmetic of its own - safe because every merged
-    # kind starts on a fresh page (assign_section_vaddrs), so the next segment begins where
-    # this one ends.
+    # covered here and stays read-only from load. p_memsz is the segment's actual extent,
+    # NOT rounded to the image max-page: a max-page larger than the runtime page (a 64 KiB
+    # aarch64 image on a 4 KiB kernel) would overshoot the mapping, and the runtime's
+    # re-protect would fault ENOMEM (briar-systems/mach#1885). the runtime rounds this
+    # extent up to the runtime page (AT_PAGESZ) - the pages the kernel actually mapped.
     val relro_seg: i64 = pie_relro_seg(segs, seg_count, dyn);
     if (relro_seg >= 0) {
         val ri: u32 = relro_seg::u32;
-        val relro_size: usize = bin.align_up(segs[ri].memsz, page_size::usize);
+        val relro_size: usize = segs[ri].memsz;
         pos = write_phdr(buf, pos, PT_GNU_RELRO, PF_R, seg_offsets[ri], segs[ri].vaddr,
                          relro_size, relro_size, 1);
     }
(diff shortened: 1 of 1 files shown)


octalide:
## Verified — full fix green end to end

- **Native arm64 (scratch CI, 4K-page runner)**: probe built with this writer + fixed mach-std → `GNU_RELRO MemSiz 0x10` (honest, == the .data.rel.ro LOAD), `mprotect(…, 4096, PROT_READ) = 0`, EXIT=42 (hardened). Was 64K/ENOMEM/exit 133.
- **`int linux-arm64` (native, this PR's CI): PASS** — pie-reloc + pie-relro-fault green with the fixed mach-std main runtime. All int legs (linux, arm64, riscv64, windows) + builds + tests pass.
- self-host fixpoint byte-identical; unit suite 511/511 (updated `emit_pie_exec:relro_splits` guard asserts the actual 16-byte extent).

mach-std#360 (runtime) is merged to mach-std dev+main and is the crash fix; this PR makes the writer's declared PT_GNU_RELRO extent honest so it matches what the runtime protects.

PR #1891, opened 2026-07-03
hotfix(link): honest PT_GNU_RELRO extent — arm64 PIE RELRO mprotect (#1885)

octalide:
Hotfix off **main (v2.14.0)** for the critical arm64 startup crash #1885. The runtime half (mach-std#360) is merged to mach-std dev+main and is the actual crash fix; this is the writer half — emit an honest PT_GNU_RELRO extent so the header matches what the runtime protects. Branched off main (not dev) so a v2.14.1 does not carry dev's debug-info work; targets main for the dual-merge.

## What

The PIE writer rounded `PT_GNU_RELRO` p_memsz up to the IMAGE max-page (64 KiB on aarch64, #1845). The kernel maps the tiny `.data.rel.ro` segment rounded to the RUNTIME page, so on a 4K/16K-page aarch64 kernel the declared region overshot the mapping and the (now-fatal, mach-std#347) re-protect faulted `ENOMEM` at startup. This emits p_memsz as the segment's **actual extent**; the runtime rounds it up to `AT_PAGESZ`.

The runtime fix (mach-std#360) alone repairs already-released v2.14.0 binaries on rebuild (it uses the `.data.rel.ro` PT_LOAD's mapped extent, robust to the over-rounded header); this writer change makes the header honest so the two agree.

## Verification

- Native arm64 (scratch CI, 4K-page runner), full fixed toolchain: `readelf GNU_RELRO MemSiz 0x10` (was 0x10000), `mprotect(…, 4096, PROT_READ) = 0` (was 65536 ENOMEM), exit 42 (was 133).
- `int linux-arm64` (native) green on the fixed stack (this branch's CI, pulling fixed mach-std main).
- self-host fixpoint byte-identical; unit suite 501/501 — `emit_pie_exec:relro_splits` updated to pin the actual extent, not page-rounded.

## Note on x86_64/riscv64

Behaviorally unaffected (int + fixpoint green; the runtime uses the PT_LOAD mapped extent regardless). The only change is the RELRO p_memsz **header field** (page-rounded → actual extent), per the actual-extent model. If strict byte-identity is preferred, the extent can round to the 4 KiB minimum page instead.

### DIFF
--- a/src/lang/target/of/elf.mach
+++ b/src/lang/target/of/elf.mach
@@ -1685,14 +1685,15 @@ fun write_pie_phdrs(buf: *u8, phdr_offset: usize, phdr_count: u32,
     # with relocated constants), when present. the region is mapped writable above for the
     # self-reloc; this header is the runtime's cue to re-protect it read-only (see the
     # `pie_seg_flags` contract). pure `.rodata` carries no base relocation, so it is never
-    # covered here and stays read-only from load. p_memsz is page-rounded so the runtime
-    # re-protects whole pages with no arithmetic of its own - safe because every merged
-    # kind starts on a fresh page (assign_section_vaddrs), so the next segment begins where
-    # this one ends.
+    # covered here and stays read-only from load. p_memsz is the segment's actual extent,
+    # NOT rounded to the image max-page: a max-page larger than the runtime page (a 64 KiB
+    # aarch64 image on a 4 KiB kernel) would overshoot the mapping, and the runtime's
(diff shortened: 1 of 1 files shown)


octalide:
Closing as duplicate — the writer fix already merged to dev via #1888 (dev-base accepted). Per the team lead, a v2.14.1 cherry-pick onto a main-based hotfix is theirs to run if octalide orders it; this PR would re-apply a change dev already has.

Issue #1904, opened 2026-07-03
debuginfo: mem2reg phi-merge dbg-value births + alloca↔source-variable link


octalide:
## What
Teach mem2reg to re-emit `OP_DBG_VALUE` births at the phi nodes it inserts, so a
source variable's debug binding is correct across control-flow merges — not just the
last-processed branch's value. This needs a new **alloca ↔ source-variable link** so
mem2reg knows which variable (name/ty/scope) a promoted alloca stands for.

## Why
#1695 emits `OP_DBG_VALUE` births at locals/params/reassignments, and #1696 maintains
them across the optimizer (RAUW moves operands, dce salvages folded definitions). But
when mem2reg promotes an alloca and inserts a phi at a dominance-frontier merge, the
variable's value at and after that merge is the **phi result**, and no birth is
re-emitted there. mem2reg cannot emit one today because it has no link from the promoted
alloca back to the source variable it represents — the `DbgVar` identity lives on the
`OP_DBG_VALUE` instructions, not on the alloca.

Consequence (current, accepted for #1696): past a merge, a variable's binding may reflect
the value from the last-processed predecessor rather than the phi. Single-value and
straight-line code is correct; only merge points are affected. This is format-agnostic
(DWARF loclists and CodeView defranges both consume the resulting bindings).

## Design (complete)
1. **alloca ↔ source-variable link.** When lowering a source local to its `alloca`,
   record the alloca instruction id → `DbgVar` (name, ty, scope) association in a
   function side table (mirrors `dbg_vars`, keyed by the alloca id). Populate it only
   under `-g`.
2. **phi births in mem2reg.** In `place_phi` (or right after a phi is materialised for a
   promoted alloca), if the alloca has a source-variable link, emit an `OP_DBG_VALUE`
   binding that variable to the phi result at the merge block's entry, with the identity
   copied from the link. Reuse the #1696 expression pool (identity expression — the phi
   result is the value directly).
3. **rename interaction.** The rename walk already threads the reaching definition per
   alloca; the phi birth uses the phi as that block's reaching def, so the binding stays
   consistent with the values flowing into later uses.

## Acceptance
- A promoted local written on both arms of an `if` and read after the merge resolves,
  in the debugger, to the merged (phi) value at and after the merge — verified by IR
  inspection of the re-emitted `OP_DBG_VALUE` at the phi, and end-to-end once the DWARF
  loclist emitter (#1594 line of work) consumes it.
[...]

Issue #1695, opened 2026-06-28
debuginfo: OP_DBG_VALUE IR entity + source-variable identity + births


octalide:
## What
Add a pure-metadata OP_DBG_VALUE IR entity (produces no value) binding a source variable — name, type, scope — to its current SSA Value. Born at lowering per decl + each reassignment; thread the variable identity down toward the owning vreg/slot. Maintained through the mutation primitives (RAUW moves dbg-values to the replacement) but WITHOUT the optimizer-specific birth/salvage logic (that is #1696).

## Why
Variable-location tiers need a source-variable<->location link that does not exist below the AST (MirVReg is only {id,class,spill_slot,assigned}; the only existing link is inline-asm AsmBinding). This entity is format-agnostic and consumed unchanged by both the DWARF and CodeView emitters.

## Acceptance
- Lowering emits OP_DBG_VALUE for locals/params carrying variable identity.
- Passes preserve it via the mutation-api primitives without affecting codegen output (verified by IR dump + byte-unchanged corpus).
- No emitter here — this is the shared IR substrate for DWARF and CodeView locations.

Depends-on: #1689, #1690. Foundation; reused by all three primes.

PR #1905, opened 2026-07-03
feat(ir): maintain dbg-values through optimization (salvage + expression pool)

### LINKED ISSUE #1696: debuginfo: maintain dbg-values through optimization (births, RAUW, salvage)
## What
Maintain OP_DBG_VALUE across the optimizer: births at mem2reg promotion (the reaching SSA def = the variable's location), RAUW moves via the mutation-api primitive, and DCE salvage — rewrite a deleted defining instruction into a recoverable expression (%c = add %a, 4 deleted -> expr %a + 4), else set 'no location' so the debugger shows *optimized out*, never garbage.

## Why
The default pipeline runs mem2reg->constfold->dce and deletes the stack slots naive debug info would point at — there is no -O0 escape hatch. Value tracking is mandatory for any location to be correct on real builds. Format-agnostic; enforced once across all passes, consumed by DWARF loclists and CodeView defranges alike.

## Acceptance
- After full optimization, dbg-values resolve to the correct current location or explicit no-location.
- A fixture confirms a mem2reg-promoted local tracks register/slot/const transitions (verified by IR dump).

Depends-on: #1695. Foundation; reused by all three primes.

octalide:
Closes #1696. Last child of the debug-info foundation epic #1688.

## What
Maintains `OP_DBG_VALUE` across the optimizer so a variable's location stays correct after mem2reg -> constfold -> dce (there is no `-O0` escape hatch):

- **RAUW** already moved dbg-value operands via the shared `Rewrite` map (#1695); nothing to add.
- **DCE salvage** (new): when a value-producing instruction bound by a dbg-value is dropped, `erase_marked` salvages the binding — `%c = add %a, 4` deleted rebinds the operand to `%a` and records a `+4` recovery step; an unrecoverable definition (e.g. a multiply) degrades to *optimized out* (a `VAL_CONST_NULL` operand), never a stale value.
- **Debug-only uses**: `dce`'s `operand_is_value_use` now excludes `OP_DBG_VALUE`, so a debug use never pins an otherwise-dead value — a `-g` build drops exactly what a plain build does (codegen identity).

## Expression seam
Salvage recovery lives in a **function-level expression pool** (`Function.dbg_exprs`): each `OP_DBG_VALUE` carries a `DbgExprId` (0 = identity, owns no slot), and steps are `plus_const` / `minus_const` with a `u64` arg (extensible enum). This is the format-agnostic seam the DWARF loclist / CodeView defrange emitters consume — not inline on `DbgVar`, since one variable carries different expressions at different points.

## Deferred (phi-merge births) — three riders
mem2reg does not yet re-emit a birth at the phi it inserts for a promoted local, so past a control-flow merge a binding may reflect the last-processed branch. Deferred with:
- follow-up issue **#1904** (alloca <-> source-variable link + phi births), filed as a child of #1688;
- a breadcrumb at the mem2reg phi-insertion site;
- a fidelity note in the `OP_DBG_VALUE` docs.

## Tests
Two pipeline-survival tests (loc-survival style, verified by IR inspection): a mem2reg-promoted local through constfold/dce keeps its binding with the correct value + `+4` expression; an unrecoverable case shows optimized-out.

## Verification
- from-source build clean; unit suite **528 passed** (526 baseline + 2 new), confirmed with a freshly-built from-source compiler.
- Remaining bar (byte-identical without `-g` on all six targets, `-g` functional, self-host fixpoint, int harness) run before marking ready.

### DIFF
--- a/src/lang/me/ir.mach
+++ b/src/lang/me/ir.mach
@@ -105,6 +105,50 @@ pub rec DbgVar {
     scope: u32;
 }
 
+# add the op's `arg` to the operand value
+pub val DBG_OP_PLUS_CONST:  u8 = 0;
+# subtract the op's `arg` from the operand value
+pub val DBG_OP_MINUS_CONST: u8 = 1;
+
+# one step of a debug-location expression, applied to an OP_DBG_VALUE operand to
+# recover the source variable's value after the instruction that defined the operand
+# was folded away (`%c = add %a, 4` deleted -> operand rebased to %a, expression
+# gains a +4 step). the op set is deliberately small and extensible: each -g location
+# emitter maps a step to its own encoding (a DWARF DW_OP_plus_uconst / minus, a
+# CodeView defrange adjust).
+# ---
+# op:  DBG_OP_* selector
+# arg: the step's u64 immediate (the folded constant's bit pattern)
+pub rec DbgOp {
+    op:  u8;
+    arg: u64;
+}
+
+# a DbgExprId of DBG_EXPR_IDENTITY names the empty expression: the OP_DBG_VALUE
+# operand is the variable's value verbatim, with no steps to apply. it owns no pool
+# slot, so a function that never salvages allocates no expression storage.
+pub val DBG_EXPR_IDENTITY: DbgExprId = 0;
+
+# an index identifying the debug-location expression an OP_DBG_VALUE applies to its
+# operand. DBG_EXPR_IDENTITY (0) is the empty expression and owns no slot; every
+# other id is 1-based into the owning function's `dbg_exprs` pool.
+pub def DbgExprId: u32;
+
+# a debug-location expression: the steps applied base-first to an OP_DBG_VALUE
+# operand to recover the source variable's value. this is the emitter-facing seam --
+# the -g location emitters (DWARF loclists, CodeView defranges) read one shared,
+# format-agnostic pool per function (Function.dbg_exprs) rather than each re-deriving
+# recovery math. an empty `ops` is the identity expression.
+# ---
+# ops:      the expression steps, applied in order (owned)
+# op_count: number of live steps
+# op_cap:   allocated capacity of `ops` (the true allocation size freed at teardown)
+pub rec DbgExpr {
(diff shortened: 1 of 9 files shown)


octalide:
## Verification (local, from-source compiler)

All checks used a freshly-built from-source compiler (the installed seed under-collects; not used for counts).

- **Unit suite:** 529 passed, 0 failed = 526 baseline + 3 new (2 dbg pipeline-survival + 1 pool-growth stress).
- **Self-host fixpoint:** stage2 == stage3, byte-identical.
- **Byte-identical WITHOUT `-g`, all six targets:** compiled a fixed input (dev source) with this branch's compiler and the `origin/dev` compiler, no `-g`; `linux-x86_64, linux-arm64, linux-riscv64, windows-x86_64, darwin-x86_64, darwin-aarch64` all **IDENTICAL** — codegen is provably unperturbed when `-g` is off.
- **`-g` build clean + functional:** built the compiler itself with `-g` (exercising dbg-value births + salvage across the whole codebase); the `-g`-built compiler runs the full 529-test suite green and carries `.debug_abbrev/.debug_info/.debug_line/.debug_str`.
- **int harness (linux leg):** all surface + regression cases pass (debug and release).

## Crash found and fixed during verification

The first `-g` self-build **segfaulted** (corrupt allocator pointer in `allocate_raw`). Root cause: the lowerer hand-rolls `Function` initialization at two sites (`lower/decl.mach`, `lower/context.mach`) instead of `function_add`, and I had added the dbg-expression pool only to `function_add`. The other two left `dbg_expr_ids` a zeroed map (nil allocator) — inert without `-g` (teardown guards tolerate nil/0), but under `-g` the dce salvage `map.insert`s through the nil allocator. Fixed by initializing the pool at both sites (mirroring how #1695 added `dbg_vars` to all three).

**Structural note (pre-existing, out of scope):** `Function` field initialization is duplicated across three sites (`function_add`, `decl`, `context`) — the exact shape that caused this bug and will recur on the next field. Worth a follow-up to fund a single `Function` initializer; I did not refactor it here to keep #1696 scoped.

Issue #1964, opened 2026-07-05
rfc: mach.toml redesign — artifacts, steps, links


octalide:
## Problem

The current manifest cannot express what C-binding and vendored-native
libraries need, especially as dependencies: there is no way to declare
system link requirements that cascade to consumers (mach-glfw's
`[os.X] libs` experiment), no build-step story for vendored C source,
and a dependency's own surface deps do not resolve in consumers
(#1927). The design was worked through against real cases (mach-glfw
system-lib bindings; a vendored-C miniz; a consumer app using both,
cross-target); the working sketch lives at
`~/dev/src/github.com/briar-systems/mach-build` (local).

> Rewritten from the original RFC text, which had drifted past the
> settled design (constraint-based version resolution, per-project dep
> scoping, toolchain resolution, sandboxing — none of that is proposed
> here; see edit history).

## Proposal

Seven sections — `[project]`, `[dep.X]`, `[link.X]`, `[artifact.X]`,
`[step.X]`, `[target.X]`, `[profile.X]` — all `[category.name]`-keyed
(TOML itself enforces name uniqueness). Convention covers absent
manifests and sections: `src/` is the module tree, `src/main.mach`
the entry, native+debug the defaults; a pure-mach single-bin project
needs no manifest at all. But a **declared** table is declared
completely: every field is required, nothing defaults (the manifest
twin of mach's explicitness philosophy). The only exception is
shape-dependence — fields whose presence follows another value (a dep
is `git` or `path`; a link entry carries `name` or `path` per its
`source`). "Any" is said explicitly with `"*"`.

The complete consumer example:

```toml
[project]
id      = "demo"
version = "0.1.0"
src     = "src"
out     = "out/{target.name}/{profile.name}"

[dep.std]
git = "https://github.com/briar-systems/mach-std"
ref = "tag/v0.17.0"

[dep.glfw]
git = "https://github.com/briar-systems/mach-glfw"
ref = "tag/v0.2.1"

[dep.mz]
git = "https://github.com/briar-systems/mach-miniz"
ref = "tag/v1.0.3"

[link.shim]
source = "local"
path   = "{project.out}/obj/platform/shim.o"
os     = "*"
isa    = "*"
abi    = "*"
export = false

[link.shim-x11]
source = "local"
path   = "{project.out}/obj/platform/x11.o"
os     = "linux"
isa    = "*"
abi    = "*"
export = false

[link.shim-win32]
source = "local"
[...]

Issue #1927, opened 2026-07-04
feat(resolver): propagate a library's transitive surface dependencies to consumers


octalide:
A library whose public surface exposes types from its own dependency forces every consumer to ALSO declare that transitive dep in its own mach.toml, even though the consumer never imports it directly. boom's surface exposes glfw/gl types, so a boom-only consumer must still declare mach-glfw and mach-gl pinned to boom's refs, or it fails with `use path 'glfw.glfw' does not name a module`.

Confirmed no library-side fix: `fwd glfw.glfw;` does not help, because boom.window's own `use glfw.glfw` resolves in the CONSUMER's flat id registry, and the resolver only registers dep ids the ROOT project declares directly. `mach dep pull` fetches glfw/gl transitively (on disk in the consumer's dep/) but they are not registered; there is no re-export manifest flag (only the `libs` cascade).

Requested: let a library propagate/re-export its surface deps so consumers do not redeclare them (a manifest re-export flag, or resolver registration of transitively-fetched dep ids used by a dependency's surface). Found via the conatus demo; boom#11 documented the current requirement as interim.

octalide:
Subsumed by #1964's design and implemented via #1968 (transitive registration into the flat id registry, hard conflict errors). The per-project scoping direction floated here was explicitly rejected in the final design — see #1964's alternatives.

PR #1986, opened 2026-07-09
feat(driver): hard-error on conflicting transitive dep ids (#1968)

### LINKED ISSUE #1968: feat(driver): register transitively fetched deps in the flat id registry
Child of #1966; design in #1964 ("Dependencies are today's system, unchanged" + transitive registration). Subsumes #1927 — its diagnosis is confirmed in code: `ProjectConfig.deps` is populated only from the root manifest's own `[deps.*]` (`resolve_deps`, driver.mach:1655), and module resolution scans exactly that array (`fqn_to_file_path` :2111, `project_module_lookup` :3295), so a dep's surface deps never resolve in consumers.

## Scope

Populate the (still flat) id registry from the **whole transitive closure**, reusing the walk that already exists on the fetch side (`enqueue_manifest`/`read_dep_manifest`/`NodeVec`, dep.mach:912-998):

- Every fetched dep's `[project].id` registers; the alias is whatever the declaring dep used (store stays alias-keyed `dep/<alias>`, driver.mach:1982).
- Unclean resolution is a hard error naming the requirer chain: same project id reached from different (source, ref) pairs, or two different projects claiming one id. Same (source, ref) unifies to one entry. Fix is fork or upstream — no remap levers (#1964).
- No version solving of any kind; today's exact-ref model.

Schema-agnostic: build against current v1 manifests on dev now. Expect a trivial rebase when the v2 parser child renames `[deps]` → `[dep]` (DepDef shape is unchanged).

## Acceptance

- The #1927 case passes: a consumer declaring only `boom`-style lib X, where X's surface exposes types from X's own dep Y, resolves `use` paths through Y without the consumer declaring Y.
- Conflict cases error with the full requirer chain, not the leaf.
- Unit tests for registration + both conflict shapes; int/ case only if it fits existing conventions (path deps; no new fixture trees outside int/ norms).
- Full bar: build from source, unit suite through the freshly built compiler, self-host fixpoint.

octalide:
Closes #1968.

The transitive closure walk already registers every fetched dep's project id into the flat id registry (the shipped registration half, #1927 acceptance passing). What was missing is the **conflict half**: a project id re-encountered in the closure was silently deduped at the `project_id` check in `resolve_deps_recursive`, so one id reached from two incompatible sources or refs quietly resolved to whichever won the walk.

## What changed (driver.mach only)
- Track each registered dep's **provenance** — its declared source label (`git <url>` / `path <path>`), its ref, and the requirer chain reaching it — in a `DepProvenance` array threaded parallel to the resolved `DepEntry` array (walk-local; freed as soon as the walk returns).
- When a project id recurs with a **different (source, ref) pair**, fail with a diagnostic that names **both full requirer chains** and what each pins, e.g.:
  ```
  error: dependency conflict: project id 'libx' is required from two incompatible sources:
      app -> foo -> deep pins path ./v/deep
      app -> bar pins path ./v/bar
      fork or upstream to reconcile; one project id must resolve to a single source and ref (no version remapping)
  ```
- **Same (source, ref) still unifies** (the legitimate diamond). Exact-ref model, no version reconciliation or remap levers (#1964). The dedup/registration behavior is otherwise unchanged.

The two conflict shapes from the issue both flow through the same check: *same id from two (source, ref) pairs* (source or ref differs) and *two different projects claiming one id* (sources differ).

## Tests (3, all under `mach.lang.driver`)
- `dep_id_conflicts_unify_and_shapes` — the unify/conflict decision: same source+ref unifies; a different ref conflicts; a different source conflicts.
- `dep_closure_conflict_names_chains` — an on-disk two-path-dep closure resolving one id from two sources errors, and the diagnostic names both requirer chains.
- `dep_closure_registers_transitive` — an on-disk `root -> a -> b` closure registers both the direct and the transitive project (the registration acceptance).

## Verification
- Build from source: clean.
- Unit suite through the fresh compiler: 539 passing, 0 failing (base 536 + these 3).
- `int/run.sh --target linux`: 42/42.
- Self-host fixpoint + end-to-end conflict demo (scratch /tmp path-dep closure): in the PR thread.

### DIFF
--- a/src/lang/driver.mach
+++ b/src/lang/driver.mach
@@ -1969,8 +1969,78 @@ fun intern_opt_unwrap(itn: *intern.Interner, text: str) intern.StrId {
     ret intern_unwrap(itn, text);
 }
 
+# the provenance of one registered dep, tracked parallel to the resolved DepEntry array
+# so a project-id collision in the closure can name the conflicting sources, refs, and
+# requirer chains rather than silently deduping the loser (#1968).
+# ---
+# source: interned source label ("git <url>" / "path <path>") the requirer declared
+# ref:    interned declared ref, STR_NIL when omitted
+# chain:  interned requirer chain ("<root> -> a -> b") ending at this dep's alias
+rec DepProvenance {
+    source: intern.StrId;
+    ref:    intern.StrId;
+    chain:  intern.StrId;
+}
+
+# a re-encountered project id unifies with its prior registration only when both the
+# declared source and ref match exactly; any difference is a hard conflict. the model
+# is exact-ref with no version reconciliation (#1964), so the two legitimate diamonds
+# (same git url+ref, same path) unify and everything else is reported.
+fun dep_id_conflicts(existing_source: intern.StrId, existing_ref: intern.StrId,
+                     new_source: intern.StrId, new_ref: intern.StrId) bool {
+    if (existing_source == new_source && existing_ref == new_ref) { ret false; }
+    ret true;
+}
+
+# the text of an interned id, or `default` when the id is STR_NIL or unrecorded.
+fun id_text_or(s: *session.Session, id: intern.StrId, default: str) str {
+    if (id == intern.STR_NIL) { ret default; }
+    val o: O.Option[str] = intern.lookup(?s.interner, id);
+    if (O.is_some[str](o)) { ret O.unwrap[str](o); }
+    ret default;
+}
+
+# a dep's declared source as an interned identity+display label: "git <url>" for a git
+# source, "path <path>" for a path source, else "<unknown source>". identity is a StrId
+# compare, so the scheme prefix keeps a git url and an identical path string distinct.
+fun dep_source_label(s: *session.Session, d: *manifest.DepDef) intern.StrId {
+    if (d.git != intern.STR_NIL) {
+        val u: O.Option[str] = intern.lookup(?s.interner, d.git);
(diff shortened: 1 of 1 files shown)


octalide:
## End-to-end conflict demo (scratch /tmp path-dep closure)

The conflict diagnostic names the **full requirer chain** (`app -> foo -> deep`, not just the leaf `deep`), and the legitimate diamond (same id from the same source) resolves without error.

```
$ tree scratch/root   # a flat closure: two projects both declare [project] id = "libx"
root/
  mach.toml            # [dep.foo] path=./v/foo   [dep.bar] path=./v/bar
  src/main.mach
  dep/foo/mach.toml    # [dep.deep] path=./v/deep
  dep/deep/mach.toml   # [project] id = "libx"
  dep/bar/mach.toml    # [project] id = "libx"

$ mach build . --target linux --profile debug
error: dependency conflict: project id 'libx' is required from two incompatible sources:
    app -> foo -> deep pins path ./v/deep
    app -> bar pins path ./v/bar
    fork or upstream to reconcile; one project id must resolve to a single source and ref (no version remapping)
# exit 1

--- legitimate diamond: the same id from the SAME source unifies (no error) ---
$ mach build . --target linux --profile debug   # foo and bar both require shared from ./v/shared
error: undefined entry symbol '_start'; ensure the target startup library is linked
# exit 1 (reaches link stage => deps resolved, no conflict)
```

## Verification summary
- build from source: clean
- unit suite (fresh compiler): **539 passed, 0 failed** (base 536 + 3 new)
- self-host fixpoint: seed→a→b→c, **cmp b c converged**; fixpoint compiler self-test 539/0
- `int/run.sh --target linux`: **42/42**

Issue #1969, opened 2026-07-05
feat(link): [link.X] entries — filtered, exportable link requirements


octalide:
Child of #1966; design in #1964. Replaces the v1 `libs` cascade on the v2 path. BLOCKED until the schema child of #1966 merges (needs the v2 structs).

## Scope

- Entry selection per build cell: `os`/`isa`/`abi` filters (`"*"` = any) evaluated against the resolved target; artifacts opt in by name via `link = [...]`; a referenced entry whose filters miss is skipped.
- `export = true` entries cascade import-driven from deps — rework the existing topological cascade (`resolve_cascade_libs` driver.mach:1692, `append_cascade_array` :1859) to read v2 `[link.X]` from dep manifests; ordering discipline unchanged (manifest before CLI inputs, topological cross-dep, dedup).
- Static/dynamic classification as today (`local` path → static input; `system`/`framework` name → dynamic/soname or framework reference per platform), feeding `link_artifact` (build.mach:1529) and the `Q_LINK_CONFIG` fingerprint unchanged.
- `local` entry paths: until the steps child lands, the path must exist on disk at link time — error names the entry. Store the entry→path association so the steps child can wire demand matching.
- `mach test`: test executables link the union of all artifacts' referenced entries plus exported dep entries, filtered to the **native** target (testing.mach / build_tests path). Symbol collisions in that union are an honest link error.

## Acceptance

- A mach-glfw-shaped v2 manifest (system lib + darwin frameworks, `export = true`) links correctly natively and cross-cell; consumer inherits by import, not declaration.
- Unit tests: filter matching (incl. `"*"` and arrays), export cascade, ordering determinism, test-union rule.
- Full bar: build from source, unit suite through the freshly built compiler, self-host fixpoint, int/ matrix.

octalide:
Audit findings against the shipped v3.0.2 state (each independently re-verified in source). What remains for this issue:

[...]

Issue #1966, opened 2026-07-05
epic: mach.toml v2 — artifacts, steps, links (build-system overhaul for 3.0.0)


octalide:
> ## Status (2026-07-09, core lane complete)
> The finish-off program's core lane is DONE: #1968, #1969, #1970, #1975, #1976, #1988, plus the related #1974/#1985→#1992/#1993 (required-path resolution), #1978 (cd.yml), #1987 (doc mkdir) — all merged with adversarial verification (the step engine survived a six-surface attack workflow plus two fix rounds: four blockers found and fixed with regression tests). conatus builds green cold and warm, from any invocation directory. The int matrix has been green since #1976. Release 3.1.0 is being cut now; it becomes the seed that gates #1971 and #1979.

Implementation umbrella for #1964 (design of record, design-complete as of 2026-07-05 — read it first; nothing here re-litigates it).

## Children

- [~] #1967 — **Schema v2 parser**: shipped; RFC-exact acceptance landed (#1983: optional name/description, {project.out}, string filters; #1990: presence semantics). Remaining half is strictness — closes with #1971.
- [x] #1968 — **Transitive dep registration**: registration + conflict hard errors with requirer chains (#1986).
- [x] #1969 — **[link.X] entries**: string/array filters, up-front local validation (cell-aware per #1988), test-link union, export cascade (#1983, #1990).
- [x] #1970 — **[step.X] engine**: demand-driven, dep export steps with root-out homing, content-fingerprint stamp caching, in-globbing; conatus green (#1989).
- [ ] #1971 — **Flag-day**: strict totality + [project] trim + template closed-set tightening (drop {name}/{ext} and bare aliases) + fixture/init re-touch. Gated on the 3.1.0 seed; implementation handoff notes on the issue.
- [x] #1972 — **Ecosystem migration (tracking)**: superseded by #1979 for the totality re-touch.
- [x] #1975 — **mach.lock legacy-section un-pin** (#1984).
- [x] #1976 — **int/dbg fixture migration** — matrix green since (#1982).
- [ ] #1977 — **Docs**: manifest.md/cli.md/help rewrite against the final strict schema; after #1971.
- [ ] #1979 — **Ecosystem totality sweep**: 13 repos to RFC-exact; gated on the 3.1.0 seed. Rides: mach-gl/mach-image changelog corrections, conatus comment fix.
- [ ] #1980 — **kind="static" phantom out**: owner decision then implement; after #1971.
[...]

Issue #1943, opened 2026-07-04
feat(build): declarative build steps in mach.toml, dependency-graph-aware


octalide:
## Problem
A dependency sometimes needs an artifact produced by a tool outside `mach` before the mach build can link/run: mach-audio's vendored miniaudio C shim (compiled with `cc`), the planned GLSL->SPIR-V shader compile for the Vulkan path (`glslc`), the mach-gl/mach-vk generators, etc. Today this is ad hoc (a per-repo Makefile) and, worse, it does NOT propagate: a consumer of boom must itself know to build boom's transitive miniaudio shim and pass `-L`. So `mach build .` is not self-sufficient for any consumer of a native-backed library.

## Direction (preferred: generalize, do NOT hardwire C)
Rather than teaching the compiler about C specifically, add a general mechanism to declare BUILD STEPS in mach.toml, scoped per-target / per-platform / per-os, that mach runs as part of the build and whose outputs feed the build (link inputs, generated sources). Keep it tool-agnostic: a step is a command that produces declared artifacts. That one mechanism covers the C shim (`make lib` / `cc -shared`), offline shader compilation (`glslc`), and codegen, without coupling mach to any of them.

Sketch (shape TBD):
```toml
[[build.step]]
os   = "linux"                 # scoped: per-os / per-platform / per-target
run  = "cc -O2 -fPIC -shared vendor/mad.c -o build/libminiaudio.so -lpthread -lm -ldl"
provides.link = ["build/libminiaudio.so"]   # cascades onto the link path
```

## The load-bearing requirement: transitive propagation
The key gap is not running a command, it's that a dependency's build steps + their provided link inputs must run and PROPAGATE when building a CONSUMER. Building boom must run mach-audio's declared steps (build the shim) and inherit its link inputs automatically, with no Makefile and no manual `-L`. This is the same class as the transitive surface-dep gap in #1927 (consumers manually declaring mach-image/mach-gltf) - both are 'mach does not carry transitive dependency requirements.' Ideally solved together.

## Interim
boom currently uses a Makefile that builds the miniaudio shim and passes `-L`; that is explicitly temporary and should be removed once this lands.

Trust/security note: running declared commands is arbitrary-command execution during build (as with cargo build.rs, npm postinstall); worth a sandbox/allow story, but that is a detail of the general mechanism, not a reason to hardwire C.

octalide:
[...]

Issue #1164, opened 2026-06-07
epic: incremental compilation — wire query.mach into the build


octalide:
`src/lang/query.mach` is already a complete revision/dependency-tracking memoization DB (kinds `Q_TOKENIZE`…`Q_CODEGEN`…`Q_LINK`) but is DEAD scaffolding — only input kinds are registered, the build never calls it, and it is in-memory only. This is an INTEGRATION job, not a rewrite: register the derived query kinds, route the driver/build stages through `query.get/compute`, record cross-query deps, and persist on-disk under `<out>`. First slice: per-module hash-and-skip (reuse an on-disk `.o` when a module's inputs are unchanged). Touches `src/lang/session.mach`, `driver.mach`, `cmd/build.mach`, `query.mach`. Complicated by the whole-graph sema snapshots (`build_sema_deps`).

## Roadmap (sub-issues)

Staged toward query-engine-driven incremental compilation. Parse caching shipped first; resolve/sema reuse needs the firewall + stable identity.

- [x] #1576 — parse-AST caching (hand-rolled `AstCache`, first increment)
- [x] #1579 — stable cross-build module identity (prereq for resolve/sema reuse)
- [x] #1580 — drive parsing through the query engine (`Q_PARSE`), retire `AstCache`
- [x] #1581 — `Q_EXPORTS` public-surface fingerprint firewall
- [x] #1582 — query-driven name resolution (`Q_RESOLVE`) — depends on #1579 #1580 #1581
- [x] #1583 — query-driven type-checking (`Q_SEMA`) — depends on #1582
- [x] #1858 — test-mode main-neutralization must not mutate query-owned IR (PR #1867)
- [x] #1859 — lowered-surface fingerprint firewall for `Q_LOWER` (back-half twin of #1581; PR #1870)
- [ ] octalide/mach-lsp#117 — drive the LSP from the query engine (set_input + re-ask)

## Status (2026-07-03)

The compiler side is COMPLETE. The back half (#1618) landed 2026-07-03 with Q_LOWER /
Q_CODEGEN / Q_LINK bound, followed the same day by diagnostics replay on reused phases
(#1586), the live-buffer source-overlay API (#1588), query-owned-IR immutability (#1858),
and the lowered-surface firewall (#1859). Remaining axes are on-disk persistence of the
query DB (not yet filed as its own issue) and downstream consumers (the LSP already
drives the front-half engine; mach-lsp#117 closed).

## Status (2026-06-26)

[...]

PR #1989, opened 2026-07-09
feat(driver): demand-driven [step.X] engine (dep steps, root homing, content cache, globs)

### LINKED ISSUE #1970: feat(build): [step.X] engine — shell commands, demand-driven, query-cached
Child of #1966; design in #1964. Subsumes #1943 (boom's interim Makefile is the removal target once this lands). BLOCKED until the schema and link-entries children of #1966 merge.

## Scope

- **Execution**: `cmd` is one string run through the platform shell (`sh -c` on posix, `cmd.exe /C` on windows) from the project root, environment inherited, via `std.process.exec` (dep/mach-std/src/process/exec.mach). No sandbox, no policing — a step may shell out to anything (#1964).
- **Demand**: a step runs only when demanded — a selected `local` link entry whose `path` matches the step's `out`, another step's `need`, or an artifact's `need`. Cycles error. Steps carry no filters; conditions live on entries. `in` globs (`*`/`**`) expand at run time, sorted; empty match = hard error.
- **Homing**: `{project.out}` resolves to the ROOT project's expanded `out` in every manifest of the closure — dep step outputs land in the consumer's out tree like compiled dep objects (`out/linux/debug/obj/std/*` precedent); dep checkouts are never written to.
- **Caching**: new query kind following the `Q_LINK_CONFIG` pattern (`fp_file_content`, driver.mach:4720): fingerprint = sorted expanded `in` paths + content hashes + expanded `cmd`; skip when unchanged AND outputs exist. Step outputs consumed as link inputs are already content-hashed by `Q_LINK_CONFIG`, so downstream invalidation is free.
- **Ordering**: demanded steps run before the demanding artifact's link, sequential in dependency order (parallel execution is an explicit follow-up, not this issue).

## Acceptance

- End-to-end: a miniz-shaped vendored-C dep (one `[step]`, one exported `local` entry) builds through a consumer with nothing in the consumer manifest; warm rebuild skips the step; touching the .c, the cmd, or deleting the .o re-runs it.
- Unit tests for demand computation, cycle error, glob expansion, fingerprint composition.
- int/ case only if it fits existing conventions (cc exists on CI runners; gate via targets.conf as the framework prescribes).
- Full bar: build from source, unit suite through the freshly built compiler, self-host fixpoint, int/ matrix.

octalide:
Re-implements the `[step.X]` build-step engine per the design in #1964 and the audit in #1970. Child of #1966; subsumes #1943.

## What changed
- **Demand-driven execution.** A step runs only when demanded — a selected, target-matching `local` link entry whose homed path matches the step's `out`, an artifact's `need`, or another step's `need` — never automatically; a `need` cycle is a hard error. `manifest.plan_steps` computes the demanded set in needs-first order; `step_producing_out` is the single match shared with the #1969 up-front local-path validation.
- **Dependency export steps.** `execute_dep_steps` runs each dep's export-demanded steps (`plan_export_steps`) from the dep checkout, homing `{project.out}` via a climb path back into the **root** out tree; `resolve_cascade_libs` homes the matching link input to the same path. Fixes the vendored-C dependency link failure (conatus).
- **Content-fingerprint caching.** A per-step stamp at `{project.out}/.steps/<owner-id>.<step>.stamp` holds the sha256 of the sorted expanded `in` paths + their content hashes + the expanded `cmd` (reused Q_LINK_CONFIG helpers). Skip iff the stamp matches and every output exists; fail toward running otherwise. Stamp not query kind because the query DB is in-memory per process today (migration to the persisted store deferred to #1164, noted in code).
- **Glob `in`.** `*`/`**` expand sorted against the project root, reusing the compiler's own directory walk; an empty match is a hard error.
- **Closed template set.** `{project.out}`/`{target.name}`/`{profile.name}` only; unknown `{a.b}` is a hard error; shell `$VAR` left to the shell. Removed the run-all loop, the mtime staleness check, and the permissive expanders.

## Verification (from-source, linux-x86_64)
- Build OK; unit suite **558 passed / 0 failed** (dev baseline 553 + 5: `plan_steps`, `plan_steps:cycle`, `plan_export_steps`, `wildcard_match`, `step_fingerprint`).
- Self-host **fixpoint byte-identical** (stage1 == stage2); **int linux 42/42**.
- Caching e2e (mach-audio): clean-run → warm-skip → touch-noop (content-based, not mtime) → content-change rerun → deleted-.o rerun → cmd-edit rerun, all as specified.
- Glob e2e: `vendor/*.c`/`*.h` expand and run; a non-matching glob errors.
- **Real-world gate: conatus builds green** with this compiler — the dep step runs, `miniaudio.o` lands in the consumer out tree (not the dep checkout), links, and warm rebuilds skip it.

Rebased onto dev `13eb7e4c` (post #1985/#1986/#1990); demand computation is filter-aware, matching #1990's `link_matches_target` guard.

Closes #1970

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+Re-implements the `[step.X]` build-step engine as demand-driven with content-based caching (#1970). A step runs only when demanded — by a selected, target-matching `local` link entry whose path matches the step's `out`, an artifact's `need`, or another step's `need` — never automatically; cycles are an error. A dependency's export steps now execute (from the dep checkout, homing `{project.out}` into the consumer out tree), so vendored-C dependencies that previously failed to link now build. Step paths and `cmd` use the closed template set `{project.out}`/`{target.name}`/`{profile.name}`; an unknown `{a.b}` is a hard error.
+
+### Added
+- steps: `in` globbing (`*`, `**`) with sorted, stable expansion; an empty match is an error.
+
+### Changed
+- steps: Demand-driven execution — a step runs only when a link entry, artifact, or another step demands it; platform-filtered steps never run on a non-matching cell.
+- steps: Content-fingerprint caching (a persisted per-step stamp over the inputs' content and the expanded `cmd`) replaces mtime staleness — a warm rebuild skips unchanged steps, a `cmd` or source edit re-runs them.
+
+### Fixed
+- driver: A dependency's `[step.X]` outputs are produced (in the consumer out tree) before the link, instead of failing with "cannot find link input".
+
 ## [3.0.2] - 2026-07-07
 
 Fixes transitive dependency resolution to resolve flat relative to the current root directory instead of nesting them.
--- a/src/lang/driver.mach
+++ b/src/lang/driver.mach
@@ -427,7 +427,8 @@ pub fun build_project(s: *session.Session, project_root: str, manifest_path: str
     sel.artifact     = "";
     sel.want_lib     = false;
     sel.has_artifact = false;
-    ret build_project_sel(s, project_root, manifest_path, ?sel);
+    # the doc/tooling entry: load the graph but never run steps (no shell in a doc pass).
+    ret build_project_impl(s, project_root, manifest_path, ?sel, false, false, false);
 }
 
 # one (target, artifact) cell a `mach build` invocation compiles.
@@ -582,7 +583,7 @@ pub fun resolve_artifact_path(alloc: *A.Allocator, project_root: str, manifest_p
 # manifest target's tuple and flips on the union load walk (see load_config /
 # walk_comptime_if_union). `all_own_src` additionally loads every own-source module
 # as a collection root (see load_all_own_src), for `mach test`.
-fun build_project_impl(s: *session.Session, project_root: str, manifest_path: str, sel: *manifest.Selection, for_union: bool, all_own_src: bool) R.Result[Project, str] {
+fun build_project_impl(s: *session.Session, project_root: str, manifest_path: str, sel: *manifest.Selection, for_union: bool, all_own_src: bool, run_steps: bool) R.Result[Project, str] {
     var p: Project;
     init_project(?p, s);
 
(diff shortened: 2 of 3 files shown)


octalide:
Adversarial verification (6 independent surfaces, each with a from-source branch build): **NOT ready to merge — two confirmed blockers**, both reproduced end-to-end. Independent reproduction (surface F) confirmed every claimed number (558/0, int 42/42, byte-identical fixpoint, conatus cold+warm green) — the blockers live in corners the acceptance cases didn't reach.

**Blockers:**
1. **`dep_out_home` overshoots on absolute project roots.** It counts every segment of `vendor_root`, which is absolute whenever the build is invoked as bare `mach build` (the default — #1985's rule-1 resolver returns the absolute cwd walk result) or with an absolute path. The climb then lands at the filesystem root: `step miniz: cc … -o ../../../../../out/linux/debug/obj/miniz/miniz.o` → fatal. The same project builds fine as `mach build .` (relative root) — which is why the acceptance runs passed. Fix directions: climb the dep path *relative to project_root* (the flat store means it's always `dep/<alias>` → fixed depth), or home outputs to an absolute path and sidestep the climb entirely.
2. **Step-output parent directories are never created, and steps run before the out tree exists.** A pristine clean build of the RFC's own miniz example (bare `cc -c -o {project.out}/obj/miniz/miniz.o`, no `mkdir`) fails at the assembler. conatus/mach-audio pass only because their `cmd` happens to embed `mkdir -p`. Fix: after the cache-miss decision, `create_dir_all` the parent of every declared expanded `out` before exec.

**To fold into the same round (small, all in touched code):**
- `mach doc` executes demanded steps (parity with dev, which ran *all* steps — so not a PR regression, but a doc/tooling load must never run shell). Gate step execution to build/test flows only.
- Glob shape gap: wildcards only work in the final segment, and `dir/**/pat` only with a single trailing pattern; other shapes silently zero-match with a misleading "matched no files". Reject unsupported shapes with an error that says what IS supported.
- Glob semantics: don't match dot-files or descend dot-directories (`**` currently walks `.git`); a `*` in a step `out` should be a strict parse error, not literal-treated.
- Wire the dead `is_valid_id` into parse for `[project].id` and step names (a `/` in either silently defeats stamp caching; `.` in an owner id makes `a.b`+`c` collide with `a`+`b.c` stamps).
- Cycle diagnostic: name the re-entered step.

Minor/accepted (no action this PR): cycle message genericity beyond naming, symlinked sources excluded by lstat (consistent with the module walk), '/'-separator globs on windows (noted for the windows track).

One more finding is being classified separately (bare `mach build .` SIGSEGV when no `[profile.release]` is declared — appears to predate this branch; will be filed independently if so).

Issue #1919, opened 2026-07-04
debuginfo: DWARF aggregate/composite type DIEs (+ DW_OP_piece fragmentation)


octalide:
## What
Give aggregate and composite-typed variables real DWARF type + variable DIEs, and — folded into the same tier — DW_OP_piece / DW_OP_bit_piece for values fragmented across multiple homes:
- `DW_TAG_structure_type` / `DW_TAG_union_type` / `DW_TAG_array_type` with `DW_TAG_member` children and byte-offset layout;
- typed pointers (`DW_TAG_pointer_type` to the pointee DIE) instead of the raw `ptr` base type;
- `DW_TAG_variable` DIEs for aggregate locals/params, with `DW_AT_type` referencing the composite DIE;
- once such a variable can be split across a register + a slot (or several registers), `DW_OP_piece` framing over the existing VarLoc channel to describe each fragment.

## Why this is the prerequisite (and why piece is folded in, not standalone)
The producer today filters variables to primitive `sem_base_type` (u8..f64, ptr), so **aggregates emit no DIE at all** and no `DW_AT_type` names a composite. `DW_OP_piece` was split out of #1706 because it had no reachable input: there is nothing to fragment when the variable doesn't exist in the debug info, and no scalar is register/slot-fragmented in the current backend. Fragmentation only becomes meaningful once aggregates are tracked with layout, so piece belongs in this tier rather than as a separate framing-only issue — my structural call is to fold it in.

This tier also **un-defers #1907's non-primitive-variable deferral**: #1704/#1705 emit variable DIEs only for primitive-typed locals; composite type DIEs are exactly what lets non-primitive locals get a DIE at all (that link is on the record from the #1704 report).

## Acceptance
- A struct/array/union local shows its members with correct names/types/values under gdb/lldb on an optimized `-g` binary; a typed pointer resolves its pointee; a fragmented aggregate renders `DW_OP_piece` per home.
- `llvm-dwarfdump --verify` clean; `-g` additivity (dbg lane step 5) preserved on all ISAs.

Child of #1594. Prerequisite surfaced by #1706 (the computed/salvaged tier, which shipped without piece).

octalide:
[...]

PR #2006, opened 2026-07-10
fix(dwarf): locate register-passed by-value aggregate parameters (#1954)

### LINKED ISSUE #1954: debuginfo: register-passed by-value aggregate parameters get no DW_AT_location
Split out of the #1919 adversarial review (finding 5).

## What
A register-passed by-value aggregate parameter (e.g. `fun take(p: Point)` with Point a <=16-byte struct passed in RDI/RSI on sysv64) gets a `DW_TAG_formal_parameter` with its composite `DW_AT_type` but **no `DW_AT_location`**, so a debugger cannot show the struct argument.

## Why
`mir/abi.emit_param_slot` keys a SLOT_SPILL directly on the param vreg `pv` (not a separate slot-key), and the dbg-value binds `pval`=that vreg. regalloc excludes slot-owner vregs from allocation, so `pv.assigned == NIL` and `pv.spill_slot == -1`; `encode.push_varloc`'s `or (vr.spill_slot >= 0)` gate is false, so it reports `VARLOC_NONE`.

This is **invariant-2-compliant** (conservative omission, never a wrong home) — #1919's aggregate-location work is correct as far as it goes — but the feature silently does not apply to register-passed aggregate params. Fixing it needs the encoder/ABI to expose the param's reconstruction slot (a real `spill_slot`) so `push_varloc` can report `VARLOC_FRAME`, at which point the existing `DwVar.byref` fbreg+deref path renders it. Out of scope for the DWARF producer alone.

Child of #1594. Surfaced by #1919.

octalide:
Closes #1954. Split out of the #1919 adversarial review (finding 5).

## Problem
A register-passed by-value aggregate parameter (e.g. a <=16-byte struct in RDI:RSI on sysv64) got a `DW_TAG_formal_parameter` with its composite `DW_AT_type` but **no `DW_AT_location`** — a debugger could not show the argument. The ABI reconstructs the parameter into a `mir.SLOT_SPILL` frame slot it owns, but regalloc excludes the slot-owner vreg (no assigned register, `spill_slot == -1`), so `encode.push_varloc` reported `VARLOC_NONE`.

## Fix (encode.mach + dwarf.mach)
- `push_varloc` now recognizes a vreg owning a `SLOT_SPILL` slot and reports a **direct frame location** (`VARLOC_FRAME`, new `off_storage` flag).
- Because that slot holds the aggregate's **bytes** (not a pointer to them, as a regalloc spill of a byref pointer would), the location is `DW_OP_fbreg <off>` **without** the byref `DW_OP_deref`. The `off_storage` flag threads through `varloc_home` → `LocRange` → `DwVar` → `build_loc_ops`. Regalloc-spilled byref pointers keep their `fbreg+deref`; named/scalar homes are byte-identical.

The honest-DWARF question (per the #1919 program rules): the reconstruction slot IS the storage, so a direct frame location is correct and cheap — no `DW_OP_piece` composition is needed (and none is emitted, since that path stays deferred).

## gdb money-proof
Fixture `fun take(pt: Point)` with `Point` a 16-byte struct:
- **pre-fix**: `pt` has no `DW_AT_location` (optimized-out).
- **post-fix**: `DW_AT_location (DW_OP_fbreg -24)`; `llvm-dwarfdump --verify` clean; at a post-prologue breakpoint gdb prints `pt = {x = 11, y = 31}` (transcript in thread).

## Tests
- `encode.push_varloc:slot_spill_owner_is_direct_frame` — the new branch reports `VARLOC_FRAME` + `off_storage`.
- `dwarf` location test extended: a `storage` byref frame home renders `DW_OP_fbreg` with no deref (contrast the pointer-spill case that keeps the deref).

## Verification
- unit suite: 571 (570 baseline + 1); dbg/verify.sh both profiles; self-additive both profiles; int linux 42/42; self-host fixpoint — in thread.

## Scoped out
gdb's default `break <fn>` prologue heuristic lands one instruction before the aggregate spill (mach emits no `prologue_end` for any function), so `print pt` there reads zeros; reading is correct at any post-prologue PC. That's a separate line-table gap — recommend a follow-up to emit `DW_LNS_set_prologue_end`.

### DIFF
--- a/src/lang/be/codegen/dwarf.mach
+++ b/src/lang/be/codegen/dwarf.mach
@@ -364,6 +364,10 @@ rec DwVar {
     range_count: u32;
     loclist_off: u32;
     byref:       bool;
+    # true when a byref aggregate's frame home addresses its storage directly (a
+    # register-passed param reconstructed into a slot it owns, #1954), so the location
+    # omits the pointer deref; false when the frame slot holds the storage pointer
+    storage:     bool;
     # the inline instance this variable belongs to (0 = the function's own body, else a
     # 1-based inline-site id); the #1707 producer nests a non-zero one under its instance
     inline_site: u32;
@@ -382,12 +386,13 @@ rec DwVar {
 # off:       frame-base offset when kind == VLOC_FRAME
 # net:       salvage net constant (0 = identity)
 rec LocRange {
-    start: u32;
-    end:   u32;
-    kind:  u8;
-    reg:   i32;
-    off:   i64;
-    net:   i64;
+    start:   u32;
+    end:     u32;
+    kind:    u8;
+    reg:     i32;
+    off:     i64;
+    net:     i64;
+    storage: bool;
 }
 
 # one inlined-subroutine instance, emitted as a DW_TAG_inlined_subroutine nested under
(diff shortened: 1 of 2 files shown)


octalide:
## gdb money-proof

Fixture: `fun take(pt: Point)` with `Point { x: i64; y: i64; }` (16 bytes, passed in RDI:RSI).

**Pre-fix:** `pt` has no `DW_AT_location` (formal_parameter carries only name + type).

**Post-fix:** `DW_AT_location (DW_OP_fbreg -24)` — a direct frame location, `llvm-dwarfdump --verify` clean. At a post-prologue breakpoint:
```
Breakpoint 1, take (pt=...) at ./src/main.mach:7
(gdb) print pt      -> $1 = {x = 11, y = 31}
(gdb) print pt.x    -> $2 = 11
(gdb) print pt.y    -> $3 = 31
```
The location is a direct `fbreg` (no `DW_OP_deref`): the slot holds the aggregate's bytes, so the deref that a regalloc-spilled byref pointer needs would be wrong here.

## Verification (fresh from-source compiler, linux-x86_64)
- unit suite: **571 passed, 0 failed** (570 baseline + 1)
- `dbg/verify.sh`: all debug-info checks passed (21 target builds — debug + release, all ELF ISAs, incl. the #1919 no-bare-register aggregate guard — so no wild register location was introduced)
- `dbg/self-additive.sh` (release capstone): -g additive, 5 PT_LOAD segments identical

PR #2008, opened 2026-07-10
fix(dwarf): emit prologue_end so break <fn> clears parameter homing (#2007)

### LINKED ISSUE #2007: debuginfo: line table emits no DW_LNS_set_prologue_end — break <fn> lands mid-prologue for aggregate params
Part of the pre-SIMD debuginfo close-out (context: #1594; found and scoped out during #1954/#2006).

mach's line program emits no `DW_LNS_set_prologue_end` for any function, so debuggers fall back to heuristics for `break <fn>` placement. gdb's heuristic happens to survive scalar-param prologues but lands one instruction BEFORE the entry spill that reconstructs a register-passed aggregate parameter — so `break take; print pt` reads zeros, while any post-prologue PC prints correctly (verified in #2006's money-proof).

## Repro

The #2006 fixture: `fun take(pt: Point)` (16-byte struct in RDI:RSI), `-g`, `break take` in gdb, `print pt` → zeros; `next` once → `{x=11, y=31}`.

## Expected

`break take` stops at the first post-prologue PC (the flagged row), where every parameter — scalar and aggregate — is readable.

## Actual

No prologue_end flag exists in the line table; the breakpoint lands mid-prologue for functions with entry aggregate spills.

## Environment

dev (post-#2006), all three ELF ISAs; format-agnostic producer issue, not ISA-specific.

## Fix

Emit `DW_LNS_set_prologue_end` on the line row at the first instruction after parameter homing/entry spills complete (the encoder knows where homing ends per function). Must hold on all three ISAs, at both profiles, and stay -g byte-additive. Falsifiable: `llvm-dwarfdump --debug-line` shows prologue_end exactly once per function at the post-homing row; the #2006 gdb repro prints real values at `break take` with no `next`.

octalide:
Closes #2007. Follow-up scoped out of #1954/#2006.

## Problem
mach's line program emitted no `DW_LNS_set_prologue_end`, so a debugger's `break <fn>` fell back to a prologue heuristic. gdb's heuristic lands one instruction before the entry spill that reconstructs a register-passed aggregate parameter, so `break take; print pt` read zeros (the #2006 money-proof needed a manual `next`).

## Fix (dwarf.mach, line program only)
`build_line`, per function:
1. Anchors a line row at the function entry (`low_pc`) — mach emitted no row over the prologue/homing, so a debugger's line-based prologue skip had nothing to anchor on and fell back to arch heuristics. The entry row uses the first row's file/line.
2. Flags the first post-homing row `DW_LNS_set_prologue_end` (opcode 10, already declared in the header's `standard_opcode_lengths`, so no header change).

`break <fn>` then stops at the prologue_end row, where every parameter (scalar and aggregate) is live. Line-table only; loadable segments untouched, so `-g` stays byte-additive.

## Falsifiable evidence
- `llvm-dwarfdump --debug-line`: `prologue_end` appears once per function, on the first row after the entry (take: entry row at `low_pc`, `prologue_end` on the next row past the RDI:RSI spill).
- gdb (the #2006 repro, default `break take`, no `next`): `print pt` -> `$1 = {x = 11, y = 31}`. `--verify` clean.

## Verification
- unit suite: 573 (572 baseline + 1 build_line test); dbg/verify.sh (both profiles, all ELF ISAs); self-additive (both profiles); int linux; self-host fixpoint - in thread.

### DIFF
--- a/src/lang/be/codegen/dwarf.mach
+++ b/src/lang/be/codegen/dwarf.mach
@@ -211,10 +211,14 @@ val NAME_OMIT: u32 = 0xFFFFFFFF;
 
 # DWARF 5 line-number program: standard opcodes, extended opcodes, and the tuned
 # special-opcode parameters this producer emits
-val DW_LNS_copy:         u8 = 0x01;
-val DW_LNS_advance_pc:   u8 = 0x02;
-val DW_LNS_advance_line: u8 = 0x03;
-val DW_LNS_set_file:     u8 = 0x04;
+val DW_LNS_copy:          u8 = 0x01;
+val DW_LNS_advance_pc:    u8 = 0x02;
+val DW_LNS_advance_line:  u8 = 0x03;
+val DW_LNS_set_file:      u8 = 0x04;
+# marks a row as the first past the prologue (parameter homing / entry spills), so a
+# debugger's `break <fn>` stops where every parameter is live (#2007). declared in the
+# header's standard_opcode_lengths (opcode 10, zero args) already
+val DW_LNS_set_prologue_end: u8 = 0x0a;
 
 val DW_LNE_end_sequence: u8 = 0x01;
 val DW_LNE_set_address:  u8 = 0x02;
@@ -1994,6 +1998,7 @@ fun build_line(s: *session.Session, funcs: *DwFunc, nfuncs: u32, rows: *enc.Line
         var cur_off:  u32 = fn.offset;
         var cur_line: i32 = 1;
         var cur_file: i64 = -1;
+        var first_row: bool = true;
 
         var r: u32 = 0;
         for (r < row_count) {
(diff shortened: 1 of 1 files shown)


octalide:
## Evidence (final build)

### llvm-dwarfdump --debug-line (take): entry row at low_pc + prologue_end
```
0x0000000000436e40      7      0      0   0             0       0  is_stmt
0x0000000000436e50      7      0      0   0             0       0  is_stmt prologue_end
```
### gdb — default `break take`, no `next`
```
Breakpoint 1, take (pt=...) at ././src/main.mach:7
$1 = {x = 11, y = 31}
```

## Verification (fresh from-source compiler, linux-x86_64)
- unit suite: **573 passed, 0 failed** (572 baseline + 1)
- dbg/verify.sh: all checks passed (21 target builds, debug + release, all ELF ISAs)
- self-additive (release capstone): -g additive, 5 PT_LOAD segments identical
- int/run.sh --target linux: all cases passed
- `--verify` clean; prologue_end appears once per function

Issue #1708, opened 2026-06-28
debuginfo: cross-TU .debug_* merge/dedup in linker


octalide:
## What
Merge/dedup .debug_* across multiple .o's in be/linker.mach (string-table merge, abbrev/type dedup, cross-unit offset relocation). For Mach-O, emit the merged, fully-relocated DWARF inline into the executable __DWARF — deliberately skipping the Apple debug-map/N_OSO + dsymutil/.dSYM path (the compiler is its own linker and relocates the DWARF itself). Can run in parallel after Tier 1.

## Why
Real programs are multi-TU; without merge/dedup, debug sections bloat and cross-unit refs break. Single-file builds already work from the earlier increments; this is the heavier dedup/merge across translation units, and it forces the Darwin executable path (#1702) onto the milestone exit.

## Acceptance
- A 2+ TU -g build produces one coherent DWARF graph; llvm-dwarfdump --verify clean on Linux and macOS executables.
- lldb/gdb debug the multi-TU executable (line stepping + locals across TUs).

Depends-on: #1700, #1702.

octalide:
Scope inherited from #1703 (subprogram + base-type DIEs): #1703 emits `DW_TAG_subprogram` DIEs **only for a module's own non-weak functions**. Weak monomorphized generic instances are deliberately excluded, because the linker deduplicates a weak function to a single winner across objects, so a non-winning object's subprogram DIE would relocate its `low_pc` outside its CU's contiguous `[low_pc, high_pc)` range — an invalid parent/child range that `llvm-dwarfdump --verify` rejects ("DIE address ranges are not contained by parent ranges"). Those instances keep their `.debug_line` coverage today but have no subprogram DIE (so they show mangled/`??` in backtraces).

This is the natural home for restoring them: cross-TU `.debug_*` merge/dedup (COMDAT-style) should keep a weak function's subprogram DIE grouped with its `.text` copy and drop both together on dedup, so exactly one CU describes the surviving copy and its range contains it. When that lands, re-enable subprogram DIEs for weak functions in the producer (`be/codegen/dwarf.mach`, the `funcs[si].weak` filter in `produce`) and verify the multi-module fixture's generic instances get named DIEs.

octalide:
## Status after PR #2009 (`.debug_abbrev` dedup) — remaining scope + `.debug_str` design spec

### Correctness finding (already held pre-PR)
[...]

PR #2009, opened 2026-07-10
fix(link): cross-TU .debug_* dedup — .debug_abbrev (#1708)

octalide:
Part of #1708 (cross-TU `.debug_*` merge/dedup). See the coordination note below — scope of the `.debug_str` string-merge is pending a decision.

## Context (measured on the -g self-host compiler, 146 CUs / 162 objects)
The linker already does **offset-correct concatenation** of `.debug_*` (per-module placements + rebased relocations); `llvm-dwarfdump --verify` is already clean on the multi-TU compiler. So #1708 is about **dedup / size**, not correctness. Current `.debug_*` breakdown:

| section | bytes | dedup potential |
|---|---|---|
| .debug_line | 1,565,337 | per-module (inherent) |
| .debug_info | 895,570 | per-CU (inherent, sans type-DIE dedup) |
| .debug_str | 405,420 | 38,585 strings, **10,297 unique** → ~250 KB |
| .debug_abbrev | 44,968 | 146 identical tables → **~44.6 KB (this PR)** |
| .debug_loclists | 117,820 | per-module (inherent) |

## This PR: `.debug_abbrev` dedup
Every module emits a byte-identical `.debug_abbrev` and every CU references it at offset 0, so 145 of 146 copies were dead bytes carried through the merge. Keep exactly one; later objects share it at offset 0.

**Result:** `.debug_abbrev` 44,968 → **345 bytes** on the -g compiler; `--verify` clean.

## Verification
- unit suite: 573/0; dbg/verify.sh (both profiles, all ELF ISAs); self-additive (5 PT_LOAD identical — non-loadable debug, so -g stays byte-additive); int linux; self-host fixpoint converged.

## `.debug_str` string-merge — design ready, scope pending
The bigger win (~250 KB). Producer emits per-module `<module>#.debug_str` anchors; strp = anchor_offset + addend. Plan: a pre-pass parses each module's `.debug_str`, builds a global dedup buffer + per-module (offset→deduped-offset) remap; the merged `.debug_str` becomes the deduped buffer; strp relocations resolve through the remap (with a content-equality check per site as the correctness guard). This is a relocation-core change; raised for a scope decision (proceed here vs. focused follow-up) rather than rushed.

### DIFF
--- a/src/lang/be/linker.mach
+++ b/src/lang/be/linker.mach
@@ -1399,6 +1399,14 @@ fun total_sections(modules: *of.ObjectImage, module_count: u32) u32 {
 fun accumulate_sections(s: *session.Session, modules: *of.ObjectImage, module_count: u32,
                         merged: *MergedSection, debug_names: *intern.StrId, debug_count: u32,
                         placements: *Placement, sec_base: *u32) R.Result[bool, str] {
+    # .debug_abbrev is byte-identical across every module (a deterministic, complete
+    # build_abbrev) and every CU references it at offset 0, so only the first copy is
+    # ever read. dedup it: a later object's table shares the first at offset 0 and adds
+    # no bytes to the merged section (#1708).
+    val an_r: R.Result[intern.StrId, str] = intern.intern(?s.interner, ".debug_abbrev");
+    if (R.is_err[intern.StrId, str](an_r)) { ret R.err[bool, str](R.unwrap_err[intern.StrId, str](an_r)); }
+    val abbrev_name: intern.StrId = R.unwrap_ok[intern.StrId, str](an_r);
+
     var flat: u32 = 0;
     var m: u32 = 0;
     for (m < module_count) {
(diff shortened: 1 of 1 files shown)


octalide:
## Rebased onto dev (post-P2a) + final verification

Rebased `fix/1708` onto `origin/dev` (a9e2e1fe, P2a merged) — clean, no conflict (the abbrev dedup lives in `accumulate_sections`, disjoint from P2a's shared/emit_shared path). Diff unchanged: `linker.mach` only (+38/−17).

**Abbrev dedup holds on P2a's new shared-object path** (which also carries debug sections through the same merge): a multi-TU `-g` shared library (`kind = "shared"`, 3 CUs) builds `ET_DYN` with `.debug_abbrev` deduped to **one 345-byte table**, `llvm-dwarfdump --verify` clean.

**Full bar, all green:**
- unit suite: **576 passed, 0 failed** (matches the new baseline; this is a linker size optimization — no unit tests added, verified end-to-end via `--verify` + section-size measurement + the bar)
- `dbg/verify.sh`: all checks passed (21 target builds, both profiles, all ELF ISAs)
- `dbg/self-additive.sh`: -g additive (5 PT_LOAD segments identical — `.debug_abbrev` is non-loadable)
- executable path (self-host compiler, 146 CUs): `.debug_abbrev` 44,968 → 345 bytes, `--verify` clean

Issue #2020, opened 2026-07-10
docs: amend types/operators/policy + manifest for shipped SIMD vectors


octalide:
Part of #1965.
Depends on #2013, #2014, #2015, #2016, #2017.

## What

Docs document shipped reality. After tier 1 lands — the gate (#2013) seeds the ten 128-bit vector types plus the sema op table and `IRT_VECTOR`, the three backends (#2014 SSE2, #2015 NEON, #2016 rv64gc scalarization) make every operator real, and the lever (#2017) adds the required `simd` profile key — four hand-written reference pages still describe vectors as "planned / not yet implemented." This child rewrites the prose to match what the compiler actually does, and nothing else: it ships no `.mach` source and no manifest fixtures.

Scope boundaries: the seeded types / sema table / `IRT_VECTOR` are #2013; the per-ISA lowering is #2014 / #2015 / #2016; the `simd` parser change and the in-tree `mach.toml` fixture sweep (28 tracked manifests) are #2013; the external 13-repo ecosystem sweep rides the SIMD release via #2017 — this docs child only edits the human-readable prose and the doc pages' own example snippets. DWARF vector DIEs (#2018), splat sugar (#2019), and the stdlib simd/math library (#2021) are their own siblings and get their own doc updates; do not describe them here as shipped.

### 1. `doc/language/types.md` — the `## SIMD vectors` section

The section (lines 22-39) is a "**Not yet implemented**" blockquote over a planned grammar. Replace it with shipped-feature prose that states:

- **The ten seeded types, exhaustively**: `f32x4 f64x2 i8x16 i16x8 i32x4 i64x2 u8x16 u16x8 u32x4 u64x2`. These are the complete compiler-seeded vector set — all 128-bit, single-`x`.
- **Grammar shape** `<u|i|f><width>x<count>`, a **single** `x`. Correct the current lines 37-39, which claim `(x<count>)*` and that `f32x4x4` is "legal grammatically" — under the settled depth-1 contract there are no multi-`x` language types; a name like `f32x4x4` is not a vector type. A matrix is an algorithm over vectors and lives in a library over vector elements (#2021), not the language.
- **Size and alignment**: every seeded vector is exactly 128 bits — `$size_of` and `$align_of` are both `16`. Arrays-of-vectors (`[4]f32x4`) and pointers-to-vectors (`*f32x4`) are ordinary composite types over a vector element.
[...]

Issue #2019, opened 2026-07-10
sema: splat sugar for vector literals


octalide:
Part of #1965.
Depends on #2013.

## What

The epic settles vector construction on **full-arity literals** —
`f32x4{1.0, 2.0, 3.0, 4.0}`, mirroring array literals — and explicitly
defers *splat sugar* (a shorthand that broadcasts one value across all
lanes) to tier 2. This child adds that shorthand. It is **frontend/sema
only**: a purely syntactic desugar to the full-arity vector literal that
#2013 already builds, checks, and lowers. It introduces **no new IR,
no IRT_VECTOR change, no isel/ABI/backend change, and no manifest/profile
interaction** — the profile `simd` lever is #2017's, IRT_VECTOR and
`infer_vector_lit` are #2013's, and the normative prose spelling of
the chosen syntax in `doc/language/types.md` §SIMD vectors lands with
#2020. Because the desugar completes before lowering, every target —
SSE2 (#2014), NEON (#2015), and the scalarized rv64 path
(#2016) — sees the identical full-arity node; splat is
target-independent.

**Open question — surface syntax needs owner sign-off at tier-2
dispatch.** The design below frames a concrete recommendation with
alternatives and is ready to ratify; the spelling is not silently chosen.

### Recommended surface syntax — the single-element brace form

`f32x4{x}` → splat `x` across all four lanes; more generally a vector
literal with **exactly one** element broadcasts that element to every
lane of the resolved vector type.

The recommendation is grounded directly in existing literal semantics.
Array and (per #2013) vector literals are **exact-arity everywhere**:
`infer_array_lit` (`src/lang/fe/sema/infer.mach:1611-1643`) enforces the
declared count exactly, reporting "too few elements" below it and "too
many elements" above it (`infer.mach:1636-1641`), and #2013's
`infer_vector_lit` mirrors that rule against the vector's lane count.
Every seeded vector has **count ≥ 2** (the narrowest are `f64x2 i64x2
u64x2`), so under exact-arity a one-element brace `f32x4{x}` is, today,
*always* the "too few elements" error — an unused syntactic slot with no
valid meaning. Repurposing exactly that slot as splat is therefore
unambiguous by construction, costs **zero new tokens**, and mirrors the
very literal it becomes.

**Desugar.** `f32x4{x}` ⇒ `f32x4{x, x, x, x}` (N = the resolved lane
count). The rewrite materializes N references to the single element into
[...]

Issue #2018, opened 2026-07-10
codegen: dwarf vector-type DIEs (DW_AT_GNU_vector) for the seeded vector types


octalide:
Part of #1965.
Depends on #2013.
Depends on #2014.

## What

Give the ten seeded 128-bit vector types (`f32x4 f64x2 i8x16 i16x8 i32x4 i64x2 u8x16 u16x8 u32x4 u64x2`) a DWARF type representation so `-g` builds describe vector-typed locals and parameters, and a debugger prints a vector lane-by-lane. This is the debug-info follow-up in the epic's tier-2 slice; it consumes the semantic vector type created by the tier-1 gate (#2013) and reads it exactly as the existing composite type-DIE path reads records/arrays.

**Encoding.** Emit each vector as a `DW_TAG_array_type` carrying `DW_AT_GNU_vector` (the established GNU encoding gdb and lldb recognize to distinguish a hardware vector from a plain array) over the element base type, with a single `DW_TAG_subrange_type` child whose `DW_AT_count` is the lane count. The vector's 16-byte size follows from element-base-size × lane count for every seeded type (4×f32, 8×i16, 16×i8, … all = 16), which is exactly the byte layout the tier-1 IR type and codegen already emit — so the DIE agrees with the emitted layout by construction. Mirror the existing plain-array abbrev, which carries no explicit `DW_AT_byte_size`, so the new abbrev differs from `ABBREV_ARRAY` only by the added `DW_AT_GNU_vector` attribute; consumers synthesize the vector's name and size from element × count.

**Where it lands (all in `src/lang/be/codegen/dwarf.mach` unless noted):**

[...]

Issue #2017, opened 2026-07-10
driver: [profile.X] simd lever semantics (scalarize/require) over the target capability fact


octalide:
Part of #1965.
Depends on #2013.
Coordinate with #2016 (detection/note seam).

## What

This is the **profile-lever child** of the SIMD program. It implements the `simd` lever's *semantics* — the meaning of `scalarize` vs `require` and how that meaning is resolved from a target capability and a project/profile setting into build behavior. The syntactic parsing of the `simd` key lands with the frontend/sema gate (#2013); this child depends on the manifest already accepting and storing the key, and owns everything downstream: the target capability fact, the resolve/threading path, the `require`-mode build error, the consumer-overrides-library effective-setting guarantee, and the record of the ecosystem sweep.

The design (settled in #1965, not re-litigated here) draws one hard line: **hardware SIMD availability is a target FACT (compiler-known); what to do on an incapable target is a project/profile LEVER.** The two must not be conflated — a target does not choose whether it has SSE2/NEON, and a profile does not choose whether an op scalarizes, only whether scalarization is *tolerated* or is an *error*.

### The capability FACT (target definition)

Capability is the target's `MachineModel.has_v128: bool` fact (`src/lang/target/isa.mach:249-278`), landed by #2013's shared scaffolding as a declared field, populated per arch in the three model blocks:

- x86_64 → `true` (SSE2 baseline). Model block in `src/lang/target/isa/x64/register.mach` (the `xmm` bank already declares `RC_FLOAT` `bit_width=128` at register.mach:214).
- aarch64 → `true` (NEON baseline). Model block in `src/lang/target/isa/arm64/register.mach` (the `vector` bank already declares `RC_FLOAT` `bit_width=128` at register.mach:158).
- riscv64 (rv64gc) → `false` (no V extension; the `fpr` bank is `bit_width=64` at `src/lang/target/isa/riscv64/register.mach:188`).

[...]

PR #2024, opened 2026-07-10
feat: SIMD vector types substrate — seeded types, lane ops, IRT_VECTOR, simd manifest key (#2013)

### LINKED ISSUE #2013: sema: 128-bit vector types — seeded set, lane ops, IRT_VECTOR substrate, and the required simd manifest key
Part of #1965.
Blocks #2014, #2015, #2016, #2017.

## What

This is the **tier-1 gate** for the SIMD vector-types program: the frontend + sema + middle-end substrate that turns the ten 128-bit vector types into real semantic types, plus the `mach.toml` parser change that reserves the strictness lever. Everything a backend or the lever attaches to — the seeded type names, the full-arity literals, comptime-constant lane access, the lane-wise operator table, comparison→mask, the `IRT_VECTOR` IR type with verifier rules and sound default passes, and the required `simd` profile key — lands here. This child does **not** select a single vector instruction on any target, does **not** enforce the lever, does **not** emit DWARF, and does **not** touch the language docs or add splat sugar. Those are the siblings named below.

The types are portable-vector-model semantic types: all ten are legal on every target regardless of hardware; whether a lane-wise op lowers to one instruction or a scalar expansion is a backend concern, and until the backend children land, a vector op that reaches isel is a defined per-target compile error (slice 7). The "seeded per-target" phrasing in #1965's child list refers to backend acceleration, not to the type surface — the type surface is seeded identically on every target.

### 1. Grammar / seeding — the ten names, seeded set only, no generative grammar

Seed exactly `f32x4 f64x2 i8x16 i16x8 i32x4 i64x2 u8x16 u16x8 u32x4 u64x2` (shape `<u|i|f><width>x<count>`, all 128-bit). There is **no** parametric `x<count>` grammar and **no** depth-2 (`f32x4x4`) — the resolver recognizes exactly these ten spellings, nothing more.

`f32x4` already lexes as a plain `KIND_IDENT` (`src/lang/fe/token.mach:13`), so the lexer is untouched; recognition happens in the resolver's primitive registry (`src/lang/fe/resolve.mach`). Today `prim_names: [11]intern.StrId` / `prim_syms: [11]SymbolId` (resolve.mach:209-210, `PRIM_COUNT = 11` at :228) are seeded by `seed_primitives` (resolve.mach:800-813) via `seed_one_primitive`, which interns each name and records a `SYM_PRIM` symbol carrying the primitive's `TypeKind` ordinal in `slot` (resolve.mach:823-835); `prim_symbol_for` (resolve.mach:844-851) linear-scans it in type position with priority over value scope (used at resolve.mach:2696, :2926). A vector is not describable by a single scalar `TypeKind` ordinal (it needs element + lane count), so add a **parallel vector registry** `vec_names[10]`/`vec_syms[10]` seeded by a sibling `seed_vectors`, and a new symbol kind `SYM_VECTOR` (next after `SYM_PRIM = 11`, resolve.mach:72) whose `slot` carries the **pre-interned semantic vector TypeId** directly (not a kind ordinal). `prim_symbol_for` consults the vector registry alongside the prim one so `f32x4` resolves in type position. In sema, `type_for_symbol` (`src/lang/fe/sema/infer.mach:1749`) gains a `SYM_VECTOR` branch returning the stored TypeId (parallel to the existing `SYM_PRIM` → `prim_or_error(sc, sym.slot::type.TypeKind)` at infer.mach:1750).

The semantic type universe (`src/lang/type.mach`) gains `TYPE_VECTOR: TypeKind = 20` (next after `TYPE_SECRET = 19`, type.mach:75) with a payload record `{ element: TypeKind; lanes: u32 }` (element constrained to the ten scalar prim kinds; lanes ∈ {2,4,8,16}), an `intern_vector(ti, element, lanes)` mirroring the compound interners, and the ten vector types pre-interned at `init` (alongside the `prims` table, type.mach:387-410) so each has a stable TypeId to seed. Add an `is_vector` predicate keyed on `TYPE_VECTOR` — vectors are **not** routed through `prim_desc` (type.mach:99-132), which stays scalar-only. Note the latent coupling flagged by the scout: `type_str_len`/`write_type_str` (`src/lang/fe/sema/check.mach:1244`, `:1294`) switch on prim TypeId constants directly, so a vector spelling arm (e.g. render `f32x4`) is added to **both** functions for diagnostics.

### 2. Full-arity literals `T{a, b, c, d}`

`f32x4{1.0, 2.0, 3.0, 4.0}` mirrors array literals: **positional**, full-arity, one element per lane. Wrong arity is a compile error (too many / too few lanes). No splat (`f32x4{1.0}` broadcast is #2019, tier 2).

Array literals are bracket-prefixed (`[N]T{...}` → `parse_array_literal`, `src/lang/fe/parser/expr.mach:792-826`, node `EXPR_KIND_ARRAY_LIT`); a named type before `{` currently routes to `parse_typed_literal`/`EXPR_KIND_STRUCT_LIT` (expr.mach:834-872), which expects named fields. A vector literal is a named type with a **positional** brace body — a shape not parsed today. Add a positional typed-literal path (distinguished at parse time by the brace body lacking `ident :` field syntax) building a new `EXPR_KIND_VECTOR_LIT` node `ExprVectorLit { ty: TypeId; elems_start; elems_len }` (mirror `ExprArrayLit`, `src/lang/fe/ast/expr.mach:190-194`). Sema adds `infer_vector_lit` mirroring `infer_array_lit` (`src/lang/fe/sema/infer.mach:1611-1643`) **exactly**: read the lane count and lane scalar from the resolved `TYPE_VECTOR`, `check.check_coercible` each element to the lane scalar, and enforce exact arity with the same too-many / too-few diagnostics (infer.mach:1636-1641).

### 3. Lane access `v[i]` — read and write, comptime-constant index, bounds-checked

`v[i]` reads and `v[i] = x` writes; the index must be a comptime constant and is compile-time bounds-checked against the lane count. A **dynamic (non-comptime) index is a defined compile error** (deferred by design: a memory round-trip on SSE2 breaks 1:1) — a clear diagnostic, never a silent fallback.

`infer_index` (`src/lang/fe/sema/infer.mach:1395-1403`) delegates to `check_index` (`src/lang/fe/sema/check.mach:286-327`), which today handles `TYPE_POINTER`/`TYPE_ARRAY` and does **no** compile-time bounds check. Add a `TYPE_VECTOR` arm returning the lane scalar, evaluating the index with `comptime.eval` and requiring `CT_KIND_INT` in `[0, lanes)` — the exact template is `resolve_type_array` (infer.mach:1862-1893): demand `cv.kind == comptime.CT_KIND_INT`, reject negative via `comptime.ct_is_negative`, reject `>= lanes`. A non-comptime index fails the `CT_KIND_INT` demand → the defined "dynamic lane index not supported" error. The write path routes through the existing index-lvalue/assignment machinery, which must accept a vector index as an assignable lvalue.

### 4. Lane-wise operator typing — the honest table

All in `infer_binary` (`src/lang/fe/sema/infer.mach:588-684`), the single per-operator rule ladder. Vector operands must be the **same shape** (exact TypeId match — no cross-shape width mixing, no implicit casts). The legal table (what is 1:1 on the SSE2/NEON baselines, and what scalarizes cleanly otherwise):

- **float lanes**: `+ - * /` (BIN_ADD/SUB/MUL/DIV), result = the operand vector type.
- **integer lanes**: `+ -` at all widths; `*` on **16-bit lanes only** (`i16x8`/`u16x8` — `pmulld` for 32-bit is SSE4.1, out of baseline); result = the operand vector type.
- **bitwise** (`& | ^ << >>` binary; `~` unary) on integer-lane vectors, same-shape operands, result = the operand vector type. The unary `~` arm lands in the unary-op checker alongside the binary arms.
- **NO vector `%`** (BIN_MOD on any vector) → compile error. **NO integer vector `/`** (BIN_DIV on integer-lane vectors) → compile error.
- **comparisons** (`== != < <= > >=`) → the **same-shape unsigned mask vector** (`f32x4`/`i32x4`/`u32x4` cmp → `u32x4`; `i8x16`/`u8x16` → `u8x16`; `f64x2`/`i64x2` → `u64x2`; `i16x8` → `u16x8`), lanes all-ones / all-zeros. This diverges from the scalar arms, which return `u8t` (infer.mach:655, :660) — the vector arm returns the interned mask vector type instead.

No implicit casts, no scalar↔vector conversions, no bitcasts in increment 1: a scalar mixed with a vector is a type mismatch, and a mask (or any vector) result must **not** silently satisfy `check_condition` truthiness (`src/lang/fe/sema/check.mach:495-509`) — a vector is not a scalar truth value. Anything outside the table (`%`, integer `/`, scalar-vector mixing) uses the existing `type_mismatch`/`context.report` diagnostic pattern (infer.mach:2171).

### 5. Size, alignment, composition

16-byte size and 16-byte alignment. Arrays-of-vectors and pointers-to-vectors are ordinary — they compose through the existing `TYPE_ARRAY`/`TYPE_POINTER` paths with no special handling. Add a `16` arm to sema `byte_size` (`src/lang/fe/sema/check.mach:559-577`, else same-size cast legality misjudges vectors) and to IR `byte_size` (`src/lang/me/ir/type.mach:374-423`, guarding against the silent `ret 0` fallthrough at :422) and `byte_align` (type.mach:436-460, guarding the silent `ret 1` at :459).

### 6. IR substrate, verifier, and sound default passes

Add `IRT_VECTOR: IrTypeKind = 8` (next after `IRT_UNION = 7`, `src/lang/me/ir/type.mach:48-67`) with an inline payload (element IrTypeId/bits + lane count), an `intern_vector` beside `intern_int`/`intern_float`/`intern_array` (type.mach:262-304), and arms in `byte_size`(16)/`byte_align`(16), `hash_ir_type`/`eq_ir_type` (the dedup key funcs, type.mach:647-717 — without these the type never interns/dedups), `is_aggregate` (**false** — a vector is register-class, not memory-by-address; type.mach:225-230), and the printer `emit_type` (`src/lang/me/ir/printer.mach:373-412`, no default arm → a missing case renders nothing).

The **AST→IR seam** is `lower_type` (`src/lang/me/lower/context.mach:898-961`): add a `TYPE_VECTOR` arm calling `ir_type.intern_vector`. Lane-wise arithmetic/bitwise ride the **existing** opcodes — `OP_ADD`/`OP_SUB`/etc. are typed by their operands (`emit_binary` sets result type = `lhs.ty`, `src/lang/me/ir/builder.mach:902-907`), so a vector op is just the shared opcode carrying a vector type; no new opcodes. The one divergence is **compare-to-mask**: `emit_compare` hardcodes an 8-bit `IRT_INT` result (builder.mach:922-940, `intern_int(...,8)` at :930). A vector comparison must instead yield the same-shape mask vector — add a builder entry (e.g. `emit_vcompare`, or a result-type parameter on `emit_compare`) that reuses `OP_CMP_*` while setting the result to the mask vector type, and select it from `lower_binary` (`src/lang/me/lower/expr.mach`) when the operands are vector-typed.

**Verifier** (`src/lang/me/ir/verify.mach`): `arity_ok` already fails loud on any opcode with no case (verify.mach:1243-1293), so reusing opcodes keeps the guardrail. Extend `check_type_agree` (verify.mach:579-615): the `is_binary_agree` rule (verify.mach:622-631) already requires binary operands to share a type id, which a same-shape vector op satisfies; add the vector-compare rule — operands same vector type, result = same-lane-count unsigned mask vector — so the non-i8 compare result is accepted rather than flagged.

**Default pipeline passes** must stay sound with vector values. Debug (`OPT_DEBUG`) runs mem2reg → constfold → dce; release adds inline/algebraic/cse (`src/lang/me/pipeline.mach`). **constfold must not fold vector ops in increment 1**: `try_fold` (`src/lang/me/pass/constfold.mach:145-205`) fires only when both operands are `VAL_CONST_INT`/`VAL_CONST_FLOAT`; since no vector-constant Value kind exists, a vector operand is always `VAL_INSTR` and every fold is already inert — the child confirms (and if needed guards) that no vector op is folded. mem2reg promotes vector allocas correctly **given the 16-byte size/align** (the binding soundness requirement — phi/spill sizing rides `byte_size`/`byte_align`, so the `16` arms in slice 5 are load-bearing here); dce handles pure vector ops unchanged; algebraic/cse/inline are inert or type-safe. Keep `--verify` clean throughout.

### 7. Defined per-target isel error until the backends land

Until #2014 / #2015 / #2016 land real selection, a vector op that reaches instruction selection on **any** target must produce a **defined, non-ICE, per-target compile error** — not a silent miscompile and not the ICE-flavored "unhandled MIR opcode" diagnostic. The clean guard site is target-generic MIR lowering: `lower_instr` / `lower_generic` in `src/lang/be/codegen/mir/lower.mach` (:819-927 / :1409-1459), before MIR is even formed. An unguarded 16-byte vector otherwise silently clamps to an 8-byte GP value (`op_width_of` clamps >8-byte types to `gpr_width`, `src/lang/be/codegen/mir/context.mach:352-358`) and misroutes off the FP bank (`reg_class_of` sends only `IRT_FLOAT` to FP, context.mach:1002-1011). Detect a vector-typed operand/result and return `R.err[..., str]("codegen: 128-bit vectors not yet supported on <arch>")`, which surfaces through `codegen_module` (`src/lang/be/codegen.mach:104-145`) as a clean error string — mirroring the loud-error pattern of `lower_div` (dispatched at lower.mach:890-899; the error body with the loud `R.err` is at :1203). Each backend child removes this guard on its target as it wires real selection.

**Shared backend scaffolding lands here** (decomposition ruling: the three backend children run in parallel and none depends on another, so every target-generic seam they share belongs to this gate):

1. `reg_class_of` gains the `IRT_VECTOR` → `fp_class_index` arm and `op_width_of` preserves the 16-byte vector width instead of clamping to `gpr_width` (`src/lang/be/codegen/mir/context.mach:1002-1011`, `:352-358`) — the per-target guard above fires after routing, before selection.
2. The `value_is_vector` predicate in `mir/context.mach` (beside `value_is_float`/`value_is_aggregate`) and the neutral classify-driver threading (the `ClassifyFn` vector flag, `src/lang/be/codegen/mir/abi.mach:185`, plus any `ParamSlot` field). Each backend child then owns only its psABI branch: SysV → #2014, AAPCS64 → #2015, lp64 memory-class → #2016.
3. The vector spill-slot alignment: `add_spill_slot` (`src/lang/be/codegen/regalloc.mach:~2395`) allocates 16-byte-aligned slots for vector-class values (`class_width` already returns 16 for the non-GP class); scalar float spills are unaffected.
4. The SIMD capability fact: `MachineModel.has_v128: bool` (`src/lang/target/isa.mach:249-278`), a declared field — never derived from the `RC_FLOAT` bank width — populated x86_64 = true, aarch64 = true, riscv64 = false. This is the single fact #2016's scalarize gate and #2017's require gate both consume.

### 8. Manifest parser — the required `simd` profile key (owner ruling 1)

`simd` becomes a **required** key on every declared `[profile.X]`, values `"scalarize"` | `"require"`. This child **parses, validates, and stores** it only; the enforcement semantics (scalarize with a build note vs. error on an incapable target) belong to #2017.

This sets the **schema-evolution precedent**: manifest additions enter as **required-everywhere** fields with a coordinated ecosystem sweep on the #1979 machinery, **never** as optional-with-default. The strictness lever is totality-pure — a declared profile is `opt` + `debug` + `simd`.

Edits in `src/lang/manifest.mach`, following the `debug` key's shape exactly:
- Allow `simd` in the `TK_PROFILE` branch of `key_known` (manifest.mach:458-464).
- Add `parse_simd_mode` mirroring `parse_debug_flag` (manifest.mach:401-412) with the same `if (v == nil && as_root)` totality gate emitting `mach.toml: [profile.<name>] is missing required key 'simd'`, and value validation (`"scalarize"`|`"require"`) emitting `mach.toml: profile '<name>': simd must be "scalarize" or "require" (got ...)` in the style of `opt_err`/`debug_err` (manifest.mach:383, :409).
- Read it in `parse_profiles` (manifest.mach:749-762) and store it as a new field on `ProfileDef` (manifest.mach:73-79) **only**. Everything from `ResolvedProfile` (manifest.mach:175-181) onward — resolution including the default-posture branch, threading into `BuildUnit` → `ProjectConfig` → `session.Session`, and all enforcement — is #2017; this child stops at the manifest store.

Because dependency manifests parse permissively (`parse(..., as_root=false)`, driver.mach:2331), a dep's `[profile.*]` is not required and won't break — the `:2557` `nontotal` off-root fixture must stay demonstrating that tolerance (the new required key is required only under `as_root=true`).

**In-tree sweep (this child), so the tree stays green:** update every tracked manifest and every source-embedded manifest string that declares a `[profile.X]` to add `simd = "scalarize"`. The tracked set is **28 `mach.toml` files**, each declaring `[profile.debug]` and `[profile.release]` (repo-root `mach.toml`; the 5 `dbg/fixtures/*`; the 6 `int/regression/*`; the 16 `int/surface/*`). The embedded strings: the `mach init` template (`src/cli/cmd/init.mach:231`, :235), the doctest manifest (`src/cli/cmd/doc.mach:766`), the driver unit-test head fixture (`src/lang/driver.mach:6002`), and the `manifest.mach` in-file fixtures (:1866, :1999-2004, :2096, :2196-2199, and the missing-required-key tests at :2498-2501). The **external 13-repo ecosystem sweep** (~2 lines per repo) is release-bundled with the SIMD release on the #1979 machinery — out of this child, but its precedent is set here.

### Boundaries (explicit, by sibling)

- **No isel / no ABI classification** — the SSE2 / NEON / rv64 selection, register-class routing, psABI vector-class plumbing, and constant materialization are #2014, #2015, #2016. This child stops at the defined isel error (slice 7).
- **No lever enforcement** — threading `simd` into the build and choosing scalarize-vs-error, plus the "skipped N target-gated modules"-style build note, is #2017.
- **No DWARF DIEs** — the `DW_TAG_array_type` + `DW_AT_GNU_vector` vector debug type is #2018.
- **No doc/language amendments** — marking `doc/language/types.md` §SIMD implemented, the `doc/language/operators.md` lane-wise tables, and `doc/language/policy.md`'s amended one-instruction rule are #2020.
- **No splat sugar** — `f32x4{1.0}` broadcast is #2019.

## Acceptance

Colocated `test "..."` unit blocks in the touched `.mach` files (parser/sema/IR-verifier), positive and negative, all passing through the freshly built HEAD compiler.

Positive:
- The ten spellings resolve as types in type position and nowhere shadow a value named the same.
- Shared scaffolding exercised: vector-typed values route to the FP bank at 16-byte width (`reg_class_of`/`op_width_of`), vector-class spill slots come back 16-byte aligned, and `MachineModel.has_v128` reads true/true/false across x86_64/aarch64/riscv64 as a declared field.
- A full-arity literal type-checks; each element coerces to the lane scalar.
- `v[i]` read and `v[i] = x` write type to the lane scalar with a comptime-constant, in-bounds index.
- Lane-wise ops per the honest table type to the operand vector type; comparisons type to the correct same-shape unsigned mask (`f32x4`→`u32x4`, `i8x16`→`u8x16`, `f64x2`→`u64x2`, `i16x8`→`u16x8`).
- `IRT_VECTOR` interns and dedups (equal shape ⇒ equal id); `byte_size`/`byte_align` = 16 at both the sema and IR layers; `lower_type` maps `TYPE_VECTOR` → `IRT_VECTOR`.
- The IR verifier accepts vector binary ops and the mask-result compare; constfold leaves vector ops unfolded.
- A vector op reaching isel returns the defined per-target error string (not an ICE, not a miscompile).
- The manifest parser requires, validates, and stores `simd` on every declared profile.

Negative (each a specific diagnostic, never an ICE or silent accept):
- Wrong-arity literal (too many / too few lanes).
- Non-comptime lane index (the defined "dynamic lane index not supported" error) and out-of-bounds constant index.
- Vector `%`; integer vector `/`.
- Scalar-vector mixing in a binary op (type mismatch); a vector used where a scalar truth value is required.
- Missing `simd` key under `as_root` (`... is missing required key 'simd'`) and an invalid `simd` value.

Program bars (implementation child):
- **-g additivity**: the compiler binary and every emitted binary are byte-identical with and without `-g`, at both debug and release profiles.
- **`llvm-dwarfdump --verify`** stays clean on the `-g` self-host build.
- **Full unit suite** runs green through the freshly built HEAD compiler (not the PATH seed), with exact suite-count accounting against the dev baseline — report "N passed (baseline B + K new)".
- **Self-host fixpoint**: gen2 byte-identical to gen1.
- **int suite green** (linux 42/42), and the 28 tracked manifests + embedded fixtures parse after the `simd`-key sweep.

octalide:
Tier-1 gate of the SIMD vector-types program (epic #1965): the frontend + sema + middle-end substrate for the ten 128-bit vector types, the required `simd` manifest key, and the shared backend scaffolding.

Closes #2013.

This is **stage γ** of the #1979 seed dance: stage α (#2027, parser accepts `simd`) shipped in **v3.3.1**, making CI's latest-release seed simd-aware; this PR flips `simd` to required-everywhere and sweeps the tree. Rebased on dev-with-α, so the manifest change here is the require-flip on top of α's accept.

## What landed

- **Semantic types** `TYPE_VECTOR` (element + lanes), dedup interning, ten seeded shapes, `is_vector`/`vector_element`/`vector_lanes`/`vec`/`vector_mask`; sema `byte_size`=16 and a `<element>x<lanes>` diagnostic spelling.
- **Resolver** seeds the ten spellings as `SYM_VECTOR`, recognised in type position with priority over value scope.
- **Full-arity literals** `f32x4{a,b,c,d}` (`EXPR_KIND_VECTOR_LIT`, positional typed-literal parse, `infer_vector_lit` with exact arity + per-lane coercion).
- **Lane access** `v[i]` read/write: comptime-const, in-bounds index → lane scalar; dynamic/OOB → defined error.
- **Lane-wise op table** (float `+ - * /`, int `+ -` all widths + `*` 16-bit-only, bitwise + unary `~` on int lanes, comparisons → same-shape unsigned mask; no vector `%`, no int vector `/`, no scalar↔vector mixing, vector never a truth value).
- **IR** `IRT_VECTOR` (dedup, 16-byte size/align, register-class), `lower_type` mapping, `emit_vcompare` mask compare, verifier acceptance, inert constfold.
- **Shared scaffolding** `reg_class_of` vector→FP, `op_width_of` 16-byte, `value_is_vector`, 16-byte spill align, declared `MachineModel.has_v128` (t/t/f), and the neutral vector-flag thread through **both** `classify_arg` and `classify_return` (zero behavior; consumers are #2014/#2015/#2016).
- **Defined per-target isel error** `codegen: 128-bit vectors not yet supported on <arch>` (x86_64/aarch64/riscv64), never an ICE.
- **Manifest** required `simd` key on `ProfileDef` + the 28-manifest / embedded-string sweep.

Out of scope (siblings): isel/ABI selection (#2014/#2015/#2016), lever enforcement (#2017), DWARF DIEs (#2018), docs (#2020), splat (#2019).

## Verification (from-source HEAD compiler, 3.3.1 seed)

- **Full unit suite:** 585 passed, 0 failed = dev baseline 578 + 7 new tests.
- **Self-host fixpoint:** gen2 == gen3 byte-identical under the 3.3.1 seed.
- **-g additivity:** PT_LOAD byte-identical, release + debug profiles.
- **llvm-dwarfdump --verify:** clean, 21 target builds across three ISAs.
- **int suite:** all cases pass on linux (debug + release).
- **Negative diagnostics:** wrong-arity literal, non-comptime/OOB lane index, vector `%`, integer vector `/`, scalar↔vector mixing, vector-as-truth-value, missing/invalid `simd` key.

### DIFF
--- a/dbg/fixtures/aggregate/mach.toml
+++ b/dbg/fixtures/aggregate/mach.toml
@@ -37,10 +37,12 @@ abi = "aapcs64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.dbgagg]
 kind = "bin"
--- a/dbg/fixtures/inline/mach.toml
+++ b/dbg/fixtures/inline/mach.toml
@@ -22,10 +22,12 @@ abi = "lp64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.inlinefix]
 kind = "bin"
--- a/dbg/fixtures/inline4/mach.toml
+++ b/dbg/fixtures/inline4/mach.toml
@@ -22,10 +22,12 @@ abi = "lp64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.inline4fix]
 kind = "bin"
--- a/dbg/fixtures/multi/mach.toml
+++ b/dbg/fixtures/multi/mach.toml
@@ -37,10 +37,12 @@ abi = "aapcs64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.dbgfix]
 kind = "bin"
--- a/dbg/fixtures/promote/mach.toml
+++ b/dbg/fixtures/promote/mach.toml
@@ -22,10 +22,12 @@ abi = "lp64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.promotefix]
 kind = "bin"
--- a/int/regression/1750-i32-signext/mach.toml
+++ b/int/regression/1750-i32-signext/mach.toml
@@ -37,10 +37,12 @@ abi = "aapcs64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.case]
 kind = "bin"
--- a/int/regression/1762-rv64-float-vararg/mach.toml
+++ b/int/regression/1762-rv64-float-vararg/mach.toml
@@ -37,10 +37,12 @@ abi = "aapcs64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.case]
 kind = "bin"
--- a/int/regression/1851-rv64-phi-signext/mach.toml
+++ b/int/regression/1851-rv64-phi-signext/mach.toml
@@ -37,10 +37,12 @@ abi = "aapcs64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.case]
 kind = "bin"
--- a/int/regression/1852-rv64-self-host/mach.toml
+++ b/int/regression/1852-rv64-self-host/mach.toml
@@ -14,10 +14,12 @@ abi = "lp64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.case]
 kind = "bin"
--- a/int/regression/1923-comptime-field-gate/mach.toml
+++ b/int/regression/1923-comptime-field-gate/mach.toml
@@ -37,10 +37,12 @@ abi = "aapcs64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.case]
 kind = "bin"
--- a/int/regression/1925-global-aggregate-byval/mach.toml
+++ b/int/regression/1925-global-aggregate-byval/mach.toml
@@ -37,10 +37,12 @@ abi = "aapcs64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.case]
 kind = "bin"
--- a/int/surface/aggregate-abi/mach.toml
+++ b/int/surface/aggregate-abi/mach.toml
@@ -37,10 +37,12 @@ abi = "aapcs64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.case]
 kind = "bin"
--- a/int/surface/control-flow/mach.toml
+++ b/int/surface/control-flow/mach.toml
@@ -37,10 +37,12 @@ abi = "aapcs64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.case]
 kind = "bin"
--- a/int/surface/elf-pie/mach.toml
+++ b/int/surface/elf-pie/mach.toml
@@ -12,10 +12,12 @@ abi = "sysv64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.case]
 kind = "bin"
--- a/int/surface/elf-relro/mach.toml
+++ b/int/surface/elf-relro/mach.toml
@@ -22,10 +22,12 @@ abi = "lp64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.case]
 kind = "bin"
--- a/int/surface/float/mach.toml
+++ b/int/surface/float/mach.toml
@@ -37,10 +37,12 @@ abi = "aapcs64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.case]
 kind = "bin"
--- a/int/surface/freestanding-aarch64/mach.toml
+++ b/int/surface/freestanding-aarch64/mach.toml
@@ -12,10 +12,12 @@ abi = "aapcs64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.case]
 kind = "bin"
--- a/int/surface/freestanding-riscv64/mach.toml
+++ b/int/surface/freestanding-riscv64/mach.toml
@@ -12,10 +12,12 @@ abi = "lp64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.case]
 kind = "bin"
--- a/int/surface/freestanding/mach.toml
+++ b/int/surface/freestanding/mach.toml
@@ -12,10 +12,12 @@ abi = "sysv64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.case]
 kind = "bin"
--- a/int/surface/macho-pie-aarch64/mach.toml
+++ b/int/surface/macho-pie-aarch64/mach.toml
@@ -12,10 +12,12 @@ abi = "aapcs64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.case]
 kind = "bin"
--- a/int/surface/macho-pie-x86_64/mach.toml
+++ b/int/surface/macho-pie-x86_64/mach.toml
@@ -12,10 +12,12 @@ abi = "sysv64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.case]
 kind = "bin"
--- a/int/surface/pe-aslr/mach.toml
+++ b/int/surface/pe-aslr/mach.toml
@@ -12,10 +12,12 @@ abi = "win64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.case]
 kind = "bin"
--- a/int/surface/pe-import-attribution/mach.toml
+++ b/int/surface/pe-import-attribution/mach.toml
@@ -12,10 +12,12 @@ abi = "win64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.case]
 kind = "bin"
--- a/int/surface/pie-reloc/mach.toml
+++ b/int/surface/pie-reloc/mach.toml
@@ -22,10 +22,12 @@ abi = "lp64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.case]
 kind = "bin"
--- a/int/surface/pie-relro-fault/mach.toml
+++ b/int/surface/pie-relro-fault/mach.toml
@@ -22,10 +22,12 @@ abi = "lp64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.case]
 kind = "bin"
--- a/int/surface/sign-extension/mach.toml
+++ b/int/surface/sign-extension/mach.toml
@@ -37,10 +37,12 @@ abi = "aapcs64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.case]
 kind = "bin"
--- a/int/surface/syscall-inline-asm/mach.toml
+++ b/int/surface/syscall-inline-asm/mach.toml
@@ -37,10 +37,12 @@ abi = "aapcs64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.case]
 kind = "bin"
--- a/mach.toml
+++ b/mach.toml
@@ -37,10 +37,12 @@ abi = "aapcs64"
 [profile.debug]
 opt = 0
 debug = false
+simd = "scalarize"
 
 [profile.release]
 opt = 2
 debug = false
+simd = "scalarize"
 
 [artifact.mach]
 kind = "bin"
--- a/src/cli/cmd/doc.mach
+++ b/src/cli/cmd/doc.mach
@@ -763,7 +763,7 @@ test "mach.cli.cmd.doc.generate:creates_missing_output_dirs" {
     if (R.is_err[bool, str](filesystem.create_dir_all(?alloc, util.cstr_str(?root[0]), 493))) { bad = 1; }
     if (R.is_err[bool, str](filesystem.create_dir_all(?alloc, src_dir, 493)))                 { bad = 1; }
 
-    val manifest_text: str = "[project]\nid = \"doctest\"\nversion = \"0.0.0\"\nsrc = \"src\"\nout = \"out/{target.name}/{profile.name}\"\n\n[target.linux]\nisa = \"x86_64\"\nos = \"linux\"\nabi = \"sysv64\"\n\n[profile.debug]\nopt = 0\ndebug = false\n\n[artifact.doctest]\nkind = \"bin\"\nentry = \"main.mach\"\nout = \"bin/doctest\"\ntargets = [\"*\"]\nlink = []\nneed = []\n";
+    val manifest_text: str = "[project]\nid = \"doctest\"\nversion = \"0.0.0\"\nsrc = \"src\"\nout = \"out/{target.name}/{profile.name}\"\n\n[target.linux]\nisa = \"x86_64\"\nos = \"linux\"\nabi = \"sysv64\"\n\n[profile.debug]\nopt = 0\ndebug = false\nsimd = \"scalarize\"\n\n[artifact.doctest]\nkind = \"bin\"\nentry = \"main.mach\"\nout = \"bin/doctest\"\ntargets = [\"*\"]\nlink = []\nneed = []\n";
     val main_text: str = "fun main() i32 { ret 0; }\n\n# a documented helper\npub fun answer() i32 { ret 42; }\n";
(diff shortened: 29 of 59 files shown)


octalide:
Coordinator review (5-dimension, adversarially verified — 23 agents). CI green ≠ merge: five findings survived verification, all fixable in one round. Manifest flip + sweep dimension: zero defects (sweep verified complete, flip error mirrors the sibling required-key idiom).

**1. [major] Positional-brace literal on a non-vector type loses its diagnostic** (`src/lang/fe/sema/infer.mach:1805`). The parser routes every positional `T{...}` to `EXPR_KIND_VECTOR_LIT`; `infer_vector_lit` silently returns the type when it isn't a vector, so `Point{1,2}` / `i32{5}` — a loud parse error before this PR — now surfaces only as a generic lower-stage "unsupported expression kind". Not a soundness hole end-to-end (it does fail), but a diagnostic regression and a layering violation vs the "specific diagnostic, never a silent accept" rule. Fix: a specific sema error on the non-vector path (the mirror-`infer_array_lit` defense fails — its equivalent branch is dead code, this one is reachable).

**2. [major] `EXPR_KIND_VECTOR_LIT` has no AST→IR lowering** (`src/lang/me/lower/expr.mach:85`). A valid `f32x4{1,2,3,4}` dies with the generic rvalue error instead of lowering to an `IRT_VECTOR` value that reaches the defined per-target isel guard. No sibling owns AST→IR lowering — backends own MIR→machine only — so without this, #2014 cannot even drive its isel with literals without editing shared middle-end code. Gate-owned; fix here.

**3. [major] Spill-slot `align = size` silently over-aligns scalar-float spills 8→16** (`src/lang/be/codegen/regalloc.mach:2397`). `class_width` returns 16 for any non-GP class, so a spilled `f32`/`f64` slot's alignment changed — codegen changes for any float-spilling program. Fixpoint/additivity stayed green only because the compiler itself spills no floats: a silent non-vector codegen change that slipped every bar. Fix: 16-byte alignment keyed on actual vector-ness (the vreg width the MIR layer already distinguishes), scalar FP spill layout byte-identical to pre-PR.

**4. [major] The slice-7 isel guard has zero test coverage** (`src/lang/be/codegen/mir/lower.mach:840`). The most load-bearing negative behavior in the gate — vector op reaching isel = defined per-target error, never an ICE — is exercised by no test; all vector tests stop at sema. Fix: drive vector source through lower+codegen and assert the defined error string.

**5. [major] The shared-scaffolding acceptance bullet has no tests beyond `has_v128`** (`src/lang/be/codegen/mir/context.mach:1030`). `reg_class_of` vector→FP, `op_width_of` 16-byte preservation, spill alignment (vector 16 / scalar-float 8 after fix 3), and the classify-thread zero-behavior claim are all explicit acceptance items with no assertions. Fix: unit tests for each.

Bundle-worthy minors from the same pass: verifier result-shape rules for vector ops (compare→correct mask shape — the issue's verifier acceptance asks for this), and sema negatives (negative lane index, reversed scalar+vector order, i16x8→u16x8 mask width).

Refuted and dropped (for the record): vector `<<`/`>>` admissibility (contract-consistent; fixed defined sequences cover SSE2 per the comparison pin), vector-param guard bypass (guard catches it), and the sema-phrased duplicate of finding 2.

octalide:
Review fix pass pushed (5 commits, `de54e163`..`f6bbbae9`). Point by point:

**1 [major] non-vector positional-brace literal diagnostic** — fixed (`de54e163`). `infer_vector_lit` now reports a specific, type-naming error on the non-vector path: `positional-brace literal requires a vector type; \`Point\` is not a vector (a record uses named fields, an array a \`[N]T\` prefix)`. Verified: `Point{1,2}` and `i32{5}` fail loudly at sema.

**2 [major] EXPR_KIND_VECTOR_LIT has no AST→IR lowering** — fixed (`2b3c488e`). `lower_vector_lit` assembles lanes through array-shaped backing storage (same 16-byte layout) and loads them back as the vector, so a valid `f32x4{1,2,3,4}` now produces an IRT_VECTOR value that reaches the per-target isel guard (verified: literal path → `codegen: 128-bit vectors not yet supported on x86_64`, not the generic rvalue error).

**3 [major] spill `align=size` over-aligns scalar-float spills** — fixed (`ae768f00`) by reverting to the pre-PR 8-byte alignment. Rationale: in this gate no vector reaches regalloc — the target-generic isel guard fires during MIR lowering, before regalloc — so the 16-byte vector spill alignment is unreachable here and only regressed floats. The vreg carries no width (MirVReg is class-only; 24 `new_vreg` sites), so keying on vector-ness would mean plumbing a width for a dead-in-gate path. Vector-16 spill alignment belongs in the backend children, where a spilled vector carries a real value width. Verified fixpoint + additivity still green (as you predicted — the compiler spills no floats — but confirmed, not assumed), and scalar-FP spill layout is now byte-identical to pre-PR. **Flag if you'd rather I plumb a vreg width now instead of deferring.**

**4 [major] verifier vector result-shape rules** — added (`4943334b`). The verifier now validates: compare → same-lane-count unsigned mask (integer lane of the operand's lane width); binary/unary → same-shape operands and result.

**5 [major] isel-guard + scaffolding test coverage** — added (`f6bbbae9`): a driver test drives vector source through lower+codegen asserting the per-target error (literal + operator paths); a `mir.context` unit covers `reg_class_of` vector→FP, `op_width_of` 16-byte, `value_is_vector`; a classify-thread zero-behavior unit asserts arg+return slots are identical with `is_vector` true/false; and sema negatives for the non-vector literal, a negative lane index, reversed scalar+vector order, and the i16x8→u16x8 mask width.

Bars re-run (from-source HEAD, 3.3.1 seed): suite **588 passed** (dev 578 + 10), fixpoint gen2==gen3 byte-identical, -g additivity both profiles, int suite green. CI re-running.

PR #2025, opened 2026-07-10
fix(link): string-merge dedup .debug_str across modules

### LINKED ISSUE #1708: debuginfo: cross-TU .debug_* merge/dedup in linker

octalide:
Closes #1708 (remaining scope — `.debug_str` string-merge; the `.debug_abbrev` half shipped in #2009).

## What

Every module emits its own `.debug_str` holding the same type/function names and source paths redundantly. The linker concatenated all of them. This dedups by content:

- **Pre-pass `build_str_merge`** interns every module's `.debug_str` strings into one `enc.ByteBuf`, storing exactly one physical copy per distinct string and recording each string's offset (`StrId -> offset`). Identical content across modules interns to the same `StrId`, so it collapses to one copy.
- **`.debug_str` merged content becomes that buffer** (special-cased in `accumulate_sections` + `build_merged_image`, mirroring the #2009 abbrev shape): no module contributes bytes; the buffer is written once.
- **strp resolution** (`patch_module_relocations` → `resolve_strp`): a relocation targeting a module's `<name>#.debug_str` anchor is remapped to its string's deduped offset instead of the concatenated placement.
- **Per-site content-equality guard** (load-bearing): before writing each strp offset, the module's own string bytes are asserted byte-equal to the deduped buffer's bytes there — a divergence is a loud linker error, since `llvm-dwarfdump --verify` cannot catch a wrong-but-valid strp offset.

Dedup runs only for the in-place-patched modes (executable, shared object); a relocatable output keeps concatenation so its carried-forward strp relocations stay valid.

## Results (self-host `-g` compiler, 146 CUs / 162 objects, linux-x86_64)

- `.debug_str`: **406,117 → 166,072 bytes** (−240,045, −59%).
- `llvm-dwarfdump --verify`: **No errors** (146/146 units).
- Named-string spot-check: the full set of resolved `DW_AT_name` strings is identical before/after except for the 18 new identifiers this PR adds to the compiler source itself — every pre-existing name resolves to the same string, confirming the remap.

Remaining verification bars (additivity both profiles, `dbg/verify.sh` both profiles, unit suite, fixpoint, int 42/42) run before marking ready.

### DIFF
--- a/src/lang/be/linker.mach
+++ b/src/lang/be/linker.mach
@@ -42,6 +42,7 @@ use O: std.types.option;
 use R: std.types.result;
 
 use mach.lang.be.obj;
+use enc: mach.lang.be.codegen.encode;
 use mach.lang.intern;
 use mach.lang.session;
 use mach.lang.target;
@@ -142,6 +143,25 @@ rec DynState {
     fixup_cap:   u32;
     base_reloc_cap: u32;
     active:      bool;
+}
+
+# the deduped `.debug_str` string table and the map from an interned string to its
+# byte offset within it, built once per link before section accumulation. every
+# module's `.debug_str` holds the same names redundantly (10k unique of ~38k on a
+# self-host `-g` build), so one physical copy per distinct string collapses the
+# section by ~60%. a strp relocation's target string interns to a `StrId` whose
+# offset here is the value written at the patch site, so identical strings across
+# modules resolve to the one copy (#1708). owns `buf` + `offs`; freed after link
+# ---
+# buf:    the deduped `.debug_str` bytes (one NUL-terminated copy per distinct string)
+# offs:   interned-string -> byte offset within `buf`
+# name:   the interned ".debug_str" section name, keying the merged slot
+# active: true when dedup is in effect (a patching-mode link carrying `.debug_str`)
+rec StrMerge {
+    buf:    enc.ByteBuf;
+    offs:   map.Map[intern.StrId, u32];
+    name:   intern.StrId;
+    active: bool;
 }
 
 # read the input object files from disk and combine them into the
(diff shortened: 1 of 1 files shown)


octalide:
## Verification results (all bars green)

All run through the freshly-built HEAD compiler (`out/linux-x86_64/debug/bin/mach`), not the PATH seed.

**`.debug_str` size (self-host `-g`, 146 CUs / 162 objects, linux-x86_64)**
- before: **406,117 bytes** (`0x63265`)
- after:  **166,072 bytes** (`0x288b8`)
- delta:  **−240,045 bytes (−59.1%)**

**`llvm-dwarfdump --verify`** on the multi-TU `-g` compiler: `No errors.` (146/146 units).

**Named-string spot-check** (the real guard against a wrong remap): the full set of resolved `DW_AT_name` strings is identical before/after except for the 18 new identifiers this PR adds to the compiler source itself (`StrMerge`, `build_str_merge`, `resolve_strp`, `debug_str_content_eq`, …). Every pre-existing name — types, functions, and source paths like `./src/main.mach` — resolves to the same string. 9,572 baseline unique names all present. The per-site content-equality guard passed on all ~38k strp sites during the build.

**`-g` byte-additivity (both profiles)** — PT_LOAD segments byte-identical between `-g` and non-`-g`:
- release: 5 PT_LOAD segments identical (`dbg/self-additive.sh`)
- debug:   5 PT_LOAD segments identical (same comparison, debug profile)

**`dbg/verify.sh`** (both profiles, all three ELF ISAs incl. the multi-TU `.debug_str` fixture): `all debug-info checks passed (21 target build(s))`.

**Unit suite** via HEAD: **576 passed, 0 failed, 576 total** — test set unchanged vs dev baseline (my diff adds/removes no `test` blocks). Linker suite `mach.lang.be.linker` 7 ok; `mach.lang.be.codegen.dwarf` 13 ok.

**Self-host fixpoint:** HEAD reproduces itself byte-identical.

**int suite** (`--target linux`): **42 PASS / 0 FAIL** (matches the 42/42 baseline). CI also green on `int linux-arm64`, `int linux-riscv64`, `int windows`, and the `debuginfo` lane.

No new `int/` fixtures or root-level `test/`.

Issue #2030, opened 2026-07-11
sema: narrow vector shifts out of the increment-1 operator table


octalide:
Part of #1965. Follow-up to #2013 (PR #2024). Blocks the guard-arm removal in #2014 and #2015.

## What

The merged tier-1 gate admits vector `<<` `>>` on integer-lane vectors (sema `infer_binary`, src/lang/fe/sema/infer.mach:739-746) and lowers them to same-shape two-vector-operand `OP_SHL`/`OP_SHR_S`/`OP_SHR_U` (src/lang/me/lower/expr.mach:1903-1907) — i.e. **per-lane variable shift**, each lane shifted by the corresponding lane of the right operand.

That contradicts the epic's honest-table premise ("what is 1:1 on the SSE2/NEON baselines"), the same criterion that kept integer `*` to 16-bit lanes (no SSE4.1 `pmulld`) and excluded 64-bit compares. Per-lane variable shift is AVX2 (`VPSLLVD`-class) on x86_64 — SSE2's packed shifts move **all** lanes by one shared count, and 8-bit lanes have no packed shift at all — so a "fixed defined sequence" would be a 4–16-lane extract/shift/insert unroll (~50 instructions for `u8x16`) on a **capable** target: scalarization-an-operator on hardware the lever machinery declares capable, far past the float-`%`/compare-sequence precedent of short fixed sequences. The gate drafter's enumeration of `<< >>` under "bitwise" over-included; the owner-signed table text says only "bitwise on integer lanes", and the table's own premise decides the ambiguity.

## Fix

Narrow increment 1 to the strict bit-parallel set: `& | ^` binary and `~` unary on integer lanes. Vector `<<` `>>` become a **specific compile error** ("vector shift is not in the increment-1 operator table" naming the type and pointing at the per-lane-vs-uniform deferral), in sema — remove the shift arms from the vector branch of `infer_binary` and the vector case from the IR shift lowering. Shifts return in a later increment as a deliberate design choice between the uniform-count shape (SSE2-1:1 with splat sugar, tier 2) and the per-lane shape (NEON-native, AVX2-gated on x86_64) — narrowing now is forward-compatible with both; admitting per-lane now would lock semantics the baseline cannot honor.

## Acceptance

- `v << w` / `v >> w` on every vector shape → the specific diagnostic; scalar shifts unaffected.
- The IR verifier/lowering no longer produce vector-typed shift ops (negative test).
[...]

Issue #1965, opened 2026-07-05
epic: SIMD vector types — portable-vector model (f32x4-class primitives)


octalide:
> ## Status (2026-07-11)
> Tier-1 gate #2013 MERGED (PR #2024, after a 5-finding review fix round;
> landed via the #1979 seed dance — α #2027, v3.3.1 seed release, require-flip).
> Both side-lanes done (#1708/PR #2025, #2022/PR #2023; RFC #2026 blocked on
> owner). All four remaining tier-1 children now dispatched in parallel:
> #2014 #2015 #2016 #2017. Scope amendment: vector spill-slot alignment moved
> from the gate to #2014 (see #2013 closing comment). Dev suite baseline: 588.

## What

First-class SIMD vector types on the **portable-vector model**:
`f32x4`-class types are *semantic* types (N lanes, lane-wise ops) legal
on every target. Backends with the hardware emit one instruction per
operator (SSE2 on x86_64, NEON on aarch64); targets without it
(rv64gc until V) get a **defined unrolled scalar expansion** of each
op. The rule is *scalarize operators, never algorithms* — the compiler
never emits an algorithm behind an operator (precedent: float `%` is
already a fixed multi-instruction lowering). doc/language/policy.md's
"one CPU instruction per operator" is amended to "one instruction on
capable targets; defined unrolled expansion otherwise, reported at
build time."

Design points (settled):

- **Seeded set: 128-bit, single-`x` only** — `f32x4 f64x2 i8x16 i16x8
  i32x4 i64x2 u8x16 u16x8 u32x4 u64x2`. Depth-1 contract: no `f32x4x4`
  language types — a matrix is an algorithm over vectors (integer
  matmul even needs widening accumulators), so matrices/quaternions
  live in a math library over vector elements, which gets clean
  operators *because* its element currency is a compiler type.
- **Literals**: full-arity `f32x4{1.0, 2.0, 3.0, 4.0}` (mirrors array
  literals); splat sugar deferred.
- **Lane access**: `v[i]` read/write with comptime-constant index,
  compile-time bounds-checked. Dynamic lane index deferred (memory
  round-trip on SSE2 breaks 1:1).
- **Comparisons** produce the same-shape unsigned mask vector
  (`f32x4` cmp → `u32x4`, lanes all-ones/all-zeros) — what the
  hardware produces; no vector-bool type. Select/blend via
  `(mask & a) | (~mask & b)` in the library first.
[...]

Issue #2015, opened 2026-07-10
codegen: aarch64 NEON isel and AAPCS64 short-vector ABI for vector types


octalide:
Part of #1965.
Depends on #2013.
Depends on #2022.

## What

The aarch64 backend half of the portable-vector program: NEON instruction selection plus the AAPCS64 short-vector ABI so the 128-bit seeded vector types (`f32x4 f64x2 i8x16 i16x8 i32x4 i64x2 u8x16 u16x8 u32x4 u64x2`) execute one instruction per operator on aarch64, exactly as the x86_64 SSE2 sibling (#2014) does on that target. This child adds only the arm64 leaf: it assumes `IRT_VECTOR` values already flow through target-generic MIR routed to the vector register file (that plumbing is #2013's — see scope below).

**The op table is the language contract, not the hardware ceiling.** The set of operators this child selects is the *honest SSE2-baseline table* chosen for cross-target parity, deliberately narrower than what NEON can do. The clearest instance: NEON `MUL` supports `.16b/.8h/.4s` integer lanes, but the language contract exposes integer `*` on **16-bit lanes only** (`i16x8`/`u16x8`), because SSE2's `pmullw` is the baseline and `pmulld` is SSE4.1 — so aarch64 must reject integer `*` on every other lane width even though the hardware could do it. This is intentional: the language op table is fixed once, at the baseline every capable target shares, so a program compiles the same everywhere. Widening the table for one ISA is a later, whole-program decision, never a per-backend liberty.

### Per-operator NEON selection table

Operands `Vn, Vm` → result `Vd`; `<T>` is the arrangement specifier chosen from the operand's `IRT_VECTOR` lane element-kind and count (`.4s` for `f32x4`/`i32x4`/`u32x4`, `.2d` for `f64x2`/`i64x2`/`u64x2`, `.16b` for `i8x16`/`u8x16`, `.8h` for `i16x8`/`u16x8`). Logical ops are byte-agnostic and always use `.16b`.

| lang op | applies to | NEON selection |
| --- | --- | --- |
| `+` float | `f32x4`, `f64x2` | `FADD Vd.<T>, Vn.<T>, Vm.<T>` |
| `-` float | `f32x4`, `f64x2` | `FSUB Vd.<T>, Vn.<T>, Vm.<T>` |
| `*` float | `f32x4`, `f64x2` | `FMUL Vd.<T>, Vn.<T>, Vm.<T>` |
[...]

PR #2034, opened 2026-07-11
feat(x86_64): SSE2 vector isel + SysV SSE-class ABI (#2014)

### LINKED ISSUE #2014: codegen: x86_64 SSE2 isel and SysV SSE-class ABI for vector types
Part of #1965.
Depends on #2013.

## What

The x86_64 tier-1 backend for the portable-vector model. This child consumes the `IRT_VECTOR` substrate, the sema op table, and the compare→mask semantics produced by #2013, and it does two things on x86_64: selects **one SSE2 packed instruction per operator** for every entry in the honest op table, and classifies vector params/returns in `xmm` per the **SysV SSE class**. The FP register file that regalloc already manages carries these values — there is **no new register class** (settled design). Sibling backends off the same gate: aarch64 NEON is #2015, rv64gc's defined unrolled scalarization plus the build-time scalarization note is #2016, and the `[profile.X]` `simd` lever is #2017. DWARF vector DIEs (#2018), splat sugar (#2019), doc amendments (#2020), and the tier-3 stdlib simd/math library (#2021) come later; #1942 (loop auto-vectorization) consumes this same isel substrate afterward.

The seeded set is the ten 128-bit single-`x` vectors from the epic: `f32x4 f64x2 i8x16 i16x8 i32x4 i64x2 u8x16 u16x8 u32x4 u64x2`. All are 16-byte size and 16-byte aligned.

### New opcode + MIR surface

No packed SSE opcode exists in the tree today — `src/lang/target/isa/x64.mach` carries only the scalar float family (`MOVSS`..`UCOMISD`, opcodes 41-67) plus `MOVQ`, `XORPS`, `MOVAPS`. This child adds packed-opcode constants there: `ADDPS SUBPS MULPS DIVPS ADDPD SUBPD MULPD DIVPD`, `PADDB PADDW PADDD PADDQ`, `PSUBB PSUBW PSUBD PSUBQ`, `PMULLW`, `PAND POR PXOR`, `CMPPS CMPPD`, `PCMPEQB PCMPEQW PCMPEQD`, `PCMPGTB PCMPGTW PCMPGTD`, and `MOVUPS`. (`MOVAPS`, `MOVQ`, `XORPS` already exist.)

The scalar float ALU already models the pattern to copy: `MIR_FADD`..`MIR_FDIV` (`src/lang/be/codegen/mir.mach`, opcodes `0x1114+`) **survive** instruction selection with `expansion_count == 1`, and the encoder materializes the SSE bytes — see `encode_float_arith`, `emit_float_load`/`emit_float_store`, `float_alu_opcode`, and `encode_sse_alu` in `src/lang/target/isa/x64/encode.mach` (`encode_sse_alu` at :1539, `encode_movaps` at :1497), with the two fixed scratch registers `FLOAT_SCRATCH0`/`FLOAT_SCRATCH1` = XMM14/XMM15. This child adds a parallel **packed** MIR family (a small set of generic packed opcodes — add/sub/mul/div, bitwise and/or/xor, compare, move/load/store, lane extract/insert) that likewise survive isel and are materialized by new packed encoders modeled on `encode_sse_alu`/`encode_movaps`. The lane element kind (int vs float) and lane byte-width (b/w/d/q) ride on the `MirInstr` so the encoder maps (op, elem-kind, elem-width) → the concrete packed opcode; every field an expander writes must be stamped into the zeroed selection buffer exactly as `select_block` already does (`src/lang/target/isa/x64/isel.mach:155`), so the self-host fixpoint stays byte-identical.

### The honest per-op mapping table (SSE2 baseline, 1:1)

Float arithmetic — `f32x4` uses the `ps` forms, `f64x2` the `pd` forms:

| lane type | `+` | `-` | `*` | `/` |
|---|---|---|---|---|
| f32x4 | ADDPS | SUBPS | MULPS | DIVPS |
| f64x2 | ADDPD | SUBPD | MULPD | DIVPD |

Integer `+` / `-` — all lane widths:

| lane width | `+` | `-` |
|---|---|---|
| 8-bit (i8x16/u8x16) | PADDB | PSUBB |
| 16-bit (i16x8/u16x8) | PADDW | PSUBW |
| 32-bit (i32x4/u32x4) | PADDD | PSUBD |
| 64-bit (i64x2/u64x2) | PADDQ | PSUBQ |

Integer `*` — **16-bit lanes only**:

| lane width | `*` |
|---|---|
| 16-bit (i16x8/u16x8) | PMULLW |

Integer `*` on 8/32/64-bit lanes is **not** in the table: `pmulld` is SSE4.1 and 8/64-bit have no baseline packed multiply. Explicitly **no `pmulld`, no SSE4.1**.

Bitwise — integer lanes, one 128-bit-wide op (width-agnostic):

| op | instr |
|---|---|
| `&` | PAND |
| `\|` | POR |
| `^` | PXOR |

Comparisons produce the same-shape unsigned **mask** vector (per-lane all-ones on true, all-zeros on false — `f32x4` cmp → `u32x4`, etc.). Float compares take an imm8 predicate via `CMPPS`/`CMPPD`; `>`/`>=` lower by swapping operands onto the LT/LE predicates:

| op | f32x4 / f64x2 | imm8 | note |
|---|---|---|---|
| `==` | CMPPS / CMPPD | 0 (EQ) | |
| `!=` | CMPPS / CMPPD | 4 (NEQ) | |
| `<`  | CMPPS / CMPPD | 1 (LT) | |
| `<=` | CMPPS / CMPPD | 2 (LE) | |
| `>`  | CMPPS / CMPPD | 1 (LT) | operands swapped |
| `>=` | CMPPS / CMPPD | 2 (LE) | operands swapped |

Integer compares use the `pcmpeq`/`pcmpgt` family (signed `>` only; `<` by operand swap):

| op | 8-bit | 16-bit | 32-bit |
|---|---|---|---|
| `==` | PCMPEQB | PCMPEQW | PCMPEQD |
| `>` (signed) | PCMPGTB | PCMPGTW | PCMPGTD |
| `<` (signed) | PCMPGTB/W/D, operands swapped |

**Fixed defined sequences** (no single SSE2 instruction exists — same status as the float-`%` precedent: a fixed multi-instruction lowering per operator, never an algorithm): integer `!=` = `PCMPEQ` + `PXOR` all-ones; unsigned `<`/`>` = sign-bias both operands (`PXOR` with the sign-bit splat) then signed `PCMPGT`; integer `<=`/`>=` = the `>`/`<` primitive followed by `PXOR` all-ones; `i64x2`/`u64x2` `==` = `PCMPEQD` + `PSHUFD(0xB1)` + `PAND` (`PCMPEQQ` is SSE4.1, out of baseline); 64-bit orderings = the canonical fixed compose over 32-bit halves (high-half compare + low-half unsigned tie-break). Every sequence is fixed-length, branch-free, and lane-exact against the scalar reference. **Every comparison operator admitted by #2013's sema table lowers here; none defers to the library** — #2021 keeps only genuine non-operators (shuffles, reductions, select/blend, saturating ops).

### Lane read/write (comptime-const index)

#2013 type-checks `v[i]` with a comptime-constant, bounds-checked index and emits a lane extract/insert into the IR. This child lowers that to a **fixed constant-offset** sequence: materialize the vector into a 16-byte-aligned frame slot with `MOVAPS`, then `MOV` the scalar lane from/to `[slot + i*elem_bytes]` (the offset is a compile-time constant, so this is a fixed-length lowering — **not** the deferred dynamic memory round-trip). A shuffle-based extract (`PSHUFD`+`MOVD`, `PEXTRW`) is a valid later optimization but is not required here; the constant-offset form is the defined baseline and always correct.

### Loads, stores, moves, spills

Register↔register 128-bit moves use `MOVAPS` (already emitted by `encode_movaps`, `encode.mach:1497`). Aligned frame-slot load/store uses `MOVAPS`; a pointer-to-vector whose alignment is not statically 16 uses a new `MOVUPS` encoder. Spill/reload go through a 16-byte-aligned spill slot with `MOVAPS`.

Reg-class routing, width preservation, and the 16-byte vector spill-slot alignment are #2013's shared scaffolding (the `reg_class_of` `IRT_VECTOR` → `fp_class_index` arm; `op_width_of` preserving 16 bytes instead of the `gpr_width` clamp; `add_spill_slot` allocating 16-byte-aligned vector slots — `class_width` already returns 16, `src/lang/be/codegen/regalloc.mach:2372`). This child replaces the gate's per-target error guard on x86_64 with real selection: `lower_instr`/`lower_generic` (`src/lang/be/codegen/mir/lower.mach:819`/`:1409`) route vector-typed `OP_ADD`/etc. to the packed selections before the generic scalar path, and `MOVAPS` reloads rely on the 16-byte-aligned slots the scaffolding guarantees.

### Vector-literal constant materialization

Scalar float constants are carried as inline immediates (`MOVABS` into R11 then `MOVQ` into an XMM). A 128-bit vector literal has no such path: emit its 16 bytes into `.rodata` at **align 16** and load with `MOVAPS [rip+disp]` (RK_PC32). `pack_globals` in `src/lang/be/codegen.mach` (rodata `sec.align` at :621-623, constant serialization at :514-556) raises a section to the strictest type/`#[align]` alignment, so the vector type declaring `byte_align = 16` (owned by #2013) is what forces the 16-aligned constant.

### SysV SSE-class ABI

A 16-byte vector is **one** SSE-class value in a single XMM register (distinct from today's two-eightbyte XMM0:XMM1 split for `{f64;f64}` aggregates). The `value_is_vector` predicate and the classify-driver vector flag (`classify_arg`/`classify_ret` invocation in `src/lang/be/codegen/mir/abi.mach:~621`) land with #2013's shared scaffolding; this child implements the SysV branch in `src/lang/target/abi/sysv.mach` (`classify_arg` at :155, `classify_ret`): a vector arg → one XMM (`CLASS_FP`, which already documents "floating-point / vector register", `abi.mach:56`) with `size = 16` when an XMM arg register is free, else a 16-byte by-value `CLASS_STACK` slot; the return goes in XMM0. The GP and XMM argument cursors stay independent, so a mixed scalar+vector signature places each on its own bank.

**Scope boundary for the ABI seam:** the shared neutral scaffolding — the `value_is_vector` predicate, the `ClassifyFn` vector flag (`abi.mach:185`), and any `ParamSlot` field — lands with #2013 and is consumed by all three backend children. This child owns **only** the SysV/x86_64 branch. AAPCS64 short-vector/HVA placement is #2015; lp64 memory/byref placement is #2016.

### Exclusions (state explicitly)

- **No dynamic lane index** — a runtime-variable memory round-trip breaks 1:1 (deferred by design).
- **No shuffles, reductions, gather/scatter, saturating arithmetic** — tier-3 #2021.
- **No autovectorization** — #1942 consumes this substrate later.
- **No SSE4.1+**: no `pmulld` (integer `*` on 8/32/64-bit lanes), no `PCMPEQQ`/`PCMPGTQ` (64-bit compares); those and the derived-mask compares are library mask algebra (#2021).
- **No casts** — no scalar↔vector implicits, bitcasts deferred (design increment 1).
- **No splat sugar** (#2019) and **no DWARF vector DIEs** (#2018) — this child adds no DIEs and keeps `-g` additive.

## Acceptance

Unit tests are `test "..."` blocks colocated in the touched `.mach` files (x86_64 is the native CI host, so they run natively). Observable criteria:

- **Every table op, per-lane exact.** For each op and each applicable lane type — float `+ - * /` (`f32x4`/`f64x2`), integer `+ -` at all four widths, `*` on 16-bit lanes, `& | ^` — a runtime test builds two vector literals, applies the op, reads every lane via comptime-const index, and checks each lane against the scalar reference computed in the same test.
- **Mask compares lane-exact.** For each supported compare (float `== != < <= > >=`; integer `==`, signed `<`/`>` at 8/16/32-bit) a test asserts the result mask is all-ones (`0xFF..` at the lane width) on true lanes and all-zeros on false lanes.
- **ABI round-trips.** Functions that take and return vectors by value across a real call, with lanes checked after the call, prove single-XMM SysV placement; a mixed scalar+vector signature (e.g. `(f64, f32x4, i32) -> f32x4`) proves the GP/XMM cursors stay independent.
- **Spill correctness under pressure.** A test with more simultaneously-live vectors than allocatable XMM registers (>14) forces spills; every lane survives the `MOVAPS` spill/reload round-trip (exercises the align-16 slot).
- **Vector-literal materialization.** A literal loaded from `.rodata` reads back lane-exact; its constant section alignment is 16.
- No vector op reaches the `write_opcode_ice` fallthrough in `select_block` — every implemented op selects a concrete instruction.

Program-wide bars (all must hold):

1. **`-g` additivity** — the compiler and every emitted binary are byte-identical with and without `-g`, at both the debug and release profiles.
2. **`llvm-dwarfdump --verify`** stays clean on the `-g` self-host build.
3. **Full unit suite** run through the freshly-built HEAD compiler (not the PATH seed), with exact suite-count accounting against the dev baseline — report `N passed (baseline B + K new)`.
4. **Self-host fixpoint** — gen2 byte-identical to gen1.
5. **int suite green** (linux 42/42).

octalide:
Implements the x86_64 tier-1 backend for the portable-vector model (epic #1965): SSE2 instruction selection for the honest op table and the SysV SSE-class ABI for the ten 128-bit vector types, on the merged `vec_lane` substrate (#2037). Rebased onto dev (169fe0d9) with NEON (#2032) and the vector-local memzero `[lanes]elem` storage (#2044) merged; the x86_64 arm joins the shared `arch_selects_vectors` seam. Vector shifts are excluded (rejected at sema by #2030).

Closes #2014.

## What's implemented
- Selection keyed on the shared `vec_lane` descriptor + the surviving generic / `MIR_SEL_*` opcode — no MIR vector opcode family. Lane-wise compute ops route through the generic path carrying `vec_lane`; the x64 isel / encoder select the packed form. `MIR_SEL_*` sit at 0x1100+, clear of the generic range and the concrete x64 opcodes (no aliasing).
- Packed encoders: float `+ − × ÷`, integer `+ −` (b/w/d/q), 16-bit `×` (PMULLW), bitwise `& | ^ ~`, and compare→mask for all six operators — including the branch-free fixed sequences SSE2 has no single form for: integer `!=`, `<=`/`>=`, unsigned orderings (sign-bias / unsigned-saturating-subtract), i64x2/u64x2 `==` (PCMPEQD+PSHUFD+PAND), and the 64-bit orderings (red-zone per-lane scalar compose).
- SysV SSE-class ABI: a vector arg in one XMM (or a 16-byte by-value stack slot), returned in XMM0; mixed scalar+vector signatures keep independent GP/XMM cursors (reuses the shared 16-byte ABI move width from #2032).
- Vector spill slots 16-byte aligned (MOVAPS reload); scalar GP/float spill layout byte-identical.
- The `encode_inst` mem-mem split stages 16-byte moves through the reserved XMM scratch, not GP R11 whose id aliases the live XMM11 at size 16 (#2034); win64 companion counts a width-16 MIR_MOV. Two encoder regression tests pin the callee and the dispatch-level staging.
- Lane access reads through the shared array-of-lanes gep (#2042/#2044); the x86_64-only gep-into-vector descent is dropped.

## Verification bars (fresh, on the rebased tree)

| bar | result |
|-----|--------|
| runtime lane-exactness, discriminating values (arith incl. u8 200+100→44 / u16 300×300→24464 cross-byte carries; all six compares × every shape incl. i64x2/u64x2, unsigned orderings, float) | ✓ 0-exit |
| high-pressure spill: 16 live vectors, balanced-tree AND standalone chain, i32x4 + f32x4, distinct lanes (#2034) | ✓ 0-exit |
| comptime-const lane read + write (array-of-lanes path) | ✓ |
| SysV ABI round-trips (vector args/returns, mixed signatures, XMM-exhausting) | ✓ |
| dispatch-level encoder regression (16-byte mem-mem MOV → XMM14 staging) | ✓ fails pre-fix, passes post-fix |
| self-host fixpoint (gen1 == gen2) | ✓ byte-identical |
| scalar byte-identity (HEAD vs dev baseline, same input) | ✓ identical, both profiles |
| -g additivity (HEAD -g output vs baseline) | ✓ identical, both profiles |
| llvm-dwarfdump --verify (-g build) | ✓ No errors |
| full unit suite via fresh HEAD | ✓ 602 / 602 (dev baseline 600 → +2 encoder regression tests) |
| int suite (linux) incl. surface/debuginfo | ✓ all cases passed |
| no new int/ fixtures or root test/ | ✓ |

Runtime lane-exactness is proven by external self-checking programs (0-exit) rather than in-tree colocated `test` blocks: the seed predates the #2013 vector types and type-checks test blocks, so f32x4-bearing colocated tests break the bootstrap until the planned v3.3.2 vector-aware re-seed. The colocated runtime per-op suite lands post-re-seed as #2038. In-tree seed-safe coverage: the `vector_isel_defined` never-ICE canary + lane-access pipeline tests (#2042) + the two encoder regression tests here.

### DIFF
--- a/src/lang/be/codegen/mir.mach
+++ b/src/lang/be/codegen/mir.mach
@@ -460,11 +460,16 @@ pub rec MirBlock {
 # class:      index into target.arch.reg_classes selecting the register bank
 # spill_slot: stack-frame slot index when spilled (-1 while unspilled)
 # assigned:   physical register id assigned by regalloc (MIR_VREG_NIL until then)
+# vector:     true when the vreg holds a 128-bit SIMD vector; distinguishes a vector
+#             from a scalar float within the shared FP bank so regalloc 16-byte-aligns
+#             its spill slot (MOVAPS reloads require it) while scalar-float slots stay
+#             at align 8 (#2014)
 pub rec MirVReg {
     id:         u32;
     class:      u32;
     spill_slot: i32;
     assigned:   u32;
+    vector:     bool;
 }
 
 # discriminant distinguishing an addressed alloca slot from a by-value spill slot
--- a/src/lang/be/codegen/mir/context.mach
+++ b/src/lang/be/codegen/mir/context.mach
@@ -148,7 +148,7 @@ pub fun ctx_setup(ctx: *LowerCtx, tgt: *target.Target, fn: *ir.Function) R.Resul
     var i: u32 = 0;
     for (i < ctx.param_count) {
         val reg_class: u32 = reg_class_of(ctx, tgt, fn.params[i].ty);
-        val res_vreg: R.Result[u32, str] = new_vreg(ctx, reg_class);
+        val res_vreg: R.Result[u32, str] = new_vreg_vec(ctx, reg_class, value_is_vector(ctx, fn.params[i].ty));
         if (R.is_err[u32, str](res_vreg)) { ret R.err[bool, str](R.unwrap_err[u32, str](res_vreg)); }
         ctx.param_vregs[i] = R.unwrap_ok[u32, str](res_vreg);
         i = i + 1;
@@ -163,7 +163,7 @@ pub fun ctx_setup(ctx: *LowerCtx, tgt: *target.Target, fn: *ir.Function) R.Resul
             # now carries the IR pointer type (its element type rides on aux_ty),
             # so it classifies general-purpose without a special case.
             val reg_class: u32 = reg_class_of(ctx, tgt, inst.ty);
-            val res_vreg: R.Result[u32, str] = new_vreg(ctx, reg_class);
+            val res_vreg: R.Result[u32, str] = new_vreg_vec(ctx, reg_class, value_is_vector(ctx, inst.ty));
             if (R.is_err[u32, str](res_vreg)) { ret R.err[bool, str](R.unwrap_err[u32, str](res_vreg)); }
             ctx.instr_vregs[i] = R.unwrap_ok[u32, str](res_vreg);
         }
@@ -265,8 +265,23 @@ pub fun new_vreg(ctx: *LowerCtx, reg_class: u32) R.Result[u32, str] {
     f.vregs[vid].class      = reg_class;
     f.vregs[vid].spill_slot = -1;
     f.vregs[vid].assigned   = mir.MIR_VREG_NIL;
+    f.vregs[vid].vector     = false;
     f.vreg_count = f.vreg_count + 1;
     ret R.ok[u32, str](vid);
+}
+
+# append a fresh vreg that holds a 128-bit vector when `vector` is set, so regalloc
+# 16-byte-aligns its spill slot for MOVAPS reloads (#2014)
+# ---
+# ctx:       the lowering context
+# reg_class: register-class index into target.arch.reg_classes
+# vector:    whether the vreg holds a 128-bit SIMD vector
+# ret:       the new vreg id, or an allocation error
+pub fun new_vreg_vec(ctx: *LowerCtx, reg_class: u32, vector: bool) R.Result[u32, str] {
+    val r: R.Result[u32, str] = new_vreg(ctx, reg_class);
+    if (R.is_err[u32, str](r)) { ret r; }
+    ctx.f.vregs[R.unwrap_ok[u32, str](r)].vector = vector;
+    ret r;
 }
 
 # the result vreg mapped to an instruction id, or mir.MIR_VREG_NIL when the id is
--- a/src/lang/be/codegen/mir/lower.mach
+++ b/src/lang/be/codegen/mir/lower.mach
@@ -841,15 +841,26 @@ fun lower_instr(ctx: *lctx.LowerCtx, fn: *ir.Function, mb: *mir.MirBlock, iid: i
     ctx.cur_vec_lane = lctx.compute_vec_lane(ctx, inst);
 
     # a 128-bit vector reaching MIR lowering needs backend selection (#1965): a
-    # target whose backend wires it (aarch64 NEON here, x86_64 SSE2, rv64
-    # scalarization) lets it through to real lowering; a target without selection
-    # yet gets a defined per-target compile error, never an ICE or the silent 8-byte
-    # GP clamp `op_width_of` would otherwise apply. phis are eliminated and
-    # dbg-values emit no code, so neither is the erroring site.
+    # target whose backend wires it (aarch64 NEON, x86_64 SSE2, rv64 scalarization)
+    # lets it through to real lowering; a target without selection yet gets a defined
+    # per-target compile error, never an ICE or the silent 8-byte GP clamp
+    # `op_width_of` would otherwise apply. phis are eliminated and dbg-values emit no
+    # code, so neither is the erroring site.
     if (inst.kind != instr.OP_PHI && inst.kind != instr.OP_DBG_VALUE
-        && instr_touches_vector(ctx, inst)
-        && !arch_selects_vectors(ctx.tgt.arch.id)) {
-        ret R.err[bool, str](vector_unsupported(ctx));
+        && instr_touches_vector(ctx, inst)) {
+        if (!arch_selects_vectors(ctx.tgt.arch.id)) {
+            ret R.err[bool, str](vector_unsupported(ctx));
+        }
+        # x86_64: a lane-wise compute op (arithmetic / bitwise / unary not / compare-to-
+        # mask) rides the generic MIR opcodes carrying `vec_lane` directly, bypassing the
+        # scalar-specific float / divide / shift routing below (a vector float divide must
+        # not reach the integer `lower_div`); the x64 isel / encoder select the packed
+        # SSE2 form. load / store / move / call / ret / gep / alloca fall through to the
+        # shared handlers carrying the 16-byte width the vreg's `vector` flag set. aarch64
+        # (NEON) selects in its isa isel and falls through here for every vector op.
+        if (ctx.tgt.arch.id == isa.ARCH_X86_64 && is_vector_compute(inst.kind)) {
+            ret lower_generic(ctx, mb, inst, iid);
+        }
     }
 
     if (inst.kind == instr.OP_CALL) {
@@ -966,10 +977,11 @@ fun instr_touches_vector(ctx: *lctx.LowerCtx, inst: *instr.Instruction) bool {
 # whether the target arch's backend selects 128-bit vector ops to real machine
 # code, so a vector-typed instruction lowers rather than hitting the defined
 # "not yet supported" guard. each backend child flips its arch true as it wires
-# selection: aarch64 (NEON) here; x86_64 (#2014 SSE2) and riscv64 (#2016
-# scalarization) add their arm as they land.
+# selection: aarch64 (NEON) and x86_64 (#2014 SSE2). riscv64 needs no arm - its
+# scalarization pass rewrites vector ops before MIR lowering, so none reaches here.
 fun arch_selects_vectors(arch_id: u32) bool {
     if (arch_id == isa.ARCH_AARCH64) { ret true; }
+    if (arch_id == isa.ARCH_X86_64) { ret true; }
     ret false;
 }
 
@@ -1629,6 +1641,25 @@ fun is_compare_kind(k: instr.InstrKind) bool {
     ret k == instr.OP_CMP_EQ || k == instr.OP_CMP_NE ||
         k == instr.OP_CMP_LT_S || k == instr.OP_CMP_LT_U ||
         k == instr.OP_CMP_LE_S || k == instr.OP_CMP_LE_U;
+}
+
+# true for a lane-wise vector compute op that the x86_64 arm routes generically
+# rather than one of the memory / call / return ops that keep their dedicated handler.
+# a vector value flows to the generic lowering as its shared opcode carrying `vec_lane`;
+# the x64 isel / encoder select the packed SSE2 form from that descriptor (#2014). shifts
+# are rejected at sema (#2030) so none reach here, but routing them generically keeps the
+# isel loud backstop the single place a stray vector shift errors.
+# ---
+# k:   the IR instruction opcode
+# ret: true for a lane-wise compute op
+fun is_vector_compute(k: instr.InstrKind) bool {
+    if (k == instr.OP_ADD || k == instr.OP_SUB || k == instr.OP_MUL) { ret true; }
+    if (k == instr.OP_DIV_S || k == instr.OP_DIV_U ||
+        k == instr.OP_REM_S || k == instr.OP_REM_U) { ret true; }
+    if (k == instr.OP_AND || k == instr.OP_OR || k == instr.OP_XOR || k == instr.OP_NOT) { ret true; }
+    if (k == instr.OP_SHL || k == instr.OP_SHR_S || k == instr.OP_SHR_U) { ret true; }
+    if (is_compare_kind(k)) { ret true; }
+    ret false;
 }
 
 # true for the width-changing integer conversions (TRUNC / SEXT / ZEXT)
--- a/src/lang/be/codegen/regalloc.mach
+++ b/src/lang/be/codegen/regalloc.mach
@@ -2374,6 +2374,18 @@ fun class_width(lr: *LiveRange) u32 {
     ret 8;
 }
 
+# the spill-slot alignment for a spilled vreg: 16 bytes for a 128-bit vector (so its
+# MOVAPS reload is aligned), 8 bytes otherwise. keyed on the vreg's `vector` flag, so
+# scalar GP / float spill layout is unchanged and byte-identical (#2014)
+# ---
+# ctx:      the allocation context
+# value_id: the spilled vreg id
+# ret:      the slot alignment in bytes (8 or 16)
+fun spill_slot_align(ctx: *AllocCtx, value_id: u32) u32 {
+    if (value_id < ctx.f.vreg_count && ctx.f.vregs[value_id].vector) { ret 16; }
+    ret 8;
+}
+
 # append a SLOT_SPILL frame slot keyed on a vreg id and return
 # its slot index. mirrors `mir`'s param-spill slot construction so frame.run and
 # the encoder resolve the slot identically
@@ -2392,7 +2404,9 @@ fun add_spill_slot(ctx: *AllocCtx, value_id: u32, size: u32) R.Result[i32, str] 
     frame.slots[si].kind   = mir.SLOT_SPILL;
     frame.slots[si].offset = 0;
     frame.slots[si].size   = size;
-    frame.slots[si].align  = 8;
+    # a 128-bit vector spill slot is 16-byte aligned so its MOVAPS reload lands on the
+    # required boundary; scalar GP / float slots stay at align 8, byte-identical (#2014).
+    frame.slots[si].align  = spill_slot_align(ctx, value_id);
     frame.slot_count = frame.slot_count + 1;
     ret R.ok[i32, str](si::i32);
 }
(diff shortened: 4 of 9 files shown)


octalide:
## Root cause of the 16-byte spill clobber: encoder mem-mem staging, not regalloc

Fresh-eyes debugging handoff result. The allocator is fully correct; the clobber is introduced at **byte-encoding time** by `encode_inst`'s generic memory-to-memory split, which stages a 16-byte move through the GP scratch id `R11` — and at size 16 that raw id re-encodes as its XMM-bank shadow, **XMM11**, a live allocatable register.

### Mechanism (exact decision sequence)

Repro: the 16-leaf balanced-tree `i32x4` sum (`spill-repro.mach`), built from this branch head (afd0dbc2), exit 3 (s = 140, want 136).

1. **regalloc (correct):** 16 simultaneously-live vectors vs 14 allocatable XMM (bank exposes XMM0–XMM13) → `p14`/`p15` spill. Under the FP-as-MEM design their operands become frame-slot MEMs; no reload temps. Post-RA MIR is coherent: `p0..p13 → xmm0..xmm13`, `p14 → [rbp-384]`, `p15 → [rbp-416]`.
2. **rewrite (correct):** `p14`'s defining load `MOV p14 <- [rax]` becomes `MOV [slot p14] <- [rax]` — a 16-byte **mem-mem** MIR MOV. By design the encoder owns staging for this shape.
3. `mir_to_inst`: opcode `x64.MOV`, `size = 16`, both operands `OPK_MEM`.
4. **`encode_inst` (`src/lang/target/isa/x64/encode.mach:1723-1742`) — the faulty decision:** the generic mem-mem split runs *before any opcode/width dispatch* and hardwires the staging register to `x64.R11` (raw GP id 11) at `mi.size`:
   ```mach
   val scratch: i32 = x64.R11;
   var ld: isa.Inst;
   ld.opcode = x64.MOV;
   ld.dst    = isa.make_reg(scratch, mi.size);   # make_reg(11, 16)
   ```
5. The recursive `encode_mov` routes `size == 16` to `encode_fp_mov` (encode.mach:559), which sees dst = REG + src = MEM and emits `emit_movups_load(dst.reg = 11, …)`. `reg_bits`/REX use only the low 4 bits of the id and nothing validates the class tag, so GP id 11 byte-encodes as **XMM11**. The store leg does the same. Emitted bytes (offsets from the repro's `main.o`):
   ```
   2db: 44 0f 10 18   movups xmm11, [rax]        ; p14 capture staging — clobbers live p11
   2df: 44 0f 11 9d…  movups [rbp-0x180], xmm11
   315: 44 0f 10 18   movups xmm11, [rax]        ; p15 capture — clobbers again ({16,…})
   319: 44 0f 11 9d…  movups [rbp-0x1a0], xmm11
   ...
   388: 66 45 0f fe f3 paddd xmm14, xmm11        ; t5 = p10+p11 reads p15's bits
   ```
   `q2 = (p8+p9)+(p10+p11)` therefore reads `{16,…}` instead of `{12,…}` → +4 per lane → 46 vs 42 in the f32 variant, 140 vs 136 here.

**Invariant violated:** encoder staging must only use registers held outside the allocatable file (R10/R11 in the GP bank, XMM14/15 in the SSE bank). The split picks a register that satisfies this in the *GP* bank, but at size 16 the id crosses into the *XMM* bank, where index 11 is allocatable and live.

### Why the prior eliminations all held

- **Liveness / assign() clean:** the allocator genuinely never errs here. The clobber is invisible at the MIR level — the post-RA listing even prints the correct single instruction `mov [rbp-384], [rax]`.
- **`encode_fp_mov` mem-mem marker never fired:** `encode_inst`'s split intercepts *every* mem-mem shape before `encode_mov`/`encode_fp_mov` run, so the instrumented branch (encode.mach:511-513, the FLOAT_SCRATCH0 staging) is unreachable through `encode_inst` — dead code for real MOVs.
- **d898eb77's unit test passes and misses the bug:** it pins the *callee* (`emit_movups_load(FLOAT_SCRATCH0)` → XMM14, which is correct) but not the *dispatch* — the split never passes FLOAT_SCRATCH0; it passes R11.

### Why scalar f64 doesn't hit it

A scalar (width 4/8) mem-mem float copy stages through R11 as a plain GP MOV: bit-exact at 8 bytes, and R11 is reserved from GP allocation — no clobber, no truncation. Only width 16 crosses banks.

### Why the left-assoc chain "passed" — coincidence, not immunity

A standalone 16-leaf left-assoc chain has the identical failure shape: all 16 leaves are simultaneously live at the first add, `c14`/`c15` spill, their captures are the same mem-mem MOVs, and `c11` sits in xmm11 and is read later (`add xmm1, xmm0, xmm11` in the emitted asm). It only "passed" inside the big vtest `main` because whatever value occupied xmm11 at that point in that larger function was dead afterward. Any shape with ≥15 simultaneously-live vectors where xmm11's occupant is read after a spilled vector's def-capture fails.

### Minimal fix (validated locally)

Select the staging register by width in the split — 16-byte moves stage through the reserved XMM scratch; scalar widths keep R11:

```mach
     if (mi.src1.kind == isa.OPK_MEM && mi.dst.kind == isa.OPK_MEM &&
         opcode != x64.NOP && opcode != x64.RET && opcode != x64.SYSCALL &&
         opcode != x64.UD2 && opcode != x64.PAUSE && opcode != x64.CDQ &&
         opcode != x64.CQO && opcode != x64.MFENCE) {
-        val scratch: i32 = x64.R11;
+        # a 16-byte (vector) memory-to-memory move must stage through the reserved
+        # XMM scratch: the GP R11 id re-encoded at size 16 lands on its XMM-bank
+        # shadow (XMM11), a live allocatable register (#2034). scalar widths keep
+        # the GP R11 staging (bit-exact, and R11 is reserved from GP allocation).
+        var scratch: i32 = x64.R11;
+        if (mi.size == 16) { scratch = FLOAT_SCRATCH0; }
```

The recursion then does the right thing end-to-end: `encode_mov` → `encode_fp_mov` (composite id `0x010E` carries its class) → `movups xmm14, [rax]` / `movups [slot], xmm14`, byte-verified in the rebuilt repro. This is not a workaround — the split's design (stage mem-mem through a reserved scratch) is sound; it only ever chose the scratch from the wrong bank for the 16-byte width, because it was written before a 16-byte operand width existed.

**Required companion (win64 only):** `function_uses_fp_scratch` (`src/lang/be/codegen/regalloc.mach:2515`) deliberately excludes both-memory copies from the "clobbers the float scratch" scan (its doc says so). With the fix, a width-16 mem-mem MOV *does* clobber XMM14, which is callee-saved under win64, so the MIR_MOV arm needs `if (mi.width == 16) { ret true; }` (inert under SysV, where the scratch mask is 0).

Validation with the fix in-tree (linux-x86_64, from-source compiler):
- balanced-tree repro: exit 0 (was 3); staging bytes now `movups xmm14`
- standalone 16-leaf left-assoc chain: exit 0
- the branch's 26-case vector suite: exit 0
- `mach test .`: 589 passed, 0 failed

### arm64 (#2015) exposure: not exposed

The NEON branch stages spilled-FP mem-mem shapes explicitly through `FP_SCRATCH0 = V31` — held outside the allocatable V0–V29 bank (V30 is also held back) — in `encode_load` / `emit_store_mem`, and `emit_fp_mem` has native Q-width (16) LDR/STR forms with unscaled and register-offset fallbacks. The GP scratches IP0/IP1 are used only for address materialization, never as a 16-byte data conduit. The generic pick-a-GP-scratch-before-width-dispatch splitter is x64-specific. (riscv64 scalarizes and never carries a 16-byte unit through regalloc.)

### Regression shape that would have caught it

- **Runtime:** an execution test with 16 distinct-valued 128-bit vector locals simultaneously live plus a reduction with an exact-result assert — both i32x4 and f32x4, both balanced-tree and chain shapes. Distinct lane values per leaf are essential (identical values mask the aliasing). This belongs with this branch's planned runtime unit tests / an int case.
- **Encoder-level:** the dispatch-level counterpart of d898eb77's test — drive a full 16-byte mem-mem `MOV` `isa.Inst` through `encode_inst` (not `emit_movups_load` directly) and assert the staged register bytes are XMM14. That test fails on the current head and passes with the fix.

Not pushing anything to this branch — the fix hunk above is validated in a detached local worktree and is the branch owner's to apply.

Issue #2038, opened 2026-07-11
test: colocated runtime vector per-op suite across capable targets (post re-seed)


octalide:
Part of #1965. Depends on #2014, #2015, and a vector-aware seed release (the planned v3.3.2 re-seed).

## What

Tier 1's runtime lane-exactness is currently proven by out-of-tree self-checking programs documented in the backend PRs, because colocated `test` blocks cannot contain vector source yet: CI (and any local bootstrap) builds the tree with the released seed, `mach build` type-checks every `test` block, and every released seed through 3.3.1 predates the vector types — a block containing `f32x4{...}` fails the seed's sema before a fresh compiler exists. This is the same seed wall the `simd` manifest key hit, resolved by the same #1979 staging: once a vector-aware seed (v3.3.2, cut after the tier-1 backends merge) is the published seed, vector source in test blocks builds everywhere.

This child lands the permanent, colocated runtime per-op suite: one shared set of self-checking `test` blocks exercising every increment-1 table operator on every legal shape with per-lane results asserted against scalar reference computation — float `+ - * /`, integer `+ -` (all widths), integer `*` (16-bit lanes), bitwise `& | ^ ~`, all six comparisons (mask lanes all-ones/all-zeros, correct mask shape per element width), full-arity literals, comptime-const lane read and write, ABI round-trips (vector args/returns across real calls, mixed signatures), and spill pressure. The suite executes natively under `mach test` on x86_64 and aarch64 CI legs. It does NOT execute on riscv64 — `mach test` cannot run under qemu-user (per-test fork/exec is not emulable, established at #2016) — so rv64 execution coverage remains the separate owner decision on a `vectors` int/ exec case (see #2016's closing comment); the scalarize pass shares this suite's source once that lane exists.

## Acceptance

- The suite builds under the published (v3.3.2+) seed and passes via the freshly built HEAD `mach test` on x86_64 and aarch64.
- Every table op × legal shape asserted lane-exact; every comparison's mask shape asserted per element width.
- Suite-count delta accounted exactly; -g additivity both profiles; dwarfdump --verify clean; self-host fixpoint; int green.

Issue #2016, opened 2026-07-10
codegen: rv64gc defined unrolled scalar expansion of vector ops + build-time scalarization note


octalide:
Part of #1965.
Depends on #2013.
Coordinate with #2017 on the scalarization detection / note seam.

## What

The rv64gc backend for the portable-vector model. rv64gc has no vector unit at baseline (the V extension is out of scope), so it is the shipping instance of the *incapable target*: it lowers every lane-wise vector operator as a **fixed, defined, unrolled per-lane scalar sequence** rather than a packed instruction. The capable backends emit one packed instruction per operator (SSE2 on x86_64 under #2014, NEON on aarch64 under #2015); this child makes the identical `f32x4`-class semantics run correctly on a target with no SIMD hardware, and adds the arch-neutral machinery that detects and reports that scalarization happened.

The governing rule is the epic's: **scalarize operators, never algorithms.** A vector `+` becomes exactly N scalar `+` (one per lane); a vector compare becomes N scalar compares each materializing a full-width all-ones / all-zeros lane. It never expands an operator into a loop or an algorithm — it is a bounded, per-operator unroll. The precedent is already in the tree: the floating-point `%` operator has no single machine form and is lowered to a fixed multi-instruction IR sequence at lower time (`lower_float_rem` in `src/lang/me/lower/expr.mach`, dispatched from the `BIN_MOD` arm of `lower_binary` at ~1849-1857, body ~1885-1910 — a divide / truncate round-trip / multiply / subtract, chosen precisely because routing `%` to the integer opcode would miscompile the IEEE-754 bit patterns). rv64 vector scalarization is the same philosophy applied per lane.

### The scalarization pass (arch-neutral)

[...]

PR #2076, opened 2026-07-11
test(int): riscv64 vectors runtime case under qemu

### LINKED ISSUE #2075: int: riscv64 vectors runtime case (scalarization leg under qemu)
Part of #1965.

## What

riscv64 lowers the ten seeded 128-bit vector types through the defined unrolled scalarization backend (#2016), but its runtime behavior has no native-execution coverage. `mach test` cannot run under qemu-user (the dispatcher's fork/exec isn't emulable), and the colocated runtime vector suite (#2038) executes only on x86_64 and aarch64. The int/ framework's rv64 leg runs cross-built binaries under qemu, so one int/ `exec` case closes the gap and gives the scalarization path permanent CI coverage.

The case is a single mach program exercising every increment-1 operator category (float/int arithmetic, int multiply, bitwise, the six comparisons, lane read/write, ABI round-trips, spill pressure) across representative float and int, wide and narrow, lane shapes, printing discriminating lane values (colliding low bytes for comparisons, cross-byte carries for arithmetic) so a wrong element-width lowering can never pass silently. Built with a profile declaring `simd = "scalarize"`. The stdout golden is target-independent and shared, complementing #2038's native legs.

Owner approved adding this one int/ fixture on 2026-07-11 (the standing "no new int/ fixtures without sign-off" rule is satisfied for exactly this case).

## Acceptance

- Case passes on the `linux-riscv64` int leg in CI (cross-built, run under qemu-riscv64).
- Golden output is stable and shared across targets; the x86_64 leg is unaffected.

octalide:
Closes #2075. Part of #1965.

Adds `int/surface/vectors`, a single `exec` case that exercises every increment-1 vector operator category — float and integer arithmetic, integer multiply, bitwise `& | ^ ~`, the six comparisons, comptime lane read/write, ABI round-trips, and spill pressure — across representative float and integer, wide and narrow, lane shapes. Every printed lane is a **discriminating value** (cross-byte arithmetic carries, colliding-low-byte comparisons, sign-straddling ordering) so a wrong element-width lowering cannot reproduce the golden. The stdout golden is target-independent and shared across legs.

The point of the case is the `linux-riscv64` leg, where the ten seeded 128-bit vector types run through defined unrolled scalarization (#2016) under qemu — the runtime coverage the native suite (#2038, x86_64 + aarch64 only) cannot reach, since `mach test` can't run under qemu-user.

## Status: draft — the case caught a real, latent rv64 miscompile on its first run

The x86_64 leg (and by construction every capable/native leg) passes. The `linux-riscv64` leg currently **fails**, and the failure is a genuine defect, not a test error: signed sub-word (i8/i16) **ordering** comparisons (`< <= > >=`) scalarize incorrectly on rv64 — a negative narrow lane is read as its unsigned bit pattern before the 64-bit compare. Equality, unsigned compares, and i32/i64 signed compares are all correct; pure-scalar narrow signed compares are correct, so the bug is scalarize-specific. Tracked as a separate issue (linked below).

This PR stays draft until that fix lands on `dev`; then this branch rebases and the rv64 leg goes green — the case is its regression test.


octalide:
Blocked (draft) by #2077 — the rv64 scalarize signed sub-word ordering-compare miscompile this case caught. Once #2077's fix merges to `dev`, this branch rebases and the `linux-riscv64` leg goes green. This case is #2077's regression test.

octalide:
Rebased onto dev (with #2077's fix). Progress + a new blocker:

- **#2077 fixed the compares** — all 8 signed sub-word ordering lines (cmp.i8/i16 lt/le/gt/ge) now match on rv64.
- **x86_64 leg: fully green** (debug + release).
- **New blocker #2086** (regression on dev, not from this case or #2077): the always-on VC_CALL_ARG_TYPE verifier (b734ac5d) rejects the scalarize pass's by-address (ptr) lowering of vector call arguments against vector-typed callee signatures. Any rv64 vector call that survives inlining fails to build — at `-O0` always, and at `-O2` when the callee doesn't inline. This case's ABI round-trip block is the first thing on dev to exercise an rv64 vector call end-to-end.

Stays draft until #2086 lands on dev; then this case builds and passes on the rv64 leg at both profiles. It's the regression test for #2086 too.

octalide:
Ready. Rebased onto dev (through #2088/#2086). Verified locally with a from-source compiler built from this branch:

- **linux-riscv64: PASS at debug AND release** — the full case, compares + ABI round-trip block, all green under qemu-riscv64.
- **linux (x86_64): PASS at debug AND release.**

Both prior blockers (#2077 compare miscompile, #2086 verifier/scalarize-call regression) are fixed on dev; this case is the regression test for both.

octalide:
## Coordinator review

**Verdict: approve — merging.** Scope is exactly the owner-approved single fixture: three files under int/surface/vectors/, no compiler changes. The case earns its permanence twice over — it caught #2077 (scalarize signed sub-word ordering) and #2086 (verifier false positive on scalarized vector calls) before it ever merged, and now regression-tests both. Verified green at debug AND release on the rv64 qemu leg and x86_64, discriminating values throughout, CI matrix green.

Issue #2022, opened 2026-07-10
link: aarch64 GOT/PIC and shared-object (.so) emission


octalide:
Part of #1980.
Blocks #2015.
Unblocks #1752 deliverable 2.

## What

PR #2004 (#1980 phase 2a) landed the arch-neutral ELF shared-object seam: `kind = "shared"` now links a `MODE_LIBRARY` artifact into a real `ET_DYN` `.so` on **x86_64** — a `.dynsym`/`.hash` of the defined globals, a `.dynamic` with `DT_SONAME`, and `R_*_RELATIVE` base relocations, all behind the format-neutral `of.ExportSym` / `of.SharedFn` / `emit_shared` vtable slot, driven by `linker.mach`'s `emit_shared_object` (`src/lang/be/linker.mach:861`) → `collect_exports` (`:946`) → `tgt.of.emit_shared`. That seam was designed arch-neutral; what it lacks on aarch64 is the GOT/PIC codegen foundation. This issue is that foundation — the **aarch64 half** of shared-object support, which is also #1752 **deliverable 2** (aarch64 GOT / PIC).

The aarch64 reloc map today returns `R_AARCH64_NONE` (`ret 0`, the writer's unsupported-kind signal) for any GOT-relative kind: `arm64_elf_reloc_type` in `src/lang/target/isa/arm64/register.mach` (map body lines 68-81; the forward-map test at `:284` currently asserts `RK_GOTPCREL → 0`). PLT calls (`RK_PLT32 → R_AARCH64_CALL26`) and absolute page+lo12 addressing (`RK_PAGE21` / `RK_ADD_LO12` / the per-width `RK_LDST*_LO12`) are mapped, so imported *function calls* are partly covered, but true PIC and imported-*data* access through the GOT are not. Note the #1752 body's "`register.mach:225`" pointer is stale — the map body is `:68-81` and the model block sits at `~:222-259`.

### The work

[...]

PR #2079, opened 2026-07-11
feat(link): imported-data GOT consumer path (#2026)

### LINKED ISSUE #2026: link: imported-data access on dynamic targets — GOT consumer path needs an owner ruling
Follow-up to #2022. Informs #1752 (deliverable 2, imported-data leg). Part of the #1980 phase-2 family. Blocked on an owner ruling — the direction below changes codegen cost or language/manifest surface, so it is not a coordinator call.

## Problem

mach can import **functions** from shared objects (PLT: `RK_PLT32` → `R_AARCH64_CALL26` / the x64 equivalent) but not **data**. #2022 (PR #2023) landed the full aarch64 GOT codegen foundation — the `RK_GOT_PAGE21`/`RK_GOT_LO12` neutral reloc kinds mapped to `R_AARCH64_ADR_GOT_PAGE`/`R_AARCH64_LD64_GOT_LO12_NC`, the `R_AARCH64_GLOB_DAT` constant, and a unit-tested GOT-indirect encode primitive (`emit_global_addr_got`, threaded via `MirOperand.sym_got`) — but ships it **dormant**, and this RFC is why.

In the per-module compilation model, every cross-module data reference is `is_extern`/`SHN_UNDEF` at compile time — `ensure_extern_global` (`src/lang/be/codegen/mir/expr.mach:587`) is the single path for any undefined-global reference. Whether the symbol lives in another object of the *same artifact* (the overwhelmingly common case) or in a *genuine dynamic import* is link-time information. So there is no compile-time signal on which isel can select the GOT path narrowly.

Wiring `is_extern` → GOT routed *every* cross-module data reference through the GOT and broke ordinary aarch64 builds (the linker has no GOT-slot builder for link-defined data: `relocation 'got_page21' not supported` cross-building the compiler itself). It was reverted in PR #2023: link-resolved data keeps `adrp+add`, which is PIC-clean in `ET_DYN` (page-relative addressing is load-base-independent), so self-contained shared objects and static/PIE executables are unaffected.

The constraint that forces a decision: mach's linker has **no relaxation**. A data import must therefore be a GOT load *at compile time* — the compiler must commit before it can know the symbol's home.

## Proposal

Three viable directions; the ruling picks one (or defers via the do-nothing alternative).

1. **GOT-always for extern data** (what `-fPIC` without relaxation does): isel routes all `is_extern` data through the GOT; the linker builds the GOT — `R_*_RELATIVE` slots for link-defined symbols, `GLOB_DAT` for true imports. Complete and honest, but permanently costs every cross-module data access an indirection + a GOT slot + a RELATIVE reloc, including in static/PIE executables (the compiler binary itself). Could be gated per artifact kind (shared-only), at the cost of kind-dependent codegen for the same source.
2. **Explicit compile-time import signal**: dynamic-imported data is declared (in the language beside `ext fun`, in the manifest dependency entry, or both); only declared imports take the GOT path. No cost to normal builds; matches mach's existing explicitness (ext funs, soname deps). Adds language/manifest surface — the schema-evolution precedent (required-everywhere + coordinated sweep, #1979 machinery) applies if the manifest carries it.
3. **Copy relocations** (executables only): the linker allocates a local copy in `.bss` + `R_*_COPY`; keeps direct addressing. Classic ELF, but exe-only (a `.so` importing data still needs the GOT), and imports the well-known COPY-reloc versioning/size pitfalls.

Coordinator lean: (2) — explicit import declarations fit the language's design posture and cost nothing on the common path; (1) is the fallback if declaring imports is judged too much surface.

Honest limits: whichever direction, the `GLOB_DAT` leg cannot be execution-proven under qemu-user + static-PIE alone; it needs a dynamic-loader leg (native aarch64 or a sysroot — #1885-adjacent). And the decision is ISA-neutral: x64 shares the same latent gap (PR #2023 records whether x64 isel ever emits `RK_GOTPCREL` for data today; expectation is no).

## Alternatives

Do nothing yet: dynamic linking stays functions-only on every ISA — now an explicitly documented boundary (PR #2023, #1752). Viable until a real consumer needs imported data.

## Open questions

- Does any near-term consumer actually need imported **data**? #1752's Android glue (`libandroid`, EGL/GLES, Bionic libc) is overwhelmingly function-surface; #1980 P2c/P2d likewise. If nothing needs it, this can sit blocked indefinitely.
- If (2): where does the declaration live — an `ext` variable form in the language, the manifest dependency entry, or both?
- If (1) gated by artifact kind: is kind-dependent codegen for identical source acceptable?
- Verification: what is the minimal honest `GLOB_DAT` proof — native aarch64 CI leg (#1885), or a hosted-sysroot qemu leg?

octalide:
## #2026 — imported-data access on dynamic targets (GOT consumer path)

Wires the imported-data GOT consumer path deferred from #2022/#2023, per the owner ruling (2026-07-11). Resolves the `is_extern` fork with **Design B** (explicit compile-time import signal), the recommendation from the [brief on #2026](https://github.com/briar-systems/mach/issues/2026#issuecomment-4947797062).

`Closes #2026`

### The fork, resolved

At per-module codegen every cross-module data reference is `is_extern`/`SHN_UNDEF` and indistinguishable from a genuine `.so` import, and mach's linker has no relaxation, so the instruction shape (`adrp`+`add` direct vs `adrp`+`ldr :got:` GOT-indirect) must be committed at codegen. The signal that disambiguates: an **explicit `ext` declaration for data** (`ext val`/`ext var`), the strict mirror of `ext fun`. Only `is_import` data lowers GOT-indirect; every same-artifact reference keeps the PIC-clean `adrp`+`add`.

### What it does

- **IR / ME:** `ir.Global.is_import` (a strict subset of `is_extern`); `ext val`/`ext var` lowers to a storage-less import global; `ensure_extern_global` takes an import flag; a cross-module reference derives it from the origin decl's `DECL_FLAG_EXT` (mirroring `names_function`).
- **Frontend:** `ext val` is exempt from the initializer requirement (the exact analogue of the body-less `ext fun`); sema rejects an initializer on `ext` data; the driver registers `ext`-data bare names and `#[library]` pins, reusing the `ext fun` scans.
- **isel:** an `is_import` global sets `MirOperand.sym_got` → aarch64 `adrp :got:` + `ldr :got_lo12:`.
- **Linker:** a GOT-kind reloc against a data import is deferred as a `GotFixup` (not rejected); init/free plumbing.
- **ELF dyn-exec writer:** data-import `.dynsym` (STT_OBJECT) / `.dynstr` / `.hash` entries, a `.rela.dyn` `R_AARCH64_GLOB_DAT` table (DT_RELA/RELASZ/RELAENT), a data GOT region in the RW segment, and `patch_got_fixups` resolving each access to its slot. **Function-only dynamic images stay byte-identical** (ndata==0 leaves every structure unchanged).

### Before / after (aarch64)

- **Before** — a same-artifact `is_extern` data ref: `adrp x0, sym` + `add x0, x0, :lo12:sym` (direct, PIC-clean; unchanged).
- **After** — an `is_import` `ext val` ref: `adrp x0, :got:sym` + `ldr x0, [x0, :got_lo12:sym]` (the `add`→`ldr` GOT dereference), object-level relocs `R_AARCH64_ADR_GOT_PAGE` + `R_AARCH64_LD64_GOT_LO12_NC`.

### Verification (from-source compiler, this branch)

- **End-to-end aarch64 GLOB_DAT (structural).** `libcounter.so` exports a data global `g_counter` (type D); a consumer reads it via `ext var g_counter` and links the `.so`. The consumer image: `ET_EXEC`/AArch64, `DT_NEEDED libcounter.so`, `BIND_NOW`, one `R_AARCH64_GLOB_DAT` binding GOT slot `0x450018` to `g_counter`. `RELASZ`=24 = exactly one dynamic reloc → **PIC-clean, no ABS64-in-text**. The consumer's `read_counter` disassembles to `adrp x0, 0x450000` + `ldr x0, [x0, #0x18]`, resolving to exactly `0x450018`.
- **Full `mach test`:** 644 passed, 0 failed — the dev baseline (643) **+1** (the added `parse_bind_decl:ext_data_import` block; exact accounting).
- **Self-host fixpoint:** gen2 == gen1, byte-identical (release).
- **x86_64 output byte-identical:** the baseline (dev) and this-branch compiler produce byte-identical output for a fixed non-`ext` program (x86_64 has no `sym_got` emit path).
- **`-g` additivity:** the compiler's release load image is byte-identical with/without `-g` (0 load-image body diffs; only the ELF header SHT-pointer fields differ, plus the appended debug SHT tail). Release image is program-headers-only.
- **`llvm-dwarfdump --verify`** on the `-g` build: **No errors.**
- **int suite (linux leg, native):** all cases pass. An existing aarch64 exec fixture runs correctly under qemu-aarch64 (rc 0).

### Seed-parseability (rider 1)

`ext var` parses under the 3.4.4 seed **unchanged** (`ext` was already a general decl flag). `ext val` (no initializer) needs this PR's parser exemption, so it is **not** seed-parseable — future in-tree `ext val` usage would need a re-seed first. The compiler itself uses no `ext` data, so no seed dance is required for this change, and the feature is proven with string-based parser tests + out-of-tree structural fixtures (not colocated `ext val` test blocks).

### Tracked residue

- **Executed `dlopen` resolution of a `.so`'s imported data** (the runtime `GLOB_DAT` bind) needs a native aarch64 loader/sysroot leg — **#1885**, absent in this environment. The slot/relocation/disasm are proven structurally here, matching the #2022/#2023 verification bar; runtime execution is the tracked follow-up.
- **Shared-object writer (backstop).** `patch_got_fixups` is consumed only by the dynamic-executable writer; `emit_shared` has no GOT-slot construction. This is **not reachable today** — a `LINK_SHARED` link never activates the dynamic-import plan (`dynamic` is executable-only), so `dyn.active` stays false, no GOT fixup is collected, and an `ext`-data reference in a `.so` errors loudly at patch time (undefined symbol, or the arm64 packer's unsupported-relocation error for a resolved GOT kind — verified). A loud hard error is nonetheless added in `emit_shared` (`got_fixup_count > 0` → "data imports in a shared object are not yet supported (#2026 is exec-only)") as defense-in-depth for when `.so`-imports-`.so` is wired later. The normal `.so` path stays byte-identical (guard inert at count 0).
- **x86_64 data imports** remain unwired (x64 has no `sym_got` emit path); the RFC records this shared latent gap. aarch64 is the delivered consumer.
- A **data-only** dynamic image (no function imports) carries a zero-size R+X PT_LOAD for the empty `.plt` — valid ELF (kernel/ld.so skip a zero `p_memsz` segment), a pre-existing degenerate-case shape now reachable via data-only imports.

### DIFF
--- a/doc/language/ext-fun.md
+++ b/doc/language/ext-fun.md
@@ -144,5 +144,6 @@ is currently implemented for the ELF (Linux) target; the PE (Windows) and Mach-O
 ## See also
 
 - [fun.md](fun.md) — regular function declarations
+- [val-var.md](val-var.md) — `ext val` / `ext var`, the data analogue (foreign data imports)
 - [visibility.md](visibility.md) — `pub` and `ext` modifiers
 - [decorators.md](decorators.md) — `symbol`, `library`, and other codegen decorators
--- a/doc/language/val-var.md
+++ b/doc/language/val-var.md
@@ -34,6 +34,29 @@ At module top level:
 
 Inside functions, they are local to the enclosing block.
 
+## `ext` — foreign data imports
+
+`ext val` / `ext var` declares a binding whose storage lives in another object,
+imported by name — the data analogue of [`ext fun`](ext-fun.md). It is a
+forward reference the linker resolves, so it is **storage-less** and carries no
+initializer:
+
+```mach
+ext var errno: i32;                        # imported mutable datum
+#[symbol("environ")] ext var env: **u8;    # renamed import
+#[library("libfoo.so")] ext val foo_flags: u32;  # library-pinned import
+```
+
+- No initializer. `ext val x: T = ...;` is an error — the definition, and its
+  value, live in the providing object (mirrors `ext fun`'s absent body).
+- The `symbol` and `library` decorators and the static/dynamic linking inputs
+  work exactly as for [`ext fun`](ext-fun.md).
+- On a dynamic target the reference is emitted GOT-indirect so the loader binds
+  it to the runtime definition (ELF: an `R_*_GLOB_DAT` GOT slot); an ordinary
+  cross-module reference to a `val`/`var` defined elsewhere in the same artifact
+  stays directly addressed. Executed dynamic-import resolution is proven on the
+  native ELF legs.
+
 ## No inference
 
 Every binding declares its type. An untyped numeric literal is checked
--- a/src/lang/be/codegen/mir/context.mach
+++ b/src/lang/be/codegen/mir/context.mach
@@ -502,15 +502,16 @@ pub fun lower_value(ctx: *LowerCtx, v: value.Value) mir.MirOperand {
     }
     if (v.kind == value.VAL_GLOBAL) {
         if (v.data.global_ix < ctx.m.global_len) {
-            # NOTE: an imported (`is_extern`) global is preemptible and would load
-            # GOT-indirect (set `o.sym_got = true` here). that wiring is withheld
-            # pending RFC #2026: with no compile-time import signal and no linker
-            # GOT relaxation, routing every `is_extern` reference through the GOT
-            # would tax all cross-module data access. `is_extern` data therefore
-            # keeps the position-independent adrp+add form for now; the GOT-indirect
-            # codegen foundation (RK_GOT_*, emit_global_addr_got, MirOperand.sym_got)
-            # ships dormant.
-            ret mir.op_sym(ctx.m.globals[v.data.global_ix].name, 0);
+            # a genuine foreign/dynamic data import (`is_import`, an `ext val`/
+            # `ext var`) is preemptible: its definition lives in another object
+            # bound at run time, so its address is loaded GOT-indirect (#2026). a
+            # plain same-artifact cross-module reference (`is_extern` but not
+            # `is_import`) is resolved within the artifact and keeps the direct,
+            # position-independent adrp+add form. the linker builds the GOT slot -
+            # GLOB_DAT for a dynamic import, a filled slot for a static definition.
+            var o: mir.MirOperand = mir.op_sym(ctx.m.globals[v.data.global_ix].name, 0);
+            if (ctx.m.globals[v.data.global_ix].is_import) { o.sym_got = true; }
+            ret o;
         }
         # an in-range global index is an IR invariant; an out-of-range one is a
         # lowering bug -- fail loudly rather than emit a STR_NIL symbol the
--- a/src/lang/be/linker.mach
+++ b/src/lang/be/linker.mach
@@ -142,6 +142,7 @@ rec DynState {
     fixup_count: u32;
     fixup_cap:   u32;
     base_reloc_cap: u32;
+    got_fixup_cap:  u32;
     active:      bool;
 }
 
@@ -2072,10 +2073,11 @@ fun patch_module_relocations(s: *session.Session, out_img: *of.ObjectImage,
             }
             or {
                 # not defined anywhere: a dynamic import, or a hard error. a
-                # pc-relative call to a FUNCTION import is deferred to the writer
-                # (only function imports get a PLT stub); any other unresolved
-                # reference - including a call kind against a data import (deferred)
-                # - is rejected as undefined.
+                # pc-relative call to a FUNCTION import is deferred to the writer as
+                # a PLT fixup; a GOT-indirect access to a DATA import is deferred as
+                # a GOT fixup (the writer builds its GOT slot + GLOB_DAT). any other
+                # unresolved reference - a call against a data import, an absolute
+                # reference to any import - is rejected as undefined.
                 val imp: R.Result[u32, str] = dynstate_import_index(dyn, r.symbol);
                 if (R.is_ok[u32, str](imp) && is_call_kind(r.kind)
                     && dynstate_import_is_func(dyn, R.unwrap_ok[u32, str](imp))) {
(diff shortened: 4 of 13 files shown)


octalide:
## Coordinator review (owner grammar approval already recorded)

**Verdict: approve pending one fix** — one silent-drop hazard on the shared-object path; everything else is clean.

### What the PR does
Completes `ext` to data declarations as the explicit dynamic-import signal (RFC #2026 Design B): parser exemption (`ext val` no-init, exact mirror of body-less `ext fun`) + sema initializer reject + driver link-name/library registration + `ir.Global.is_import` (strict subset of `is_extern`) threaded from decl lowering and cross-module reference sites → `MirOperand.sym_got` only for imports → linker defers GOT-kind relocs against data imports as `GotFixup` → ELF exec writer lays out `.rela.dyn` GLOB_DAT + a data GOT region, extends `.dynsym`/`.dynstr`/`.hash`, and patches the `adrp`/`ldr` halves.

### Correctness — checked and sound
- `is_import` propagation is monotonic (`ensure_extern_global` upgrades an existing extern, never downgrades); same-artifact refs keep `adrp+add`.
- `.dynsym` index math: data entries follow function entries; `.rela.dyn` `d_sym = nimp+1+ordinal` matches the write order; GOT slot ordinal = `data_import_ordinal` in the same plan order. Consistent.
- `ndata == 0` leaves every layout size untouched → function-only images byte-identical (CI-proven).
- Data GOT sits in the RW segment (loader-writable), `.rela.dyn` in RO; `DT_RELA/RELASZ/RELAENT` only when `ndata > 0`; BIND_NOW arches unaffected.
- `patch_got_fixups` bounds-checks segment index and offset, and fails loudly on an unexpected kind — `RK_GOTPCREL` is in `is_got_kind` but x86_64 has no emit path, so the loud reject is the right honest gap.
- Memory: `got_fixups` grow/free paired via `got_fixup_cap` in `free_dynstate`.

### The one required fix
`patch_got_fixups` is called only from `emit_dyn_exec`. `emit_shared` (P2a: self-contained, no imports) never consumes `dyn.got_fixups`. If a `kind = "shared"` link can reach `dynstate_add_got_fixup` (a .so referencing an `ext var`), the fixups are **silently dropped** and the image ships `adrp`/`ldr` with zero immediates — a silent mispatch, the exact class this program exists to kill. Either:
1. confirm the path is unreachable (the driver/linker rejects a shared output with dynamic imports before patching) and say where, **and**
2. add the loud backstop in `emit_shared` regardless: `got_fixup_count > 0` (or `import_count > 0`) → hard error "data imports in a shared object are not yet supported (#2026 exec-only)". Precedent: the missing-validator and out-of-table isel backstops.

### Minor (no action required)
- `data_import_ordinal` is O(n) per fixup → O(n²) overall; fine at these import counts and consistent with `func_import_count` style.
- `lower_global`'s ext branch pins `is_pub = false`; matches the `ext fun` posture (the import is module-local surface), noting it here so it's a recorded choice.

Tests: parser round-trip + ext-only exemption negative is right; e2e structural proof + CI 14/14 (incl. native aarch64 full suite) is the agreed bar; executed-dlopen residue correctly tracked at #1885.

Issue #2064, opened 2026-07-11
codegen: duplicate same-name Function entries per generic instance (extern decl + definition); hides instance bodies from the inliner


octalide:
Surfaced while fixing #2035. A referencing module carries two `Function` entries for the same generic instance: the signature-only extern declaration created at the reference site, and the materialized WEAK definition emitted by the drain (visible in emitted IR as `fn @...(): void extern` at reference position plus a full definition later). Beyond the redundancy, the extern-decl shape means instance BODIES are invisible to the inliner in referencing modules (they're extern at inline time), so cross-instance inlining opportunities are silently lost.

Structural; the owner may want it addressed with care. Queued on the program backlog for post-#2035 owner review. Related to the sig-carrying work in the sibling follow-up.

Issue #2058, opened 2026-07-11
rfc: win64 vector ABI conformance — should ext-boundary calls marshal __m128 by reference?


octalide:
Part of #1965. Follow-up to #2055 (which classified win64 vectors as XMM to fix the shipped CLASS_GP miscompile). **Blocked on owner** — a language-level ABI commitment, not to be designed under a release clock.

## The fork

#2055 shipped mach's win64 vector ABI as **internal XMM**: a 128-bit vector is passed in an XMM argument register and returned in XMM0, identical to the SysV / AAPCS64 by-value-in-a-vector-register convention and consistent across all four ABIs. It reuses the merged #2053 capture path with zero new machinery and is wine-proven (full windows suite green).

The **Microsoft x64** convention is different on the argument side: `__m128` / `__m128i` / `__m128d` are **never passed by immediate value** — the caller allocates a copy and passes a pointer (by reference). The return side agrees (XMM0). So mach's internal-XMM deviates from MS x64 only for **arguments**.

This is harmless today: mach's `f32x4` etc. are its own types with no C FFI surface — every vector call is mach→mach, where internal-XMM is correct. It becomes a conformance question only if a windows `ext` fun ever takes or returns a vector-typed value that must bit-compatibly match a C `__m128` boundary.

## What MS-conformant by-reference would cost (scoping from #2055)

The argument side needs real work, because a vector is register-class (unlike an aggregate, which is already memory-resident):
- **Caller**: materialize a register-resident vector to a temp stack slot before taking its address (a value-vreg is not an address; `addr_to_vreg` treats it as one today).
- **Callee**: a by-reference vector parameter arrives as a pointer; the callee needs a by-ref-vector value model (load the vector from the pointer) distinct from how it treats an aggregate-by-ref param.

That two-sided machinery is its own scoped PR if this is ruled in.

## Design thought worth recording

If C vector FFI ever matters, the mach-consistent answer is likely **conformant marshalling at explicit `ext` boundaries only** — keep the internal convention XMM everywhere (mach→mach), and marshal to/from the MS `__m128` by-reference shape only when crossing a declared `ext` boundary. That matches the explicitness philosophy in RFC #2026 (import-signal lean): ABI conformance is a property you opt into at the boundary, not a tax on every internal call.

## Ask

[...]

Issue #2035, opened 2026-07-11
codegen: -O2 miscompiles inlined accept-loop callee; fresh local Vector reads garbage .len (SIGSEGV)


octalide:
-O2 miscompiles a binary whose `main` inlines a callee containing an infinite `tcp.accept` loop: a freshly `vector.init`'d local `Vector` reads a garbage stack address as `.len` (e.g. `140737195795104` instead of `0`), causing an out-of-bounds element access and SIGSEGV on the first iteration. The identical code is correct at `-O0`, and the same library code is correct at `-O2` when *not* inlined into `main` (a 39-test suite including a threaded end-to-end broker exercise passes under `--profile release`). Found while implementing `brokerd` in briar-systems/mach-mqtt (PR briar-systems/mach-mqtt#2): the daemon binary is the only shape that miscompiles.

## Repro

Two-file project, `mach build . --profile release`, run the binary.

`src/svc.mach`:

```mach
use allocator: std.allocator;
use page: std.allocator.page;
use vector: std.collections.vector;
use tcp: std.net.tcp;
use print: std.print;
use std.types.size.usize;
use std.types.result.Result;
use std.types.result.is_err;
use std.types.result.unwrap_ok;
use std.types.string.str;

rec ConnCtx { fd: i32; }

fun reap(conns: *vector.Vector[*ConnCtx]) {
    print.printlnf("DBG reap len={}", vector.length[*ConnCtx](@conns)::u64);
    var i: usize = 0;
    for (i < vector.length[*ConnCtx](@conns)) {
        val ctx: *ConnCtx = @unwrap_ok[**ConnCtx, str](vector.get[*ConnCtx](conns, i));
        i = i + 1;
    }
}

pub fun serve(ln: *tcp.Listener) {
    var pa: allocator.Allocator;
    page.make(?pa);
    var conns: vector.Vector[*ConnCtx] = vector.init[*ConnCtx](?pa);
    for {
        reap(?conns);
        val ar: Result[tcp.Stream, str] = tcp.accept(ln);
        if (is_err[tcp.Stream, str](ar)) { brk; }
        val stream: tcp.Stream = unwrap_ok[tcp.Stream, str](ar);
        val ca: Result[*ConnCtx, str] = allocator.allocate[ConnCtx](?pa, 1);
        if (is_err[*ConnCtx, str](ca)) { cnt; }
        val ctx: *ConnCtx = unwrap_ok[*ConnCtx, str](ca);
[...]

PR #2080, opened 2026-07-11
fix(me): carry real IRT_FN signatures on generic-instance callee references (#2063)

### LINKED ISSUE #2063: codegen: generic-instance callee references carry no signature (ABI classified per-operand)
Follow-up to #2035 (which shipped the companion inliner fix). The architectural invariant behind that bug remains: a direct call's ABI should classify from the callee's declared signature, but generic-instance callee references are typed bare `ptr` (`lower_generic_callee` at src/lang/me/lower/expr.mach:1738-1740, plus the value-instance :1607-1609 and pack :1822-1824 paths), so `ensure_extern_function` records `sig = ptr`, `emit_call` stamps a degenerate `aux_ty`, and every generic call falls to the per-operand classification fallback (`lower_call`, mir/abi.mach). That fallback is correct only because the FE stamps arg operands with the declared aggregate type — a fragile contract (the #2035 inliner clobber is what exposed it).

## The right end state
Give instance callee references their true `IRT_FN` (decl param/ret sema types substituted with the concrete type-args, lowered and interned — the `pack_instance_sig` machinery in decl.mach), so calls classify from the declared signature, inlined or not.

## Implementation requirement (why the naive fix regresses)
A straightforward caller-side computation — `context.lower_type(generics.substitute(decl_type_of(callee), type_args))` or the equivalent temp-subst — REGRESSES correct code. Evidence: with that sig, a `Result[usize,str]` = `{i8,{i64,ptr}}` param comes out pointer-width and the call passes it in a register while the callee homes it by value from the stack. Root cause (per the #2035 debugger, corroborated): the nested-generic param needs the callee decl's type-param binding to resolve; resolved in the CALLER's context that binding is absent, so `lower_type` takes its documented pointer-width erasure fallback (context.mach:889-897). A concrete param (`Trip{i64,i64,i64}`) resolves fine; only nested generics erase.

The ABI classifier itself is SOUND — not the culprit: a non-generic function taking a `{i8,{i64,ptr}}`-shaped 24-byte aggregate by value is correct at -O0 and -O2 (sysv `classify_arg` at target/abi/sysv.mach:159-165 is shape-blind and has no by-reference form for args). So this is purely about computing the reference sig correctly: derive it in the callee's lowering context, or reuse the enqueued-instance record materialization already builds, rather than re-deriving caller-side.

No release gate — the companion fix (#2035) closes the runtime miscompiles; this is the durable correctness/architecture cleanup.

## Durable guard (lands with this fix)
Once instance callee references carry real IRT_FN signatures, the ALWAYS-ON exit verifier can assert each call's argument operand type against the callee's declared parameter type — at which point the entire #2035 defect class (a by-value aggregate arg mis-typed to a pointer) becomes structurally impossible to ship. This is an additional argument for doing this work: the fix and its permanent tripwire land together. (A --verify-ir check of the same invariant cannot be added ahead of this issue, because generic callees have no signature to check against today — that is the very gap this issue closes.)

octalide:
Closes #2063.

Follow-up to #2035 / #2062. The architectural invariant behind the #2035 miscompile: a call's ABI must classify from the callee's declared signature. Generic, comptime-value, and comptime-pack **instance** callee references were typed as a bare `ptr`, so `ensure_extern_function` recorded `sig = ptr`, `emit_call` stamped a degenerate `aux_ty`, and every such call was classified per-operand from its argument stamps — the fragile contract the #2035 inliner clobber exposed.

## Fix
Each of the three instance callee-reference paths (`lower_generic_callee`, `lower_value_callee`, `lower_pack_callee`) now carries the instance's true `IRT_FN`, computed at the reference site by:
- binding the **origin module's** typing + the concrete substitution env (exactly as the instance drain does), then
- lowering the template's declared function type through the **identical path** the drained definition uses.

A reference's signature is therefore byte-identical to its definition's, so every fixed argument classifies from the declared parameter type — inlined or not. `function_return_ir` and `pack_instance_sig` moved to `context.mach` so the definition and the reference-signature helper share one implementation.

### Why not a caller-side `lower_type`
`lower_type(substitute(decl_type_of(callee), type_args))` **regresses** correct code: a nested generic param (`Pair[T]` where `T` is the callee's own type parameter) needs the callee decl's type-parameter binding, absent in the caller's context, so `lower_type` takes its pointer-width erasure fallback and the param comes out 8 bytes wide. Binding the origin context avoids this.

## Per-operand classification is now scoped, not a fallback
The per-operand argument-classification path is **not** removed — but its role is now precise: it applies only at positions where **no declared signature exists**, never as a fallback for a direct call whose signature merely went uncomputed (that gap is exactly what this fix closes). Concretely it is reached at:
- the **variadic tail** past the declared parameter count, and
- a callee whose IR value type is **not** an `IRT_FN` — an erased / untyped function pointer.

**A typed function-pointer indirect call is NOT such a position** (stated for the record per review): the loaded fn-pointer value carries the pointee `IRT_FN` as its IR type (`TYPE_FUN` lowers to `IRT_FN`), so `emit_call` stamps a real `aux_ty` and the call classifies from the pointee signature. Verified end-to-end: a typed `fun(Big) i64` pointer called indirectly with a 32-byte aggregate emits `%N = call i64 %fp: fn({i64,i64,i64,i64}) -> i64, %arg: {i64,i64,i64,i64}` and stages the aggregate to the outgoing stack (CLASS_STACK). The compiler's own backend ABI vtable (`abi.ArgPassingFn = fun(...AggLayout) ParamSlot`, an aggregate-in/aggregate-out typed fn pointer) is exactly this case and is covered. The one adjacent position genuinely not covered is a call through an *erased/untyped* `ptr` (none arise in well-typed mach today) — noted as adjacent to #2064 / the win64 ext-marshaling ruling (#2058), **not** widened here.

## Durable guard: always-on `VC_CALL_ARG_TYPE`
Now that every direct call and every typed-fn-pointer indirect call carries a real `IRT_FN`, the verifier asserts each fixed argument operand type against the callee's declared parameter. An aggregate parameter must receive an aggregate operand — a pointer/scalar there is the #2035 mis-typing, now unrepresentable. (The verifier has no target, so it cannot compare two distinct aggregate ids' ABI classification; both flowing by address is the load-bearing, target-independent invariant.)

**Promoted to always-on** (entry + exit verifier and every `--verify-ir` pass), **not** the opt-in `repr` family. Evaluation: clean across the whole compiler self-build, all unit-test modules at -O2, the stdlib, and mach-mqtt; **no measurable build-time delta** (release self-build steady at ~13s with and without it) — the check is a handful of type-table lookups per call argument. Kept out of `repr` because it needs no target and does not false-positive. Colocated test drives it directly (opt-independent → guards in CI's default `mach test`).

## Verification (from-source fix/2063 compiler, never the PATH seed)
- `mach test` **644/644** at **-O0 / -O1 / -O2** (643 baseline + the colocated verifier test; incl. the 36-test `vector_runtime` suite).
- Self-host **fixpoint g2 == g3** byte-identical. g1 != g2 — the fix corrects the compiler's own generic-instance ABI classification (expected, surgical).
- **ASM**: a 32-byte by-value aggregate arg copies to the outgoing stack (`[rsp..+24]`, CLASS_STACK); a nested-generic `Pair[T]`->`Pair[i64]` 16-byte arg passes in two GP registers (rdi/rsi) — refuting the pointer-width erasure regression, not an 8-byte register pointer.
- **mach-mqtt brokerd @ -O2** (#2035 field case): listens, blocks in accept, no SIGSEGV; mqtt suite 46/46 at -O2.

## Follow-up (not this PR)
The dual same-name Function entries per instance (#2064) are intentionally untouched; this fix makes both entries carry the real sig, so calls resolving to either stay consistent.

### DIFF
--- a/src/lang/me/ir/verify.mach
+++ b/src/lang/me/ir/verify.mach
@@ -82,6 +82,16 @@ pub val VC_DANGLING_ID:     VerifyCheck = 12;
 # const-int block id). opt-in alongside the other `repr` checks
 pub val VC_TYPE_AGREE:      VerifyCheck = 13;
 
+# a direct call's fixed argument operand type disagrees with the callee's declared
+# parameter type: an aggregate parameter received a non-aggregate (pointer/scalar)
+# operand - the #2035 class, a by-value aggregate arg mis-typed to a pointer - or a
+# scalar parameter received an incompatible operand. only checked when the call
+# carries a real IRT_FN signature (every direct call, since generic instances now
+# do too); an indirect / untyped-pointer call and the variadic tail are not
+# checked. always on - the durable tripwire that keeps the #2035 defect class from
+# shipping
+pub val VC_CALL_ARG_TYPE:   VerifyCheck = 14;
+
 # one invariant failure, located as precisely as the failing check allows
 # ---
 # function: index into Module.functions
@@ -192,6 +202,7 @@ pub fun describe(check: VerifyCheck) str {
     if (check == VC_TYPE_RESOLVE)    { ret "type resolution: operand or aux type does not resolve"; }
     if (check == VC_DANGLING_ID)     { ret "dangling id: a block list names an out-of-range instruction id"; }
     if (check == VC_TYPE_AGREE)      { ret "type agreement: operand types violate the opcode's typing contract"; }
+    if (check == VC_CALL_ARG_TYPE)   { ret "call argument typing: a call argument's type disagrees with the callee's declared parameter"; }
     ret "unknown verification check";
 }
 
@@ -526,6 +537,12 @@ fun check_one_instr(m: *ir.Module, fn: *ir.Function, iid: id.InstructionId, fi: 
         val res_agree: R.Result[bool, str] = check_type_agree(m, ins, fi, bid, iid, r);
         if (R.is_err[bool, str](res_agree)) { ret res_agree; }
     }
+    # always on: a direct call's fixed argument types must agree with the callee's
+    # declared IRT_FN signature. now that generic instances carry real signatures,
+    # this makes the #2035 defect class (a by-value aggregate arg mis-typed to a
+    # pointer) structurally impossible to ship.
+    val res_call: R.Result[bool, str] = check_call_arg_types(m, ins, fi, bid, iid, r);
+    if (R.is_err[bool, str](res_call)) { ret res_call; }
     ret R.ok[bool, str](true);
 }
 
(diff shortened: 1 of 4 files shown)


octalide:
### Verification (rebased on dev @ #2078; from-source fix/2063 compiler)

**Core fix**
- `mach test` **644/644** at **-O0 / -O1 / -O2** (643 baseline + the new colocated verifier test; incl. the 36-test `vector_runtime` suite).
- Self-host **fixpoint g2 == g3** byte-identical. g1 ≠ g2 — the fix corrects the compiler's own generic-instance ABI classification (expected).
- **ASM** (generic aggregate args at -O2, non-inlined cross-module instance calls): a 32-byte by-value aggregate copies to the outgoing stack (`[rsp..+24]`, CLASS_STACK); a **nested-generic** `Pair[T]`→`Pair[i64]` 16-byte arg passes in two GP registers (rdi/rsi) — refuting the pointer-width erasure regression, not an 8-byte register pointer.
- **mach-mqtt brokerd @ -O2**: listens, blocks in accept, no SIGSEGV; mqtt suite 46/46 at -O2.

**Durable guard (`VC_CALL_ARG_TYPE`, always-on)**
- Asserts each direct-call fixed-arg operand type against the callee's declared IRT_FN parameter. An aggregate parameter must receive an aggregate operand — a pointer/scalar there is the #2035 mis-typing, now unrepresentable.
- **Promoted to always-on** (entry + exit verifier, every `--verify-ir` pass), not the opt-in `repr` family. Proven clean across the whole compiler self-build, all unit-test modules at -O2, the stdlib, and mach-mqtt; no measurable build-time delta.
- Not gated as `repr` because it needs no target and does not false-positive; the opt-in `repr` checks stay opt-in. Colocated test drives it directly (opt-independent → guards in CI's default `mach test`).

**On "kill the per-operand ABI fallback"**: the per-operand path is retained for *true varargs tails* and *indirect calls through untyped function pointers* (no declared param types exist there). What the fix eliminates is the fallback applying to a *direct* call's fixed args; the always-on guard makes a direct-call aggregate-vs-pointer mismatch unrepresentable.

octalide:
## Coordinator review

**Verdict: approve — merging.** The load-bearing structural fix for the #2035 defect class, done the right way.

### Correctness
- `instance_ref_sig` / `pack_instance_ref_sig` bind the **origin module's** AST + typing + concrete subst env and lower through the identical path the drained definition uses — the only construction that can't diverge from the definition (the naive caller-side `lower_type` regression is refuted by the nested-generic ASM proof). State restore is unconditional on every exit path, including errors.
- The three callee-path rewrites correctly reorder deallocation so `args`/`types` stay alive until the signature is computed — the previous free-before-use shape would have been a use-after-free with the new call.
- `pack_instance_sig` is called inside the origin swap, so its `lc.a.typed_names` reads hit the origin's param pool — load-bearing and correct.
- `VC_CALL_ARG_TYPE` leniency table is exactly the invariant and nothing more: aggregate param ⇒ aggregate operand (by-address flow, target-less verifier rightly declines ABI-class comparison); pointer param accepts any address form; `IRT_NIL` carries no information; exact match otherwise. Skips (indirect/untyped callee, variadic tail, arity-short) are each the intended scope. Always-on with measured-negligible cost across the self-build is the right promotion.
- Shared-helper move (`function_return_ir`, `pack_instance_sig` → context.mach) removes the definition/reference divergence risk permanently.

### On the record (no action)
- The retained per-operand classification is now **scoped to genuinely signature-less positions** (variadic tails, calls through untyped function pointers) — it is no longer a fallback any direct call can reach. Deleting it outright would break those positions; retention is correct.
- **Typed function pointers do not exist in today's IR** (function pointers are bare `ptr`), so ALL indirect calls are outside the guard. When function-pointer types gain IRT_FN (adjacent: #2064, the #2058 ext-marshaling ruling), `VC_CALL_ARG_TYPE` extends to them for free via `aux_ty`. Recorded here so it's a known boundary, not a discovery.
- Verification bar met: 644/644 (dev 643 + the new colocated verify test) at -O0/-O1/-O2, fixpoint g2==g3 (g1≠g2 expected — the fix corrects the compiler's own instance ABI), nested-generic ASM proof, brokerd + mqtt 46/46 at -O2 (the #2035 field case).

PR #2083, opened 2026-07-11
fix(codegen): aarch64 vector `:~` bitcast selects FMOV, not a GP move (#2082)

### LINKED ISSUE #2082: codegen: aarch64 elides vector bit-reinterpret (:~) — result reads a never-populated register
Relates to #2015 (aarch64 NEON isel). Found by #2021 (blocks briar-systems/mach-std#372's aarch64 leg — 9 test failures).

On aarch64, a vector bit-reinterpret (`f32x4 :~ u32x4`) is elided as a no-op, but the
result vreg is assigned a different NEON register with **no copy inserted** — the
consumer reads a never-populated register. Lane reads return garbage; combined with a
following vector bitwise op it SIGSEGVs (exit 139). x86_64 lowers the same program
correctly. Shipped in v3.4.0 through v3.4.4.

This blocks the load-bearing primitive of the tier-3 stdlib (#2021): float and signed
select/blend must move vectors through the unsigned mask domain via `:~`
(`doc/language/operators.md`; `::` is numeric conversion, `:~` is the bitcast).

## Repro

```mach
pub fun bc(f: f32x4) u32 {
    val u: u32x4 = f:~u32x4;
    ret u[0];
}
```

Build for aarch64-linux (v3.4.4 seed or dev HEAD) and inspect the emitted asm.

## Expected

`u` is bit-identical to `f`; `u[0]` returns f's lane-0 bit pattern. A same-bank vector
bitcast may be a rename, but only if the vreg mapping is unified or a copy is emitted.

## Actual

The argument arrives in `q0` (AAPCS64), but the body does `str q1, [slot]` — `q1` is
never populated — then loads lane 0 from the slot. The reinterpret was dropped while
its result vreg got a fresh NEON register with no copy.

Boundary (all verified):

- scalar `:~` (`f32 :~ u32`) is **correct** on aarch64;
- pure-unsigned vector ops with no reinterpret (u32x4 select, u8x16 saturating add)
  are **correct** on both targets;
- so the defect is specifically vector `:~`/`::` lowering on aarch64.

A second suspected manifestation (possibly the same root, to be confirmed by the fix):
a function calling two helpers that share a scalar-`:~` Newton rsqrt bit-seed (quat
`length` + `normalize`) returns a wrong normalized lane on aarch64 (0.191 vs 0.200)
while x86_64 is exact — a reduced repro is available from the #2021 worker and will be
attached below.

Prior art for the class — a "free-looking" move dropped instead of materialized:
riscv64 dropped a same-bank float bitcast (#1762), and regalloc's `rewrite_instr`
dropped `vec_lane` (fixed in #2032). Suspect: NEON isel treats the same-bank vector
bitcast as a pure rename without unifying the vregs.

## Environment

mach 3.4.4 (also dev HEAD), linux-x86_64 host, aarch64-linux target (qemu and native
CI legs), also observed under the 3.4.4 seed by mach-std's test suite.

octalide:
## Summary

Fixes the shipped aarch64 miscompile in #2082: a vector bit-reinterpret
(`f32x4 :~ u32x4`) was elided and its result vreg assigned a fresh NEON register
with no copy, so consumers read a never-populated register (wrong lane reads; SIGSEGV
with a following vector bitwise op). Shipped v3.4.0–v3.4.4. x86_64 was correct.

## Root cause

A same-bank vector `:~` has **both** operands in the NEON/vector bank. The arm64 isel
`MIR_BITCAST` handler selected `FMOV` only when the banks *differ*
(`operand_is_fp(dst) != operand_is_fp(src)`), so a vector bitcast stayed `MIR_BITCAST`.
The encoder lowers a surviving `MIR_BITCAST` through `encode_convert` as a **GP** `ORR`
of the register *index* (`req_gpr` ignores class), writing a general-purpose register
and leaving the 128-bit vector destination unwritten.

The `MIR_MOV` handler already uses the correct test (`||`: either operand FP-bank →
FMOV). This aligns `MIR_BITCAST` with it. A same-bank vector bitcast then routes to
`encode_fmov`'s existing width-16 path (`MOV Vd.16b, Vn.16b`), and regalloc coalesces
it away when the vregs don't interfere. GP↔GP reinterprets still stay `MIR_BITCAST`;
cross-bank scalar `:~` is unchanged.

## Before / after (`f32x4 :~ u32x4`, arg in `q0`)

```
BEFORE (buggy)                          AFTER (fixed)
  sub  x0, x29, #0x14                     sub  x0, x29, #0x14
  mov  x1, x0        ; GP move (wrong)    str  q0, [x0]      ; stores the argument
  str  q1, [x0]      ; q1 never written   ldr  w0, [x1]
  ldr  w0, [x1]                           ...
```

The FMOV coalesced away entirely, so the argument `q0` is stored directly.

## Tests

- `arm64.isel:bitcast_bank_selects_fmov` — unit guard: vector/cross-bank `:~` → FMOV, GP↔GP stays `MIR_BITCAST`.
- Five `:~` round-trip blocks in the #2038 runtime vector suite (`vector_runtime.mach`): f32x4/f64x2 bit round-trips, float select/blend via the u32 mask domain, fabs via sign-bit clear, signed/unsigned int reinterpret. Discriminating bit patterns (sign-straddling floats, colliding low bytes, following vector bitwise op).

## Verification (built from this branch, verified with that binary)

- All 5 reinterpret tests **fail on the buggy seed** under qemu-aarch64 (one via signal 11), **pass with the fix**.
- Full vector suite (41) green on x86_64 and aarch64 (qemu) legs.
- Full `mach test` green at `-O0/-O1/-O2` (649 tests).
- Self-host fixpoint: stage2 == stage3, byte-identical.

Unblocks briar-systems/mach-std#372's aarch64 leg.

Closes #2082

### DIFF
--- a/src/lang/be/codegen/vector_runtime.mach
+++ b/src/lang/be/codegen/vector_runtime.mach
@@ -558,3 +558,73 @@ test "mach.lang.be.codegen.vector_runtime.spill:chain_deep" {
     if (acc[3] != -2147483549) { ret 2; }
     ret 0;
 }
+
+# bit reinterpret (`:~`) round-trips through the unsigned mask domain: the tier-3
+# stdlib's load-bearing float select/blend primitive (`doc/language/operators.md`;
+# `::` is numeric conversion, `:~` is the same-bit-pattern reinterpret). a same-bank
+# vector `:~` is a register bit copy the aarch64 NEON isel once dropped, assigning the
+# result a fresh vector register with no copy so a consumer read a never-populated one
+# (#2082, wrong lane / SIGSEGV). DISCRIMINATING bit patterns: sign-straddling floats
+# (negatives, -0.0), colliding low bytes with differing high bytes, and a following
+# vector bitwise op (the crash trigger).
+
+# f32x4 <-> u32x4 exact round-trip: each lane's IEEE-754 bits survive both directions.
+# 1.0 = 0x3F800000; -2.0 = 0xC0000000 (sign bit set); 3.5 = 0x40600000; -8.0 = 0xC1000000.
+test "mach.lang.be.codegen.vector_runtime.reinterpret:f32x4_u32x4_roundtrip" {
+    var f: f32x4 = f32x4{ 1.0, -2.0, 3.5, -8.0 };
+    var u: u32x4 = f:~u32x4;
+    if (u[0] != 0x3F800000 || u[1] != 0xC0000000 || u[2] != 0x40600000 || u[3] != 0xC1000000) { ret 1; }
+    var g: f32x4 = u:~f32x4;
+    if (g[0] != 1.0 || g[1] != -2.0 || g[2] != 3.5 || g[3] != -8.0) { ret 2; }
+    ret 0;
+}
+
+# f64x2 <-> u64x2 exact round-trip. 1.5 = 0x3FF8000000000000; -0.5 = 0xBFE0000000000000.
+test "mach.lang.be.codegen.vector_runtime.reinterpret:f64x2_u64x2_roundtrip" {
+    var f: f64x2 = f64x2{ 1.5, -0.5 };
+    var u: u64x2 = f:~u64x2;
+    if (u[0] != 0x3FF8000000000000 || u[1] != 0xBFE0000000000000) { ret 1; }
+    var g: f64x2 = u:~f64x2;
+    if (g[0] != 1.5 || g[1] != -0.5) { ret 2; }
+    ret 0;
+}
+
+# float select/blend entirely in the u32 bit domain, then reinterpret back: `:~` into
+# the mask domain, a vector bitwise blend (& | ~), `:~` out. a dropped `:~` on any of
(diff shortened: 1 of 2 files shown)


octalide:
## Coordinator review

**Verdict: approve — merging.** The minimal correct fix at the exact decision point: MIR_BITCAST selects FMOV whenever either operand is float/vector-bank (`||`, mirroring the MIR_MOV bank test), so a same-bank vector `:~` reaches encode_fmov's width-16 path instead of the GP ORR that wrote the aliasing register index. The case analysis in the comment is complete (scalar float↔float impossible by width, 128-bit GP nonexistent, GP↔GP correctly untouched), the isel guard covers all three bank combinations, and the five runtime blocks use the discriminating patterns (sign-straddling floats, -0.0, colliding low bytes, the SIGSEGV blend shape) that fail on the buggy seed. Verified rebased over #2079: disasm before/after, 41/41 vector suite both legs, 652/652, fixpoint, CI 14/14.

PR #2084, opened 2026-07-11
fix(scalarize): sign-extend narrow signed vector ordering compares (rv64)

### LINKED ISSUE #2077: riscv64 scalarization: signed sub-word (i8/i16) vector ordering compares read the lane unsigned
Part of #1965. Found by the #2075 int case (PR #2076) on its first run on the linux-riscv64 leg.

## Symptom

A signed **ordering** comparison (`< <= > >=`) of an `i8x16` or `i16x8` vector produces the wrong mask on any lane whose signed value is negative. The negative narrow lane is compared as its **unsigned** bit pattern.

For `i16x8{266,-1,5,5,...} < i16x8{10,0,5,6,...}`, lane 1 is `-1 < 0`:
- correct (x86_64 SSE2 / aarch64 NEON, and the #2038 native suite): `true` → all-ones mask lane
- rv64 scalarization: `false` (`0xFFFF = 65535 < 0` is false unsigned)

Same on `i8x16` lane `-1 vs 0` for `<= > >=`. Concretely the int case shows (golden = correct, rv64 = observed):
```
cmp.i16.lt  0 65535 0 65535   ->  0 0 0 65535
cmp.i16.le  0 65535 65535 65535 -> 0 0 65535 65535
cmp.i16.gt  65535 0 0 0       ->  65535 65535 0 0
cmp.i16.ge  65535 0 65535 0   ->  65535 65535 65535 0
cmp.i8.lt   0 255 255 0       ->  0 255 0 0
cmp.i8.le   255 255 255 255   ->  255 255 0 255
cmp.i8.gt   0 0 0 0           ->  0 0 255 0
cmp.i8.ge   255 0 0 255       ->  255 0 255 255
```

## Scope (what is NOT affected)

- Equality (`== !=`) on every shape — bit-equality doesn't depend on sign extension. Correct.
- **Unsigned** ordering (u8/u16/... ) — correct.
- Signed ordering on **i32x4 / i64x2** — correct (register-width lanes are canonically sign-extended on rv64).
- All arithmetic (`+ - *`), bitwise, lane r/w, ABI, spill — correct across every shape.
- **Pure-scalar** narrow signed compares (`fun f(x: i16) { x < 0 }`) — correct on rv64. So this is **scalarize-specific**, not a general backend defect.

## Root cause (analysis, for the fix owner to confirm)

`me/pass/scalarize.mach` `expand_compare` loads each lane at the vector element type (`emit_load`, `op.elem` = i8/i16) and emits a scalar `OP_CMP_LT_S`/`LE_S` directly on those loaded values. On rv64 the compare is a 64-bit `slt`, so both operands must be sign-extended to the register width for a signed compare. A sub-word load does not leave the value sign-extended into bits 63..w for the signed compare (i32/i64 escape because `lw`/register form is already canonical). The scalar path works because the i16 value there flows through the backend's normal canonicalization; the pass's freshly-emitted load + signed compare does not get the same extension.

Likely fix: in `expand_compare`, for signed ordering kinds (`CMP_LT_S`/`CMP_LE_S`, and the `> >=` forms that lower to swapped `LT/LE`), sign-extend the loaded narrow lane operands to a canonical width before the compare (and zero-extend for unsigned), rather than comparing the raw sub-word load. Alternatively fix at the rv64 isel so a signed compare re-extends sub-word operands per its signedness. The fix owner should pick the layer; the scalarize-local extension keeps the fault contained to the pass that introduces the pattern.

## Acceptance

- The #2075 int case (`int/surface/vectors`, PR #2076) passes on the linux-riscv64 leg (it is the regression test; it currently reproduces this exactly, at both -O0 and -O2).
- x86_64 / aarch64 unaffected; no change to the compiler binary's self-host byte-identity.

octalide:
Closes #2077.

## Problem

On a scalarized target (rv64gc, `has_v128 == false`), `me/pass/scalarize.mach`
`expand_compare` loaded each vector lane with the width-driven integer loads and
fed it straight to a scalar signed compare. rv64's sub-word loads (`lbu` / `lhu`)
**zero-extend**, so a negative `i8` / `i16` lane arrived in the register as its
large unsigned bit pattern and the 64-bit signed `slt` read `-1 < 0` as false.
Only signed **ordering** (`< <= > >=`) of `i8x16` / `i16x8` was affected;
equality, unsigned relations, and `i32` / `i64` lanes were already correct
(`lw` / `ld` are sign-canonical, and unsigned relations want the zero-extended
form the loads already give).

The pure-scalar case (`fun f(x: i16) { x < 0 }`) is correct because the rv64 ABI
sign-extends narrow integer *arguments* into the register — a vector lane is
*loaded* from a stack slot and gets no such extension, so the fix belongs in the
pass that emits the load + compare.

## Fix

In `expand_compare`, sign-extend both operands of a signed sub-word ordering
compare (`CMP_LT_S` / `CMP_LE_S`, `op.bits < 32`) to a 32-bit register-canonical
width via `OP_SEXT` before the scalar compare (the shift-pair fills the full
64-bit register). Everything else is unchanged. The fix stays inside the pass's
defined-expansion model — it scalarizes the operator, adding no algorithm.

## Test

A colocated unit guard (`expand_compare:sign_extends_narrow_signed_lanes`) runs
the pass over synthetic compares and counts emitted `OP_SEXT` — the pass's only
source of that opcode — asserting two per lane for signed sub-word ordering
(`i16x8 LT_S` → 16, `i8x16 LE_S` → 32) and none for equality, unsigned, or
register-width (`i32x4`) lanes. Reverting the fix fails it.

## Verification

- **`mach test`**: 644 passed, 0 failed, via a from-source compiler built on this branch.
- **int/ vectors case (`feat/2075` PR #2076) on `linux-riscv64` under qemu:**
  - **before (dev):** FAIL — the 8 signed-narrow lines (`cmp.i8`/`cmp.i16` `lt`/`le`/`gt`/`ge`) mismatch at both `-O0` and `-O2`.
  - **after (this branch):** PASS at both `-O0` and `-O2`; every other line byte-identical.
- **Capable targets untouched:** the dev compiler and this branch's compiler emit **byte-identical** x86_64 output for the same source (identical SHA-256) — scalarize is a no-op on `has_v128` targets.

Rides a v3.4.5 patch release; unblocks PR #2076 (rebases once this is on `dev`).

### DIFF
--- a/src/lang/me/pass/scalarize.mach
+++ b/src/lang/me/pass/scalarize.mach
@@ -400,6 +400,20 @@ fun expand_compare(c: *Ctx, blk: *ir.Block, kind: instruction.InstrKind, dst: va
     if (R.is_err[ir_type.IrTypeId, str](ram)) { ret R.err[bool, str](R.unwrap_err[ir_type.IrTypeId, str](ram)); }
     val arr_mask: ir_type.IrTypeId = R.unwrap_ok[ir_type.IrTypeId, str](ram);
 
+    # a signed ordering compare of a sub-word lane must sign-extend both operands to a
+    # register-canonical width before the scalar compare: rv64's sub-word integer loads
+    # zero-extend (lbu / lhu), so a negative i8 / i16 lane would otherwise be read as its
+    # large unsigned bit pattern and `-1 < 0` would evaluate false (#2077). i32 / i64
+    # lanes load sign-canonical (lw / ld), and equality / unsigned relations read the
+    # zero-extended lane correctly, so neither needs the extension.
+    val need_sext: bool = (kind == instruction.OP_CMP_LT_S || kind == instruction.OP_CMP_LE_S) && op.bits < 32;
+    var ext_ty: ir_type.IrTypeId = ir_type.IRT_NIL;
+    if (need_sext) {
+        val rx: R.Result[ir_type.IrTypeId, str] = ir_type.intern_int(?c.m.types, 32);
+        if (R.is_err[ir_type.IrTypeId, str](rx)) { ret R.err[bool, str](R.unwrap_err[ir_type.IrTypeId, str](rx)); }
+        ext_ty = R.unwrap_ok[ir_type.IrTypeId, str](rx);
+    }
+
     var k: u32 = 0;
     for (k < op.lanes) {
         val pa: R.Result[value.Value, str] = lane_ptr(c, blk, a, arr_op, k, loc);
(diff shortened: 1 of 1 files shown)


octalide:
## Coordinator review

**Verdict: approve — merging.** The fix is exactly the issue's root cause and stays inside the expansion model: OP_SEXT both operands of a signed sub-word ordering compare to the 32-bit register-canonical width (rv64 lbu/lhu zero-extend; lw/ld are sign-canonical, so i32/i64 rightly untouched; equality/unsigned rightly untouched; GT/GE covered via LT/LE canonicalization, proven by the int case's all-four-relations pass). The OP_SEXT-count unit guard discriminates precisely (16/32 for narrow signed ordering, 0 for eq/unsigned/i32), and the sentinel-on-setup-failure detail is the right paranoia. Verification bar met: int case before/after at -O0 and -O2, x86_64 byte-identical (scalarize no-op on has_v128), 644/644, CI 14/14.

Issue #2055, opened 2026-07-11
codegen: win64 ABI never classifies 128-bit vectors (mis-passed as CLASS_GP); by-value vector args/returns miscompile on windows


octalide:
Part of #1965. Distinct defect class from #2053 (which is the SysV/NEON capture-width regression): the win64 vector ABI was **never implemented**. Surfaced by the colocated ABI round-trip test added in PR #2054 (gated off windows there so the #2053 critical patch can ship; ungate when this lands).

## Bug

`src/lang/target/abi/win64.mach` is the **only** ABI classifier without an `if (is_vector)` branch — `sysv.mach`, `aapcs64.mach`, and `lp64.mach` each have one. A 128-bit vector has `is_aggregate == false` and `is_float == false`, so in win64's arg and return classification it falls through to `CLASS_GP` and is mis-passed in a single 8-byte GP register (or `CLASS_STACK`). x64's `has_v128` is `true` and is keyed on the ISA, so vectors are **native** on windows-x86_64 — the ABI simply never handled them.

Result: any function that passes or returns a 128-bit vector by value on windows-x86_64 miscompiles. This shipped broken in v3.4.0 **and** v3.4.1 (the #2053 patch fixes SysV/NEON only). It is win64-CI-provable only (no local windows execution here).

## Correct convention (Microsoft x64 — get exact, do not mirror-image the two sides)

- **Arguments**: `__m128` / `__m128i` / `__m128d` (any 16-byte vector) are **never passed by immediate value** — the caller allocates a copy and passes a **pointer** to it. This is by-reference: the `CLASS_BYREF` path win64 already uses for a >8-byte aggregate (the `is_by_value_size(size)` == false branch → `CLASS_BYREF` / `CLASS_STACK_BYREF`).
- **Return**: `__m128` / `__m128i` / `__m128d` are **returned in XMM0** — NOT by-reference. So the return side is `CLASS_FP` (XMM0), asymmetric with the arg side. Do not apply the by-ref arg rule to the return.

## Fix direction

Add an `if (is_vector)` branch to win64's arg classifier (route to the by-ref `CLASS_BYREF` / `CLASS_STACK_BYREF` path, as a >8-byte aggregate) and to its return classifier (route to `CLASS_FP` in XMM0). Verify on the windows-x86_64 CI runner with a by-value vector round-trip (arg + return, discriminating lanes).

## Coordination

[...]

Issue #1852, opened 2026-07-03
infra: riscv64 self-host smoke — the compiler never executes on riscv64 in CI


octalide:
Depends on #1851. Closes the coverage gap surfaced by #1742.

## What

CI was fully green while the riscv64-hosted compiler could not compile a two-block function (#1851). The gap: the riscv64 unit lane runs the test suite under qemu but exercises regalloc only at helper granularity, and the int/ riscv64 leg runs *produced* binaries under qemu — the compiler itself never executes on riscv64 anywhere in CI.

Add a riscv64 self-host smoke: run the cross-built riscv64 compiler under qemu-riscv64 against a small multi-block input and assert it produces a working artifact. The int/ harness is the natural home (targets.conf stays the source of truth; placement per int/ conventions), but this is a new kind of case — it executes the compiler under emulation rather than a produced binary — so the exact shape needs octalide's sign-off before implementation.

## Acceptance

- A CI-visible check executes the riscv64-hosted compiler (not merely its outputs) under qemu on at least one input with cross-block liveness.
- The check fails on dev @ 96358fe8 (pre-#1851) and passes once #1851 lands.
- No new root-level test/ directory; lives within the existing int/ harness and lanes.

octalide:
Unblocked: #1851 is fixed and riscv64 self-host is verified (see #1742). Shape still needs octalide's sign-off — this would be a new kind of int/ case (executing the compiler under qemu rather than a produced binary).

octalide:
Shape approved as proposed (octalide, 2026-07-03): an int/-hosted case that executes the cross-built riscv64 compiler under qemu-riscv64 against a small multi-block input and asserts it produces a working artifact. Dispatching.

PR #2087, opened 2026-07-11
feat(abi): win64 marshals ext-boundary vector args by reference (#2058)

### LINKED ISSUE #2058: rfc: win64 vector ABI conformance — should ext-boundary calls marshal __m128 by reference?

### LINKED ISSUE #2086: verify: always-on VC_CALL_ARG_TYPE rejects scalarized (rv64) vector calls — vector param vs by-address ptr arg
Regression on dev. Found by the #2075 int case (PR #2076) after rebasing onto current dev; it is not caused by that case or by the #2077 fix.

## Symptom

On rv64 (the scalarized target), building any program that makes a **vector-typed call which survives inlining** fails:
```
error: call argument typing: a call argument's type disagrees with the callee's declared parameter
```
Minimal repro (fails at rv64 `release` because it resists inlining; a small callee fails only at `debug` and inlines away at `-O2`):
```mach
fun accum(a: i32x4, n: i64) i32x4 {
    if (n <= 0) { ret a; }
    var one: i32x4 = i32x4{ 1, 1, 1, 1 };
    ret accum(a + one, n - 1);
}
# main calls accum(a, 5) -> build error on --target linux-riscv64
```
x86_64 / aarch64 (capable targets, no scalarization) are unaffected: the vector argument stays vector-typed and matches the callee signature.

## Root cause

Bisected to `b734ac5d feat(verify): always-on VC_CALL_ARG_TYPE tripwire for call/callee ABI mismatch` (the promotion of the aggregate call-arg check to always-on).

The scalarize pass (`me/pass/scalarize.mach`, #2016) lowers a vector on rv64 to its memory (by-address) class: it retypes vector **call arguments** (and body params) to `ptr`, while **deliberately keeping the callee's IRT_FN signature params vector-typed** for ABI classification (documented in `run()`, ~line 754). The new always-on check (`check_call_arg_types` → `call_arg_type_ok`, `verify.mach`) compares each call argument type `q` against the callee signature's declared parameter type `p`:
```
fun call_arg_type_ok(m, p, q) bool {
    if (p == q)                         { ret true; }
    if (p == IRT_NIL || q == IRT_NIL)   { ret true; }
    if (is_aggregate(p))                { ret is_aggregate(q); }   # vector p falls through or requires aggregate q
    if (is_pointerish(p) && is_pointerish(q)) { ret true; }         # vector p is not pointerish
    ret false;                                                      # -> violation
}
```
For a scalarized vector call, `p` = vector (from the signature), `q` = `ptr` (scalarize's by-address arg). A vector `p` is not equal to `ptr`, is not pointerish, and (if it isn't classed aggregate) falls through to `ret false`; if it *is* classed aggregate, it demands an aggregate `q` and rejects the `ptr`. Either way → VC_CALL_ARG_TYPE violation on legal IR.

The tripwire simply doesn't model the scalarize representation: on a SIMD-incapable target a vector parameter flows by address exactly like an aggregate, so a `ptr` (or aggregate-by-address) argument against a vector-typed signature param is correct.

## Why it shipped undetected

Nothing on dev exercised an rv64 vector call through a full build until this int case: `mach test` can't run under qemu-user, and the #2038 runtime suite runs only on x86_64/aarch64 (capable targets, no scalarization). The #2075 int case is the first end-to-end rv64 vector-call coverage.

## Fix direction (fix owner's call)

In `call_arg_type_ok`, treat a **vector** declared parameter like an aggregate one: accept an argument that flows by address (pointerish / aggregate-by-address), since that is the scalarized representation. Mirror the existing aggregate branch (line 706) for `type.is_vector(p)`. Capable targets are unaffected (there `p == q` vector). Alternatively, exempt the check when the target scalarizes — but the verifier has no target, so the type-level allowance is cleaner.

## Acceptance

- rv64 builds of vector-calling programs succeed at `-O0` and `-O2` (recursive/non-inlined included).
- The #2075 int case (PR #2076) builds and passes on the linux-riscv64 leg at both profiles.
- x86_64 / aarch64 unaffected; verifier still catches the genuine #2035-class aggregate/pointer mis-typing it was added for.

octalide:
## #2058 — win64 vector-argument marshaling at `ext` boundaries only

Implements the owner ruling (2026-07-11) on RFC #2058: on `x86_64-windows`, a call to a declared `ext` function passes each 128-bit vector argument per the Microsoft x64 convention — the caller stores the vector to a 16-byte-aligned temporary and passes its address in the parameter's GP register / stack slot. Internal (mach→mach) calls keep mach's convention (vectors in XMM, #2055), byte-identical on every target. Vector **returns** already conform (XMM0 under both conventions) and are untouched.

`Closes #2058`

### Design

- **Classification signal is the declared `ext` fact, not `FN_FLAG_EXTERN`.** A new `ir.Function.is_import` (the exact function twin of `Global.is_import` from #2079) is a strict subset of `FN_FLAG_EXTERN`: true only for a declared `ext fun`. A same-artifact cross-module reference — a generic instance drained in another object, a forward reference — is `FN_FLAG_EXTERN` but **not** `is_import`, so it never marshals. Derived at every function reference from the origin decl's `DECL_FLAG_EXT`, mirroring `references_import_data`.
- **One classification rule in the win64 module.** An `is_ext` boundary fact is threaded through the shared `ArgPassingFn` (the same neutral-substrate pattern `is_vector` used); only win64's `classify_arg` consumes it. `ext`-ness is **not** baked into `IRT_FN` type identity — the callee's `is_import` is read at the call site.
- **Distinct placement class, so rv64 is untouched.** A win64 ext vector is `CLASS_VEC_BYREF` / `CLASS_VEC_STACK_BYREF` — a *register-resident value* passed by reference, so the caller must spill it to an aligned temp before its address exists. This is deliberately distinct from `CLASS_BYREF`: a scalarized rv64 (lp64) vector is already memory-resident (its operand IS a pointer) and stays `CLASS_BYREF`. Reusing `CLASS_BYREF` would have corrupted rv64's existing vector lowering.
- The variadic tail follows the same rule (MS passes `__m128` varargs by reference too). `lower_params`/`lower_ret` pass `is_ext=false`: a compiled mach function is never itself `ext`.

### Verification (from-source compiler on this branch, never the PATH seed)

_Rebased onto current `dev` (v3.5.0, carrying #2083/#2084/#2088/#2089) and re-verified end to end. The rebase composes cleanly with #2089's new `AbiVTable.fp_callee_saved_full_vector` — an orthogonal callee-saved-width field on the vtable rec / registration sites, distinct from this PR's `ArgPassingFn` threading._

- **Full `mach test`:** 656/656 pass (adds a win64 classifier test for the register, later-slot, and exhausted cases, plus a vector-return XMM0 conformance test).
- **`mach test --target windows-x86_64` under wine:** 656/656 pass (includes the vector-runtime lane-exactness tests executed under wine).
- **Structural conformance (disasm of an `ext` vector call, 5 vector args).** The internal call keeps XMM; the ext call marshals each vector by reference per MS x64:
  ```
  movups XMMWORD PTR [rbp-0x40], xmm0   ; store vector to 16-byte-aligned temp
  lea    rcx, [rbp-0x40]                ; arg0 address -> RCX
  movups XMMWORD PTR [rbp-0x60], xmm6
  lea    rdx, [rbp-0x60]                ; arg1 -> RDX
  ... -> R8, R9 ...
  movups XMMWORD PTR [rbp-0xc0], xmm0
  lea    rbx, [rbp-0xc0]
  mov    QWORD PTR [rsp+0x20], rbx      ; arg4 address spilled to stack (above shadow space)
  call   c_sink
  ```
  Temps are 16-byte aligned (rbp is 16-aligned on entry; offsets are multiples of 16); stores are genuine 128-bit `movups`. The internal `local_use` call uses `movaps xmm0/xmm1` (mach convention, unchanged).
- **Byte-identity (base compiler at the merge-base vs this branch, same fixed corpus):** object trees **byte-identical on all four targets** — `linux-x86_64`, `linux-arm64`, `linux-riscv64`, and `windows-x86_64` (a corpus with no ext-vector calls). A **vector-bearing** program (`vecext`, internal + ext vector calls) is likewise byte-identical on `linux-x86_64`, `linux-arm64`, and `linux-riscv64` — the direct proof that the distinct `CLASS_VEC_*` class leaves rv64's memory-resident vector lowering untouched.

### Honest limits

- **No true MSVC-compiled callee tested** (no MSVC toolchain locally). A *self-consistent* wine round-trip is inherently infeasible in mach's type system: the caller must see a vector signature (to trigger marshaling) while the callee reads a pointer, and aliasing one symbol to both signatures is rejected by the always-on call-arg-type verifier. The structural disasm above is the conformance proof.
- **riscv64 (lp64):** untouched *by construction* (it produces `CLASS_BYREF`/`CLASS_STACK_BYREF`; this PR's new branch keys only on the win64-only `CLASS_VEC_*` classes, and its classifier ignores `is_ext`) and **machine-confirmed**. Cross-building the compiler to riscv64 is a first-class path (rv64 self-host #1742, CI-proven under qemu #1852). On the rebased head the from-source x86_64 compiler cross-builds the **compiler itself for `linux-riscv64` (163 objects)** and the object tree is byte-identical to the base — as is the vector-bearing `vecext` program. (A cross-build against my *original* merge-base failed with an always-on call-arg verifier error; that was the since-fixed #2086 false-positive on scalarized vector call args, which predates #2088 — not this change and not an unsupported path.)
- Function *pointers* are bare `ptr` in today's IR, so an ext-typed function pointer cannot carry the rule (the recorded #2058 boundary). Direct `ext` calls — the actual C-interop surface — are covered.
- **Sequencing:** #2064 (dual same-name instance Function entries) has not yet merged to `dev`; this branch composes with its future in-place extern reuse (the `is_import` assignment at the `ext fun` decl is authoritative and independent of how `append_function` allocates the entry).

### DIFF
--- a/doc/language/ext-fun.md
+++ b/doc/language/ext-fun.md
@@ -77,6 +77,24 @@ not bound to a single dependency, so `library` has no effect on the emitted
 binary — the value is still validated against the link's dependencies, but every
 dependency is searched at load time regardless.
 
+## Vector arguments and the C ABI
+
+A 128-bit vector (`f32x4`, `i32x4`, …) crossing an `ext` boundary follows the
+target's **C** vector convention, not Mach's internal one, so the call is
+bit-compatible with a C `__m128` parameter:
+
+- **x86_64-windows (Microsoft x64):** the caller passes each vector argument **by
+  reference** — it stores the vector to a 16-byte-aligned temporary and passes that
+  temporary's address in the parameter's integer register (RCX/RDX/R8/R9) or, once
+  those are exhausted, on the stack. A variadic vector argument is passed the same
+  way. Vector **returns** ride XMM0, which both conventions already agree on.
+- **Every other target** (System V, AAPCS64, RISC-V lp64d): the C convention already
+  matches Mach's internal one, so nothing special happens at the boundary.
+
+Only a *direct* call to a declared `ext fun` marshals — an ordinary Mach→Mach call
+always keeps the internal convention (a vector in a vector register). A call through
+a function *pointer* does not yet carry the `ext` fact, so it is not covered.
+
 ## Linking external objects
 
 An `ext fun` is only a forward reference — its definition must be supplied at
--- a/src/lang/be/codegen/mir/abi.mach
+++ b/src/lang/be/codegen/mir/abi.mach
@@ -494,13 +494,28 @@ fun round_up_i64(value: i64, align: u64) i64 {
 # <=8-aligned, so this is always eight and the round-up is identity; AAPCS64 honors
 # a 16-byte-aligned by-value argument.
 # ---
-# class: the slot's classification (CLASS_STACK or CLASS_STACK_BYREF)
+# class: the slot's classification (CLASS_STACK, CLASS_STACK_BYREF, or CLASS_VEC_STACK_BYREF)
 # align: the value's natural alignment in bytes
 # ret:   the slot's stack alignment in bytes
 fun stack_arg_align(class: abi.ParamClass, align: u64) u64 {
-    if (class == abi.CLASS_STACK_BYREF) { ret 8; }
-    if (align < 8)                      { ret 8; }
+    # both by-reference stack spills carry only an 8-byte pointer, so they round to 8
+    # regardless of the referent's alignment (CLASS_VEC_STACK_BYREF's referent is the
+    # spilled vector temp, aligned at its own frame slot, not this outgoing pointer slot).
+    if (class == abi.CLASS_STACK_BYREF)     { ret 8; }
+    if (class == abi.CLASS_VEC_STACK_BYREF) { ret 8; }
+    if (align < 8)                          { ret 8; }
     ret align;
+}
+
+# whether a placement occupies an outgoing stack slot the signature walk assigns and
+# advances a stack offset for (a by-value stack argument or either by-reference stack
+# spill). the register and register-plus-stack-split placements are handled off the
+# register-cursor path instead
+# ---
+# class: the slot's classification
+# ret:   true when the slot is a pure stack placement
+fun is_stack_slot_class(class: abi.ParamClass) bool {
+    ret class == abi.CLASS_STACK || class == abi.CLASS_STACK_BYREF || class == abi.CLASS_VEC_STACK_BYREF;
 }
 
 # classify a whole function ABI in one left-to-right walk into a `SigLayout`
@@ -533,10 +548,15 @@ fun stack_arg_align(class: abi.ParamClass, align: u64) u64 {
 # ctx:      the active lowering context (supplies the IR type oracle and target)
 # sig:      the function signature (an IRT_FN id; non-IRT_FN yields zero fixed params)
 # ret_type: the declared return type to classify the return / sret reservation from
+# is_ext:   true when classifying a call to a declared `ext` (foreign C) function, so
+#           each argument follows the foreign convention at the boundary (win64
+#           marshals a 128-bit vector by reference, #2058); false for an internal call
+#           and for a callee's own parameter / return classification (a compiled mach
+#           function is never itself `ext`)
 # out:      receives the assembled SigLayout (`params` is owned - free it with
 #           `sig_layout_free`)
 # ret:      ok(true) on success, or a loud error from the ABI vtable surface
-pub fun classify_signature(ctx: *context.LowerCtx, sig: type.IrTypeId, ret_type: type.IrTypeId,
+pub fun classify_signature(ctx: *context.LowerCtx, sig: type.IrTypeId, ret_type: type.IrTypeId, is_ext: bool,
                            out: *abi.SigLayout) R.Result[bool, str] {
     val rok: R.Result[bool, str] = require_abi(ctx.tgt);
     if (R.is_err[bool, str](rok)) { ret rok; }
@@ -620,7 +640,7 @@ pub fun classify_signature(ctx: *context.LowerCtx, sig: type.IrTypeId, ret_type:
         var pagg: abi.AggLayout = abi.agg_none();
         if (wants_agg) { pagg = agg_layout(ctx, pty, pag); }
         val pvec: bool = context.value_is_vector(ctx, pty);
-        var slot: abi.ParamSlot = classify_arg(i::i32, psz, pal, pfl, peb, pag, pvec, gp_used, fp_used, phm, phe, pagg);
+        var slot: abi.ParamSlot = classify_arg(i::i32, psz, pal, pfl, peb, pag, pvec, is_ext, gp_used, fp_used, phm, phe, pagg);
 
         # record the real outgoing stack offset on the slot (the ABI classifier
         # returns a sizing-only slot with offset 0); the param / call lowering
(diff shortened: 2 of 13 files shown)


octalide:
## Coordinator review

**Verdict: approve. Merge is HELD briefly for release sequencing** — the #2092 shipped-critical fix cuts v3.5.1 as a pure patch first; this PR merges immediately after the tag (it's a behavior change at ext boundaries and belongs in the next minor, not an emergency patch).

The implementation is the ruling, exactly:
- `Function.is_import` is the strict-subset ext fact (twin of `Global.is_import`), set at decl lowering and derived at cross-module references from the origin decl's flag — a drained generic instance stays extern-but-mach and never marshals. Monotonic upgrade in `ensure_extern_function` mirrors the data path.
- `is_ext` threads through the shared ArgPassingFn — one added parameter, every non-win64 ABI documents ignoring it; own-param/return classification and instance references explicitly pass false. The varargs tail honors it (MS passes `__m128` varargs by reference too).
- `CLASS_VEC_BYREF`/`CLASS_VEC_STACK_BYREF` correctly model the novel placement (register-resident value whose address must be materialized) as distinct from `CLASS_BYREF` (already memory-resident) — which is why rv64's scalarized vector path is byte-identical, machine-confirmed.
- `spill_vector_to_temp`: 16-aligned frame temp, full-width store, LEA — the MS shape, disasm-confirmed through the 5th-arg stack case above the shadow space.
- Docs updated with the honest fnptr boundary (no callee identity in today's IR; recorded, consistent with #2080's note).
- rv64 caveat fully resolved post-#2088: compiler cross-build succeeds, byte-identity on all four targets, and the merge composed cleanly with #2089's vtable field.

Honest limit accepted: no true MSVC-compiled callee locally; structural conformance (disasm vs the published convention) + wine execution is the agreed bar, recorded in the PR body.

PR #2088, opened 2026-07-11
fix(verify): accept a by-address argument for a vector parameter (rv64 #2086)

### LINKED ISSUE #2086: verify: always-on VC_CALL_ARG_TYPE rejects scalarized (rv64) vector calls — vector param vs by-address ptr arg

octalide:
Closes #2086. Urgent — hard-gates v3.4.5.

## Problem

The always-on `VC_CALL_ARG_TYPE` tripwire (b734ac5d, PR #2080) checks each call
argument's operand type against the callee's declared `IRT_FN` parameter type,
with no allowance for a **vector** parameter. On a scalarized target
(rv64, `has_v128 == false`, #2016) the pass retypes a vector call argument to a
plain `ptr` while **deliberately keeping the callee signature's parameter
vector-typed** for the lp64 BYREF/SRET ABI classification. So `call_arg_type_ok`
saw `p` = vector, `q` = `ptr`, matched no leniency branch (`is_aggregate(vector)`
is false, a vector is not pointerish), and fell through to `ret false` — a
`VC_CALL_ARG_TYPE` violation on **correct** IR. Every rv64 program with a vector
call that survives inlining failed to build, at both `-O0` and `-O2`. Capable
targets (x86_64 / aarch64) were unaffected: there the argument stays vector-typed
and `p == q`.

Regression on dev, bisected to b734ac5d; not caused by the #2075 int case or the
#2077 fix — that case is simply the first end-to-end rv64 vector-call coverage
(`mach test` can't run under qemu, and the #2038 runtime suite runs only on
capable targets).

## Fix

In `call_arg_type_ok`, add a vector-`p` branch before the aggregate branch: a
vector parameter accepts a by-address argument (pointerish or aggregate-by-address).
Design framing: a vector's passing model is target-dependent — register-class on a
capable target (caught by the `p == q` vector match), memory-class under
scalarization where the argument flows by address — and the verifier is
target-less, so it must accept the union. The relaxation is **by-address only**:
`is_address_typed` blesses `IRT_PTR` / aggregate but not a bare scalar, so the
#2035 aggregate discipline the tripwire was added for is untouched.

## Test

Colocated `check_call_arg_types:vector_param_accepts_by_address_arg`: a
`fn(i32x4) -> i64` callee with (1) a vector argument → no violation (capable
target), (2) the argument retyped to `ptr` → no violation (the fix; scalarized
rv64), (3) the argument retyped to a bare `i64` → violation still fires (bounded
relaxation). The existing `aggregate_arg_pointer_mistype` (#2035) test is unchanged
and still passes.

## Verification

- `mach test`: full suite green via a from-source compiler on this branch.
- int/ vectors case (`feat/2075`, PR #2076) on `linux-riscv64` under qemu: build-fails on the ABI block **before**; **after**, builds and passes at both `-O0` and `-O2`.
- Capable targets untouched: dev and branch compilers emit **byte-identical** x86_64 output for the same source (identical SHA-256).

### DIFF
--- a/src/lang/me/ir/verify.mach
+++ b/src/lang/me/ir/verify.mach
@@ -703,6 +703,12 @@ fun check_call_arg_types(m: *ir.Module, ins: *instruction.Instruction, fi: u32, 
 fun call_arg_type_ok(m: *ir.Module, p: type.IrTypeId, q: type.IrTypeId) bool {
     if (p == q)                                 { ret true; }
     if (p == type.IRT_NIL || q == type.IRT_NIL) { ret true; }
+    # a vector parameter's passing model is target-dependent: register-class on a
+    # capable target (the `p == q` vector match above) or memory-class under
+    # scalarization (#2016), where the argument is retyped to flow by address. the
+    # verifier is target-less, so accept the union - a by-address argument (pointerish
+    # or aggregate-by-address) against a vector parameter is the scalarized form (#2086).
+    if (type.is_vector(?m.types, p))            { ret is_pointerish(m, q) || type.is_aggregate(?m.types, q); }
     if (type.is_aggregate(?m.types, p))         { ret type.is_aggregate(?m.types, q); }
     if (is_pointerish(m, p) && is_pointerish(m, q)) { ret true; }
     ret false;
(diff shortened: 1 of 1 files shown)


octalide:
## Coordinator review

**Verdict: approve — merging.** One branch, exactly the bounded relaxation: a vector parameter accepts a by-address argument (the scalarized memory-class form, #2016) or the vector itself (the capable register-class form via `p == q`) — the union a target-less verifier must accept. The unit guard proves all three states: capable match, scalarized ptr accepted, bare scalar still rejected. The #2035 aggregate discipline is untouched (existing test passes). Verified before/after on the exact trigger (the #2076 ABI block: dev build-fails at -O0, branch green at both profiles), x86_64 byte-identical, 647/647, CI 14/14. Author-ack from the tripwire's author (sig-2063) invited post-merge; the relaxation's boundedness is proven in-diff so the release train doesn't wait.

Issue #2021, opened 2026-07-10
stdlib: simd + math library over the vector types (select/blend, shuffles, reductions, matrices)


octalide:
Part of #1965.
Depends on #2014, #2015, #2016.
Nice to have: #2019 (splat sugar) — ergonomic broadcast/constant construction for the library's builders, not a blocker.

## What

Tier-3 of the SIMD program: the **library over the compiler-known vector
types** — the long tail of SIMD operations plus a small math library
(matrices, quaternions) built on `f32x4`-class elements. Nothing here is a
compiler change; it is plain Mach with per-arch `asm` bodies, sitting on the
substrate the tier-1 gate and backends provide.

This is exactly the split `doc/language/policy.md` already draws. Its
`## Stdlib handles` bullet (lines 36-38) names "the long tail of SIMD ops —
shuffles, reductions, gather/scatter, saturating arithmetic, specialized
math. Functions over the compiler-known SIMD types," and the dividing rule
(lines 45-48): "Stdlib handles things that map 1:1 to specific instruction
sequences with predictable lowerings." The compiler seeds the vector *types*
and the core lane-wise *operators* (policy.md lines 14-17); everything below
is this child.

**Why a library, not language types.** The epic's depth-1 contract: there are
no `f32x4x4` language types — a matrix is an *algorithm* over vectors, and
integer matmul even needs widening accumulators (AMX/SME-class, firmly stdlib
`asm` territory). So `Mat4` / quaternions are records over vector elements
(the `[4]f32x4` shape), and they get clean `+ - *` on their element currency
*because* that element is a compiler type. This is the whole point of seeding
the vectors: the library is readable Mach, not intrinsics.

Design for this slice:

- **select / blend** — `(mask & a) | (~mask & b)`, pure lane-wise bitwise ops
  over the unsigned mask vectors the compiler produces from a comparison
  (`f32x4` cmp → `u32x4`, lanes all-ones/all-zeros, per the epic). The epic
  fixes this as "in the library first" — no compiler `select` op in this
  program. This is the load-bearing primitive the rest of the library builds
  on.
- **shuffles, reductions, gather/scatter, saturating arithmetic** — via
  per-arch `asm` bodies dispatched by `$if` / `$or` on `$project.target.arch`
  (the established stdlib idiom, `doc/language/policy.md` "See also" →
  `asm.md`, `comptime-control.md`). These are the ops with no 1:1 language
[...]

PR #2089, opened 2026-07-11
fix(codegen): aarch64 vector live across a call corrupted (#2085)

### LINKED ISSUE #2085: codegen: aarch64 float live across a call is corrupted (two-call-site scalar rsqrt path)
Distinct from #2082 / PR #2083 — confirmed empirically by running the repro against the #2083-fixed compiler (unchanged) and by inspection (the repro contains **no vector `:~`**, only scalar `f32:~u32` / `u32:~f32`; scalar `:~` codegen is byte-identical before/after #2083). Blocks briar-systems/mach-std#372's quaternion normalize/nlerp aarch64 leg (2 tests). Found by #2021, characterized jointly by the std-2021 and neonbc-2082 workers.

On aarch64, a float value is not preserved across an intervening call: when two helpers
that both route through the same scalar-`:~` Newton rsqrt are called from one function
(`z_length4` then `z_normalize4`), the SECOND call site returns a corrupted rsqrt.
x86_64 is exact. Shipped in the 3.4.4 seed (and likely earlier — the path is scalar
float codegen, not vectors).

## Repro

Self-contained `z_*` snippet: `scratchpad/quat-2082-repro.mach` (attached below by the
workers; a runnable std-linked project also exists). Two test prints:

- "two call sites": `z_length4` then `z_normalize4`, both calling `z_rsqrt` (Newton
  refinement seeded via scalar `f32:~u32` bit tricks).
- "normalize only": the single-call-site control.

Run on aarch64 (qemu or native) and x86_64.

## Expected

Normalized lane0 = 200 (fixed-point print of 0.200) on both prints, both targets —
matching x86_64.

## Actual

aarch64: "two call sites" = **191**, "normalize only" = 200. x86_64: 200 / 200.
The 0.191-vs-0.200 shape is a Newton refinement operating on a corrupted intermediate
(`half`/`y`) — a float live across the intervening call comes back wrong.

## Suspected area

A callee-saved FP register (v8–v15 lower 64 bits) not saved/restored, or a float
spill slot clobbered across a call — start from the frame/regalloc callee-saved-FP
path, NOT isel bitcast (ruled out by the #2083 control run). Note the aarch64 native
test suite passes, so the trigger is narrow: likely the specific
live-across-call + scalar-`:~`-heavy shape.

## Environment

mach 3.4.4 seed and dev HEAD, aarch64-linux (qemu + native), host linux-x86_64.

octalide:
## Summary

Fixes #2085: on aarch64, a 128-bit vector live across a call came back with corrupted
upper lanes (2–3). x86_64 was correct. Distinct from #2082/#2083 — this is a
register-allocation / ABI-modeling bug, not the vector `:~` selector.

## Root cause

AAPCS64 callee-saves only the **low 64 bits** of V8–V15; their upper halves are
caller-saved. The register allocator modeled V8–V15 as *whole-register* callee saves
(`aapcs64.mach` even notes the simplification), so a 128-bit vector parked in one
across a call kept lanes 2–3 in a register the callee freely clobbers. On x86_64/SysV
no XMM is callee-saved, so every FP value already spills across a call — hence x86_64
was unaffected. win64 is the opposite: it preserves the **full** 128 bits of XMM6–15,
so a vector there must *not* be force-spilled.

## Fix

Model the fact as an ABI vtable flag `fp_callee_saved_full_vector`:
- AAPCS64 → `false` (only low 64 preserved)
- win64 → `true` (full 128-bit XMM6–15 preserved)
- SysV / lp64d → `true` (moot: no callee-saved FP, or vectors scalarized)

In `spill_across_calls`, a vector live across a call is spilled out of a callee-saved
FP register when the flag is `false` — reusing the exact path a caller-saved value
already takes.

## Before / after (`z_normalize4`-style caller holding `q` across a call)

```
BEFORE: ldr q8, [q]        ; q parked in v8 (callee-saved)
        bl  callee          ; callee clobbers v8 upper 64 -> lanes 2,3 lost
AFTER:  stur q31, [slot]   ; q spilled to a 16-byte slot around the call
        ldur q0,  [slot]   ; reloaded intact after the call
```

## Tests

`vector_runtime.call:vector_live_across_call` — a vector parameter used in a vector op
after a call whose callee writes V8–V15's upper half; a dropped upper half fails the
exact per-lane compare. Fails on the buggy compiler under qemu-aarch64 (lanes 2–3
corrupted), passes with the fix; inert on x86_64.

## Verification (built from this branch, verified with that binary)

- The std-2021 quat repro (`z_length4` + `z_normalize4`) returns **200/200** on aarch64 with the fix (was 191/200); x86_64 unchanged at 200/200.
- Repro disasm confirms the vector is spilled around the call and reloaded intact.
- Full vector suite (37) green on x86_64 and aarch64 (qemu) legs.
- Full `mach test` green at `-O0/-O1/-O2` (647 each).
- Self-host fixpoint: stage2 == stage3, byte-identical.

Unblocks briar-systems/mach-std#372's last 2 aarch64 quaternion tests.

Closes #2085

### DIFF
--- a/src/lang/be/codegen/regalloc.mach
+++ b/src/lang/be/codegen/regalloc.mach
@@ -2633,9 +2633,14 @@ fun spill_across_calls(tgt: *target.Target, ctx: *AllocCtx) R.Result[bool, str] 
                 # callee-saved XMM survives the call once the prologue preserves it).
                 # under SysV every XMM is caller-saved, so this still spills every
                 # FP value across a call. an asm barrier additionally spills a
-                # callee-saved XMM the body writes.
+                # callee-saved XMM the body writes. a 128-bit vector is the exception on
+                # an ABI that callee-saves only the low 64 bits of its FP registers
+                # (AAPCS64's V8-V15): the callee clobbers the vector's upper half, so it
+                # must spill around the call like a caller-saved value (#2085).
                 val xidx: u32 = isa.regid_index(mv.assigned::i32)::u32;
                 if (xidx >= ctx.fp_count || !ctx.fp_callee_saved[xidx]) {
+                    must_spill = true;
+                } or (mv.vector && !tgt.abi.fp_callee_saved_full_vector) {
                     must_spill = true;
                 } or (is_asm && (asm_fp & (1::u32 << xidx)) != 0) {
                     must_spill = true;
--- a/src/lang/be/codegen/vector_runtime.mach
+++ b/src/lang/be/codegen/vector_runtime.mach
@@ -628,3 +628,34 @@ test "mach.lang.be.codegen.vector_runtime.reinterpret:u32x4_i32x4_roundtrip" {
     if (back[0] != 0x80000001 || back[1] != 0xFFFFFFFF || back[2] != 266 || back[3] != 0x7FFFFFFF) { ret 2; }
     ret 0;
 }
+
+# a 128-bit vector live across a call must survive all four lanes. AAPCS64 callee-saves
+# only the low 64 bits of V8-V15, so lanes 2-3 (the vector's upper half) are lost if a
+# caller parks the vector in a callee-saved FP register across a call instead of spilling
+# it around the call (#2085). the callee must actually write V8-V15's upper half to expose
+# the drop: `xc_clobber` parks a float in a callee-saved FP register across its own inner
+# call (an `fadd s8, ...` zeros V8's upper 64), and `xc_survive` uses its vector parameter
+# in a vector op AFTER the call, forcing it register-resident across the call. the +1 makes
+# every lane distinct so a lost upper half fails the exact per-lane compare.
+fun xc_leaf() f32 { ret 1.0; }
+
+fun xc_clobber(a: f32x4) f32 {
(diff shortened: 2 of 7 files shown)


octalide:
## Coordinator review

**Verdict: approve — merging.** The root cause is the AAPCS64 fact itself (V8-V15 callee-save covers only the low 64 bits) and the fix models it as exactly that — a documented ABI vtable fact (`fp_callee_saved_full_vector`) consumed by one condition in shared regalloc's spill-across-calls path. No arch branching in shared code; win64 correctly keeps full-width XMM6-15 preservation (a force-spill there would be a pessimization bug); SysV/lp64d moot-true with the reasoning stated. The regression test is the strong form: the callee deliberately writes s8 to zero V8's upper half, so the test fails deterministically on the buggy compiler instead of depending on incidental clobbers, and distinct lane values catch a partial loss. Verified: quat repro 200/200 (was 191), suite 42/42 both legs, 654/654 at three opt levels, fixpoint, disasm before/after. This closes the second of the two aarch64 SIMD criticals the #2021 library exposed.

PR #2093, opened 2026-07-11
fix(ir): preserve use-site operand type stamps in replace_uses

### LINKED ISSUE #2092: me: value-substitution primitives clobber use-site type stamps (aggregate deref-copy of an inlined callee's returned pointer miscompiles at -O1+)
Surfaced while implementing #2064 (unify the dual same-name Function entries per generic instance so instance bodies become visible to the inliner). #2064's unification is correct, but enabling the inliner to see monomorphized instance bodies exposes a **pre-existing latent miscompile** in the inline path.

## Symptom (verified, deterministic)
On a from-source compiler carrying the #2064 unification: full `mach test` is **644/644 at -O0** but **643/644 at -O1 and -O2** — `mach.lang.me.pass.mem2reg.run:phi_dbg_birth` fails. The failure is a **silent wrong value**, not a crash.

- Passes on plain dev (no #2064) at every opt level.
- Fails only at opt≥1 (the inliner runs at -O1+).
- Split the failing test's asserts by exit code: it fails at `!found` (the phi promotion itself is correct — the merge phi IS created; the **dbg-value birth** for it is missing or mis-recorded).

## Bisection (deterministic)
Gating `should_inline` to skip weak (instance) callees by instruction count brackets the trigger exactly:
- skip all weak inlining → **passes**
- inline weak callees only if `callee_instr_count < K`: K≤11 passes, **K=12 fails, K=13 fails** → the trigger is inlining a **12-instruction weak (instance) body**.

## It is NOT a single instance / NOT reproducible in isolation
- The first weak-12 body inlined in the mem2reg module is `std.types.result.unwrap_err[*u32,*u8]`, but skipping **only** that still fails → multiple weak-12 instances are involved.
- **Every isolated reproduction inlines correctly.** I built and inspected the IR for: the real `unwrap_err` (with its `panic`-arm branch); tagged-union / `Result`-shaped aggregates; the `is_err`→`unwrap_err` error-propagation flow; 16-byte two-register-class and 24-byte stack-class by-value aggregate params; multi-block callee bodies. All produce correct inlined IR in a simple caller. The fault only appears inside mem2reg's full `run` path.
- Ruled out: `clone_instruction` deep-copies operand arrays (no original/clone aliasing).

**Working hypothesis:** this may not be an inliner *clone* defect per se, but **cumulative instance inlining** bloating a mem2reg function enough to trip a **downstream pass** (post-inline mem2reg2 / dce / codegen); the weak-12 size may simply be what pushes one function over an edge. Needs full-function IR tracing across the inline→mem2reg2→codegen boundary to localize.

## Relation to #2035 / #2062
Same *flavor* as #2035 (silent wrong value around by-value aggregates through generics), but a **different path**: #2062/#2063 fixed the **call-site ABI** (per-operand stamps, the callee signature). This is inside the inline **body-clone / post-inline** path, where no call remains — so the always-on `VC_CALL_ARG_TYPE` verifier cannot see it (inlining removed the call).

## Reachability (release-gating input)
The **trigger** is dev-reachable **today**: a concrete, non-forward-referenced function with a by-value aggregate parameter **does inline on dev** (demonstrated: `fun geterr(r: Res) i64 { ret r.pay.err; }` inlines into its caller on the dev compiler). What is **not** demonstrated is a dev program that actually miscompiles — every dev-reachable shape built inlines correctly; the only confirmed miscompile needs mem2reg's full context, reached via #2064's instance inlining. So: **trigger reachable, no demonstrated dev miscompile.** Whether the reachable trigger alone warrants gating a release is a maintainer call.

## Blocks
#2064 stays parked until this is fixed (it must not ship a compiler that miscompiles its own mem2reg at -O2).

## Regression test target
An opt-independent in-vitro test over `inline.run` (the PR #2062 precedent) proving the defect without #2064's unification, once the exact context/defect is isolated. Until then, `mem2reg.run:phi_dbg_birth` built on the #2064 branch at -O1/-O2 is the reproducer.

octalide:
Closes #2092

## The defect

In this IR an aggregate value flows by address with the aggregate type stamped on the operand: a store whose value operand is a ptr-producing instruction stamped `{agg}` means "copy the aggregate", not "store the pointer". The use-site stamp is the slot's semantic contract, independent of the named value's own def type.

The batched RAUW facility (`Rewrite` / `rewrite_resolve`) preserves that contract explicitly — it retypes every resolved replacement to the slot's type, documented as "the aggregate-pointer identity mem2reg relies on". Its immediate sibling `mutate.replace_uses` — used only by the inliner's `finish_returns` to wire a call result to the callee's return value (both the single-return arm and the multi-return result-phi arm) — substituted the replacement **wholesale**, clobbering the use-site stamp with the replacement's bare def type.

Concretely: `var x: Rec = @f();` lowers to `store %call:{Rec-stamp}, %x_slot` where `f` returns `*Rec`. Once `f` is inlined, the store's operand becomes the return value typed `ptr` — the backend stores 8 bytes of pointer bits into `x` instead of copying the record. Silent wrong values.

## Reachability

**Not gated on #2064.** Plain current dev and the shipped 3.5.0 release miscompile this 20-line program at -O1/-O2 (7/8/9 expected):

```
rec R { a: i64; b: i64; c: i64; }
var g: R = R{ a: 7, b: 8, c: 9 };
fun make() *R { ret ?g; }
...
var x: R = @make();      # O2: a=4481064 b=0 c=<stack garbage>
```

Any release-profile program deref-copying the result of an inlinable (small or single-use, same-module) pointer-returning function is affected. **Verified affected: every shipped release** — v1.0.0 (earliest, exit-code probe) and v3.0.0 empirically, v3.5.0 and dev directly; structurally the substitution is byte-identical from pre-v1.0.0 (06cb9b84) through the mutate migration (1f6c2581) to today. Details on the issue. #2064's instance unification merely widened the set of inlinable callees (`O.unwrap[*T]` et al.) until the compiler's own `emit_phi_dbg_births` hit the shape — that is the whole `phi_dbg_birth` mystery: `dv.name` became the low 32 bits of a heap pointer.

## The fix

`replace_uses` retypes the replacement to each rewritten slot's type, exactly mirroring `rewrite_resolve`. One-line semantic change; the substitution primitives now uniformly own the stamp-preservation invariant (the #2035 fix did the same for VAL_PARAM substitution in `remap_clone_operands`).

## Tests

Two opt-independent in-vitro tests over `inline.run` (the PR #2062 precedent), pinning both `finish_returns` arms:
- `preserves_aggregate_result_operand_type` — single-return callee; the surviving store keeps the `{agg}` stamp.
- `preserves_aggregate_result_operand_type_phi` — two-return callee; the store names the result phi and keeps the `{agg}` stamp.

Both fail (exit 2) without the one-line fix and pass with it.

## Verification (from-source compiler, in progress)

- [x] new tests fail-before / pass-after
- [x] full `mach test` at -O0/-O1/-O2
- [x] repro program correct at -O0/-O1/-O2
- [x] self-host fixpoint
- [x] in-vivo: #2064 unification applied on top — `phi_dbg_birth` green at -O1/-O2

- [x] int/regression/2092-inline-result-agg-copy: release leg fails on unfixed compiler, both profiles pass on fixed

### DIFF
--- a/int/regression/2092-inline-result-agg-copy/expect.txt
+++ b/int/regression/2092-inline-result-agg-copy/expect.txt
@@ -1,0 +1,3 @@
+make=24
+pick1=24
+pick0=60
--- a/int/regression/2092-inline-result-agg-copy/mach.toml
+++ b/int/regression/2092-inline-result-agg-copy/mach.toml
@@ -1,0 +1,57 @@
+[project]
+id = "case2092"
+version = "0.0.0"
+src = "src"
+out = "out/{target.name}/{profile.name}"
+
+[target.linux]
+isa = "x86_64"
+os  = "linux"
+abi = "sysv64"
+
+[target.linux-arm64]
+isa = "aarch64"
+os  = "linux"
+abi = "aapcs64"
+
+[target.linux-riscv64]
+isa = "riscv64"
+os  = "linux"
+abi = "lp64"
+
+[target.windows]
+isa = "x86_64"
+os  = "windows"
+abi = "win64"
+
+[target.darwin-x86_64]
+isa = "x86_64"
+os  = "darwin"
+abi = "sysv64"
+
+[target.darwin-aarch64]
+isa = "aarch64"
+os  = "darwin"
+abi = "aapcs64"
+
+[profile.debug]
+opt = 0
+debug = false
+simd = "scalarize"
+
+[profile.release]
+opt = 2
+debug = false
+simd = "scalarize"
+
+[artifact.case]
+kind = "bin"
+entry = "main.mach"
+out = "bin/case"
+targets = ["*"]
+link = []
+need = []
+
+[dep.mach-std]
+git = "https://github.com/briar-systems/mach-std"
+ref = "branch/main"
--- a/int/regression/2092-inline-result-agg-copy/src/main.mach
+++ b/int/regression/2092-inline-result-agg-copy/src/main.mach
@@ -1,0 +1,43 @@
+# regression pin for #2092: an aggregate deref-copy of an inlined call's result.
+#
+# `var x: R = @f()` lowers to a store whose value operand is the ptr-returning
+# call stamped with the aggregate type (a by-address aggregate). when f is
+# inlined, replace_uses substitutes the callee's return value into that slot;
+# clobbering the use-site stamp with the return value's bare ptr type degraded
+# the record copy to an 8-byte pointer-bits store, so the release profile
+# printed pointer fragments while debug printed the record. covers both
+# finish_returns arms: the single-return callee (direct value) and the
+# two-return callee (result phi in the continuation).
+
+use std.runtime;
+use std.types.size.usize;
+use p: std.print;
+
+rec R { a: i64; b: i64; c: i64; }
+
+var g:  R = R{ a: 7, b: 8, c: 9 };
+var g2: R = R{ a: 10, b: 20, c: 30 };
+var sel: i64 = 1;
+
+# single return: the call result is wired directly to the returned pointer.
+fun make() *R { ret ?g; }
+
+# two returns: the call result is wired to a result phi over both pointers.
+fun pick(c: i64) *R {
+    if (c != 0) { ret ?g; }
+    ret ?g2;
+}
+
+#[symbol("main")]
+fun main(argc: usize, argv: **u8) i64 {
+    var x: R = @make();
+    p.printlnf("make={}", x.a + x.b + x.c);       # 24
+
+    var y: R = @pick(sel);                        # sel is a mutable global, not folded
+    p.printlnf("pick1={}", y.a + y.b + y.c);      # 24
+
+    sel = 0;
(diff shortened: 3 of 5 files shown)


octalide:
## Coordinator review

**Verdict: approve — merging; cuts v3.5.1 immediately.** The fix is the primitive-level invariant, not a call-site patch: `replace_uses` retypes the substituted value per use slot (`value.retype(replacement, op.ty)`), mirroring the batched `rewrite_resolve` convention, with the slot-semantic contract now documented at the primitive. The audit of every other wholesale operand-assignment site (rewrite_apply, dbg salvage, scalarize resolve, phi compactions, constfold CBR) is the part that makes this class dead rather than moved — recorded on #2092. Tests pin both finish_returns arms in-vitro (fail-before/pass-after verified) plus the owner-approved end-to-end release-profile int case. Verification: 656/656 at three opt levels, fixpoint, and the in-vivo proof that #2064's blocker vanishes with this fix. Root-cause work of the highest order.

Issue #1859, opened 2026-07-03
incremental: lowered-surface fingerprint firewall for Q_LOWER


octalide:
Part of #1164. Depends on #1618. The back-half analogue of #1581's `Q_EXPORTS` public-surface firewall, surfaced by PR #1850.

## What

`Q_LOWER(m)` records deps on `Q_SEMA` of m's transitive import closure — required for soundness, because generic/comptime instances from dependencies monomorphize into m's IR and no per-hop lowering firewall exists. The consequence: an impl-only edit to a dependency invalidates every importer's `Q_LOWER` (and so `Q_CODEGEN`) even when the material those importers actually consume is unchanged.

Introduce a per-module lowered-surface fingerprint: a hash of exactly the material that can flow into an importer's IR. Importers' `Q_LOWER` then deps on the dependency's fingerprint instead of its full `Q_SEMA`, so impl-only edits stop cascading through the back half, mirroring how #1581 stopped the same cascade through resolve/sema.

## The lowered surface (code-accurate)

The material that can monomorphize into an importer's IR, and therefore MUST be in the fingerprint:

- generic and comptime-param template bodies (monomorphized as weak local definitions in the importer);
- comptime-evaluable values (constant-folded into the importer);
- ALL type layouts, including private ones (they embed into monomorphized code);
- every symbol's mangled name, signature, and flags (a monomorphized body calls them by mangled name).

Excluded (does NOT flow, verified from emitted IR): the bodies of regular non-generic functions — including plain `#[inline]` functions. mach does **not** inline across module boundaries: a cross-module callee is emitted **extern** in the importer (`@_M...N4fn(): i32 extern`), and the inline pass skips extern callees (`should_inline` returns false on `FN_FLAG_EXTERN`). So editing a regular or plain-inline export's body changes no importer's IR — it is an impl-only edit. (An earlier draft of this issue listed "inline-eligible exports" as flowing surface; that described cross-module inlining, which does not exist today. Plain-inline bodies would join the surface only if/when cross-module inlining is added — pinned by a driver test and a comment at the inline pass's extern-skip.)

## Acceptance

- An impl-only edit to a dependency (a regular- or inline-function body; unchanged lowered surface) reuses every importer's `Q_LOWER`/`Q_CODEGEN` on a warm session (`last_changed`-probe driver test).
- A surface-affecting edit (generic body or comptime value) still invalidates importers.
[...]

PR #2096, opened 2026-07-11
fix(me): unify the dual same-name Function entries per generic instance (#2064)

### LINKED ISSUE #2064: codegen: duplicate same-name Function entries per generic instance (extern decl + definition); hides instance bodies from the inliner

octalide:
Closes #2064.

A forward reference or a generic-instance reference creates a signature-only `FN_FLAG_EXTERN` Function entry at the call site (`ensure_extern_function`), and every such call is captured as `value.fn(that-index)`. Materialising the definition into a **fresh** entry orphaned the extern: the call kept pointing at the bodiless extern, so the inliner (which never inlines an extern callee) could never see the body, and a redundant extern decl shipped alongside every instance/forward definition.

## Fix
`append_function` now **upgrades an existing extern entry of the same name in place** — its index, and thus every already-emitted call, now names the definition — unifying the two entries and exposing instance and forward-referenced bodies to the inliner. Only an extern is reused; a genuine cross-module import is never defined locally, so `append_function` is never called for it and its extern stays extern (the `should_inline` extern skip and the #1859 incremental-fingerprint invariant are preserved).

Composes with #2087's `Function.is_import`: for concrete/`ext` funcs `lower_fun` sets `is_import` *after* `append_function`, so the reuse can't lose it; for instances the reference passes `import_flag=false` and the reuse preserves it (matching a fresh entry's default), so an instance is never mis-marked an import.

The second commit is comment-only framing polish (`abi.mach`/`verify.mach`) stranded off #2063 by a merge-ordering crossing; it rides here.

## Inliner win (the point of #2064)
A generic-instance call and a forward-referenced call that previously pointed at bodiless externs — never inlinable — now inline. Demo (`-O2`): `fun main() { ret (early() + id[i64](7)) }` folds to **`ret 13`** (both `id[i64]` the instance and `late` via `early` inlined+folded), where before both remained `call`s to externs.

## Verification (from-source fix/2064 compiler, never the PATH seed)
- `mach test` **659/659** at **-O0 / -O1 / -O2**, including `mem2reg.run:phi_dbg_birth` **green at all levels** (the #2092 miscompile this unification exposed is fixed on dev by #2093) and the 42-test vector runtime suite.
- Self-host **fixpoint g2 == g3** byte-identical.
- Inliner-win demo as above (`ret 13`).

## Note
This unification is what surfaced #2092 (a pre-existing, dev-/3.5.0-reachable inliner call-result RAUW stamp clobber, fixed by #2093 and shipped in v3.5.1). #2064 was the flashlight, not the cause.

### DIFF
--- a/src/lang/be/codegen/mir/abi.mach
+++ b/src/lang/be/codegen/mir/abi.mach
@@ -782,9 +782,13 @@ pub fun lower_call(ctx: *context.LowerCtx, mb: *mir.MirBlock, inst: *instruction
     # fixed argument read their placement from this single layout so the call and
     # the callee's `lower_params` can never disagree - in particular a by-value
     # aggregate argument is classified from its DECLARED parameter type, not from
-    # the 8-byte pointer its operand carries. an indirect call through an untyped
-    # function pointer yields zero fixed params, so every argument falls back to
-    # per-operand classification below.
+    # the 8-byte pointer its operand carries. every call with a declared signature
+    # takes this path - a direct call to a named function or monomorphized instance,
+    # and an indirect call through a TYPED function pointer (its callee carries the
+    # IRT_FN). only a signature-less position yields zero fixed params and defers to
+    # per-operand classification below: the variadic tail, or a callee with no
+    # IRT_FN type (an erased / untyped function pointer). this is never a fallback
+    # for a direct call whose signature merely went uncomputed.
     val sig: type.IrTypeId = inst.aux_ty;
     # the declared return type comes from the callee signature when one is known,
     # not `inst.ty`: on a scalarized target a vector-returning call's result type is
--- a/src/lang/me/ir/verify.mach
+++ b/src/lang/me/ir/verify.mach
@@ -645,14 +645,17 @@ fun check_type_agree(m: *ir.Module, ins: *instruction.Instruction, fi: u32, bid:
     ret R.ok[bool, str](true);
 }
 
-# VC_CALL_ARG_TYPE: a direct call classifies its ABI from the callee's declared
-# IRT_FN signature (`aux_ty`), so each fixed argument's operand type must agree with
-# the declared parameter type. checked only when `aux_ty` is a real IRT_FN - every
-# direct call to a named function or (now) a monomorphized instance; an indirect
-# call through an untyped function pointer carries no declared parameter types and
-# is skipped, as is the variadic tail past the declared parameter count (its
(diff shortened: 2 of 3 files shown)


octalide:
### Verification (from-source fix/2064 compiler, rebased on dev @ #2087)
- `mach test` **659/659 at -O0/-O1/-O2**; `mem2reg.run:phi_dbg_birth` **green at all levels** (was the #2092 signature that #2093 fixed); vector runtime suite 42/42.
- Self-host **fixpoint g2==g3** byte-identical.
- **Inliner win**: `main` folds to `ret 13` at -O2 (generic instance `id[i64]` and forward-ref `late` both inlined; previously both stayed `call`s to bodiless externs).
- **#2087 `is_import` composition checked**: concrete/`ext` funcs — `lower_fun` sets `is_import` after `append_function`, so the reuse can't drop it; instances — reference passes `import_flag=false`, reuse preserves it. Full suite (incl. win64 marshal tests) green corroborates.

octalide:
## Coordinator review

**Verdict: approve — merging.** The unification is the minimal correct mechanism: `append_function` upgrades a prior extern entry in place (index-stable, so every captured `value.fn` now names the definition) rather than orphaning it, with the boundary conditions documented — genuine redefinitions still append and surface downstream, and true cross-module imports never reach this path because they are never defined locally. The #2087 is_import composition is verified both ways (assigned after append for ext decls; preserved on reuse for instance refs). The cherry-picked polish lands the durable framing: per-operand classification is scoped to genuinely signature-less positions, and typed function pointers are covered by the guard via the pointee IRT_FN — recorded in the code, not just a PR thread. Verification bar met: 659/659 at three opt levels with phi_dbg_birth green everywhere (the #2092 in-vivo proof), fixpoint, 42/42 vectors, and the ret-13 inliner win that was this issue's whole point.

PR #2097, opened 2026-07-12
fix(sema): fold imported pub val in type positions (#2081)

### LINKED ISSUE #2081: sema: imported pub val constant rejected as array length
A `pub val` integer constant defined in one module and imported into another is rejected when used as an array length, even though a local `val` with the same value is accepted. This forced a small capacity constant to be duplicated locally instead of shared.

## Repro

Module `a.mach`:

```mach
pub val CAP: usize = 16;
```

Module `b.mach` (same project):

```mach
use a: proj.a;
fun f() {
    var buf: [a.CAP]u8;   # rejected as an array length
}
```

A local `val CAP: usize = 16;` in `b.mach` used as `[CAP]u8` compiles.

## Expected

An imported `pub val` compile-time integer constant should be usable as an array length, the same as a local `val`.

## Actual

The cross-module `pub val` is not accepted in array-length position (const-eval / import does not treat it as a compile-time constant there).

## Environment

- mach 3.4.4, linux/x86_64
- Found in briar-systems/bramble-hub (registry store capacity constant).

Minor severity: worked around by duplicating the constant, but it defeats sharing a single capacity definition across modules.

octalide:
Fixes #2081.

## Root cause

An imported `pub val` integer constant used as an array length (`var buf: [a.CAP]u8;`) was rejected while an identical local `val` compiled.

Array lengths — and every syntactic type node, including function-body-local ones — are resolved in sema's `resolve_all_types` pass, which evaluates the length through `comptime.eval`. That pass never installed the module-member resolver (`member_fn`): only lowering wired it up. So a cross-module `alias.CONST` reference in a type position could not be folded to its comptime value and `resolve_type_array` reported "array length is not a comptime constant". A local `val` works because it is bound by name into the module's comptime ctx during load.

## Fix

Wire a sema-side module-member resolver into the comptime context used for type positions:

- `sema/context.mach`: add `resolve_module_member_const`, mirroring the lowering-installed resolver but reading the `*SemaContext` (`sc.rr`, `sc.s`). It resolves the MEMBER symbol, and for an imported module-level constant returns the pre-folded value from the declaring module's ComptimeCtx. A miss (function, `var`, non-constant `val`, unexported/unknown member) returns none, which `comptime.eval` turns into the existing loud diagnostic.
- `sema.mach`: install the resolver on `ctx` around `resolve_all_types` only, and clear it after. That pass is the single authoritative site for type-position comptime evaluation; expression positions in bodies stay unresolved at sema and continue to fold against the lowering-installed resolver, so expression-position comptime semantics are unchanged.

Topo order guarantees the origin module's comptime ctx is registered by the time a dependent's types resolve, so the fold produces exactly the value lowering would — identical before and after lowering.

## Tests

Colocated two-module sema test: an imported `pub val` used as an array length type-checks to an array of the right size; the negative (imported `var`, non-comptime initializer) is still rejected with the existing diagnostic.

### DIFF
--- a/src/lang/driver.mach
+++ b/src/lang/driver.mach
@@ -6807,6 +6807,62 @@ test "mach.lang.driver:vector_negative_diagnostics" {
     ret bad;
 }
 
+# build a two-module project (main.mach + a.mach) from source, run resolve + sema,
+# and return the project so a test can inspect p.s.diags. the caller tears down the
+# project and session
+fun ut_sema2(s: *session.Session, alloc: *A.Allocator, main_src: str, a_src: str) R.Result[Project, str] {
+    var names: [2]str;
+    names[0] = "main.mach"; names[1] = "a.mach";
+    var texts: [2]str;
+    texts[0] = main_src; texts[1] = a_src;
+    val pr: R.Result[Project, str] = ut_build(s, alloc, ?names[0], ?texts[0], 2);
+    if (R.is_err[Project, str](pr)) { ret pr; }
+    var p: Project = R.unwrap_ok[Project, str](pr);
+    val se: R.Result[bool, str] = run_sema_pass(?p);
+    if (R.is_err[bool, str](se)) { dnit_project(?p); ret R.err[Project, str](R.unwrap_err[bool, str](se)); }
+    ret R.ok[Project, str](p);
+}
+
+# an imported `pub val` integer constant folds in a type position (an array length)
+# exactly as a local `val` does, so `[a.CAP]u8` compiles and sizes to the constant's
+# value; an imported `var` - not a comptime constant - is still rejected (#2081)
+test "mach.lang.driver:imported_pub_val_array_length" {
+    var alloc: A.Allocator;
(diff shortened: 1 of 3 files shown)


octalide:
## Coordinator review

**Verdict: approve — merging.** Fixed at the cause with no position special-casing: sema's `resolve_all_types` — the single authoritative site for type-position comptime eval — gains the module-member resolver its lowering twin always had, scoped to exactly that pass with an explicit clear (no dangling ctx pointer). The resolver is properly guarded at every step (nil ctx, unresolved expr, non-imported symbol, missing origin ctx, lookup miss) and every miss becomes the existing loud diagnostic rather than a silent fold. Expression-position semantics untouched by construction. The test proves the fold's *value*, not just absence of errors (a 16-element literal only checks against `[a.CAP]u8` if CAP folded to exactly 16), plus the imported-`var` negative. Verified: 660/660, fixpoint, int harness both profiles, the bramble-hub repro compiles on the fix and still fails on the seed.

PR #2098, opened 2026-07-12
feat(manifest): step template target-tuple vars + MACH_TARGET env injection (#1964)

### LINKED ISSUE #1964: rfc: mach.toml redesign — artifacts, steps, links

octalide:
Implements point 7 of RFC #1964 (the last undelivered point; merging fully reconciles and closes the RFC, per the owner-approved 2026-07-11 reconciliation comment).

## What

- **Template variables.** The closed, final manifest template set grows by exactly three keys — `{target.isa}`, `{target.os}`, `{target.abi}` — resolved from the active build cell. The full set is now `{project.out}`, `{target.name}`, `{target.isa}`, `{target.os}`, `{target.abi}`, `{profile.name}`. They resolve everywhere `manifest.expand` runs (artifact `out`, local link `path`, step `in`/`out`/`cmd`).
- **Env injection.** Every build-step process now inherits `MACH_TARGET_ISA`, `MACH_TARGET_OS`, `MACH_TARGET_ABI` for the active target, appended to the inherited environment at exec time.

`[project]` parsing is untouched — point 1 of the finalized-design comment is superseded by the shipped stricter surface, per the reconciliation.

## How

The target/profile identity a template expands against is bundled into a single `manifest.TmplVars` record (`target`/`isa`/`os`/`abi`/`profile`) threaded through `expand` and its callers, so the six-key set is uniform and future keys are one field rather than another positional argument at every call site. `{project.out}` stays a separate positional argument since it varies by what a template homes against (and the `[project].out` self-reference guard is unchanged). Driver-side, `cell_tmpl_vars(p)` builds the bundle from the active `TargetEntry`; the same three values feed both cmd expansion and the injected env.

## Tests

- `expand:project_out` and `expand:closed_set` exercise the three new keys resolving (`x86_64-linux-sysv64`) and the dropped `{name}`/`{ext}` aliases still erroring.
- Existing manifest/driver suites updated for the new signature.
- Full suite: **660 passed, 0 failed** via the from-source compiler. All CI lanes green (build/test/int/release across x86_64, aarch64, riscv64, windows).

Docs updated in `doc/manifest.md`.

## End-to-end proof

Scratch project with a step whose `cmd` uses all three keys plus `{project.out}`, and a script echoing the three env vars, built against two targets with the from-source compiler:

```
########## TARGET linux-x86_64 ##########
step probe: sh scripts/probe.sh x86_64 linux sysv64 out/linux-x86_64/debug/probe.txt
  TEMPLATE keys : isa=x86_64 os=linux abi=sysv64
  ENV inherited : MACH_TARGET_ISA=x86_64 MACH_TARGET_OS=linux MACH_TARGET_ABI=sysv64

########## TARGET linux-arm64 ##########
step probe: sh scripts/probe.sh aarch64 linux aapcs64 out/linux-arm64/debug/probe.txt
  TEMPLATE keys : isa=aarch64 os=linux abi=aapcs64
  ENV inherited : MACH_TARGET_ISA=aarch64 MACH_TARGET_OS=linux MACH_TARGET_ABI=aapcs64
```

Both the `{target.*}` template expansion (visible in the emitted `cmd`) and the injected `MACH_TARGET_*` env resolve from the active build cell and differ correctly per target. As a control, the released seed compiler (3.5.1) rejects the same manifest with `mach.toml: unknown template variable '{target.isa}'` — the feature is what makes these keys valid.

Closes #1964

### DIFF
--- a/doc/manifest.md
+++ b/doc/manifest.md
@@ -88,7 +88,7 @@ ref = "branch/main"
 | `id`      | string | Root segment of every module path the project exposes: a file at `<src>/foo/bar.mach` is reachable as `<id>.foo.bar`. Must be a plain identifier — letters, digits, `_`, `-` — since it names the dependency store and keys step stamp files. Read by `$project.id`. |
 | `version` | string | Project version. Read by `$project.version`; the source of truth a Go-style `tag/<version>` acquisition checks. |
 | `src`     | string | Source root, project-root-relative. Module paths resolve under it. |
-| `out`     | string | The output-path template root, referenced as `{project.out}` by artifact `out`, step paths, and `cmd`s. Expanded over `{target.name}`/`{profile.name}` (see [Path templates](#path-templates)). |
+| `out`     | string | The output-path template root, referenced as `{project.out}` by artifact `out`, step paths, and `cmd`s. Expanded over `{target.name}`/`{target.isa}`/`{target.os}`/`{target.abi}`/`{profile.name}` (see [Path templates](#path-templates)). |
 
 `[project]` is exactly these four keys; `name`, `description`, and any other key
 are unknown-key errors in a root manifest.
@@ -261,6 +261,13 @@ outputs still exist is skipped; change an input or the command and it re-runs.
 consumer's output tree — exactly as a dependency's compiled modules do — and the
 dependency's own checkout is never written to.
 
+**Target environment.** Every step process additionally inherits the active
+build cell's target tuple as `MACH_TARGET_ISA`, `MACH_TARGET_OS`, and
+`MACH_TARGET_ABI`, so a `cmd` or the script it invokes can branch on the target
+without threading it through the template — e.g. `cc --target=$MACH_TARGET_ISA-…`.
+The same three values are available in the `cmd` template as the `{target.*}` keys
+(see [Path templates](#path-templates)).
+
 ## `[dep.<alias>]`
 
 Each `<alias>` names a dependency materialised under `dep/<alias>/`. The build
@@ -297,12 +304,18 @@ entries demand. Nothing else in a dependency's manifest applies to consumers.
 
 ## Path templates
 
-Paths and `cmd`s expand over a closed, final set of three variables:
+Paths and `cmd`s expand over a closed, final set of six variables:
 
 - `{project.out}` — the **root** project's expanded `[project].out`, in every
   manifest of the closure.
 - `{target.name}` — the resolved target name (never the literal `native`).
+- `{target.isa}` — the resolved target's `isa` (e.g. `x86_64`).
+- `{target.os}` — the resolved target's `os` (e.g. `linux`).
+- `{target.abi}` — the resolved target's `abi` (e.g. `sysv64`).
 - `{profile.name}` — the selected profile name.
+
+The three `{target.*}` tuple keys are also exported to every step process as
+`MACH_TARGET_ISA`/`MACH_TARGET_OS`/`MACH_TARGET_ABI` (see [build steps](#stepname--build-steps)).
 
 An artifact's `out` is relative to the expanded project `out` and is rooted there
 automatically — write `bin/demo`, not `{project.out}/bin/demo`. Step `out` lists
--- a/src/lang/driver.mach
+++ b/src/lang/driver.mach
@@ -1567,14 +1567,16 @@ fun load_config(p: *Project, project_root: str, t: *toml.Table, sel: *manifest.S
     }
     val bu: manifest.BuildUnit = R.unwrap_ok[manifest.BuildUnit, str](ur);
 
+    # the closed template-variable set for this build cell (#1964): the target tuple
+    # and profile every {target.*}/{profile.name} path and cmd expands against.
+    var cell_v: manifest.TmplVars = manifest.tmpl_vars_of(?p.s.interner, bu.target,
+        O.unwrap[str](intern.lookup(?p.s.interner, bu.profile_name)));
+
     # honest collision guard over every artifact for the selected target/profile.
-    val pn_opt: O.Option[str] = intern.lookup(?p.s.interner, bu.profile_name);
-    if (O.is_some[str](pn_opt)) {
-        val cr: R.Result[bool, str] = manifest.check_collisions(p.s.alloc, ?p.s.interner, ?m, bu.target, O.unwrap[str](pn_opt));
-        if (R.is_err[bool, str](cr)) {
-            manifest.dnit(?m, p.s.alloc);
-            ret R.err[bool, str](R.unwrap_err[bool, str](cr));
-        }
+    val cr: R.Result[bool, str] = manifest.check_collisions(p.s.alloc, ?p.s.interner, ?m, ?cell_v);
+    if (R.is_err[bool, str](cr)) {
+        manifest.dnit(?m, p.s.alloc);
+        ret R.err[bool, str](R.unwrap_err[bool, str](cr));
     }
 
     if (bu.warned) { emit_native_warning(p, bu.target_name); }
@@ -1631,10 +1633,8 @@ fun load_config(p: *Project, project_root: str, t: *toml.Table, sel: *manifest.S
     p.config.step_order       = nil;
     p.config.step_order_count = 0;
 
-    val tn_s: str = O.unwrap[str](intern.lookup(?p.s.interner, bu.target.name));
-    val pn_s: str = O.unwrap[str](intern.lookup(?p.s.interner, bu.profile_name));
     val out_tmpl_s: str = O.unwrap[str](intern.lookup(?p.s.interner, m.out_tmpl));
-    val res_project_out_str: R.Result[str, str] = manifest.expand(p.s.alloc, out_tmpl_s, "", tn_s, pn_s);
+    val res_project_out_str: R.Result[str, str] = manifest.expand(p.s.alloc, out_tmpl_s, "", ?cell_v);
     if (R.is_ok[str, str](res_project_out_str)) {
         val project_out_str: str = R.unwrap_ok[str, str](res_project_out_str);
         val res_po_id: R.Result[intern.StrId, str] = intern.intern(?p.s.interner, project_out_str);
@@ -1644,7 +1644,7 @@ fun load_config(p: *Project, project_root: str, t: *toml.Table, sel: *manifest.S
         # up-front local link path validation (#1969): a `local` entry's path must
         # match a step's `out` (demanding it) or exist on disk; anything else errors
         # here naming the entry, instead of flowing to the linker to fail at link time.
-        val vr: R.Result[bool, str] = validate_local_links(p, project_root, ?m, bu.target, project_out_str, tn_s, pn_s);
+        val vr: R.Result[bool, str] = validate_local_links(p, project_root, ?m, bu.target, project_out_str, ?cell_v);
         if (R.is_err[bool, str](vr)) { str_free(p.s.alloc, project_out_str); manifest.dnit(?m, p.s.alloc); ret R.err[bool, str](R.unwrap_err[bool, str](vr)); }
 
         # `mach test` links the union of every artifact's referenced entries (native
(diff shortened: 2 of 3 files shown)


octalide:
## Coordinator review

**Verdict: approve — merging; this fully reconciles and closes RFC #1964.** The TmplVars bundle is the right mechanism: the closed set becomes one record threaded uniformly through every expand path (artifact out, local link, step in/out/cmd), so future keys are a field rather than a 19-site positional-arg sweep — and the diff is net-simpler for it (the repeated interner lookups collapse into cell_tmpl_vars). {project.out} correctly stays positional (per-template variance + the self-ref guard). Env injection at the exec seam, inherited environment preserved. The end-to-end proof has the right control: two targets producing different expansions AND the 3.5.1 seed rejecting the same manifest with the unknown-key error — proving the feature is the enabler, not a coincidence. Point 1 of the finalized comment stands superseded per the reconciliation; with point 7 merged the RFC's ledger is square.

Issue #1825, opened 2026-07-02
epic: build performance — parallelize the backend, share work across the target matrix


octalide:
> ## Status (2026-08-07): every child is closed; the epic is complete pending an owner call to close it
>
> All five children are merged. Re-measured today on `dev` @ `7ca88f1e` with the
> compiler built from that commit, 16-core host, clean tree (`mach clean` first),
> single self-build, debug:
>
> | | audit (#1803, 159 modules, serial) | today (227 modules, parallel) |
> |---|---:|---:|
> | wall | 6.7 s | **5.1 s** |
> | CPU utilisation | 99 % of one core | **221 %** |
>
> More modules, less wall time, and the build genuinely spans cores, which is the
> thing this epic existed to fix. The out-of-scope items are all closed too (#1164,
> #1618, #1773), so nothing this epic defers is still pending.
>
> **One defect found while re-measuring, not yet filed.** The `-v` phase readout is
> now wrong about where the time goes: it reports `codegen  222 modules  231us`, which
> cannot be true for 222 modules producing a 5 MB binary, and it omits the `optimize`
> phase entirely even though the table below measures it. Codegen runs inside the
> parallel pool the lower phase opens, so the codegen timer measures the join rather
> than the work. The number the audit in this epic was built on is no longer readable
> off the tool that produced it. That is #2288's defect class exactly, and it should be
> filed against the readout rather than fixed here.
>
> Not closing this on my own. Every child being done is a fact; whether the epic's goal
> is met is a judgement.
>
> ---

## What

The `mach build` audit (#1803) measured where a build's wall time actually goes and
found the same shape the test-harness overhaul (#1788) started from: the dominant cost
is **per-artifact work that is done serially or redundantly**, not intrinsic compile
work. This epic umbrellas the build's performance follow-ups the way #1788 umbrellaed
the test harness's.

Measured on the compiler itself (159 modules, current `dev`, release-built
compiler-under-test, 16-core host), single self-build:

| phase    | debug (`-O0`) | release (`--release`) |
|----------|--------------:|----------------------:|
| load     |        457 ms |                457 ms |
| resolve  |        180 ms |                180 ms |
| sema     |        186 ms |                186 ms |
| lower    |        1.5 s  |                1.5 s  |
| optimize |        594 ms |                1.3 s  |
| codegen  |        3.4 s  |                5.4 s  |
| link     |        331 ms |                395 ms |
| **wall** |    **6.7 s**  |            **9.5 s**  |

[...]

PR #2099, opened 2026-07-12
feat(build): parallelize per-module codegen across cores

### LINKED ISSUE #1826: build: parallelize the per-module backend (lower/optimize/codegen) across cores
Part of #1825.

## Problem

The back half of a build — lowering, optimization, codegen — runs strictly serially on a
single core. `compile_modules` in `src/cli/cmd/build.mach` walks `p.topo` three times
(one `for` per phase), each iteration handling one module in-process, and the process
never leaves one core: a compiler self-build reports `6.31s user / 6.80s wall`, 99 % of
one CPU, with 15 cores idle on the host.

That back half is where the time is. Measured on the compiler (159 modules, current
`dev`):

- debug: lower 1.5 s + optimize 0.59 s + codegen 3.4 s = **5.5 s of 6.7 s wall (~83 %)**
- release: lower 1.5 s + optimize 1.3 s + codegen 5.4 s = **8.2 s of 9.5 s wall (~86 %)**

codegen alone is the single largest phase (~50–57 % of wall), and it is **embarrassingly
parallel**: each module is lowered, optimized, and codegen'd into its own independent
`of.ObjectImage` (`codegen.codegen_module(p.s, target, ir, asm_out)`), with no
cross-module dependency in the back half — the topo order matters for the *link*, not for
producing each object. This is the build's version of the serial-test-execution problem
the harness overhaul (#1788) solved with a parallel runner; the win here is the same shape.

## Direction

Run the per-module back half across a worker pool sized to `os.cpu_count()` (the same
primitive the parallel test runner uses), producing the `images[]` / `irs[]` arrays out
of order but assembling the link in collection order so output stays byte-identical.

The correctness gate is shared session state: codegen reads the interner, the query DB,
and the monomorphizer caches off `p.s`. Two paths, to be decided in design:

- **In-process threads** over codegen, if the back-half reads of the session can be made
  read-only / per-worker (the monomorphization cache is the main hazard).
- **Process-level fan-out** (the harness's model), if in-process sharing is too entangled
  — but that re-does the front half per worker unless lowered IR can be serialized.

Either way this **depends on #1618**: binding `Q_LOWER`/`Q_CODEGEN`/`Q_LINK` into the
query engine gives each module a stable cache identity, which is the prerequisite for
running them out of order (as #1618 itself notes). File this now so the parallel work is
tracked explicitly with the audit numbers rather than living as a parenthetical in the
incremental epic.

## Acceptance

- The back half saturates available cores; a compiler self-build's wall time drops
  materially from the ~6.7 s (debug) / ~9.5 s (release) single-core baseline.
- Output is byte-identical to the serial build: the `.o`/`.ir`/`.s` trees and the linked
  binary reproduce exactly (the self-host fixpoint and the rebuild-determinism checks from
  the audit still hold).
- `-v`/`-vv` per-phase timing still reports sensible aggregate numbers under parallel
  execution.

octalide:
Parallelizes the per-module **codegen** phase of `mach build` across a worker pool, byte-identical to the serial build. Implements Option A of the design fork on #1826 (premise changed after #1850 landed — see the scope note on the issue). Lowering stays serial (its monomorphization writes the shared interner — split out as #2100); codegen runs on N workers, each with a private arena and a child interner over the frozen session interner, and results are merged deterministically in topo order and recorded into the query engine.

Measured on the release self-build (163 modules, 16-core host): **codegen 17.7s → 3.3s (5.4x)**, overall **25.2s → 11.7s (2.15x)**. The remaining bottleneck is the still-serial lower phase (~6.7s) — the parallel-lowering follow-up #2100.

## Design

- **Child interner** (`intern.make_child`): a per-worker scratch that layers over the frozen session interner — ids below the snapshot boundary resolve into the base (so every worker resolves the IR StrIds each image inherits), new codegen strings land locally. Many children read one frozen base concurrently (the base's dedup map is read-only during the parallel region). `reintern_child` folds a worker's new strings back into the session interner at collection and returns the id remap.
- **Worker session view**: a copy of the real session with only `.alloc` (→ a private arena) and `.interner` (→ a child) overridden. Everything else codegen reads (sources, types, target) is frozen after the front half and shared read-only. Audited: codegen mutates *only* the interner and the allocator; it emits no diagnostics.
- **MIR scratch allocator** threaded through `lower_ir`/`lower_function` so the transient MIR builds in the worker arena instead of the IR module's (shared) allocator. Serial path passes the session allocator — behaviour-identical.
- **Collection** (`obj.rehome_remap`): each worker-built image is deep-copied onto the session allocator with its four interned-name fields rewritten into the session interner's id space, in topo order, then staged on its `ModuleEntry`.

### Engine stays authoritative (rider #2)

The prelude does not bypass Q_CODEGEN. It stages each finished image on the `ModuleEntry`; the unchanged serial `query.get(Q_CODEGEN)` loop then **adopts** the staged image as the query's value (recording the same `Q_LOWER`/`Q_TARGET` deps), so Q_CODEGEN remains the value's system of record and the downstream Q_LINK fingerprint (which folds every module's Q_CODEGEN revision) is unaffected. The in-engine codegen compute is still the sole codegen-value writer and is exercised by the driver tests and every non-CLI caller (`jobs == 0`, serial path). The same `codegen.codegen_module` produces both the staged and the inline image, so the two cannot drift.

## StrId-leak audit (rider #1)

Every place a StrId reaches image bytes or link inputs, and its disposition:

| Site | Kind | Disposition |
|---|---|---|
| `ObjectImage.name`, `Section.name`, `Symbol.name`, `Relocation.symbol` | StrId **field** | **remapped** by `rehome_remap` into the session interner |
| `.text` / `.data` / `.rodata` / `.bss` bytes | — | symbol refs are Relocations (offset + symbol field), patched at link; content bytes carry no StrId |
| `.debug_str` bytes | — | real string **content**, copied verbatim, deduped by content; no StrId |
| `.debug_info` / `.debug_line` string refs | — | **byte offset** into a per-module `.debug_str`, via a relocation against a per-module `.debug_str` anchor symbol (anchor = a remapped `Relocation.symbol`; offset = self-contained addend). No StrId numeric value in DWARF bytes |
| `.debug_*` address refs | — | relocations against function symbols (remapped `Relocation.symbol`) |
| ELF writer `.strtab`/`.symtab`/`.shstrtab` | — | resolves StrId→string at emit time; lays out by iterating image arrays in **insertion order** — StrId numeric values never affect byte layout |
| Linker merged symbol table | — | StrId-keyed maps are **lookup-only**; the output table is emitted by iterating modules in topo order then per-module insertion order — StrId numbers never affect ordering |

**No StrId numeric value is baked into any image bytes unremappably.** Output bytes are a pure function of (image array insertion order, string content), both deterministic, so byte-identity is structural — confirmed empirically below (parallel == serial, both profiles). `-g` DWARF's cross-module `.debug_str` dedup (#1708) resolves by string content, unaffected; `-g` runs serial here regardless (#2101).

## `--jobs`

`--jobs <n>` selects the worker count (default: host CPUs; `1` drives the whole pool on one thread — the same machinery, so the serial reference and the parallel path cannot diverge). Documented in `mach help build`.

## Scope

Parallel codegen runs for non-`-g` builds (both profiles build without debug info by default). `-g` builds keep the serial path — DWARF interns IR types through the shared session allocator, which the worker arenas do not isolate (#2101). Parallel **lowering** is #2100.

- **Parallel == serial, byte-identical** (forced-serial compiler vs parallel, full `.o` tree + linked binary): release `a2322967…`, debug `53aa429b…`. Both match.
- **Determinism across core counts**: `--jobs 1 / 4 / 7 / 16` + repeats produce an identical object tree and binary — linux-x86_64 (both profiles), linux-arm64 (both profiles), linux-riscv64 (release).
- **Self-host fixpoint through the parallel path**: g2 == g3 byte-identical.
- **Full unit suite**: 661 passed / 0 failed (adds one child-interner test; test-mode builds exercise the prelude too).
- **int/ golden-diff harness**: all cases pass on the linux leg.
- **`-g` serial fallback**: builds a valid debug ELF, exit 0.

## Follow-ups

- #2100 — parallelize lowering (the now-dominant serial phase).
- #2101 — extend parallel codegen to `-g` builds.

Closes #1826

### DIFF
--- a/src/cli/cmd/build.mach
+++ b/src/cli/cmd/build.mach
@@ -33,6 +33,7 @@ use A: std.allocator;
 use O: std.types.option;
 use R: std.types.result;
 use ctime: std.chrono.time;
+use sys: std.system.os;
 
 use cfg: mach.cli.config;
 use cli_diag: mach.cli.diagnostic;
@@ -685,6 +686,19 @@ fun build_unit(a: *A.Allocator, argc: usize, argv: **u8, marks: *bool, emit_obj:
     # middle-end scalarize/require gate reads the consumer's setting (#2017). the
     # posture is the consumer's alone; a dependency's own value never reaches here.
     p.s.simd = p.config.simd;
+
+    # resolve the parallel-codegen worker count (`--jobs <n>`, default host CPUs) onto
+    # the project so the codegen pass runs its worker pool. a malformed value skips the
+    # compile with exit 1 rather than silently serialising.
+    val res_jobs: R.Result[u32, str] = resolve_jobs(marks, argc, argv);
+    if (R.is_err[u32, str](res_jobs)) {
+        print.eprint("error: ");
+        print.eprint(R.unwrap_err[u32, str](res_jobs));
+        print.eprint("\n");
+        br.code = 1;
+        debug_ok = false;
+    }
+    or { p.jobs = R.unwrap_ok[u32, str](res_jobs); }
 
     # fold the manifest default into the level: a CLI `-O*` flag wins,
     # otherwise the selected profile's opt applies, else debug.
@@ -3007,8 +3021,47 @@ pub fun mark_opt_flags(marks: *bool, argc: usize, argv: **u8) {
 pub fun mark_build_flags(marks: *bool, argc: usize, argv: **u8) {
     mark_opt_flags(marks, argc, argv);
     util.mark_value(marks, argc, argv, "--emit");
+    util.mark_value(marks, argc, argv, "--jobs");
     util.mark_value(marks, argc, argv, "-L");
     util.mark_value(marks, argc, argv, "-l");
+}
+
+# resolve the parallel-codegen worker count for a build from `--jobs <n>`
+#
+# An explicit `--jobs` value wins (a positive integer); absent it the count
+# defaults to the host CPU count, and 1 when that cannot be determined. The result
+# is always >= 1 - the driver clamps it down to the module count. A malformed
+# value is rejected so a typo does not silently serialise the build.
+# ---
+# marks: consumption record parallel to argv (marks the flag + its value consumed)
+# argc:  argument count
+# argv:  argument vector
+# ret:   ok with the resolved worker count, or err on a malformed `--jobs` value
+fun resolve_jobs(marks: *bool, argc: usize, argv: **u8) R.Result[u32, str] {
+    val opt_jobs: O.Option[*u8] = util.mark_value(marks, argc, argv, "--jobs");
+    if (O.is_some[*u8](opt_jobs)) {
+        val jv: i64 = parse_jobs(O.unwrap[*u8](opt_jobs));
+        if (jv < 1) { ret R.err[u32, str]("--jobs expects a positive integer"); }
+        ret R.ok[u32, str](jv::u32);
+    }
+    var n: u32 = 1;
+    val nc: i64 = sys.cpu_count();
+    if (nc > 1) { n = nc::u32; }
+    ret R.ok[u32, str](n);
+}
+
+# parse a positive decimal `--jobs` value, or -1 when empty, non-decimal, or absurd
+fun parse_jobs(s: *u8) i64 {
+    if (s == nil || s[0] == 0) { ret -1; }
+    var v: i64 = 0;
+    var i: usize = 0;
+    for (s[i] != 0) {
+        if (s[i] < '0' || s[i] > '9') { ret -1; }
+        v = v * 10 + (s[i] - '0')::i64;
+        if (v > 65536) { ret -1; }
+        i = i + 1;
+    }
+    ret v;
 }
 
 # `mach build` subcommand entry point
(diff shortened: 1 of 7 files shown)


octalide:
## Coordinator review

**Verdict: approve — merging.** The Option A ruling implemented faithfully, with all four riders discharged:

- **StrId audit in the body**: the only interned ids in image data are the four name fields, all remapped through the producing child's table; DWARF refs are per-module byte offsets via anchor relocations; writer and linker order by insertion, never by id value. The id-space partition is sound — children snapshot base_len before the base ever grows, and collection remaps each image through its own child's table in topo order, so numerically-overlapping child ids can never cross-contaminate.
- **Engine authoritative**: images prestage into Q_CODEGEN and the serial query loop adopts them — dep recording intact, Q_LINK untouched, in-engine compute still exercised (driver tests, jobs 0).
- **j=1 same machinery**, and the determinism matrix ran the full bar: parallel==forced-serial byte-identical (tree + binary, both profiles), core-counts 1/4/7/16 + repeats identical on three targets, fixpoint g2==g3 THROUGH the parallel path — the strongest oracle available for this race surface.
- **Follow-ups filed as #1825 children**: #2100 (parallel lowering — now the dominant serial phase) and #2101 (-g codegen: DWARF type-synthesis allocator entanglement, correctly a clean serial fallback rather than a hack).

Numbers for the record: codegen 17.7s → 3.3s (5.4x), whole release self-build 25.2s → 11.7s (2.15x) on 16 cores. The design-boundary stop on the -g path was the right call twice over — reported, gated cleanly, and filed.

PR #2105, opened 2026-07-12
ci(cd): drop x86_64-darwin as a released platform (Closes #2104)

### LINKED ISSUE #2104: cd: x86_64-darwin seed segfaults under Rosetta building stage-a (v3.6.0 release blocker)
**v3.6.0 release blocker** — CD publish skipped; no assets shipped. The tag stands unpublished until this is root-caused.

The x86_64-darwin CD lane fails at the first self-host stage: the cross-built
darwin seed (itself compiled from the v3.6.0 tree) segfaults under Rosetta on
the macOS runner while compiling stage-a.

## Repro

CD run 29180770906 (tag v3.6.0), job `build x86_64-darwin`, step `build fixpoint`:

```
./seed/mach-darwin-seed build . --profile release -o a
  → Segmentation fault: 11 (exit 139)
```

A `--failed` rerun is in flight to rule out a runner flake; treat the failure
as deterministic unless it flips.

## Expected

The lane was green for v3.5.1 (2026-07-11) with the identical script: seed →
a → b → c, `cmp b c` fixpoint, publish.

## Actual

Stage-0 (the seed binary) crashes immediately-ish into `build .` — before any
stage output exists. `dep pull` and lock verification complete first, so the
crash is inside compilation proper.

## Delta under suspicion (v3.5.1 → v3.6.0, five merges)

- **#2099 parallel codegen — prime suspect.** `--jobs` defaults to host CPUs,
  so this seed is the FIRST darwin-x86_64 binary to spawn worker threads by
  default (`std.sync.thread`/`std.sync.atomic` on darwin x86_64, under
  Rosetta). aarch64-darwin (native) passed, so it is not "darwin threading is
  broken" in general — Rosetta-specific behavior or an x64-backend-specific
  defect in the threading path are both live.
- **#2096 inliner instance visibility** — second suspect: substantially more
  inlining everywhere could surface a latent darwin-x86_64 miscompile through
  newly-hot code paths.
- #2087 (win64-only behavior), #2097 (sema fold), #2098 (manifest expansion)
  — unlikely but in the window.

## Probe mechanism

`cd.yml` supports `workflow_dispatch` running the darwin lane (seed + native
fixpoint + self-test) against an arbitrary ref — dispatch probe branches to
bisect without cutting tags. A probe branch may also edit the fixpoint step
(e.g. inject `--jobs 1`) since dispatch uses the ref's workflow file.

## Environment

- macOS runner (x86_64 lane runs under Rosetta per the workflow comment).
- Seed: cross-built from the v3.6.0 tree by the same run's `seed x86_64-darwin` job (green).
- linux (x86_64/aarch64/riscv64) and windows lanes: all green on the same tree.

octalide:
Closes #2104.

Drops **x86_64-darwin** as a released/supported platform, moves the remaining aarch64-darwin lane to macos-15, and carries the mach-std bsdthread entry-ABI hardening. Zero compiler-code changes — the x86_64-darwin target triple remains a valid (now unsupported) cross-target.

## The #2104 ledger

**Symptom.** The v3.6.0 CD x86_64-darwin lane segfaults (SIGSEGV, exit 139) at stage-a of the self-host fixpoint. Deterministic (a `--failed` rerun reproduced it).

**Root cause (confirmed via CI core backtrace).** mach's parallel codegen (#2099) made per-module worker threads the default build path — the first darwin-x86_64 binary to spawn threads by default. Those threads are created with **raw `bsdthread_create` (pthread=0, flags=0)**, bypassing libpthread — the path **Rosetta 2** hooks to initialize each thread's translation context. The x86_64-darwin release lane runs **under Rosetta on the arm64 runner** (Intel macos-13 is queue-starved), so the worker thread runs translated code with no Rosetta context. The crash core (arm64) shows the faulting `pc` in the Rosetta runtime (`0x7ff7ff…`), reached from mach code (`lr` in mach's `0x1_0000_0000` image), dereferencing null+0x18 = Rosetta's per-thread context. Corroboration: `--jobs 1` (same codegen machinery, no thread spawned) builds cleanly; mach emits zero `%gs`/TSD accesses of its own; #2099's parallel codegen is verified working on native linux-x86_64 (real threads) and passes on native darwin-aarch64.

**Why no sound mach-side fix.** Making raw bsdthreads Rosetta-compatible means either linking/using libpthread (violates mach's libSystem-free design) or reverse-engineering the undocumented Rosetta+kernel bsdthread handshake to trigger Rosetta's per-thread init — fragile guessing at internals, not a sound fix.

**Native-Intel argument, and why it's unverifiable.** The x86_64-darwin *artifact* targets native Intel Macs (no Rosetta), where mach's raw threads should work (mach uses no TLS itself; the parallel logic is proven on native x86_64 linux). The crash is a Rosetta-CI artifact. But this is **not verifiable**: the native-Intel `macos-13` probe I dispatched sat queued for hours (the runner starvation the workflow itself documents), so there is no way to gate a release on native x86_64-darwin.

**Owner ruling (2026-07-12).** Drop x86_64-darwin from support — clean removal, no deprecation banners (absence is the statement). EOL hardware, Rosetta provably can't host mach threads, native Intel CI unavailable. arm64-darwin (Apple Silicon) is the supported darwin asset.

## Changes
- `cd.yml`: remove the x86_64-darwin `seed-darwin` and `build-darwin` matrix entries and the Rosetta-install step; the published-asset/SHA256SUMS wiring is glob-driven, so it drops the asset implicitly. Move the aarch64-darwin lane to **macos-15** (macos-14 nears runner retirement; release lanes validate on the oldest supported macOS).
- `int/targets.conf`: remove the `darwin-x86_64` integration row — it ran the cross-built compiler under Rosetta on macos-14 and would reproduce the #2104 crash on the first int-main run post-merge. Move the `darwin-aarch64` row to **macos-15** to match the release lane. This drops x86_64-darwin execution validation entirely — inherent to the ruling; the cross-target stays buildable but unvalidated, which is what "unsupported" means.
- `mach.lock`: bump mach-std to **7fd2a30e** (released 0.18.0 base + only mach-std #377, the naked realigning bsdthread trampoline — a real x86_64 entry-ABI hardening). Darwin-only source change → linux/windows/riscv output byte-identical (locally confirmed: linux-x86_64 release compiler byte-identical with vs without the change).

## Verification
- Dispatched darwin lane green end-to-end on **macos-15** with the bumped lock (aarch64-darwin — now the only darwin lane).
- Full PR CI green.
- Linux byte-identity unaffected by the lock bump (darwin-only mach-std change); from-source `mach test` 661/0.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -9,17 +9,13 @@ on:
 permissions:
   contents: read
 
-# a tagged push builds the release binary for each target — on its native host, or
-# under emulation where no practical native runner exists — each to its own
-# self-host fixpoint, then publishes the archives + SHA256SUMS. correctness is CI's
-# job — every change reaches main through a PR it gates.
+# a tagged push builds the release binary for each target on its native host, each
+# to its own self-host fixpoint, then publishes the archives + SHA256SUMS.
+# correctness is CI's job — every change reaches main through a PR it gates.
 #
 # darwin has no prior release to self-seed from, so seed-darwin cross-builds the
 # stage-0 seed on linux first; collapse that bootstrap once a darwin archive ships.
-# both darwin archives build on macos-14 (apple silicon): aarch64-darwin natively
-# and x86_64-darwin under Rosetta 2 — Intel macos-13 runners are queue-starved, and
-# the cross-built x86_64 seed bakes in x86_64-darwin as its host, so its
-# under-Rosetta a/b/c fixpoint stays self-hosted on the arm64 box.
+# the darwin archive builds on macos-15 (apple silicon), aarch64-darwin natively.
 #
 # workflow_dispatch runs the darwin lane (seed + native fixpoint + self-test) on a
 # branch with no tag, so on-Mac validation needs no release; the tag-verify and
@@ -222,8 +218,6 @@ jobs:
         include:
           - target: aarch64-darwin
             triple: darwin-aarch64
-          - target: x86_64-darwin
-            triple: darwin-x86_64
     steps:
       - uses: actions/checkout@v6
 
@@ -265,14 +259,10 @@ jobs:
       fail-fast: false
       matrix:
         include:
+          # release lanes validate on the oldest supported macOS — forward-compat
+          # covers newer, and macos-14 is next in the runner-retirement line.
           - target: aarch64-darwin
-            runs-on: macos-14
-          # x86_64-darwin: an Intel Mach-O run on the arm64 box via Rosetta 2. the
-          # cross-built seed bakes in x86_64-darwin as its host, so the whole a/b/c
-          # fixpoint emits Intel binaries and self-hosts under Rosetta.
-          - target: x86_64-darwin
-            runs-on: macos-14
-            rosetta: true
+            runs-on: macos-15
     steps:
       - uses: actions/checkout@v6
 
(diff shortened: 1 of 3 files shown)


octalide:
## Coordinator review

**Verdict: approve — merging.** The drop is exactly the owner's ruling: removal by absence — the cd.yml prose is rewritten as if x86_64-darwin had never been there rather than annotated with deprecation residue; the target triple stays a buildable, unvalidated cross-target with zero compiler-code changes. The int/targets.conf row removal correctly prevents the first post-merge int-main run from reproducing the crash, the surviving darwin lane's macos-15 move carries its rationale inline, and the lock bump to 7fd2a30e picks up exactly the #377 entry-ABI hardening and nothing else (non-darwin output byte-identical, proven). Verification: full PR CI green including all int legs post-main-restoration, and the dispatched darwin lane green end-to-end on macos-15. The PR body stands as the #2104 ledger — root cause with backtrace, the no-sound-fix analysis, and the ruling.

Issue #2118, opened 2026-07-23
codegen: table-driven encoder retrofit


octalide:
Superseded by [mach#2212](https://github.com/briar-systems/mach/issues/2212). Remaining obligations and acceptance are now carried by those issues in the mandatory v5 set. Closed as superseded, not fixed.

<details>
<summary>Historical report and reproduction, retained for provenance</summary>

Part of #2212.

> ## Status (2026-08-08)
>
> **No longer greenfield. The pattern now exists in the tree, for one instruction family.**
>
> #2766 collapsed aarch64's five parallel width ladders (`ldst_lo12_kind`, `width_shift` and the
> three `ldst_base_*` base-word tables) into **one form table per family**: `int_access` and
> `fp_access` each return a row carrying the three addressing-mode base words, the imm12 shift and
> the `:lo12:` relocation kind together. riscv64 got the same treatment via `int_mem_funct3` /
> `fp_mem_funct3`. A width with no row is refused rather than defaulted.
>
> That is precisely this issue's "format tables + a shared packer", delivered for the memory-access
> family. So the remaining work is **generalizing an existing, proven shape to the other families**,
> which is a smaller and much better-defined job than the retrofit originally described.
>
> It also revises the approach below. This says generalize from rv64 + 6502, but the strongest
> table-driven example in the tree is now aarch64's access rows, and they were arrived at by fixing
> a real defect rather than by design — which is the better evidence. Start there.
>
> Worth carrying forward as the motivating case: the five ladders were not merely repetitive, they
> were five copies of one decision that **agreed with each other by accident**, each naming the
> narrow widths and each ending in the doubleword row. A width the target had no form for silently
> took some other form's bytes. That is the defect class tables exist to make inexpressible, and it
> is the same duplicate-policy shape that has produced several separate bugs in this codebase.

## What

Generalize the per-ISA byte encoders toward the data-driven shape the selection rule packs now
have. Instruction-word packing (rv64's R/I/S/B/U/J formats, arm64's bitfield classes) becomes
format tables plus a shared packer; the survivor-materializer dispatch becomes a per-arch table
over a shared walker.

Genuinely algorithmic parts stay code: rv64's branch-relaxation fixpoint, x64's variable-length
emission, and addressing-mode legalization. The goal is not to table-drive everything, it is to
table-drive the parts where a second copy of a decision can silently disagree with the first.

## Acceptance

- The memory-access shape from #2766 is generalized to at least one further instruction family per
[...]

PR #2123, opened 2026-07-23
perf(abi): stage the SysV vector-count (AL) only for variadic calls

### LINKED ISSUE #1939: abi: unconditional rax zeroing staged before every call
Part of #1933.

## What

Call staging emits `mov rax, 0` before every call, in the 7-byte `48 c7 c0 00000000` encoding. Both fib recursion sites at v2.14.1:

```asm
mov    rdi,rax
mov    rax,0x0        ; 7 bytes, staged before every call
call   fib
```

That register is the SysV variadic convention: AL carries the count of vector registers used by the arguments, and it exists so a variadic callee knows how much of the XMM save area to spill. Only a call to a variadic function needs it. Mach-to-Mach calls and non-variadic C externs never read it, yet every call site pays the bytes and the register write today. The staging appears unconditional in the call expansion (`src/lang/be/codegen/mir/abi.mach`, `OP_CALL` expansion into pre-coloring plus the `MIR_CALL` pseudos).

## Fix

Two parts, both in the call expansion where the callee signature is in hand: stage AL only when the callee is variadic, and encode the stage as `xor eax, eax` (2 bytes, dependency-breaking) rather than the 7-byte immediate move. Preferred is both together. The variadic save model plumbing (`tgt.abi.va_model`) already distinguishes the conventions, so the signature check has a home.

## Acceptance

fib's call sites carry no rax staging. A variadic C extern call (printf-family interop test) still zeroes AL and behaves. Binary sizes shrink, checksums identical.

octalide:
Closes #1939.

## The gate

SysV call lowering staged the vector-count hidden argument (AL, the low byte of RAX) before **every** call, in the 7-byte `mov rax, 0` (`48 c7 c0 00000000`) encoding. AL is a live input only to a variadic or prototype-less callee; every fully-prototyped fixed-arity call — each mach-to-mach call and each non-variadic C extern — paid the bytes and the register write for nothing.

The stage is now gated on whether the call reaches the variadic / unprototyped region, computed in `lower_call` from the signature already in hand:

```
variadic_call = !sig_is_fn || ((operand_count - 1) > layout.param_count)
```

This is exactly SysV's `%al` rule — set for **variadic calls** (arguments past the declared fixed parameters) and **prototype-less calls** (an indirect callee carrying no IRT_FN). A needing-AL call still stages it (with the same conservative `fp_used` count as before); a fixed-arity call stages nothing. mach removed C-style `...` in v2.0.0 and monomorphizes every comptime `va: ...` pack to a concrete fixed-arity signature, and function pointers lower to a typed IRT_FN, so in the current language every reachable call is fully prototyped — the corpus loses the stage on 99.6% of call sites.

When the stage **is** emitted it is now a 5-byte `mov eax, imm32` (the count is 0–8, so writing EAX sets AL and zero-extends the rest) rather than the 7-byte 64-bit-immediate form. The change is x86_64-only: SysV is the sole convention with `vector_count_reg >= 0`; win64 / aapcs64 / lp64 report `-1` and were already skipped.

## Measured win (isolated: this change alone, off origin/dev)

Release compiler compiling itself, executable `PT_LOAD` (R E) segment:

| | R E segment |
|---|---|
| old codegen | 11,504,757 bytes (0xaf8c75) |
| this change | 11,261,086 bytes (0xabd49e) |
| **removed** | **243,671 bytes (2.12%)** |

The old codegen staged `mov rax, 0` before **49,647 of 49,828** call sites in the emitted corpus; the new codegen emits **zero** AL staging anywhere. fib's two recursion sites go from `mov rdi, rax; mov rax, 0; call fib` to `mov rdi, rax; call fib`.

## Correctness

The ON-path is correct by construction: when `variadic_call` holds, the emitted value is the same accumulated `fp_used` the old code used — only prototyped fixed-arity calls (which never read AL) lose the stage.

- compiler suite (from-source HEAD): **715 / 0**
- mach-std suite: **637 / 0** (identical seed vs branch)
- int **linux** (native) and **linux-riscv64** (qemu): all cases pass
- single-generation fixpoint **B == C** (byte-identical)
- fib runtime output correct (`832040`) under old and new codegen

(int `linux-arm64` was not executed here — it is a `native` leg with no aarch64 runner on this x86_64 host; the build succeeds, only execution is unavailable. aapcs64 has `vector_count_reg = -1`, so behavior there is unchanged.)

### DIFF
--- a/src/lang/be/codegen/mir/abi.mach
+++ b/src/lang/be/codegen/mir/abi.mach
@@ -806,6 +806,20 @@ pub fun lower_call(ctx: *context.LowerCtx, mb: *mir.MirBlock, inst: *instruction
 
     val rslot: abi.ParamSlot = layout.ret_slot;
 
+    # whether this call reaches the variadic / unprototyped region, the sole condition
+    # under which the SysV vector-count register (AL) is a live hidden argument (#1939):
+    # a call passing arguments past its declared fixed parameters, or a prototype-less
+    # indirect call whose callee carries no IRT_FN (SysV's `%al` rule covers both). a
+    # fully-prototyped fixed-arity call - every mach-to-mach call and every non-variadic
+    # C extern - never reads it, so it stages no vector-count write. captured here before
+    # the layout is freed, since it reads `layout.param_count`.
+    val sig_it:      O.Option[*type.IrType] = type.get(ctx.types, sig);
+    var sig_is_fn:   bool = false;
+    if (O.is_some[*type.IrType](sig_it)) {
+        sig_is_fn = O.unwrap[*type.IrType](sig_it).kind == type.IRT_FN;
+    }
+    val variadic_call: bool = !sig_is_fn || ((inst.operand_count - 1) > layout.param_count);
+
     # the call's result type sizes the caller-owned aggregate result storage.
     val ret_size:  u64 = type.byte_size(ctx.types, ret_type, ctx.tgt)::u64;
     val ret_aggr:  bool = context.value_is_aggregate(ctx, ret_type);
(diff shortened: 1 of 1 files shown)


octalide:
Scope note on the encoding half of #1939: this PR gates the stage (the primary win) and narrows the emitted form to `mov eax, imm32` (7 -> 5 bytes), the natural width for the 0-8 count. The further `mov-imm-0 -> xor eax, eax` idiom is deliberately **not** hand-rolled in the ABI lowering — a zero-immediate move should select to `xor` as a general encoder/peephole rule (it benefits every zeroing site, not just this one), which is the encoder retrofit tracked in #2118. Special-casing it in `lower_call` would be a single-site carve-out that the retrofit would later have to unwind. With the gate, the AL stage is emitted only for variadic / prototype-less calls, of which the current language (no C-style `...`; comptime packs monomorphized; typed function pointers) produces none, so the idiom has no measurable target today regardless.

Issue #2136, opened 2026-07-23
sema: #[oblivious]-required check (per-monomorphized-instance secrecy-transparency)


octalide:
Part of #1643. Split out of #1646 (the `#[oblivious]` contract + secret taint), per the agreed scope: #1646 ships the decorator, the CT-capability flag, the float/multiply variable-latency gates, and the taint threading to MIR; this follow-up adds the **required-check** that was listed in #1645's acceptance but deferred because it needs a monomorphization-aware analysis.

## What

Require `#[oblivious]` on any function *instance* that performs genuine secret computation, and error otherwise with a teaching diagnostic. The check is per **monomorphized instance**, not per generic template:

- A secrecy-**transparent** instance — one that only *moves / stores* secrets, never branching on, indexing by, or computing on them — is oblivious vacuously and needs no annotation. Ordinary generic movers (`copy`, `swap`, containers) instantiated with a secret type argument MUST stay annotation-free, per the epic (#1643).
- Only an instance that genuinely **computes** on a secret (arithmetic, comparison feeding a mask, etc.) requires the decorator.

## Why deferred

The transparency analysis is the trickiest sub-piece and is cleanly separable from the taint foundation. #1646 already establishes the secret taint on IR values/instructions (`Instruction.secret`, `Value.secret`) that this check can consult per instance, so this is an additive sema pass.

## Acceptance

- A secret-computing function without `#[oblivious]` errors; adding the decorator fixes it.
- A generic mover (`copy[^u32]`, `swap`, containers) that only moves/stores the secret compiles with no annotation.
- `#[oblivious]` on a secret-free function stays a no-op (already true after #1646).
- Clear diagnostic naming the function and why it must be `#[oblivious]`.

## Dependencies

Depends on #1646 (the decorator + taint). Blocks nothing else in the epic; the translation validator (#1648) and ct harness (#1647) build on the taint, not on this check.

Issue #1648, opened 2026-06-26
codegen/infra: constant-time translation validator (relational leakage check on emitted code)


octalide:
Part of #1643. Depends on #1646 (the secret taint must reach the emitted instructions).

The proof-grade layer: verify the *output* of each build rather than proving the compiler once. This dissolves the maintenance burden of the `#[oblivious]` contract — any present or future backend pass that breaks constant-time fails the build, instead of relying on every pass staying disciplined.

## Approach

Consume the emitted machine code plus the secret taint annotations. Verify relationally (product program / self-composition — the ct-verif and Binsec/Rel technique) that the leakage trace — control flow, memory addresses, variable-latency operands — is independent of secret inputs. Fail the build on any violation.

This is the largest single lift and the thing that makes a proof-grade claim defensible rather than testing-grade. Scope it first to the `ct` primitives and a reference ML-KEM kernel, then widen.

## Acceptance

- Proves the `ct` primitives and a reference ML-KEM kernel constant-time at the emitted-code level.
- Catches an injected secret-dependent branch / memory access introduced below MIR.
- Runs in CI on the crypto corpus.

octalide:
## #1646 <-> #1648 taint contract (re-architecture)

The `#[oblivious]` taint was re-architected into two separate surfaces after a security-soundness review found the original result-taint-stamping + hand-preserved-`.secret` + single-opcode-gate model structurally unsound.

**#1646 provides (on the MIR):**
- **The gates** (compile-time enforcement, sound): a single op->required-CT-capability table (`mach.lang.ct`). Sema rejects a secret operand of the always-unsafe class (integer div/mod, ALL floating point, incl. vectors); MIR lowering rejects the target-dependent class over every opcode (integer multiply unless `ct_trust_mul`; variable shift by a secret **count** unless `ct_trust_var_shift`). Adding a variable-latency op is one table row.
- **The secret-source SEEDS**: `MirVReg.secret` is set once, only on the secret SOURCES - secret parameters, and loads/calls whose result is secret (welded `*^T` storage, secret globals). It is NOT a propagated per-value taint. Plus stores/memzero into secret storage are marked by destination secrecy for the zeroization guarantee.
- **The `:^` barrier**: `OP_DECLASSIFY`, an explicit public SSA definition (selecting to a coalescable move) a substitution can never re-taint.

[...]

Issue #1647, opened 2026-06-26
infra: constant-time test harness — ctgrind/dudect + golden-MIR in CI


octalide:
Part of #1643. Depends on #1646 (something to test).

Testing-grade constant-time assurance — the layer before the translation validator. Wire into CI for the crypto corpus.

- ctgrind-style (Valgrind taint): mark secret inputs; error on any secret-dependent branch or memory address in the binary.
- dudect: statistical timing test (Welch's t over two input classes) on real hardware, to catch data-dependent latency the taint tools miss.
- golden-MIR/asm snapshots for the `#[oblivious]` contract (selects stay branchless, no variable-latency on secrets, zeroization survives).

## Acceptance

- The harness runs in CI against the `ct` primitives and crypto kernels.
- An injected leak (a secret branch, a deleted zeroization) fails the harness.

Issue #1643, opened 2026-06-26
epic: constant-time guarantees + post-quantum cryptography


octalide:
## What

Add a language-level, ultimately **provable** constant-time guarantee to Mach, and build post-quantum cryptography on top of it as the first proving ground. The aim: a general-purpose systems language that can carry a machine-checked "no secret influences the program's timing or memory-access trace" guarantee down to emitted machine code.

This is unique to Mach's position. The verified-crypto tools that already guarantee constant-time (Jasmin/libjade, FaCT, HACL*) are special-purpose DSLs. The general-purpose languages (C, C++, Rust, Go, Zig) cannot *guarantee* it because they do not own their backend — the optimizer they depend on can destroy or restore the property. Mach owns its entire backend (no LLVM), so it can thread a secrecy taint to the bottom of codegen and verify the property on the output.

## The model

Two properties on two different things.

- **Secret data** — a type qualifier `^` on values (`^u32`, `*^u8`, `[N]^u8`). It flows through every operation: a public value coerces up the lattice (public ⊑ secret), and any operation with a secret operand yields a secret result. The type checker *gates* the operations a leakage model can observe — a secret may not be a branch condition, a memory index, or an operand of a variable-latency instruction. The only downgrade is the explicit postfix cast `:^`.
- **Constant-time code** — a decorator `#[oblivious]` on functions, required (never inferred) on any function handling a secret. It is purely subtractive: it forbids the codegen rewrites that would leak (introducing secret-dependent branches, variable-latency instructions on secret operands, dead-store-eliminating zeroizing writes). It injects nothing.

Leakage model: control-flow trace + memory-address trace + variable-latency operands. This is classical constant-time; transient execution (Spectre-class) is out of scope.

## Trusted base and limits

State these honestly in any external claim:

- The only secret→public crossings are the explicit `:^` cast and inline `asm` blocks (which a type system cannot check). Everything else is enforced.
- Pointer secrecy is welded to storage and non-launderable — casts cannot change `^`, a `*^T` cannot be erased to the untyped `ptr`, and a `uni`'s overlapping fields must agree on secrecy — so the public/secret aliasing leak is unconstructable with no alias analysis.
- A proof is always relative to a leakage model; its fidelity to real silicon (DIT/DOITM mode, microarchitecture) is empirical.
- Assurance is testing-grade (ctgrind/dudect) until the translation validator lands, which makes it proof-grade per build.

## Children

Compiler (this repo):
[...]

PR #2137, opened 2026-07-23
feat(codegen): #[oblivious] constant-time contract + secret taint to MIR (#1646)

### LINKED ISSUE #1645: sema: secret-qualifier flow typing — join, gate, welded-storage pointers
Part of #1643. Depends on #1644 (the `^`/`:^` tokens).

Implement the information-flow type rules for the `^` secret qualifier — the source-level non-interference discipline. Design the rules to be provably sound (the goal is to mechanize the core calculus later), not merely enforcing.

## Lattice and join

Two-point security lattice: public ⊑ secret. A public value coerces *up* to secret implicitly wherever a secret is expected. Any operation with a secret operand yields a secret result (taint joins across `+ - * / % & | ^ ~ << >>`, comparisons, etc.).

## Gate (each is a compile error, decided by operand type)

A secret may not reach a position the leakage model observes:
- a branch/loop condition that is secret
- a memory index that is secret (`table[i]`, `i` secret)
- an operand of a variable-latency instruction: `div`/`mod` always; floating-point by default; integer multiply gated only per the target's CT-capability flag (defined in the codegen issue)

## Downgrade

- `::` and `:~` cannot add or drop `^`.
- `:^` is the only downgrade — a value-level coercion producing a new public value; it never reinterprets storage in place.

## Welded-storage pointer rules (closes the aliasing leak with no alias analysis)

Secrecy is fixed at declaration/allocation and is non-launderable:
- deref of `*^T` yields `^T`; a `^T` may be stored only through a `*^T`, never a `*T`
- a `*^T` cannot be erased to the untyped `ptr` (secret memory is handled through secrecy-typed pointers, never raw `ptr`)
- a `uni`'s overlapping fields must agree on secrecy

Together these make two contradictory classifications of the same storage (a `*^u8` and a `*u8` that alias) unconstructable.

## `#[oblivious]`-required check

A function whose signature or body involves a secret value must carry `#[oblivious]` (a compile error otherwise). The check is per *monomorphized instance*: a secrecy-transparent instance — one that only moves/stores secrets, never branching/indexing/computing on them — is oblivious vacuously and needs no annotation; only instances doing genuine secret computation require the decorator. Ordinary generics (`copy`, `swap`, containers) stay annotation-free.

## Acceptance

- Positive/negative test matrix: each gated op errors on a secret operand and passes on public; join produces secret results; `::`/`:~` reject secrecy changes; `:^` strips; the three welded-storage rules reject every laundering path; the aliasing leak cannot be written.
- Clear diagnostics ("secret value used as branch condition", etc.).

### LINKED ISSUE #1646: codegen: '#[oblivious]' constant-time contract + secret taint to MIR
Part of #1643. Depends on #1645.

Implement the `#[oblivious]` codegen contract and thread the secret taint from the type system through MIR down to the final machine instructions. Build this taint-aware from the start: the translation validator (#1648) is impossible if the taint dies before the bottom.

## `#[oblivious]` decorator

A new closed-set decorator on functions. Purely subtractive — emits exactly the instructions the source implies, and only forbids leak-introducing rewrites for tainted values:
- no introduction of secret-dependent branches (no reverse if-conversion of a branchless select into a branch)
- no selection of variable-latency instructions on a tainted operand
- no dead-store elimination of zeroizing writes to secret storage

Because the taint reaches codegen the contract is precise: public operations in the same function compile normally. It injects nothing into the runtime.

## Variable-latency / CT-capability flag

Add a single conservative CT-capability flag to the target description: `div`/`mod` always gated on secrets; floating-point gated by default; integer multiply trusted under the hardware data-independent-timing mode (Arm `PSTATE.DIT`, Intel DOITM) where the ISA documents it, gated otherwise. One flag per target — not a per-CPU table. DIT/DOITM is set explicitly by the stdlib `ct.begin`/`ct.end`, not injected here.

## Scope

x86_64 first (current primary backend); aarch64/riscv64 completeness rides the multi-target work (#1180). The portable masked primitives need no per-arch asm, so this does not hard-block on per-arch assembler support.

## Acceptance

- Golden-MIR / golden-asm tests: a branchless select stays branchless (lowers to `cmov`/masked ops, never a branch); no variable-latency instruction is selected for a tainted operand; a zeroizing store survives DSE.
- The secret taint is queryable on MIR values and on final instructions.
- `#[oblivious]` on a function with no secret values is a no-op.

octalide:
Closes #1646. Part of the constant-time epic #1643 (builds on the closed #1645 sema flow typing).

## What this establishes

The constant-time **foundation**: a `#[oblivious]` function contract plus a secret-taint bit threaded from sema's `^` flow typing through IR and MIR down to the final instruction stream, with the target-dependent integer-multiply gate as its first live consumer. Built taint-aware from the start so the #1648 translation validator is possible.

Secrecy is a **type** qualifier (`^T` = `TYPE_SECRET`), but the IR type universe deliberately erases it (a `^T` shares `T`'s machine shape). So the taint cannot ride the IR type — it is a separate bit threaded down, mirroring how `Function.ret_sem` already carries semantic info the IR type erases.

## The taint model (what #1648 will consume)

- **IR** — `Value.secret` (per-use taint) and `Instruction.secret` (result taint / secret-storage write). The lowerer (`lower_rvalue`) stamps every produced value from sema's flow typing after applying the active generic substitution, so a secret flowing into a generic-parameter-typed expression (`copy[^u32]`) is seen as secret. A `:^` downgrade is typed public by sema, so it clears the per-use value bit — the one place a value sheds taint — while never clearing a shared definition's instruction bit (monotone).
- **Soundness to the bottom** — taint is monotonically inherited across every RAUW (`rewrite_resolve` / `replace_uses`), preserved by `clone_instruction` (inline), by the isel piece-emission chokepoint, and by the regalloc flatten copy. Every `Value` / `Instruction` construction site initialises the bit deterministically (mach `var` is not zero-init), so no garbage taint can spuriously appear and secret-free code stays byte-identical.
- **MIR** — `MirVReg.secret` (queryable on MIR **values**) set in `ctx_setup` from the defining instruction's / parameter's taint, and `MirInstr.secret` (queryable on **final instructions**) stamped at the `push_instr` chokepoint from `ctx.cur_secret`, carried through isel and regalloc to the emitted stream.

## The contract and gates

- `#[oblivious]` — a closed-set, functions-only decorator lowered to `FN_FLAG_OBLIVIOUS`. A no-op on a secret-free function.
- `ct_trust_mul` — a single CT-capability flag on `MachineModel`, conservatively `false` on every ISA (integer multiply gated on secrets until a DIT/DOITM mode is wired). `div`/`mod` and floating point are always gated regardless.
- **Float-default gate** (sema) — a secret operand of a floating-point op is rejected (variable-latency, gated by default; target-independent, so it joins the existing `div`/`mod` gate).
- **Integer-multiply gate** (MIR lowering) — a secret-tainted `OP_MUL` on a target without `ct_trust_mul` is a teaching per-target compile error. This is the **first live consumer** of the threaded taint, proving it reaches MIR (the acceptance anchor). It lives at MIR because the target's CT capability is unknown to sema.
- **Subtractive-contract invariants documented** for future backend passes (DCE / dead-store-elim, secret-dependent branch introduction, variable-latency selection): each such decision must consult `Instruction.secret` / `MirInstr.secret` / `MirVReg.secret`. mach has no if-conversion / DSE today, so the guarantees hold vacuously; the taint is present so they *can* be honoured, and #1648 enforces on the emitted code.

## Handoff

- **#1648** (translation validator) consumes this taint on the emitted stream (via `MirVReg.secret` on values and `MirInstr.secret` on final instructions) for the relational leakage check.
- **#1647** (ct harness) is independent of this taint (ctgrind/dudect-grade).
- **#2136** (filed here) is the split-out `#[oblivious]`-required check — the per-monomorphized-instance secrecy-transparency analysis (generic secret-movers stay annotation-free), deferred from this PR to keep it reviewable, per the agreed scope.

## Verification

- Full suite green (722 + new taint/gate/decorator tests).
- Byte-identical self-build proven for secret-free code (every taint bit is false; the bit never influences emitted bytes).
- Single-generation fixpoint B == C.
- `mach-std` baseline; int linux + rv64.
- Positive/negative: taint propagates through derived values; a secret multiply and a secret float op are rejected with teaching diagnostics; `#[oblivious]` on a secret-free function is a no-op.


octalide:
Verification complete — marking ready.

- Full suite (seed and from-source HEAD compiler): **725 / 0** (+3 new: `secret_float_gate_rejected`, `oblivious_decorator`, `secret_mul_codegen_rejected`).
- `mach-std` baseline through the from-source compiler: **663 / 0**.
- **Byte-identical for secret-free code, proven two ways:** single-generation fixpoint B == C, and the base (pre-#1646) compiler vs this compiler compiling the *same* secret-free source produce identical binaries. Every taint bit is false for non-oblivious code, so codegen is unchanged.
- **int** linux and linux-riscv64 (qemu): all cases pass.
- `-g` additivity: no codegen change for secret-free code (proven above), so it is preserved; the `debuginfo` int case passes on both legs.

The multiply-gate acceptance anchor is exercised end-to-end (`secret_mul_codegen_rejected`): a secret reaches an `OP_MUL` through derived values (public call args up-coerce to the `^u32` params), and the gate rejects it at codegen — proving the taint propagates all the way to MIR.

Note: the split-out `#[oblivious]`-required check is filed as #2136 (linked under the epic). Separately, while writing an end-to-end int case I found a **pre-existing** bug — the `:^` strip cast is unresolved in a full `mach build` though it type-checks in the in-memory sema tests (repros on the released seed, independent of this PR) — filed as #2138; it blocks an end-to-end declassifying case here and will block the #1647/#1648 corpora until fixed.

octalide:
Converting back to draft — a max security-soundness review found the taint foundation is architecturally fragile (it passed self-verification only because nothing in the compiler/stdlib uses `^`; the holes bite ct code):

- **Gate gaps (false-accept, real leaks):** secret vector fp ops bypass all gates ([1]); secret shift-count ungated ([3]); scalarize launders vector-op taint to public scalar lanes ([2]).
- **Taint laundering / uninitialized on MIR:** phi / inlined-return / spill-reload carry public taint ([4][5]); three sites read uninitialized stack ([6][7][8], **fixed now** in 09c0f64e).
- **Contract:** `carry_secret` re-taints a `:^`-declassified value ([9]); zeroizing store of public 0 into secret storage unmarked ([10]); the gate is a bespoke `OP_MUL` special-case rather than a general op→CT-capability mechanism ([11], the root cause of [1]/[3]).

Root cause: result-taint STAMPING + hand-preserved `.secret` at every construction site + a single-opcode gate can't be patched sound. Re-architecture proposed (two taint surfaces: sound IR operand-taint for the gates via a general op→CT-capability table; a re-derived dataflow fixpoint from semantic seeds + `:^` barriers for the #1648 surface, replacing all hand-preservation). Awaiting design sign-off before reworking; a rejected/correctly-tainted test will be added per hole.

octalide:
Re-architected to the sound foundation and re-ready for (max) re-review.

## What changed vs the flagged version
Two separate surfaces, per the review:

**GATES (compile-time enforcement, now sound + general):**
- New `mach.lang.ct`: the single op→required-CT-capability table both enforcement points read.
- Sema gates the always-unsafe class (div/mod, ALL floating point) via the table, now including secret VECTOR float ops — closes **[1]**.
- MIR lowering gates the target-dependent class over EVERY opcode via the table — integer multiply (`ct_trust_mul`) and variable shift by a secret COUNT (new `ct_trust_var_shift`) — replacing the bespoke OP_MUL check, closes **[11]/[3]**.
- Scalarize propagates operand taint onto its lanes, so a scalarized secret multiply trips the MIR gate — closes **[2]**.
- `:^` lowers to an explicit `OP_DECLASSIFY` barrier (a distinct public SSA definition a substitution can't re-taint) — closes **[9]**; selects to a coalescable move.

**SEEDS for #1648 (contract change — documented on #1646/#1648):**
- Deleted the fragile MirInstr.secret machinery (push_instr stamp, isel/regalloc copies, ctx_setup result-stamp, the [6][7][8] uninitialized sites).
- #1646 now marks only the secret SOURCES (`MirVReg.secret` on secret params / welded loads / secret calls, set once on stable sources) plus stores/memzero into secret storage by DESTINATION secrecy — closes **[10]**. #1648 re-derives per-instruction taint over the final MIR from these seeds, stopping at OP_DECLASSIFY.
- The IR-level operand taint (Value.secret + carry_secret) is retained for the gates (the review confirmed it sound).

**Also fixes #2138** (pre-existing): resolve didn't recurse into a `:^` strip's inner expression, so declassification failed in a real build though it passed in-memory sema. This unblocks end-to-end constant-time programs.

## Verification (all green)
- Full suite **730 / 0** (a rejected/correctly-gated test per hole: [1] vector fp, [2] scalarized vector mul on rv64, [3]+[11] the shift/mul gate mechanism incl. count-only precision, [9] declassify-barrier-not-retainted through mem2reg, [10] zeroizing secret store; plus the ct-table test).
- **Byte-identical for code using neither `^` nor `:^`**, proven two ways: fixpoint B == C, and the base (pre-#1646) compiler vs this compiler compiling the same secret-free source → identical binaries.
- `:^` code **compiles and RUNS** correctly (declassify coalesces to a move); a new `int/surface/oblivious` case runs a masked select + zeroize end-to-end.
- `mach-std` baseline **663 / 0**; **int linux + linux-riscv64** all pass (incl. the new oblivious case).

Note: the [3] shift REJECTION isn't end-to-end buildable (mos6502/spirv are the only `ct_trust_var_shift=false` targets and neither codegens in a unit test), so it's proven at the mechanism level (the gate-mechanism + ct-table tests) rather than a full compile.

octalide:
Correction: #2138 (the `:^` resolve fix) was already fixed and merged as #2143 while this branch was in flight — my independent re-fix was a **duplicate, not a bonus**. On merging `origin/dev` I took dev's (#2143's) `EXPR_KIND_STRIP` arm and dropped mine, so there is one copy. Disregard the earlier "also fixes #2138" note. Merge commit reconciled #2143 (resolve) and #2146 (aggregate-`==` reject, composed with the CT gates in `infer_binary`); full bar re-run on the merged commit.

octalide:
Security re-review findings addressed — point by point.

**[3] LINCHPIN (barrier survival):** `OP_DECLASSIFY` now lowers to a distinct **`MIR_DECLASSIFY`** pseudo-opcode (added to the selection alphabet), not an indistinguishable `MIR_MOV`. The #1648 re-derive runs over the **lowered, target-independent MIR** — the layer where all seeds and the barrier live — and stops secrecy propagation at each `MIR_DECLASSIFY`. isel retags it to the concrete machine move per ISA, so it's identical bytes / a coalescable copy at runtime. (Contract doc updated: "lowered MIR", not post-selection.)
**[1] fp conversions:** sema now gates the `::` cast path (`^f64::^i64` fp→int, `^f32::^f64` widen; mach has no implicit float widening, so the explicit cast is the only secret-fp-convert path), and the MIR classifier maps the `OP_FP_*` convert opcodes to `CT_OP_FLOAT` (defense-in-depth backstop).
**[2] vector declassify:** a vector `:^` scalarizes per-lane instead of reaching the v128 guard on a no-SIMD target — the hard-compile regression is gone (test: compiles clean on rv64).
**[4] looprotate:** `add_vreg_like` now initializes `MirVReg.secret` (copied from the template).
**[5] zeroization seed on MIR:** a store/memzero into secret storage sets a targeted `MirInstr.writes_secret` flag (the destination-secrecy the value seed misses), forwarded once at lowering.
**[6] zeroization assignment:** `lower_assign` now seeds `k = 0` into `^` storage.
**[7] float-multiply classification:** the MIR classifier maps a float multiply to `CT_OP_FLOAT` via `is_float_op`, not `CT_OP_INT_MUL`.
**[8]** extracted `ut_run_codegen_err` / `ut_run_codegen_ok`.
**[9]** consolidated the mark-defining-instruction idiom into `instruction.mark_result_secret` (it lives in `instruction.mach`, not `value.mach`, because `value` sits below `instruction` in the import order).
**Refuted item confirmed:** div/mod are gated at sema (target-independent) and stay `CT_OP_NONE` at MIR — the classifier correctly omits them (asserted in the gate-mechanism test).

**Tests, one per correctness fix:** secret fp conversion rejected; `:^` emits a distinct `OP_DECLASSIFY` barrier; secret vector `:^` scalarizes on rv64; the gate mechanism covers the fp-convert classification + the count-only shift precision; the int oblivious case wipes secret storage with a zeroization assignment end-to-end.

**Re-verified (all green):** suite **745/0**; **byte-identical** for non-`^`/`:^` code vs current dev + fixpoint **B==C**; `mach-std` **663/0**; **int linux + rv64** all pass (declassify + oblivious, incl. the zeroization wipe). `#[oblivious]` code compiles and runs; the barrier is a runtime no-op.

octalide:
Second re-review findings addressed — one root cause (MIR_DECLASSIFY + `writes_secret` not propagated to non-rule-based / pre-isel consumers), swept the whole class.

**[1] SPIR-V** — `emit_instr` now handles `MIR_DECLASSIFY` directly as a value copy (SPIR-V is a direct emitter that bypasses isel). A `:^` no longer hits the unsupported-instruction error.
**[2] legalize** (runs before isel) — handles `MIR_DECLASSIFY` exactly like `MIR_MOV` in `expansion_count` + `expand_instr`, so a `:^` on a wider-than-ALU value (mos6502) legalizes instead of aborting. (The remaining mos6502 "wide memory access" error is a pre-existing limitation on wide *loads*, past the declassify legalization.)
**[3]** the explicit-initializer declaration store (`var z: ^u32 = 0`) now marks destination-secrecy, symmetric with `lower_assign`.
**[4]** `looprotate.make_instr` carries `writes_secret` onto rotation clones (was uninitialized; inert today, but no UB and keeps the seed if rotation ever peels a store).

**Root-cause sweep:** `MIR_DECLASSIFY` is now handled everywhere `MIR_MOV` is — spirv emit, legalize, the four rule packs, the x64 printer — and added to the selection alphabet. `writes_secret` is carried through every MirInstr copy that survives to later stages (regalloc flatten, isel piece emission, legalize piece). Confirmed spirv bypasses looprotate/legalize/isel (emit_module path), so those are machine-path-only.

**Tests:** `declassify_edge_consumers` — `:^` emits cleanly on SPIR-V and legalizes past the declassify on mos6502. (looprotate stays inert — determinism covered by the fixpoint + byte-identical.)

**Re-verified (all green):** suite **746/0**; **byte-identical** for non-`^`/`:^` code vs current dev + fixpoint **B==C**; `mach-std` **611/0** (matches the seed compiler on the pinned commit — the earlier 663 was the shared dep drifting, not a regression); **int linux + rv64** all pass.

Folded in the staged contract-doc tradeoff note for #1648's designer (lowered-MIR validation trusts isel/regalloc/encode; emitted-code translation-validation needs the seeds + barrier to survive to post-regalloc MIR).

Note on the generic-opcode-coverage test you suggested for SPIR-V: the `declassify_edge_consumers` test covers this specific gap; a full per-opcode coverage harness for the direct emitters (needs a synthetic Emit/VMaps context) is a good permanent guard but larger infra — flagging as a recommended follow-up rather than folding it into this security PR.

Issue #2141, opened 2026-07-23
codegen: element-wise loop vectorization (increment 1) + opt-out surface


octalide:
Part of #1942. Depends on #2139 and the versioning framework.

## Scope
The first vectorizing transform — release pipeline (`-O2`) + capable targets only (`tgt.model.has_v128`; rv64 stays scalar):
- Recognize the element-wise shape `dst[i] = src0[i] OP src1[i]` (and unary) over the counted loop, for element types that fit a 128-bit lane (i8/i16/i32/i64, unsigned, f32/f64).
- Rewrite the versioned clone's body to `IRT_VECTOR` ops: vector load/op/store by the lane count (`16/elem_bytes`), stepping the IV by the lane count, with a scalar remainder loop for the trip-count tail. Consumes the SIMD op layer already shipped (SSE2/NEON isel).
- Cost model read off `tgt`: vectorize only when profitable (lane count >= 2, pure element-wise body).
- Opt-out surface (required by the #197 direction): a per-build manifest profile key (schema-sweep, like the `simd` key — name TBD, propose to owner) and a `#[no_vectorize]` function decorator; both honored.

## Acceptance
Element-wise f64/f32/integer loops emit packed SIMD at `-O2` on x64/arm64, bit-identical to the scalar reference (element-wise reordering is exact), with a correct scalar remainder at non-multiple trip counts; measured throughput win (feasibility proven: hand-vectorized f64x2 1.86x, f32x4 5.3x, both bit-identical). Opt-out flag and decorator suppress vectorization. rv64 output unchanged. New int case + vector-runtime coverage green; single-gen fixpoint.

Issue #1942, opened 2026-07-04
epic: loop auto-vectorization (SIMD)


octalide:
Epic. Part of #1933.

Auto-vectorization: transform scalar element-wise / reduction loops into 128-bit SIMD at `-O2` on capable targets. Promoted from a single issue to an epic after a scoping pass showed it needs the standard vectorizer stack (loop analysis, dependence + runtime alias checks, the transform, and reductions) — three-plus issues of its own, as this issue's original text anticipated.

## The codegen layer is already proven ready
The SIMD op layer shipped with the #1965 campaign (`IRT_VECTOR`, SSE2/NEON isel, the 42-test `vector_runtime` suite). A hand-vectorized `a[i] = b[i] + c[i]` with explicit vector types + a scalar remainder was built at `-O2` and run:
- **f64x2** (2 lanes): bit-identical to scalar at an odd trip count (remainder exercised), **1.86x**.
- **f32x4** (4 lanes): bit-identical, **5.3x**.
- Emitted asm confirmed: packed ops (`add xmm,xmm,xmm`) + full-width unaligned `movups` in the vector body; scalar `fadd`/`movsd` in the remainder. x64 already emits unaligned vector load/store for pointer-to-vector access (`encode.mach`), so 8-aligned array data is safe — no alignment wall.

So children below are **execution, not research**. rv64 stays scalar throughout (`has_v128 == false`; the existing scalarize path is unaffected).

## Design boundary
- Recognized shape: innermost counted loops, unit-stride contiguous accesses, element types that fit a 128-bit lane; element-wise (children 1-3) and reductions (child 4).
- Correctness over coverage: pointer-parameter loops can't be proven non-aliasing at compile time (no `restrict`), so a runtime alias-check + loop versioning is the enabling mechanism; if a loop can't be proven safe, it stays scalar.
- Vectorize the versioned clone; a scalar remainder covers the trip-count tail.

## f64-reduction policy (owner ruling 2026-07-23)
IEEE-strict is the default: **never reassociate f64**. Integer reductions are associative/exact, so they vectorize **bit-identically by default** (child 4's baseline acceptance). f64/f32 reduction vectorization reassociates, so it is gated behind a project-config profile key (a fast-math-family manifest key, like `simd`, off by default; final name proposed to the owner when child 3/4 designs the profile surface). This dissolves the original "packed f64 matmul AND identical bit-pattern checksum" contradiction.

[...]

PR #2145, opened 2026-07-23
feat(me): loop-analysis infrastructure for auto-vectorization (#2139)

### LINKED ISSUE #2139: codegen: loop-analysis infrastructure (dominators, natural loops, induction vars)
Part of #1942.

First child of the auto-vectorization epic. Pure middle-end analysis, no codegen change (byte-identical compiler output). Reusable by the vectorizer (children below) and future loop passes (LICM, unrolling).

## Scope
A new analysis module `src/lang/me/analysis/loops.mach` providing, over an `ir.Function`:
- Dominator tree (Cooper-Harvey-Kennedy over reverse-postorder) with a `dominates` query.
- Natural-loop / back-edge detection; a loop forest with nesting (parent + depth) so the vectorizer can select innermost loops.
- Per-loop structure: header, unique preheader, unique latch, unique exit / exiting block (`BLOCK_NIL` when not unique — the conservative signal children consume).
- Canonical induction-variable recognition: a header phi `{init, +, step}` add-recurrence with a loop-invariant (constant) step.
- Counted-loop recognition (SCEV-lite): the head-controlled exit test `cmp(iv, bound)` with a loop-invariant bound and a predicate (LT/LE, signed/unsigned) — the trip-count shape.

Unwired in this increment (nothing calls it) so the compiler binary is byte-identical; child 3 consumes it when it wires vectorization into the pipeline. Independent of mem2reg's private dominator copy; a later cleanup may unify them (out of scope here, to preserve byte-identity).

## Acceptance
Unit tests construct loop IR and assert dominators, natural-loop membership/nesting, preheader/latch/exit identification, the IV `{init, +, step}`, and the counted-loop bound/predicate — including negative cases (multi-latch / non-canonical loops return the conservative NIL / none). Full suite green; compiler output byte-identical (pure analysis).

octalide:
Closes #2139. First child of the auto-vectorization epic #1942.

Pure middle-end analysis, no codegen change. Adds `src/lang/me/analysis/loops.mach` (`mach.lang.me.analysis.loops`), a read-only analysis over an `ir.Function`:

- **Dominator tree** — Cooper-Harvey-Kennedy over reverse postorder, with a reflexive `dominates(a, b)` query.
- **Natural loops** — back edges (edges whose head dominates their tail) collected into a forest with nesting (`parent` + `depth`), so a consumer can select the innermost loop.
- **Per-loop structure** — `header`, the unique `preheader` / `latch` / `exit` / `exiting` block, each `BLOCK_NIL` when not unique (the conservative signal downstream passes gate on).
- **Induction variables** — a canonical header-phi `{init, +, step}` add-recurrence with a loop-invariant step.
- **Counted loops** (the trip-count / SCEV-lite shape) — a head-controlled `cmp(iv, bound)` conditional branch with a loop-invariant bound and an LT/LE predicate, recording the bound, predicate, orientation, and which branch edge re-enters the loop.

Predecessors are derived internally from the terminator successor edges, so the analysis never mutates the function and is correct regardless of `Block.preds` consistency. It is independent of mem2reg's private dominator implementation (a later cleanup may unify them; done here would change mem2reg's compiled bytes and break byte-identity).

### Not wired in
Nothing imports the module this increment, so `mach build` never reaches it and the compiler binary is unchanged. The element-wise vectorization transform (#2141) wires it into the release pipeline.

### Verification
- **Byte-identical compiler output** — built the compiler with and without the new file; both are the same SHA256 (`ba714c79…`). Proven, per the child-1 bar.
- **Full suite** — `mach test .` (release) = **727 passed / 0 failed** (723 dev baseline + 4 new loop-analysis tests): the counted-loop structure end-to-end (dominators, preheader/latch/exit, IV `{0,+,1}`, `i < n`); a reduction loop with two header phis, asserting the counter is selected over the accumulator; a two-latch loop (unioned body, latch not unique, recognition conservatively declines); and a nested loop (parent/depth/containment, both counted).
- **Single-gen self-host fixpoint** — B == C byte-identical.
- **int** — `linux` and `linux-riscv64` (qemu) both all-pass; rv64 unaffected (this touches no codegen).

### DIFF
--- a/src/lang/me/analysis/loops.mach
+++ b/src/lang/me/analysis/loops.mach
@@ -26,6 +26,8 @@ use std.types.bool.true;
 use std.types.size.usize;
 use std.types.string.str;
 
+use std.system.panic.panic;
+
 use A: std.allocator;
 use O: std.types.option;
 use R: std.types.result;
@@ -42,8 +44,10 @@ pub val NONE_U32: u32 = 0xFFFFFFFF;
 # ---
 # phi:    the header phi instruction defining the induction variable
 # init:   the value entering the phi from the preheader edge
-# step:   the loop-invariant step added each iteration (a constant in the counted
-#         form; any loop-invariant value in the general form)
+# step:   the step added each iteration. read `step.data.int_lit` as a constant
+#         ONLY under the counted guarantee (Loop.is_counted), which enforces a
+#         non-zero VAL_CONST_INT; a general (has_iv but not is_counted) IV carries
+#         any loop-invariant value here, which may not be a constant
 # update: the `phi + step` instruction on the latch edge
 pub rec InductionVar {
     phi:    id.InstructionId;
@@ -53,6 +57,12 @@ pub rec InductionVar {
 }
 
 # the recognized counted-loop exit test over an induction variable
+#
+# Valid only when Loop.is_counted. Its guarantees, all fully enforced before
+# is_counted is set, are the contract a trip-count consumer relies on: the header
+# is the loop's sole exiting block; the induction variable's step is a non-zero
+# VAL_CONST_INT whose sign drives the variable across `bound` so the test
+# terminates; and `bound` is loop-invariant.
 # ---
 # bound:            the loop-invariant value the induction variable is compared against
 # pred:             the comparison opcode (OP_CMP_LT_S / _LT_U / _LE_S / _LE_U)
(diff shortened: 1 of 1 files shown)


octalide:
Review-fix round (guard hardening): a positive signal is now emitted only when it fully holds; otherwise the conservative signal.

- `is_counted` requires the header to be the **sole** exiting block (mid-body early exit → not counted).
- `preheader` requires the unique out-of-loop predecessor's **sole** successor be the header (guard block that also skips the loop → BLOCK_NIL).
- `is_counted` requires the step **sign** to match the exit predicate (negative step under LT → not counted).
- the counted form requires a **non-zero constant** step (param/global/zero step → not counted).
- `exit` is set only when **dedicated** (no predecessors from outside the loop; shared join → BLOCK_NIL).
- `compute_nesting` / `dominates` now **panic loudly** on the by-construction-impossible cyclic chain instead of silently truncating.

Six negative-case tests added; nested test uses a dedicated inner preheader. Re-verified: byte-identical compiler output (SHA `ba714c79…`), suite 733/0, fixpoint B==C, int linux + linux-riscv64 green.

PR #2148, opened 2026-07-23
feat(me): loop dependence analysis + runtime alias-check + versioning (#2140)

### LINKED ISSUE #2140: codegen: loop dependence analysis + runtime alias-check + loop versioning
Part of #1942. Depends on #2139.

## Scope
Consuming the loop analysis (#2139):
- Memory-access collection over a counted loop body: base pointer + index expression per load/store, classified as unit-stride contiguous vs. other.
- Dependence test proving no cross-iteration dependence for the element-wise shape `dst[i] = f(src0[i], src1[i], ...)`.
- Runtime alias-check synthesis + loop versioning: emit a guard comparing the `dst` pointer range against each `src` range; branch to a vectorizable clone when disjoint, else the original scalar loop. The CFG/phi surgery framework (clone a loop, rewire preds/phis, keep the verifier's reachability/dominance invariants) lands here.

Pointer-parameter loops (axpy/matmul) cannot be proven non-aliasing at compile time (mach has no `restrict`), so runtime versioning is the enabling mechanism; without it, only compile-time-distinct memory is safe.

## Acceptance
The versioning framework produces a guarded scalar/clone pair for a counted element-wise loop; the clone is a faithful copy (still scalar at this stage) selected only when the runtime disjointness check passes; verifier-clean; the scalar path is taken when ranges overlap. Tests cover runtime overlap and disjoint. Output bit-identical for loops the legality test rejects.

octalide:
Closes #2140. Second child of the auto-vectorization epic #1942.

The safety machinery child 3 (element-wise vectorization) will consume: three callable APIs over child 1's loop analysis, all **dormant** (nothing wired into the pipeline), so the compiler binary is unchanged.

### 1. Dependence analysis — `mach.lang.me.analysis.dependence`
Over a counted loop, a deliberately **conservative** verdict:
- `DEP_INDEPENDENT` only when every memory access is a non-volatile load/store through `gep(base, iv)` — a single index equal to the induction-variable phi (offset 0, unit stride) over a loop-invariant base — the induction variable is the loop's **sole** register recurrence, and there is no opaque memory effect (call / asm / memzero).
- `DEP_DEPENDENT` for everything else (a loop-carried `a[i]=a[i-1]`, a reduction's second phi, an unknown access, a non-counted loop).

Collects the recognized `base[iv]` accesses (base, per-element stride type, element type, store/load) for the alias guard.

### 2. Runtime alias-check — `versioning.emit_alias_guard`
Synthesizes the runtime disjointness guard: for each write base against every distinct-base access range, `(end_w <= start_o) || (end_o <= start_w)` over per-access byte ranges `[base+init*sz, base+bound*sz)` with unsigned pointer compares, AND-ed across pairs. No distinct-base pair (no writes, single base, all-equal bases) yields a constant-true guard (no hazard).

### 3. Loop versioning — `versioning.version_loop`
Clones a well-formed, dependence-independent loop into a runtime-guarded **fast** copy plus the original **scalar** fallback, with the CFG / phi surgery: dedicated guard block (alias check → `cbr` fast/scalar) and merge block, preheader→guard rewire, header phi-incoming relabeling, and exit routing through the merge. Both copies are scalar this increment, so the transform is **behavior-preserving**; child 3 replaces the fast copy's body with vector ops. **Refuses (no mutation)** any loop that is not counted, not well-formed (relying on child 1's dedicated preheader/latch/exit guarantees), dependence-dependent, or that defines a **live-out** SSA value (deferred to child 4's reduction-aware merging).

### Conservative-safety argument
Pointer-parameter loops (axpy/matmul) can't be proven non-aliasing at compile time (mach has no `restrict`), so static independence is discharged at runtime by the alias guard + versioning; a loop that can't be proven safe stays scalar. Every positive signal is fully guarded (child 1's lesson): the dependence verdict clears only the proven element-wise shape, and versioning proceeds only on child-1-well-formed loops with no live-out.

### Verification
- **Byte-identical compiler output** — built with and without both new files; same SHA256 (`f4ddf123…`). Both modules are orphan (unimported).
- **Full suite** — `mach test .` = **742 / 0** (9 new tests): dependence independent / loop-carried / reduction; the alias-guard instruction shape (6 geps + 4 compares + 2 ors + 1 and for `a[i]=b[i]+c[i]`); **versioning is verifier-clean** (a test runs the IR verifier over the versioned function — reachability + dominance enforced), two loops after versioning, guard `cbr` targets; and refusal of dependent and live-out loops without mutation.
- **Single-gen fixpoint** B == C byte-identical.
- **int** `linux` and `linux-riscv64` all-pass; **mach-std** 663 / 0 under the from-source compiler. rv64 stays scalar (no codegen change).

### DIFF
--- a/src/lang/me/analysis/dependence.mach
+++ b/src/lang/me/analysis/dependence.mach
@@ -1,0 +1,465 @@
+# mach.lang.me.analysis.dependence: cross-iteration memory-dependence analysis
+#
+# Over a counted loop recognized by `mach.lang.me.analysis.loops`, decide whether
+# the loop is a unit-stride, offset-0 element-wise map -- `dst[i] = f(src0[i], ...)`
+# -- carrying no cross-iteration memory dependence, and collect its memory accesses.
+#
+# The verdict is deliberately conservative (child 1's lesson): a positive
+# INDEPENDENT verdict is returned ONLY when every memory access in the loop is a
+# non-volatile load/store through `gep(base, iv)` -- a single index equal to the
+# loop's induction variable (offset 0, unit stride) over a loop-invariant base --
+# and the loop carries no register recurrence beyond the induction variable and no
+# opaque memory effect (call / asm / memzero). Anything unproven is DEPENDENT.
+#
+# A DEPENDENT verdict means "do not vectorize". An INDEPENDENT verdict means "safe
+# to vectorize IF the runtime alias check discharges base aliasing" -- distinct
+# pointer bases cannot be proven disjoint at compile time (mach has no `restrict`),
+# so the caller pairs this with the runtime alias guard in the versioning module.
+#
+# This module is a callable analysis; it is not wired into the pipeline (nothing
+# imports it there), so it does not affect emitted code.
+
+use std.types.bool.bool;
+use std.types.bool.false;
+use std.types.bool.true;
+use std.types.size.usize;
+use std.types.string.str;
+
+use A: std.allocator;
+use O: std.types.option;
+use R: std.types.result;
+
+use mach.lang.me.ir;
+use mach.lang.me.ir.id;
+use mach.lang.me.ir.instruction;
+use mach.lang.me.ir.value;
+use ir_type: mach.lang.me.ir.type;
+use loops: mach.lang.me.analysis.loops;
+
+# one recognized affine memory access `base[iv]` in a counted loop
+# ---
+# instr:     the load or store instruction
+# gep:       the OP_GEP computing `base[iv]` that the access dereferences
+# base:      the loop-invariant base pointer (the gep's first operand)
+# stride_ty: the gep's source pointee type -- the per-element stride, reused to
+#            compute the accessed byte range for the alias guard
+# elem_ty:   the loaded / stored value type
+# is_store:  true for a store, false for a load
+pub rec MemAccess {
+    instr:     id.InstructionId;
+    gep:       id.InstructionId;
+    base:      value.Value;
+    stride_ty: ir_type.IrTypeId;
+    elem_ty:   ir_type.IrTypeId;
+    is_store:  bool;
+}
+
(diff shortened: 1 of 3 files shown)


octalide:
Review-fix round (byte-footprint soundness): the analysis now reports INDEPENDENT/DISJOINT only when the true byte ranges are provably non-overlapping.

**Correctness**
1. **LE off-by-one** — child 1's `CountedInfo` now carries `bound_inclusive`; the alias guard derives the exclusive high index via one shared `versioning.emit_exclusive_bound` (LE materializes `bound+1`), which child 3's vector trip will also use — no consumer re-normalizes the raw bound. Fixes an `i<=n` guard range that was one element short (a[n] top-overlap read as disjoint).
2 + 3. **stride/width footprint** — INDEPENDENT now requires every access be a contiguous non-overlapping unit map (gep stride == access width, a supported scalar footprint) **and** all same-base accesses share an identical width, so an i8-load + i64-store on one base, or a 1-byte stride feeding an 8-byte store, is DEPENDENT.

**Cleanups**
4. One shared value-identity predicate (`dependence.equal`, mirroring constfold/cse + the const kinds), both modules routed through it — kept in the orphan module rather than `ir/value.mach` **specifically to preserve byte-identity** (an emitted `ir/value.mach` helper grows the compiler binary; verified it diverged from dev, so it migrates there with child 3 when the modules become reachable).
5. Range geps emitted lazily (only for accesses in a distinct-base pair).
6. Removed the numbered phase-marker comments.

**Re-verified:** compiler **byte-identical to dev** (SHA `f4ddf123`); single-gen fixpoint B==C; suite **745/0** (3 new footprint tests: LE exclusive bound, same-base different width, wider-than-stride); mach-std 663/0; int linux + linux-riscv64 green. The IR-verifier versioning test still passes.

PR #2150, opened 2026-07-23
feat(codegen): element-wise loop auto-vectorization (increment 1) (#2141)

### LINKED ISSUE #2141: codegen: element-wise loop vectorization (increment 1) + opt-out surface

octalide:
Closes #2141. Third child of the auto-vectorization epic #1942 — and the **first codegen-changing** piece: it vectorizes loops, so output is intentionally not byte-identical for vectorizable code.

## The transform
Wires child 1 (loop analysis) + child 2 (dependence + alias-check + versioning) into the vectorizer. On a capable target (`has_v128`; SSE2 x86_64 / NEON aarch64) in the **release** pipeline, for a counted, well-formed, dependence-independent, unit-stride element-wise map `dst[i] = f(src0[i], ...)`:
1. `version_loop` (child 2) emits a runtime **alias-guarded fast copy + scalar fallback**;
2. a dedicated vector preheader is inserted (so the fast loop re-analyzes as counted), then the fast copy is rewritten to **128-bit vector ops** over the lane count (`16/elem_bytes`), stepping the IV by the lane count;
3. a **scalar remainder loop** handles the trip-count tail, entered from the vector loop's final IV.

At runtime the alias guard routes overlapping arrays to the scalar fallback. **rv64** (`has_v128 == false`) never enters the pass and is byte-identical to before.

**Conservative legality:** only add/sub/negate (int + float), float mul/div, and integer bitwise ops vectorize (the lane forms selectable on the SSE2/NEON baseline), only when every store's value is a pure tree of those ops over the loads (no scalar/splat operand), uniform supported element width, lane count ≥ 2.

## Opt-out surface — names flagged for owner sign-off
- **Manifest profile key** `vectorize = true|false` — optional, defaults on; `false` skips the pass at release. Threaded through the resolved profile and the request fingerprint. Made **optional-with-default** (not a required-everywhere key), so it needs no coordinated 28-manifest sweep with a provisional name; the required-everywhere sweep (the `simd`-key precedent) can follow once the owner fixes the final name.
- **`#[no_vectorize]` function decorator** (`FN_FLAG_NO_VECTORIZE`) — skips the function; the inliner also declines to inline it, so the per-function opt-out survives inlining (the flag would otherwise be dropped when the body moves into an unflagged caller).

Proposed names: profile key `vectorize`, decorator `no_vectorize`. **Please take these to the owner** (he names user-facing profile keys).

## Verification — runtime correctness is the bar
- **Runtime golden (int/surface/vectorize + local matrix):** every width (i8/i16/i32/i64, unsigned, f32/f64), each op (add/sub/mul/div/and/or/xor), at trip counts that are **multiples of the lane count and not** (exercising the remainder), plus an **aliasing hazard** (dst overlaps src by one element → the guard routes to scalar) — all **bit-identical** to a `#[no_vectorize]` scalar reference. The int case (`mismatches=0`) runs on linux + linux-riscv64, debug (scalar) and release (vectorized), same golden.
- **Three-generation fixpoint:** the seed lacks this pass so A≠B is legitimate; B==C **byte-identical** (self-applied vectorization is a fixpoint — vectorizing the compiler's own loops does not miscompile).
- **-g additivity:** the vectorized `.text` is byte-identical with and without `-g`.
- **rv64 unaffected:** byte-identical object to the dev-baseline compiler; non-vectorizable x86_64 code likewise unchanged.
- Full suite **748/0** (incl. the 42-test vector-runtime suite); mach-std **663/0**; int linux + linux-riscv64 green.
- **Measured win** (seed-scalar vs vectorized, cache-resident N=1024): **f64 2.33x, f32 4.95x**.

## Notes for review
- The pass is gated in the release block on `has_v128 && vectorize_enabled`; the always-on exit verifier is the safety net (it caught a CFG-surgery bug during development — no malformed vectorized IR can reach the backend).
- Increment-1 boundary (reductions are child 4; no splat of scalar operands; op set limited to baseline-selectable lane ops) is enforced conservatively — anything unproven stays scalar.


octalide:
Review-fix round (max review, 4 confirmed + 4 cleanups) + the owner's `#[scalar]` rename, one pass.

**Confirmed miscompiles — all fixed conservatively (decline when in doubt):**
1. **Uniform element TYPE** — legality requires every access share one scalar element type (not merely one width); the vset additionally requires each op's result type equal the element type. A same-width i32/f32 mixed loop now declines.
2. **Vector negate removed** from the supported op set (no vector-negate lowering exists; a vector float neg would integer-negate IEEE bits). Negate loops stay scalar.
3. **Single-block body** — exactly header + one body/latch block, no internal control flow, and every access in that body block. Conditional stores (and header accesses) decline.
4. **`ProjectConfig` copies `bu.vectorize`** alongside `bu.simd` (metadata-drop family closed).

**Cleanups:** dependence runs once per candidate (collect-phase DepInfo kept — valid across appends) + the unavoidable post-version re-check; `VersionResult.bound_ex` materialized once in the guard block and reused by the vector preheader (`emit_alias_guard` now takes the materialized bound); ONE shared `versioning.clone_region` scaffold (vec_lane lesson); the exit-test rewrite no longer assumes `continue_on_true` — the shape gate declines instead. A post-mutation decline rebuilds preds (verifier-consistent); the fast-loop re-check correctly does not demand a dedicated exit (the merge has the scalar path as a second pred — found when the first cut of the re-check declined everything).

**Rename:** `#[scalar]` everywhere (FN_FLAG_SCALAR); mechanism unchanged (vectorizer skips + inliner declines). `vectorize` manifest key confirmed final.

**Re-verified:** int case extended with mixed-type, float-negate (incl. -0.0 sign), conditional-store — all bit-identical to `#[scalar]` references, release/debug goldens agree (`mismatches=0`); per-kernel IR audit: 6 plain kernels vectorize (incl. i8x16), all 4 hazard kernels decline. Suite 748/0; three-gen B==C; -g additivity (.text identical); rv64 .o byte-identical vs dev-baseline; mach-std 611/0; int linux + linux-riscv64 full harness green; opt-outs proven both ways (key off → 0 vector ops; default → vectorized); win re-measured: **f64 2.07x, f32 4.90x** (seed-scalar vs vectorized, N=1024).

Issue #2127, opened 2026-07-23
sema: aggregate value ==/!= type-checks but lowers as a non-structural int/address compare (silent miscompile)


octalide:
Surfaced while implementing std.derive (mach-std #285), which exists partly *because* of this.

## What

`record == record` and `array == array` on VALUES type-check but lower to an integer/address comparison, not a structural (field/element-wise) one. So `a == b` on two record values silently returns the wrong answer — it compares representation words/addresses, not contents. No diagnostic; it just compiles to something meaningless.

## Options

1. **Reject in sema (conservative, recommended):** `==`/`!=` on aggregate (record/union/array) operands is a compile error with a teaching message pointing at a field-wise compare or `std.derive.eq`. In mach's explicit spirit (no hidden potentially-expensive structural compare behind `==`), and it eliminates the silent-wrong with minimal surface. Risk: rejects currently-accepted (but already-wrong) code — almost certainly none relies on the meaningless current behavior.
2. **Implement structural compare:** `==` on aggregates lowers to a field/element-wise compare (recursively). Makes `==` "just work" but hides a potentially large operation behind the operator, and needs a per-type lowering.

This is a **language-semantics decision** — flagging for the owner's call before implementing.

PR #2155, opened 2026-07-24
feat(sema): #[oblivious]-required check + per-instance CT re-validation (#2136, #2157)

### LINKED ISSUE #2136: sema: #[oblivious]-required check (per-monomorphized-instance secrecy-transparency)

### LINKED ISSUE #2157: sema: generic ==/!= secrecy computed at template time — secret-dependent branch compiles clean (CT false-negative)
## Security false-negative in the constant-time guarantee (upstream of #2136)

Found by adversarial review of #2155, proven end-to-end (differential compile + disassembly of the PR compiler).

### The hole
```
fun cmp_branch[T](a: T, b: T) u32 { if (a == b) { ret 1; } ret 0; }  // @ T = ^u32
```
compiles CLEAN and emits a real secret-dependent branch — disassembly of the `^u32` instance shows `cmp %esi,%edi; je …`. The non-generic twin `fun(a: ^u32, b: ^u32)` is correctly rejected at sema ("secret value used as a branch condition"). Only the generic collapse differs.

### Root cause
- `==`/`!=` is the ONE compute sema permits on an opaque generic `T`: `operands_ok = (lt == rt) || both_int`, and `T == T` satisfies `lt == rt` (fe/sema/infer.mach ~688-691). Arithmetic/bitwise/shift/ordering-compares/index all require provable int-ness and ARE rejected on opaque `T`.
- The compare result is `u8`; its secrecy = `join_secret` computed at TEMPLATE-check time over the opaque (non-secret) `T` → recorded as PUBLIC `u8`. The recorded type contains no `T`.
- At lowering the `^u32` instance, `stamp_secret` (me/lower/expr.mach:126) keys taint off the recorded result type via `type_is_secret`→`substitute_type`; substituting into a `u8` with no `T` is a no-op → the CMP's `.secret` is never set.
- Consequences: the required-check #2136 (keys on `Instruction.secret`) doesn't flag it; the sema #1645 branch gate ran on the TEMPLATE (public) so didn't fire on the instance; and because it's unflagged, the #1648 validator's #[oblivious]-scoped checks never run on it. Three defenses, one root cause.
- The `^u32`/`u32` instances also mangle-collapse to ONE weak symbol carrying only public taint — the exact collapse #2136's design dismissed as "not a hole".

### Fix direction (design fork — needs investigation)
The gap is structural in secrecy flow typing / monomorphization, not patchable in oblivious.mach (the taint simply isn't there). Options:
1. **Re-derive secrecy per monomorphized instance** — join over the instance's concrete operand types; the sema gates then fire on the instance directly (rejects at the earliest point). Most structural, biggest.
2. **Operand-derived compute taint** — for compute opcodes, stamp `.secret` from the OPERANDS' instance secrecy, not the recorded result type (aligns with how #1648's re-derive already works). Targeted; closes the flagging/validation path.
3. **Forbid `==`/`!=` on an unconstrained generic `T`** — language restriction; smallest but changes what compiles (an owner call).

Any of these also closes the escalated secret-branch leak. Security-critical: this is a false-negative in the CT guarantee shipped in #2137. Blocks #2136's stated soundness (#2155).

octalide:
Closes #2136. Closes #2157. Part of the constant-time epic #1643 (builds on #1646).

Folds the `#[oblivious]`-required check (#2136) together with the upstream generic-secrecy fix (#2157) an adversarial review of this PR uncovered, since #2136's soundness depends on it.

## #2157 — per-instance constant-time re-validation of generic secrecy (the security fix)

`==`/`!=` is the one compute sema permits on an opaque generic `T` (`operands_ok = (lt==rt)||both_int`). Its `u8` result secrecy was join'd at generic-**template** time over the non-secret parameter, so a secret instantiation that branched / indexed / aggregate-compared through the result compiled **clean** — a proven CT false-negative shipped in #2137 (`cmp_branch[^u32]` emitted a real `cbr.eq` on secret operands). The monomorphization worklist + mangling also erase secrecy, collapsing `foo[^u32]` / `foo[u32]` to one lowered body, so no post-lowering analysis could see the secret instance.

**Fix (owner-approved, no ABI / mangling change)** — the owner's `^`-is-a-type-wrapper model: a generic operation on `T` behaves at each instantiation exactly as the concrete operation would.

- `infer_generic_call` records each instantiation `(origin, decl, concrete args)` to a worklist, keyed by the concrete type-args resolved **before** lowering's mangle collapse — so the secret instance is distinct from a public twin. Pruned to secret / aggregate args; deduped for termination.
- After all bodies are typed, the worklist drains: each instance body is re-inferred under its substitution into **scratch** typing arrays (the template arrays lowering reads stay intact → byte-identical), so `join_secret` over concrete operands, `check_condition` / `check_index`, `gate_ct_always`, and the #2127 aggregate-`==` rejection fire on the instance. Nested generic calls enqueue their own instances (the transitive case).
- Substitution injects at the sema type-source leaves (`resolved_type_of` / `resolve_type_ref` / `generic_param_type_for`) via a function pointer on `SemaContext`, breaking the `context`→`generics` import cycle.

Result: `cmp_branch[^u32]` is rejected with the **same diagnostic** as the concrete twin (`secret value used as a branch condition`); `eq[u32]` still compiles; the twin case rejects the secret one despite the symbol collapse; transitive forwarding rejects; a generic `==` at `T`=record is rejected (extends #2127 to generics).

## #2136 — the `#[oblivious]`-required check

Per-monomorphized-instance secrecy-transparency gate: a function instance that **computes** on a secret (arithmetic / bitwise / shift / comparison / negation) must carry `#[oblivious]`; a **transparent** instance that only moves / stores / declassifies secrets stays annotation-free. Runs over the freshly-lowered IR before optimization; predicate `instruction.op_is_secret_move`.

Two lowering changes make it correct for the generic case:
- **operand-derived comparison taint** (`stamp_secret`): recovers a comparison's frozen `==` result secrecy from its operands, so a **legal** secret comparison mask in a generic instance carries `Instruction.secret` and #2136 flags it;
- a generic instance now **inherits its template's `#[oblivious]` flag** (previously dropped), so annotating the template clears the check.

Both are inert on secret-free code.

## Verification

- Full suite **762/0** (new `generic_instance_ct_reinfer` covering: secret branch rejected same-diagnostic; public `eq[u32]` clean; twin rejected despite collapse; transitive rejected; aggregate-`==` rejected; a legal secret mask flagged by #2136; `#[oblivious]` clears it).
- **Byte-identical** for secret-free / non-generic-`==` code: this compiler and an independent no-change baseline compiler produce byte-identical output for the same base source. **Fixpoint B == C**.
- **mach-std 611/0** (no spurious re-validation on the generic-heavy stdlib). **int linux + rv64** green.

Security-critical (the CT primitive) — the per-instance gating is exhaustive across branch / index / div / float / aggregate-`==` and the transitive + twin-collapse cases.

### DIFF
--- a/int/surface/declassify/src/main.mach
+++ b/int/surface/declassify/src/main.mach
@@ -11,7 +11,10 @@ use std.runtime;
 use std.types.size.usize;
 use p: std.print;
 
-# add two secrets and declassify the result with the bare `:^` strip
+# add two secrets and declassify the result with the bare `:^` strip. it COMPUTES on
+# secrets (the `a + b`), so it carries `#[oblivious]` - required per instance (#2136) of
+# any function that arithmetically derives from a secret rather than only moving it.
+#[oblivious]
 fun secret_add(a: ^u32, b: ^u32) u32 {
     val s: ^u32 = a + b;
     ret s:^;
@@ -20,13 +23,17 @@ fun secret_add(a: ^u32, b: ^u32) u32 {
 # double a secret and declassify with the typed `:^u32` strip (names the
 # operand's own stripped public type). doubles with `a + a`, not `a * 2`: a secret
 # multiply is a variable-latency leak the #1646 constant-time gate rejects, while a
-# secret add is constant-time - the discipline this whole feature enforces.
+# secret add is constant-time - the discipline this whole feature enforces. it too
+# computes on a secret, so it is `#[oblivious]` (#2136).
+#[oblivious]
 fun secret_double(a: ^u32) u32 {
     val s: ^u32 = a + a;
     ret s:^u32;
 }
 
-# up-coerce a public into a secret, then strip it straight back
+# up-coerce a public into a secret, then strip it straight back. this only MOVES the
+# value (a public up-coerce and a declassify), never computing on the secret, so it is
+# secrecy-transparent and needs no `#[oblivious]` (#2136).
 fun roundtrip(x: u32) u32 {
     val s: ^u32 = x;
     ret s:^;
--- a/src/lang/driver/passes.mach
+++ b/src/lang/driver/passes.mach
@@ -44,6 +44,7 @@ use mach.lang.fe.token;
 use mach.lang.intern;
 use mach.lang.manifest;
 use mach.lang.me.ir;
+use mach.lang.me.analysis.oblivious;
 use mach.lang.me.lower;
 use mach.lang.me.lower.testrunner;
 use mach.lang.me.pipeline;
@@ -959,6 +960,21 @@ pub fun q_lower_compute(db: *query.QueryDb, key: u64, alloc: *A.Allocator) R.Res
     val res_lower: R.Result[ir.Module, str] = lower.lower_module(s, ?p.target, a, m.fqn, mid, sr, m.resolve_result, ?m.ctx, ctx.req.debug);
     if (R.is_err[ir.Module, str](res_lower)) { ret R.err[query.QueryOutput, str](R.unwrap_err[ir.Module, str](res_lower)); }
     var irmod: ir.Module = R.unwrap_ok[ir.Module, str](res_lower);
+
+    # the `#[oblivious]`-required check (#2136): over the freshly-lowered IR, before
+    # optimization, so a dead secret compute cannot be optimized away first and the
+    # requirement is independent of the optimization level. every monomorphized instance
+    # is visible here with its concrete secret taint, which sema could not see through a
+    # generic template. a filed diagnostic makes the failure a reported user error.
+    val res_obl: R.Result[bool, str] = oblivious.check_module(?irmod, ?s.interner, ?s.diags, s.alloc);
+    if (R.is_err[bool, str](res_obl)) {
+        ir.dnit(?irmod);
+        ret R.err[query.QueryOutput, str](R.unwrap_err[bool, str](res_obl));
+    }
+    if (!R.unwrap_ok[bool, str](res_obl)) {
+        ir.dnit(?irmod);
+        ret R.err[query.QueryOutput, str]("oblivious-required check: a function computes on a secret without `#[oblivious]`");
+    }
 
     # optimize in place: the cached artifact is the post-pipeline IR the object is
     # built from, so an unchanged optimized module reuses its codegen output.
--- a/src/lang/driver/tests.mach
+++ b/src/lang/driver/tests.mach
@@ -4737,8 +4737,10 @@ test "mach.lang.driver:oblivious_required" {
 
     # a GENERIC MOVER instantiated at a secret type moves the secret through pointers and
     # is NOT flagged - the epic's core requirement that generic copy/swap/containers stay
-    # annotation-free (a generic body cannot compute on its opaque `T`, so a secret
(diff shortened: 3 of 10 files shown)


octalide:
## HELD pending #2157 — soundness-claim correction

Adversarial review found a **proven security false-negative upstream of this check** (filed as #2157). It falsifies the "mangling secrecy-collapse is not a hole" argument in this PR's description:

- `==`/`!=` **is** a compute sema permits on an opaque generic `T` (`operands_ok = (lt==rt)||both_int`; `T==T` satisfies `lt==rt`). Its result secrecy is join'd over the non-secret parameter at **template-check time** → recorded public `u8`; the per-instance substitution can't recover it, so `Instruction.secret` is never set for it.
- The monomorphization worklist **dedups by mangled name**, and mangling erases secrecy, so `foo[^u32]` and `foo[u32]` collapse to **one lowered body** carrying only the public taint (confirmed: a public+secret twin emits a single `…N8eqI3u32E`).

Result: `fun cmp_branch[T](a: T, b: T) u32 { if (a==b) {ret 1;} ret 0; }` at `T=^u32` compiles clean and emits a real secret-dependent branch (`cbr.eq rdi, rsi`) — this check does not see it because the taint isn't there.

This check is **correct for correctly-stamped taint**; the gap is upstream. Held until #2157 lands, at which point a generic-secret-compute regression test (`cmp_branch[^u32]` rejected/flagged) is added here. The commit `69c23d06` retracts the falsified claim in the module doc + test comment and references #2157.

PR #2164, opened 2026-07-24
fix(sema): reject a secret value passed to a variadic pack (#2162)

### LINKED ISSUE #2162: sema: a secret (^T) value can be formatted/printed without declassification — undeclassified disclosure
## Security: direct confidentiality leak in the secret/constant-time feature

Found while investigating #2157 (it is a DISTINCT root cause, not the generic-== gap — packs gate correctly).

`printlnf("{}", s)` with `s: ^u32` (a SECRET) **compiles clean and prints the secret value** to stdout, with no `:^` declassification anywhere. This defeats the core purpose of `^`: the type system is supposed to force an explicit `:^` before a secret can be observed. Along the way the format path also runs secret `x/10` / `x%10` (variable-latency div on a secret), but the headline is the plain confidentiality disclosure — the secret's VALUE reaches stdout undeclassified.

## Root cause (to confirm in the fix)
A secret reaches a public-typed `format_uint`-style parameter (or silently widens to public) somewhere in the print/format call chain, WITHOUT tripping the secret→public coercion gate. Today `check_secrecy_cast` only permits public→secret up-coercion and requires `:^` for secret→public — so the format/print path is bypassing that check (a variadic/format-specific path, or a silent widening). The fix is to enforce the existing secret→public `:^` requirement through the format/print chain so `printlnf("{}", s)` is a compile error unless the caller writes `printlnf("{}", s:^)`.

## Severity / disposition
Release-blocking for any release that advertises secret/constant-time support (a secret must not be printable without `:^`). Serialized behind #2157 (same sema secrecy-flow surface) — assigned to the same worker to avoid a parallel collision. Surfaced by the #2157 investigation.

octalide:
Closes #2162. Security fix in the constant-time feature (epic #1643).

## The leak
`printlnf("{}", s)` with `s: ^u32` compiled clean and printed the secret's **value** to stdout — no `:^` anywhere — plus a secret `x/10` in the itoa path. An undeclassified confidentiality disclosure that defeats the point of `^`.

## Root cause — variadic pack-instance secrecy collapse (NOT a coercion bypass)
Root-caused, correcting the issue's hypothesis: the coercion gates are **sound** (verified in isolation — `a::u64` on `^u32` and a `^u64`→public-`u64` param are both rejected, including inside pack bodies). The real cause is that a pack's monomorphized instance name erases secrecy — `pack_instance_name` mangles element types through the same `write_encoded_type` that drops `TYPE_SECRET` — so `printlnf[^u32]` and `printlnf[u32]` collapse to **one** emitted body. The pack worklist dedups by name; when a public instance wins (std's own internal public prints guarantee one), the secret call reuses the **public** body: the element is public `u32`, `$type_of` matches the int arm, `arg::u64` is public→public, itoa's `x/10` is a public div. Same secrecy-erasing collapse as #2157, here for variadic packs.

Proven: with an explicit twin, `which(pub); which(s)` → both `matched-u32` (secret reuses public body); reversed → both `matched-none`. And `write_encoded_type` demonstrably strips `^`.

## Fix (Option C — owner-approved)
A **general** sema call-site gate: a secret-carrying value passed to a `...` variadic pack is rejected, requiring `:^` first. A variadic is an observation boundary (format / log / print) and the mangling cannot carry secrecy, so a secret must be declassified before it enters. Placed at the pack-arg check in `check_call`, so it fires uniformly for `printlnf`, log-style, and any user variadic — not a print special-case. A `va...` spread is skipped (its elements were gated at their own entry). Once a secret can't enter the pack, the secret div in itoa is unreachable.

This is the epic's "`:^` before observation" rule applied at the variadic boundary. It forbids nothing that works — a constant-time variadic over secrets is already broken by the collapse (it shares the public body), so requiring `:^` removes a footgun, not a capability. No ABI / mangling change (Option A, the secrecy-aware mangling, was gate-kept for #2157 and would only yield a runtime `ERR_BAD_FORMAT`, not the compile error the guarantee needs).

## Verification
- `printlnf("{}", s)` (s: `^u32`) → **rejected** with the teaching diagnostic; `printlnf("{}", s:^)` → compiles + prints `42`; public args + declassified secrets unaffected.
- **General**: a user `fun logf(va: ...)` called with a secret is rejected; a secret bound to a **fixed** (non-variadic) parameter is unaffected (ordinary assignability governs it).
- Full suite **777/0** (new `secret_variadic_rejected` test).
- Byte-identical for secret-free code + fixpoint B==C; mach-std 611/0 (pinned); int linux + rv64. (running)

Security fix → adversarial review when green.

### DIFF
--- a/src/lang/driver/tests.mach
+++ b/src/lang/driver/tests.mach
@@ -4821,6 +4821,17 @@ test "mach.lang.driver:secret_variadic_rejected" {
 
     # a PUBLIC record (no `^` anywhere in its field graph) is NOT over-rejected.
     if (!ut_sema_clean("fun logf(va: ...) u32 { ret 0; }\nrec Q { x: u32; y: u32; }\nfun f(a: Q) u32 { ret logf(a); }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+
+    # a GENERIC INSTANCE wrapping a secret is rejected regardless of whether its field
+    # table was materialized before the check (a bare local never accessed, and a call
+    # RESULT, both reach the boundary un-elaborated) - the deep walk materializes the
+    # instance's fields itself. Nested instances too; a public instance is not
+    # over-rejected (via call result and bare local).
+    if (!ut_sema_rejects("fun logf(va: ...) u32 { ret 0; }\nrec Box[T] { v: T; }\nfun mk[T](x: T) Box[T] { var b: Box[T]; b.v = x; ret b; }\nfun f(s: ^u32) u32 { ret logf(mk[^u32](s)); }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    if (!ut_sema_rejects("fun logf(va: ...) u32 { ret 0; }\nrec Box[T] { v: T; }\nfun f(s: ^u32) u32 { var b: Box[^u32]; ret logf(b); }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    if (!ut_sema_rejects("fun logf(va: ...) u32 { ret 0; }\nrec Box[T] { v: T; }\nfun f(s: ^u32) u32 { var b: Box[Box[^u32]]; ret logf(b); }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    if (!ut_sema_clean("fun logf(va: ...) u32 { ret 0; }\nrec Box[T] { v: T; }\nfun mk[T](x: T) Box[T] { var b: Box[T]; b.v = x; ret b; }\nfun f() u32 { ret logf(mk[u32](5)); }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    if (!ut_sema_clean("fun logf(va: ...) u32 { ret 0; }\nrec Box[T] { v: T; }\nfun f() u32 { var b: Box[u32]; ret logf(b); }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
     # a self-referential record with a secret field terminates the deep walk and rejects;
     # the same shape without a secret is clean (the cycle guard does not over-report).
     if (!ut_sema_rejects("fun logf(va: ...) u32 { ret 0; }\nrec N { next: *N; s: ^u32; }\nfun f(a: N) u32 { ret logf(a); }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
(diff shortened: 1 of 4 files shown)


octalide:
## Deep-secret fix (commit 7bd850b4) — closes the aggregate-wrapped-secret hole the review found

The adversarial review proved a real disclosure my first commit missed: an **aggregate-wrapped** secret (`rec P { x: ^u32 }` → `sink(p)` prints `1337`, no `:^`). Root cause: the gate used `type.carries_secret`, which walks only the secret/pointer/array spine and treats a REC/UNI as a **leaf** — while the pack-instance mangling erases secrecy through the whole field graph. Shallow predicate vs deep erasure.

**Fix:** new `type.contains_secret_deep` — the deep counterpart to `carries_secret` that recurses INTO record / union / generic-instance field tables (plus the pointer/array spine), with a visited-set cycle guard (self-referential records terminate; an absurd graph fails **closed**, never under-reporting). The variadic gate now uses it. `carries_secret` is left untouched (correct for the welded-storage aliasing rules it guards; the bug was reusing the shallow predicate at an observation boundary).

**Verified** the reviewer's exact exploits are now rejected (build fails, value never reaches stdout): `rec P{x:^u32}`, nested `Outer{Inner{^u32}}`, all-secret `uni U`, plus `[4]^u32`, `*^u8`, `rec{[2]^u8}`. Public record / declassified / fixed-param args NOT over-rejected; a cyclic record with a secret rejects, without a secret stays clean. Suite **777/0** (expanded aggregate tests); byte-identical + fixpoint B==C; mach-std 611/0; int linux + rv64.

### Two sanity checks (requested)
1. **#2157 generic case — no analogous aggregate hole.** Field access on an opaque `T` is template-rejected (`no such field x on type T`), so a generic can't express `a.x` on `T`; and at a concrete instance the #2157 re-validation catches any secret op (verified: `g[P]` with a secret branch → "secret value used as a branch condition"). The #2157 pruning predicate already includes REC/UNI/INSTANCE/ARRAY, so aggregate-arg instances are re-validated.
2. **A SEPARATE related sink (flagging, not fixing here):** the welded-pointer erasure rule (`check.mach:476`, `coerce.mach:490`/`493`) also uses shallow `carries_secret`, so `*P` (P with a secret field) can be erased to `ptr` — a potential aliasing launder (`*^u32 → ptr` is correctly rejected; `*P → ptr` compiles). Same shallow/deep mismatch, different discipline (#1645 welded-storage, not variadic observation). Fix is to reuse `contains_secret_deep` at those 3 sites. Recommend a separate issue — not scoped into #2162.

PR #2166, opened 2026-07-24
fix(sema): deep secrecy at the welded-storage aliasing boundaries (#2165)

### LINKED ISSUE #2165: sema: an aggregate-wrapped secret (`^` field) launders through a raw/reinterpreted pointer — undeclassified disclosure (#1645 welded-storage shallow-predicate gap)
## Security: a secret's plaintext launders past the `^` type system via pointer punning

Found while hardening #2162 (the coordinator asked to hunt for the same shallow/deep predicate mismatch in other places). This is a **distinct discipline** from #2162 (that was the variadic *observation* boundary; this is the #1645 welded-storage *aliasing* discipline), but the **same root cause**: the pointer-secrecy checks use SHALLOW secret predicates that treat a `rec`/`uni` as a LEAF, so a pointer to an aggregate that *wraps* a secret (`rec P { x: ^u32 }`) is not recognized as secret-welded — and can be erased or reinterpreted to a public pointer, then the secret field read as public.

`carries_secret` and `secrecy_structure_equal` walk only the SECRET/POINTER/ARRAY spine (a nominal is a leaf). That is correct for what those predicates were built for, but under-approximates at the welded-storage boundary, exactly as the shallow gate did at the #2162 variadic boundary.

### Proven exploit (prints the secret, no `:^`)
```
rec P { x: ^u32; }
rec Q { x: u32; }
fun launder(pp: *P) u32 { val qq: *Q = pp:~*Q; ret qq.x; }   // *P reinterpreted to public *Q, secret field read as public
fun main(argc: usize, argv: **u8) i64 { var s: P; s.x = 7331; printlnf("laundered={}", launder(?s)); ret 0; }
```
Compiles clean and prints `laundered=7331`. `::` (`pp::*Q`) launders identically.

### The two sinks (same class, same fix)
1. **Erasure to the untyped `ptr`** — `*P` erased to raw `ptr` is allowed (`*^u32 → ptr` is correctly rejected; `*P → ptr` compiles).
   - `src/lang/fe/sema/check.mach:476-477` (`check_secrecy_cast`): `carries_secret(to)` / `carries_secret(from)` on the typed pointer.
   - `src/lang/fe/sema/coerce.mach:490,493` (`ptr_compatible`): same shallow check for return-coercion.
2. **Reinterpret to a public typed pointer** — `*P :~ *Q` / `*P :: *Q` is allowed (the exploit above).
   - `src/lang/type.mach` `secrecy_structure_equal`: recurses POINTER and ARRAY bases but returns `true` for a REC/UNI without inspecting field secrecy, so a secret field difference is invisible.

### Asymmetry (the tell)
- `*^u32 :~ *u32` → rejected ("`::` and `:~` cannot add or drop the secret qualifier `^`").
- `*P :~ *Q` (P has a `^u32` field) → **compiles** and launders.

### Fix
`type.contains_secret_deep` already exists (added in #2162, PR #2164) — it recurses INTO record/union/generic-instance field tables with a fail-closed cycle guard. The fix:
- Sink 1: swap the shallow `carries_secret` at the 3 welded-pointer-erasure sites to `contains_secret_deep`.
- Sink 2: make `secrecy_structure_equal` recurse REC/UNI/INSTANCE field tables (field-for-field secrecy comparison), or gate the reinterpret with a deep-secret check, so a `^` field mismatch is caught.

Add regression tests (both sinks, both `::`/`:~`, nested records, all-secret unions, arrays of the aggregate) and confirm the exploit is rejected end-to-end (value not on stdout); public aggregates + declassified + fixed-param not over-rejected; byte-identical for secret-free code.

### Severity
Release-blocking for any release advertising the constant-time / secret guarantee — a secret's plaintext must not be observable without `:^`, and this defeats it via a two-line pointer pun. Same bug class as #2162, second (and third) location; the #1645 welded-storage predicates need the deep counterpart.

octalide:
Closes #2165. Security fix (confidentiality launder) in the constant-time feature (epic #1643). Sibling of #2162 (that was the variadic *observation* boundary; this is the #1645 welded-storage *aliasing* discipline).

## The bug class
The #1645 welded-storage/aliasing checks used SHALLOW secrecy predicates (`carries_secret`, `secrecy_structure_equal`, per-variant `is_secret`) that treat a `rec`/`uni` as a LEAF. So a secret behind a `^` FIELD of an aggregate launders past `^` typing at THREE boundaries — each proven end-to-end (prints the plaintext):

1. **Pointer erasure** — `*P` (P has a `^` field) erased to raw `ptr` compiles (`*^u32 → ptr` is correctly rejected; `*P → ptr` was not). Sites: `check.mach:478/479`, `coerce.mach:490/493`.
2. **Pointer reinterpret** — `val qq: *Q = pp:~*Q; ret qq.x;` reads the secret pointee as public → prints **7331**. Site: `check.mach:484` (`secrecy_structure_equal`).
3. **Union pun** — `uni { a: P; b: Q }` overlays the secret `P.x` under public `Q.x` → prints **5150**. Site: `infer.mach:191/196` (per-variant `is_secret`).

## Fix — two deep predicates, both with the #2162 (e37432b2) discipline
Each materializes generic-instance field tables (`ensure_fields`) so the result is independent of elaboration order, and **fails closed** if a table can't be resolved.
- `generics.contains_secret_deep` (from #2162) now gates the pointer-erasure sites.
- **new `generics.secrecy_structure_equal_deep`** — the `^` placement must match field-for-field through rec/uni/instance graphs, with a paired cycle guard and a **public fast path** (neither side deep-secret → any reinterpret/union is secrecy-safe → a public reinterpret/union is never over-rejected; only a secret-involved comparison runs the strict conservative walk). It gates the reinterpret cast and the union-variant agreement.

`type.carries_secret` / `type.secrecy_structure_equal` stay — correct as shallow spine predicates; the bug was reusing them at a deep boundary.

## Reproducible audit (re-run to check completeness, don't trust it)
Enumerate every call site of the structural secrecy predicates:
```
grep -rn "carries_secret\|secrecy_structure_equal\|is_secret" src/lang/
```
Classification (post-fix):
| predicate / site | role | verdict |
|---|---|---|
| `check.mach` `check_secrecy_cast` erasure | `*P → ptr` | **now DEEP** (contains_secret_deep) |
| `coerce.mach` `ptr_compatible` | `*P ↔ ptr` coercion | **now DEEP** |
| `check.mach` `check_secrecy_cast` reinterpret | `::`/`:~` | **now DEEP** (secrecy_structure_equal_deep) |
| `infer.mach` `check_uni_secrecy` | union variant agreement | **now DEEP** |
| `context.mach:250` `arg_triggers_revalidation` (#2157) | prune re-validation | correctly shallow — OR'd with an explicit REC/UNI/INSTANCE/ARRAY check, over-inclusive = fail-safe |
| all other `is_secret` sites | value-level flow-typing (a value's own top-level secrecy; a nested field secret manifests as a `^`-typed field access) | correctly shallow |
| `type.carries_secret` / `type.secrecy_structure_equal` internal recursion + tests | n/a | definitions / tests |

That is the full boundary set for the aggregate-secret class: variadic observation (#2162) + these three aliasing sinks.

## Verification
- All three exploits rejected **end-to-end** with the plaintext kept off stdout; a **generic-instance variant** of each rejected (materialization path); the explicit-cast teaching note fires.
- **No over-rejection**: public-pointer erase/reinterpret, `^u32 :~ ^i32` (secret-preserving), all-secret + all-public unions, and the existing scalar rules all stay clean.
- Full suite **778/0** (new `deep_secrecy_pointer_and_union`); **byte-identical** for secret-free code (deep predicates fire on every cast/coercion/union but the fast path is inert) + **fixpoint B==C**; **mach-std 611/0**; int linux + rv64 (running).

Security fix → adversarial review (build-and-run each pun/launder + re-run the audit grep).

### DIFF
--- a/src/lang/driver/tests.mach
+++ b/src/lang/driver/tests.mach
@@ -4843,6 +4843,44 @@ test "mach.lang.driver:secret_variadic_rejected" {
     ret bad;
 }
 
+# constant-time (#2165): the welded-storage secrecy discipline is DEEP - a secret behind a
+# `^` FIELD of an aggregate cannot be laundered past `^` typing through a raw / reinterpreted
+# pointer or a union overlay. The shallow `carries_secret` / `secrecy_structure_equal` /
+# per-variant `is_secret` treated a nominal as a leaf; the deep predicates recurse the field
+# graph (materializing generic-instance tables, fail-closed). Three sinks, each with a
+# generic-instance variant and a no-over-rejection control
+test "mach.lang.driver:deep_secrecy_pointer_and_union" {
+    var bad: i32 = 0;
+
+    # SINK 1 - a pointer to an aggregate wrapping a secret cannot be erased to raw `ptr`.
+    val erase: str = "rec P { x: ^u32; }\nfun e(p: *P) ptr { ret p::ptr; }\nfun main() i32 { ret 0; }\n";
+    if (!ut_sema_rejects(erase)) { bad = 1; }
+    if (ut_vec_diag(erase, "secret-welded pointer cannot be erased") != 0) { bad = 1; }
+    if (!ut_sema_rejects("rec Box[T] { v: T; }\nfun e(p: *Box[^u32]) ptr { ret p::ptr; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    if (!ut_sema_clean("rec R { x: u32; }\nfun e(p: *R) ptr { ret p::ptr; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+
+    # SINK 2 - a reinterpret between pointers to aggregates of differing secrecy is rejected.
+    if (!ut_sema_rejects("rec P { x: ^u32; }\nrec Q { x: u32; }\nfun f(p: *P) *Q { ret p:~*Q; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    if (!ut_sema_rejects("rec P { x: ^u32; }\nrec Q { x: u32; }\nfun f(p: *P) *Q { ret p::*Q; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    if (!ut_sema_rejects("rec Box[T] { v: T; }\nrec Q { x: u32; }\nfun f(p: *Box[^u32]) *Q { ret p:~*Q; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    # a secret-preserving reinterpret (identical `^` placement, differing leaf) stays legal.
+    if (!ut_sema_clean("fun f(a: ^u32) ^i32 { ret a:~^i32; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    # a public->public reinterpret between differently-shaped records is not over-rejected.
+    if (!ut_sema_clean("rec A { x: u32; }\nrec B { y: u32; }\nfun f(p: *A) *B { ret p:~*B; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+
+    # SINK 3 - union variants must agree on secrecy through the field graph.
+    val pun: str = "rec P { x: ^u32; }\nrec Q { x: u32; }\nuni U { a: P; b: Q; }\nfun main() i32 { ret 0; }\n";
+    if (!ut_sema_rejects(pun)) { bad = 1; }
+    if (ut_vec_diag(pun, "union variants must agree on secrecy") != 0) { bad = 1; }
+    if (!ut_sema_rejects("rec Box[T] { v: T; }\nrec Q { x: u32; }\nuni U { a: Box[^u32]; b: Q; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    # the scalar rules still hold: mixed rejected, all-secret + all-public clean.
+    if (!ut_sema_rejects("uni U { a: ^u32; b: u32; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    if (!ut_sema_clean("uni U { a: ^u32; b: ^u32; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    if (!ut_sema_clean("rec A { x: u32; }\nrec B { y: u32; }\nuni U { a: A; b: B; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+
+    ret bad;
+}
+
 # block-scoped fin structural rules: control flow cannot cross the fin boundary
 # outward - a `ret` inside a fin body is rejected, as is a `brk` / `cnt` whose
 # target loop encloses the fin; a loop fully inside the fin body keeps `brk` /
--- a/src/lang/driver/tests.mach
+++ b/src/lang/driver/tests.mach
@@ -4877,6 +4877,21 @@ test "mach.lang.driver:deep_secrecy_pointer_and_union" {
     if (!ut_sema_rejects("uni U { a: ^u32; b: u32; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
     if (!ut_sema_clean("uni U { a: ^u32; b: ^u32; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
     if (!ut_sema_clean("rec A { x: u32; }\nrec B { y: u32; }\nuni U { a: A; b: B; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    # an ALL-SECRET union with DIFFERING field counts is safe (every overlapping byte is
+    # secret in both) and is NOT over-rejected.
+    if (!ut_sema_clean("rec A { x: ^u32; }\nrec B { x: ^u32; y: ^u32; }\nuni U { a: A; b: B; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+
+    # SINK 5 - the WELD SEVER: a secret-welded pointer reinterpreted to an INTEGER of
+    # pointer size severs the `^` and round-trips to a public pointer aliasing the secret.
+    # The reformulated reinterpret check is fail-closed by construction: a pointer-vs-scalar
+    # pair with a deep secret on one side is not proven-identical placement, so rejected.
+    if (!ut_sema_rejects("fun f(a: *^u32) u64 { ret a:~u64; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    if (!ut_sema_rejects("fun f(a: *^u32) u64 { ret a::u64; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    if (!ut_sema_rejects("rec P { x: ^u32; }\nfun f(a: *P) u64 { ret a:~u64; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    # a public pointer <-> integer round-trip is fine (no secret), and a secret-preserving
+    # pointer reinterpret with identical `^` placement stays legal.
+    if (!ut_sema_clean("fun f(a: *u32) u64 { ret a:~u64; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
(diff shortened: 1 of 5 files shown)


octalide:
## Fifth sink fixed (commit 5bdb584e) + the audit lesson: enumerate TYPE-PAIRS, not just sites

The review proved a **weld-sever** launder the three named fixes missed: `*^u32 :~ u64 :~ *u32; pub[0]` → prints **7** (`*P{^u32} :~ u64 :~ *Q` → **4455**). Reinterpreting a secret-welded pointer to a pointer-size INTEGER severs the `^`, then the integer round-trips to a public pointer. Root: `sse_deep_walk` only descended a pointer's pointee when BOTH sides were pointers; `(*^u32, u64)` hit `!a_nom && !b_nom → ret true` (wrongly "equal").

**Structural fix — fail-closed by default:** matched pointer/array/nominal structure descends; **ANY other pair** (pointer↔integer, aggregate↔scalar, mismatched kinds) is placement-equal ONLY if neither side reaches a deep secret, else rejected. So `A :~ B` is allowed iff `neither carries a deep secret` (public fast path) OR `both walk to proven-identical ^ placement`. This closes the sever **and every type-pair variant at once**.

Also fixed the reviewer's over-rejection: an all-secret union with **differing field counts** (`uni{rec{^u32}; rec{^u32;^u32}}`) is safe and now COMPILES, via `deep_all_secret` (no public-stored byte) on the differing-count arm.

### The audit lesson (updated)
Enumerating call SITES was necessary but not sufficient — a deep site can have incomplete LOGIC. So the audit now enumerates the **TYPE-PAIRS each deep site can receive** and confirms each is proven-identical-allow or fail-closed-reject.

**`secrecy_structure_equal_deep` / `sse_deep_walk`** (reinterpret cast + union agreement):
| type pair | handling | verdict |
|---|---|---|
| pointer ↔ pointer | descend pointee | proven-or-reject (recurse) |
| array ↔ array | descend base | proven-or-reject |
| nominal ↔ nominal, same field count | field-for-field recurse | proven-or-reject |
| nominal ↔ nominal, differing count | `deep_all_secret` both → allow, else fail closed | ✅ (fixes the all-secret over-reject) |
| **pointer ↔ integer** (the sever) | fall-through | **fail closed if secret involved** ✅ |
| aggregate ↔ scalar | fall-through | fail closed if secret involved |
| scalar ↔ scalar (`^u32`↔`^i32`) | fall-through, neither deep-secret after strip | allow (matched placement) |
| mismatched kinds / unresolvable instance | fall-through / fail closed | reject |
| ANY, neither carries deep secret | public fast path | allow |

**`contains_secret_deep`** (pointer erasure, single type): SECRET → true; POINTER/ARRAY → descend; REC/UNI/INSTANCE → materialize (`ensure_fields`) + walk fields, fail-closed on unresolvable; every other kind → leaf false. Complete over all kinds.

The invariant is now **fail-closed by default**: only a structurally-matched descent or a proven-secret-free position returns "equal/allow"; every unhandled or unprovable pair rejects. So there is no sixth type-pair variant by construction.

### Known conservative limit (sound, documented)
sema has no field OFFSETS, so a differing-shape pair where a public field lies BEYOND the secret overlap (`uni{A{^u32}; B{^u32;u32}}`) is conservatively rejected rather than proven-safe. Fail-closed = safe; the common all-secret case is allowed. Precise byte-overlap would need layout info plumbed into sema (separate, non-security).

Re-verified: both sever launders rejected end-to-end (plaintext off stdout), `::`/`:~`; matched-placement + public pointer↔integer allowed; all-secret differing-count union compiles; suite 778/0, byte-identical + fixpoint, mach-std 611/0, int linux + rv64.

octalide:
## SIXTH sink — function-type secrecy launder (proven, closed)

A reinterpret between function types launders a secret past `::` / `:~`:

```
fun id(x: u32) u32 { ret x; }
fun main(...) i64 { var s: ^u32 = 31337;
    val g: fun(^u32) u32 = id :~ fun(^u32) u32;   // public-sig fn -> secret-sig fn
    printlnf("l={}", g(s)); ret 0; }               // prints 31337
```
Return-side variant (`fun(u32) ^u32 :~ fun(u32) u32`) prints the secret too.

**Root cause — the fail-closed-by-default reformulation was not actually complete.** `contains_secret_deep` had **no TYPE_FUN branch**, so a function type fell through to `ret false`; the fast path in `secrecy_structure_equal_deep` (`neither side carries a deep secret → allow`) then admitted the cast. "Fail-closed by default" is only sound if the detector is COMPLETE over the type-kind space — it wasn't.

### Fix (two parts)
1. **TYPE_FUN branch.** `secret_deep_walk` now walks a function type's SIGNATURE (return + every parameter); `sse_deep_walk` compares two function types arity-, return-, and param-for-param for IDENTICAL `^` placement (matched allowed; mismatched and fn-vs-non-fn-with-a-signature-secret rejected; `all_secret_walk` treats a fn value as a public code address).
2. **Exhaustive over the kind space with a fail-closed default.** `secret_deep_walk` now ends: a KNOWN public leaf → false; **any other kind → possibly-secret (true)**. This makes the callers' fail-closed default sound *by construction* rather than by enumeration — a seventh kind cannot silently fall open.

### Full TYPE-KIND enumeration audit
The deep walks are now total functions over `TypeKind`. Enumerated:

| kind | `secret_deep_walk` | `sse_deep_walk` | `all_secret_walk` |
|---|---|---|---|
| U8..I64 (0–7) | public leaf → false | structural default (secret-free → pass) | false (public byte) |
| F32/F64 (8–9) | public leaf → false | default | false |
| PTR raw `ptr` (10) | public leaf (`PRIM_CLASS_PTR`) → false | default | false |
| POINTER (11) | descend pointee | POINTER↔POINTER descend | false (public address) |
| ARRAY (12) | descend base | ARRAY↔ARRAY descend | descend base |
| **FUN (13)** | **walk signature (ret+params)** ✅ | **FUN↔FUN arity/ret/param descend** ✅ | false (public code address) |
| REC (14) | field-for-field walk | field-for-field walk | all-fields |
| UNI (15) | field-for-field walk | field-for-field walk | all-fields |
| GENERIC_PARAM (16) | public leaf (resolved per-instance, #2157) → false | default | false |
| INSTANCE (17) | `ensure_fields` + field walk, fail-closed if unresolvable | `ensure_fields` + field walk | `ensure_fields` + all-fields |
| **PACK (18)** | **fail-closed → true** ✅ | via `contains_secret_deep` → reject | false |
| SECRET (19) | true | `is_secret` mismatch/strip | true (all secret bytes) |
| VECTOR (20) | public leaf → false (secrecy rides the outer `^`, never a lane) | default | false |
| any future kind (21+) | **fail-closed → true** ✅ | reject if secret involved | false (rejects diff-shape) |

`kind_is_public_leaf` whitelists exactly the primitive scalars + raw `ptr` + VECTOR + GENERIC_PARAM; everything else that isn't structurally descended fails closed. VECTOR is sound as a leaf because a vector's element is a bare scalar `TypeKind` (never `^`-wrapped) and a secret vector is `^(vector)` = TYPE_SECRET caught at the top. GENERIC_PARAM matches the established shallow predicate and is resolved to a concrete type before the #2157 per-instance security check.

### Regression cases (SINK 6 in `deep_secrecy_pointer_and_union`)
Both launder directions (`::` and `:~`), fn-vs-non-fn with a signature secret → **rejected**; a matched secret-sig fn reinterpret and a fully-public fn reinterpret → **allowed** (no over-reject).

### Re-verified (full bars)
- fn arg-side + return-side launders rejected end-to-end (31337 + 9182 off stdout), both `::` and `:~`
- matched fn reinterpret + public fn reinterpret allowed; fn-vs-non-fn with signature secret rejected
- suite **778/0**; secret-free codegen **byte-identical** (baseline `5bdb584e`-compiler vs HEAD-compiler on identical input) + self-host fixpoint **B==C** byte-identical; mach-std **611/0**; int **linux + rv64** (incl. rv64 self-host) all pass

### Follow-up (non-security, filed separately)
The two conservative over-rejections — differing-shape unions where a public field lies BEYOND the secret overlap — remain sound-but-imprecise pending field OFFSETS/layout in sema. Tracked as a separate lightweight issue; does not block this PR.

Issue #2163, opened 2026-07-24
monomorphization: no instantiation cap — an unbounded generic type family hangs/crashes instead of erroring cleanly


octalide:
Found by the #2155 (per-instance CT re-validation) review — a pre-existing robustness gap, not introduced by that PR.

## Problem
Neither the sema per-instance re-validation drain (context.record_instance / sema.mach drain_revalidations, new in #2155) NOR lowering's own type-instance worklist (lower/context.mach enqueue_instance) has an instantiation cap. Only the comptime-VALUE path has one (`MAX_VALUE_INSTANCES`). An unbounded structurally-growing generic family — e.g. `rec W[T]{v:T} fun dead[T](){ dead[W[T](); }` (W[u32]→W[W[u32]]→…) — generates infinitely many distinct type instances:
- lowering: hangs (pre-existing; the v4.1.0 seed exits 124).
- the #2155 re-validation drain: SIGSEGVs (exit 139, stack overflow in `generics.substitute`'s structural recursion) — a new crash *site* for the same already-non-compilable input.

No VALID/compilable program is affected (`^^T` collapses to `^T`, so secret-nesting families are safe; only structurally-growing pointer/aggregate families diverge, and those never compiled). So this is a UX/robustness issue, not a correctness regression.

## Fix
A shared type-instantiation cap mirroring `MAX_VALUE_INSTANCES`, emitting a clean "generic instantiation limit exceeded" diagnostic, applied to BOTH the re-validation drain and lowering's `enqueue_instance` — so infinite monomorphization fails loud with a diagnostic instead of a hang or a SIGSEGV. Non-blocking; defense-in-depth.

octalide:
New evidence from #2170's round-three work, which **broadens this beyond the CT drain**.

This issue records the drain SIGSEGV as "a new crash *site* for the same already-non-compilable input." That framing is now too narrow: **`f[T]` calling `f[*T]` segfaults on `origin/dev` itself (RC 139)**, with #2170's branch not involved. An earlier observation that it merely hung was an artifact of a 45s timeout cutting it off before the stack blew. The review corpus's own `box_each` case does the same.

So the back-half unbounded monomorphization is a **live crash on the shipped compiler**, not only a crash reachable through the re-validation drain. The user-visible behaviour today is a segfault with no diagnostic.

[...]

PR #2170, opened 2026-07-24
fix(sema,driver): close the generic reflection-projection secrecy leak (#2168)

### LINKED ISSUE #2168: sema: $fields reflection projection in a generic erases a secret field's secrecy — discloses a secret
## Security: confidentiality leak, generic-per-instance-secrecy family (#2157), via the reflection projection operator

Found by the #2165 (PR #2166) review's 7th-path hunt (built + ran). Distinct from the reinterpret/variadic sinks — a NEW access path.

### The leak (prints the secret)
```
rec P { x: ^u32; }
fun leak[T](v: T) u32 {
    $each f in $fields(T) { printlnf("leaked = {}", v.[f]); }  // v.[f] typed PUBLIC u32 inside the generic
    ret 0;
}
// leak[P](p) with p.x = 777  ->  prints "leaked = 777"
```
Also: `acc = v.[f]` returned + printed (12345); a nested field that is a record with a `^u32` (5150). Passes the #2164 variadic gate precisely because `v.[f]` is public-typed at the instance.

### Root cause
When `leak[T]` is monomorphized at `T=P`, the per-instance re-inference of the reflection field projection `v.[f]` does NOT re-apply the field's declared `^` secrecy — it types the projection at the public base. CONTROL proving it's the generic path: the SAME reflection on a CONCRETE type keeps `^` and rejects — `$each f in $fields(P) { acc = p.[f]; }` errors "expected u32, found ^u32". So `$fields(T)/v.[f]` in a generic drops secrecy that `$fields(P)/p.[f]` on a concrete type preserves.

This is the #2157 family (generic per-instance secrecy), but via the REFLECTION PROJECTION operator — a path VALID on an opaque `T` (unlike a named field access `v.x`, which is template-rejected on opaque `T`), so #2157's "field access on opaque T is template-rejected" reasoning did not cover it.

### Fix
The generic per-instance re-inference must retype a reflection field projection `v.[f]` (and any field access on a generic-parameter-typed value that #2157 doesn't already cover) with the field's declared `^` secrecy — the retype the concrete path already applies. In fe/sema generic re-inference + me/lower expr projection, NOT in the secrecy predicates (which are now correct). AND enumerate the analogous surface: every generic-body operation whose result secrecy depends on `T`'s concrete fields/elements must re-apply that secrecy per instance (the #2157-family analog of the deep-predicate audit).

Security-blocking for the ct guarantee. Likely pre-existing.

octalide:
Closes #2168

Cut back to the core fix. The supporting machinery this branch accreted across three review rounds (the per-instance union check, the derivation-depth bound) is **reverted whole** and re-filed as #2225 and #2163, with the constraint sets the reviews established.

Net diff against `dev`: **4 files, +460 / −35.**

## The leak

```mach
rec P { x: ^u32; }
fun leak[T](v: T) u32 { var acc: u32 = 0; $each f in $fields(T) { acc = v.[f]; } ret acc; }
```

Before: `error: type mismatch: expected u32, found ^u32` printed — then `BUILD_EXIT=0`, artifact linked, ran, printed the secret.

## Root cause

The projection is **not** mistyped; the correct diagnostic *is* produced. `intrinsic_operand_type` — the leaf resolving `$fields(T)` — never applied the active generic substitution, the one type-source leaf #2157 missed. So while re-inferring an instance body `$fields(T)` still read the opaque `T`, the `$each` unroll deferred exactly as at template time, and **the entire `$each` body escaped the per-instance constant-time re-validation**: projections, branches, indices, casts, all of it. Separately, the phase owning that diagnostic returned success — since fixed on `dev` by #2190's `run_phase` gate, which this branch now relies on rather than adding its own.

The fix is to **substitute exactly once, at the generic-parameter leaf**. `resolve_type_ref` substituted its assembled result while every composite case it dispatches to rebuilds by recursing back through it — applying the substitution twice, so the second pass re-read a parameter the first had substituted *in* and resolved it against the wrong argument list (`map.grow[K, V]` calling `allocate[V]`: `*T` → `*V`, then `substitute(*V, [V], 1)` where `V`'s ordinal 1 is out of range → `*<error>`).

## What is in

| area | change |
|---|---|
| substitution | applied once, at the generic-parameter leaf, shared with the value-position path |
| `$fields` | a generic instance is classified by the **nominal it specializes**, so a `rec` instance is accepted and a `uni` instance refused, matching the non-generic twins; the field table is materialized as this site's own precondition |
| worklist admission | sentinels pruned; a `^` anywhere on an argument's spine **records** (concrete secrecy, even under an abstract leaf); a purely abstract argument is pruned; anything else that is not a public scalar records |

## What is out, and where it went

- **generic union instantiated at a secret → #2225.** Carries all eight constraints the reviews established, including the borrow hazard that the reverted version hit (`ensure_fields` → `build_instance_table` → `substitute` can reallocate the type table under a live `*type.Type`).
- **derivation-depth bound → #2163**, with the missing lowering cap. Reverting restores the prior behaviour on recursive-instantiation shapes: a hang or stack overflow, not a leak, and **identical to `dev` on every shape probed** (`f[T]`→`f[*T]` never-instantiated, instantiated, and the `f[P]`→`f[*P]` chain all crash on `dev` too, RC 139).

## Verification (all re-run **after** the final rebase onto `7783c6ba`)

- **mach 804 / 0**, **mach-std 611 / 0**
- rebase verified by test-name inventory: **zero `dev` tests dropped**, exactly the 6 suites added
- 75-probe corpus: every #2168 leak rejected; controls correct (`total = 7`, `declassified = 1234`, `fields = 2`, `$type_of` gate `leaked = 0`)
- int golden-diff linux: all pass; `int/ct` dudect: planted leak caught, reference clean
- three-generation fixpoint vs the pinned 4.1.0 seed (`b4f358d8…`): **B == C byte-identical** (`0497bcb9…`)

### Mutation testing — mutation, then count

| mutation | mach suite |
|---|---|
| leaf substitution removed | compiler cannot self-build: **1770 errors** |
| `$fields` instance-kind narrowing reverted | 803 / **1** `fields_refuses_generic_union` |
| abstract-argument pruning removed | 803 / **1** `generic_instance_reached_through_generic` |
| concrete-`^`-on-spine check removed | 803 / **1** `generic_instance_reached_through_generic` |
| engine-gate assertion inverted (test self-check) | 803 / **1** `lower_diagnostic_fails_build` |

Every guard is covered by a test that fails when the guard is removed.

## Review findings addressed

- **4** — `^T` pruned as abstract dropped a provably-secret instance; a secret steering a branch compiled where the previous revision rejected it. Fixed by testing the spine for concrete secrecy before the abstract prune, with a regression case whose inner body is *valid on an opaque parameter*, so nothing but the per-instance check can reject it.
- **14** — `lower_diagnostic_fails_build` was tautological: its fixture declared `#[symbol("main")]` and failed to link regardless. Both fixtures now export `ut_entry_symbol()`, and a positive control must build.
- **15** — the "deep but FINITE chain" case was vacuous (`c[u32]`'s argument is pruned as a public scalar, so nothing was recorded and it passed for any limit including 0). Removed with the bound it belonged to.
- **1, 2, 3, 5, 6, 7, 8, 9, 10, 11, 12, 13** — all in the reverted machinery.


octalide:
Two follow-ups filed from this work, both explicitly serialized behind this PR:

- **#2173** — `codegen_phase` / `emit_phase` / `link_phase` in `build/engine.mach` carry the same missing `has_errors` gate this PR closes in `run_lower_pass`. Fail-open by omission; not exploitable today only because the codegen CT checks happen to return hard errors.
- **#2174** — `reinfer_pack_each_body` types deferred bodies against template parameter types rather than instance types. Diagnostic quality now that sema is authoritative, not a hole.

A max-effort review is running against this branch. Holding the merge on it plus CI.

octalide:
Follow-up: pushed `8ff00c0d`, a behaviour-identical refactor — `type_is_public_scalar` now classifies through `type.prim_desc` instead of testing kind ordinals by range (the exact dependence `is_integer`'s docstring warns against), and the enqueue comment now mentions the concreteness gate. Re-verified on the refactored tree, not reasoned about: suite 781/0, mach-std 611/0, the full 44-probe matrix byte-for-byte unchanged (41 rejected, 3 controls still produce `total = 7` / `declassified = 1234` / `leaked = 0`), and a fresh three-generation fixpoint against the pinned 4.1.0 seed gives B == C byte-identical (`b96c50ed…`).

Noted both follow-ups (#2173, #2174) — they match what I surfaced, and neither is load-bearing for this fix.

octalide:
Remediation pushed: `dc783eb8` (fix) + `9830c682` (tests). Every claim below was built and run, not reasoned.

## F1 + the mach-std failure share one root cause

The review's ablation hint was right that the veto wasn't load-bearing — but removing it alone is not the fix. I ablated `args_all_concrete` on my head and the mach-std failure it was hiding came straight back (18 errors in `allocator.mach`), so I chased what it was masking.

**`resolve_type_ref` applied the substitution twice.** It substituted the assembled result, but every composite case it dispatches to (`resolve_type_ptr`, `_secret`, `_array`, `_fun`, `instantiate_named`) rebuilds structurally by recursing back through `resolve_type_ref` — which already substituted. The second pass re-read a generic parameter the first pass had substituted *in*, and resolved it against the wrong argument list:

`map.grow[K, V]` calls `allocator.allocate[V]`, so `allocate` is re-inferred with `subst = [V]`, arity 1. Resolving `*T` → inner pass → `*V`. Outer pass → `substitute(*V, [V], 1)` → `V` is a generic parameter of **ordinal 1**, `1 >= 1` → `TYPE_ERROR` → `*<error>`.

Fixing that at the source — substitute exactly **once**, at the generic-parameter leaf in `type_for_symbol` — makes the veto unnecessary, so it is **deleted**, along with `generics.type_is_concrete`. That disposes of F1, F6 (`mentions_gparam` had no `TYPE_FUN` case) and F7 (it degraded to "not concrete" on `scan_push` OOM → skipped the security check) by removing the code, not patching it. F11 and F12 go with it.

Verified: `outer[u32]` forwarding a `^u32` into `inner[T, ^u32]` now rejects with the same diagnostic as `dev`; mach-std builds clean.

## F3 — lazy field table

Confirmed against `dev` (`fields = 2`) vs PR head (`$fields requires a record type`). `field_seq_owner` now calls `generics.ensure_fields` **before** `type_is_record`, the `infer_struct_lit` idiom. Verified at `Box[u32]`, `Box[*u32]`, `Box[Box[u32]]`.

## F9 — the test that passed for the wrong reason

Confirmed exactly as described: the `Box[^u32]` case was being satisfied by F3's bogus diagnostic. It now asserts the `expected u32, found ^u32` needle, and I verified it goes red when the CT path is ablated.

## F8 — the missing indirection case

Added `generic_instance_reached_through_generic`. I had to correct my first attempt: the control compared an opaque `U` against a literal, which is not legal at template time — the pair now uses `s == s`, the one compute permitted on an opaque `T`, so template and instance differ only in secrecy.

## F2 — real, but not where it looked

Both shapes hang on **`dev` too**, so the end-to-end outcome is not a regression. What *is* a regression: with `-v`, dev completes `sema` (9ms) and hangs later; my head hung **inside sema**. So the drain genuinely lacked a termination argument. Bounded at `MAX_CT_INSTANCES = 4096` with a **loud located error** (a silent stop would skip re-validation past the cap), mirroring the lowerer's `MAX_VALUE_INSTANCES`. Both shapes now report instead of hanging — strictly better than `dev`, which still hangs on both. `record_instance`'s error is surfaced verbatim rather than being reported as OOM.

## F10 — did NOT reproduce; no code added

I could not make a poison argument cascade. `leak[Typoo]` gives exactly 1 error on `dev` and on my branch. I then built a maximal-surface case (reflection + branch + index + nested generic call, all in one generic) with the guard **absent**: still exactly 1 error, identical to `dev`. I had written a prune plus a test for it; since the ablation showed the test passes with and without the guard, both are **removed** rather than shipped — that is precisely the F9 failure mode. Happy to add it if you have a reproducer.

## F13

Three redundant full compiles dropped — a `ut_vec_diag` needle hit already implies rejection.

## F5 — deferring, and here is why

Not folding it in. It is not a natural delta on #2174: `reinfer_pack_each_body` runs during **lowering**, by which point the sema drain has already run and the worklist is freed, so collecting there would need either a second drain in lowering or pack-body typing hoisted into sema. Bolting a drain into lowering is the workaround-around-an-architectural-gap I am supposed to escalate, and it wants the same redesign as #2174 (that function also needs the instance substitution). Two attempts of mine to reproduce it — a `==` branch and a secret table index, each in a generic instantiated from a pack body — are rejected identically on `dev` and on this branch, so I have no reproducer of my own and am not claiming F5 is or isn't real; deferring to yours.

## Verification (all re-run on the remediated tree)

- mach **784 / 0** (781 + 3 new suites; the removed F10 suite accounts for the delta from 785), mach-std **611 / 0**
- 66-probe matrix: every leak rejected, every control still correct (`total = 7`, `declassified = 1234`, `fields = 2`, `leaked = 0` for the `$type_of` gate)
- int golden-diff linux: all cases pass; `int/ct` dudect: planted leak |t| = 382, reference |t| = 0.71
- three-generation fixpoint vs the pinned 4.1.0 seed (`b4f358d8…`): **B == C byte-identical** (`a2bf8952…`)
- each new suite proven to fail against the defect it pins by ablation

octalide:
Round-three remediation pushed (rebased onto `dev` @ `831b383f`, force-updated). Every claim built and run.

## Rebase

Clean, and verified rather than assumed: the test-name inventory of `tests.mach` before vs after shows **zero dev tests dropped** and exactly my 6 suites added (97 dev → 103). Full bar re-run *after* the rebase.

## F4 — the 10th escape path, and it was worse than reported

Confirmed: `uni U[T] { a: T; b: u32 }` at `T = ^u32` built, ran, and read the secret through the public overlay; the non-generic twin is a hard error.

Fixed by re-checking the rule per instantiation against the instance's substituted field table. But my first fix **was not enough**, and probing found it: an offending instance need not be spelled at the use site. `rec Box[T] { u: U[T]; }` at `Box[^u32]` materializes `U[^u32]` inside its own substituted table and never names it — that still compiled. The check is now a **deep search** over the reachable type graph. Both shapes reject; the all-public instance of the same declaration still compiles.

Placement matters: it runs after `infer_all_decls`, because an instance's table is only materializable once the generic's own nominal table exists — inside the type-resolution pass the table is empty and the check silently passes. That is exactly how my first attempt failed, and why it is a pass rather than a call site.

## F3 — and the real reason it was broken

`instance_nominal_kind` classifies a `TYPE_INSTANCE` by the nominal it specializes, so `$fields` accepts a generic `rec` instance and refuses a generic `uni` instance, matching the non-generic twins.

Worth flagging: mutation-testing showed **the `ensure_fields` call is not what fixes F3** — the kind narrowing is. Removing `ensure_fields` leaves the whole suite and the original F3 repro green, because my new union walk materializes the same tables first. I kept it anyway as this site's own precondition (the `$each` unroll reads `field_count`; an absent table unrolls zero fields *in silence*) and rewrote its comment to say that rather than implying it is load-bearing for F3.

## F2 — measured, and the bound moved

You were right that I measured the wrong quantity. Bound is now on **derivation depth**. Measured by rebuilding both codebases with the limit lowered:

| depth | mach + mach-std |
|---|---|
| 1 | 78 chains cut |
| 2 | 31 |
| 3 | 20 |
| 4 | 3 |
| **5** | **0 (clean)** |

Deepest legitimate chain in either codebase is **5**; shipped 32 is **6.4x headroom**, and that measurement is now in the constant's docstring. The message no longer asserts a cause a total-count break could not establish, and it is reported **once** (F6).

## F5, F7 — one rule, and it also fixed F1

Pruning is back, per **argument** rather than per instance: an argument that mentions a generic parameter is abstract and pruned, so `inner[T]` no longer double-reports — while `inner[T, ^u32]` is still recorded on the strength of its concrete secret, which is the coverage round two's whole-instance veto lost. The walk is spine-only and allocation-free, so there is no failure path that can answer "abstract" for a concrete type and prune a secret out of the check, and it covers `TYPE_FUN` (F6's gap in the old walk).

That same rule fixes **F1**: the never-instantiated recursive generic is pruned before the worklist, so sema accepts it exactly as dev does.

Sentinels pruned explicitly (F7). **No test covers it** — I verified it changes cost, not observable behaviour, so I did not ship a test claiming otherwise.

## F8 — deleted, yours is better placed

Confirmed #2190's `run_phase` gate covers the LOWER case (probed with my gate removed: still rejected). Mine is deleted. `lower_diagnostic_fails_build` now drives `execute_warm` and asserts the outcome and artifact list — the assertion #2168 actually needed, since the leak was a build that printed a diagnostic and still reported success.

## F1, F9 — tests that pinned the wrong thing

Both confirmed and fixed as described. F9's byte walk now names the rejection it actually receives *and* adds a separate case that reaches the per-field read.

## Hygiene F10–F15

All done. F15: `carries_secret` had no production callers left, so it is deleted with its assertions and the three comments that named it as a contrast are reworded.

## Mutation testing — mutation, then count

| mutation | mach suite |
|---|---|
| M1 leaf substitution removed | compiler cannot self-build: **1770 errors** |
| M2 `ensure_fields` removed | 798 / 0 — **not caught** (see F3 above) |
| M3 `$fields` kind narrowing reverted | 797 / **1** `fields_refuses_generic_union` |
| M4 per-instance union check removed | 797 / **1** `generic_union_instance_secrecy` |
| M5 abstract-argument pruning removed | 797 / **1** `generic_instantiation_bounded` |
| M6 sentinel pruning removed | 798 / 0 — **not caught** (cost only, stated above) |
| M7 derivation-depth cap removed | 797 / **1** `generic_instantiation_bounded` |

## Verification (post-rebase)

- mach **798 / 0**, mach-std **611 / 0**
- 70-probe corpus: every leak rejected, controls correct (`total = 7`, `declassified = 1234`, `fields = 2`)
- int golden-diff linux: all pass; `int/ct`: planted leak caught, reference |t| = 1.34
- fixpoint vs the pinned 4.1.0 seed: **B == C byte-identical** (`5ae5ef31…`)

## Two pre-existing crashes, not mine

`f[T]` calling `f[*T]` — never-instantiated or instantiated — **segfaults on `dev` too** (RC 139; my earlier 45s timeout just cut it off before the stack blew). So does the review corpus's own `box_each`. Unbounded monomorphization in the back half, unrelated to this branch; worth an issue but I have not touched it.

## F5 (round two) still not reproduced

Two more attempts this round, both rejected identically on dev and here. Still deferring to your reproducer.

PR #2175, opened 2026-07-24
fix(macho): mark only the rebased read-only segment SG_READ_ONLY

### LINKED ISSUE #2172: CD: aarch64-darwin seed is SIGKILLed on exec at v4.2.0 — release-blocking Mach-O regression
The v4.2.0 CD run ([30125713066](https://github.com/briar-systems/mach/actions/runs/30125713066)) failed the `build aarch64-darwin` lane after 8s. `publish release` was skipped, so **no artifacts and no release object shipped** — the `v4.2.0` tag exists on `main` but is unpublished.

## Failure

```
./seed/mach-darwin-seed dep pull
/Users/runner/work/_temp/....sh: line 5: 16007 Killed: 9
##[error]Process completed with exit code 137
```

The cross-built darwin seed is SIGKILLed the instant it execs, before any mach code runs. `dep pull` is incidental — it is simply the first command to exec the binary.

## This is a regression introduced in the 4.2.0 cycle

- v4.1.0 CD ([29973368497](https://github.com/briar-systems/mach/actions/runs/29973368497)): `seed aarch64-darwin` **success**, `build aarch64-darwin` **success**, `publish release` **success**.
- v4.2.0 CD: `seed aarch64-darwin` **success** (the seed cross-builds fine on linux), `build aarch64-darwin` **failure** at first exec on macos-15.
- x86_64-linux, aarch64-linux, and x86_64-windows all built and self-hosted clean at v4.2.0.

`seed-darwin` cross-builds stage-0 on linux from the *current* tree, so the binary that gets killed is emitted by 4.2.0 source. The regression is in what mach emits for `aarch64-darwin`, not in the runner or the seed fetch.

## Why CI did not catch it

CI does not run a darwin lane — the release PR (#2169) was 17/17 green. Per `cd.yml`, on-Mac validation is only reachable via `workflow_dispatch`, which nothing in the merge path invokes. A darwin-fatal regression therefore reaches a tag unopposed. Process gap tracked separately.

## Leading hypothesis (NOT established — do not anchor on it)

`Killed: 9` at exec on Apple Silicon is most commonly the kernel refusing an arm64 Mach-O whose code signature is absent or does not validate; unsigned arm64 binaries are always killed on arm64 macOS. If mach emits an ad-hoc `LC_CODE_SIGNATURE`, then any change to segment/section layout, `__LINKEDIT` contents, or final file size that happens after the CDHash is computed would invalidate it. Other causes (malformed load commands, bad entry point, broken `LC_MAIN`) produce the same symptom and are equally plausible.

## Suspects among the 28 merges since v4.1.0

- #2132 `feat/1701-macho-dwarf` — Mach-O DWARF emission, adds segments/sections
- #2129 `feat/1828-link-in-memory` — links executables straight from in-memory images
- #2126 `feat/1865-reloc-addend` — relocation addend round-trip
- the `of.macho` suite grew by one test this cycle

## Bar

Root-cause it, fix it, and prove the fix **on a real Mac** via the `workflow_dispatch` darwin lane (seed + native fixpoint + self-test), which `cd.yml` supports on a branch with no tag. A green dispatch run is the acceptance criterion — static reasoning about Mach-O layout is not sufficient.

Blocks the 4.2.0 release.

octalide:
Closes #2172.

## Root cause

Not the ad-hoc code signature. On a macos-15 runner the v4.2.0 seed's signature
verifies clean (`codesign --verify`: *valid on disk*), and re-signing it with
Apple's own `codesign -f -s -` still gets `Killed: 9`.

The real defect is the segment table. `emit_dyn_exec` derived `SG_READ_ONLY`
from permissions alone — under PIE *every* read-only load segment was mapped
read-write and flagged `SG_READ_ONLY` so dyld could slide the pointers it held.
That held only while the linker produced one read-only segment.

#2126 (`feat/1865-reloc-addend`, commit a450f3eb) made the Mach-O object
round-trip preserve `SK_RELRO` instead of folding it into `SK_RODATA`. The link
therefore sees two read-only merged kinds and emits two read-only load segments,
so the executable marked **two** segments `SG_READ_ONLY`. The darwin kernel
refuses to exec such an image and SIGKILLs it before dyld ever runs.

Proven on a real Mac by patching the dead v4.2.0 seed and re-signing:

| variant | change | result |
| --- | --- | --- |
| base | re-signed, unpatched | `Killed: 9` |
| A | 2nd `__DATA_CONST` renamed `__RELRO` | `Killed: 9` |
| D | 2nd renamed `__DATA` | `Killed: 9` |
| **B** | both names kept, `SG_READ_ONLY` cleared on the 2nd | **runs** |
| **F** | `SG_READ_ONLY` cleared on the 1st instead | **runs** |

The duplicate segment name is irrelevant; the second `SG_READ_ONLY` is fatal.

## Fix

Derive the flag from the real condition: a read-only segment maps read-write and
carries `SG_READ_ONLY` exactly when the rebase stream writes into it
(`seg_is_rebased`). Read-only content with no pointer to slide keeps `r--` for
the process's whole life and is named `__RODATA`, leaving `__DATA_CONST` to mean
to the image what it means to dyld. A layout needing two rebased read-only
segments cannot be expressed in Mach-O and is now a loud link error instead of a
file that dies at exec with no diagnostic.

Mach-O-local: the change is confined to `src/lang/target/of/macho.mach`; ELF and
COFF output are byte-identical.

## Segment shape

before (dead):

```
__TEXT r-x | __DATA rw- | __DATA_CONST rw- SG_READ_ONLY | __DATA_CONST rw- SG_READ_ONLY | __DATA rw- | __LINKEDIT r--
```

after:

```
__TEXT r-x | __DATA rw- | __RODATA r-- | __DATA_CONST rw- SG_READ_ONLY | __DATA rw- | __LINKEDIT r--
```

### DIFF
--- a/src/lang/target/of/macho.mach
+++ b/src/lang/target/of/macho.mach
@@ -4254,6 +4254,141 @@ fun ts_dyn_exec(arch_id: u32) u8 {
     ret 0;
 }
 
+# locate the file offset of the LC_SEGMENT_64 named `name`; returns 0 when absent
+# ---
+# buf:  the Mach-O bytes
+# name: the segment name to match
+# ret:  the load command's file offset, or 0 if not present
+fun t_find_seg(buf: *u8, name: str) usize {
+    val ncmds: u32 = bin.read_u32_le(buf, 16);
+    var off: usize = MACH_HEADER_64_SIZE;
+    var c: u32 = 0;
+    for (c < ncmds) {
+        if (bin.read_u32_le(buf, off) == LC_SEGMENT_64 && fixed16_equals(buf, off + 8, name)) { ret off; }
+        off = off + bin.read_u32_le(buf, off + 4)::usize;
+        c = c + 1;
+    }
+    ret 0;
+}
+
+# count the LC_SEGMENT_64s flagged SG_READ_ONLY
+# ---
+# buf: the Mach-O bytes
+# ret: how many segments carry the flag
+fun t_count_sg_read_only(buf: *u8) u32 {
+    val ncmds: u32 = bin.read_u32_le(buf, 16);
+    var off: usize = MACH_HEADER_64_SIZE;
+    var n:   u32 = 0;
+    var c:   u32 = 0;
+    for (c < ncmds) {
+        if (bin.read_u32_le(buf, off) == LC_SEGMENT_64
+            && (bin.read_u32_le(buf, off + 68) & SG_READ_ONLY) != 0) { n = n + 1; }
+        off = off + bin.read_u32_le(buf, off + 4)::usize;
+        c = c + 1;
+    }
+    ret n;
+}
+
+# a PIE image whose link carries two read-only segments - plain rodata and the
+# rebased rel.ro - marks exactly one of them SG_READ_ONLY: the one the rebase stream
+# writes into, which is the only segment that may be named __DATA_CONST. The other
+# stays r-- under __RODATA and is never mapped writable.
+#
+# Falsifies the permission-derived rule that flagged every read-only segment: that
+# emitted two SG_READ_ONLY segments, which the darwin kernel refuses to exec,
+# SIGKILLing the image before dyld runs (#2172). Also pins the rejection of a layout
+# with base relocations in two read-only segments, which Mach-O cannot express
+fun ts_dyn_exec_relro_segments(arch_id: u32) u8 {
+    var alloc: A.Allocator;
+    page.make(?alloc);
+    var itn: intern.Interner = intern.init(?alloc);
+
+    var isa_vt: isa.IsaVTable;
+    isa_vt.id = arch_id;
+
+    var text_bytes: [16]u8;
+    var ro_bytes:   [16]u8;
+    var relro_bytes: [16]u8;
+    var data_bytes: [16]u8;
+    var k: usize = 0;
+    for (k < 16) {
+        text_bytes[k] = 0; ro_bytes[k] = 0; relro_bytes[k] = 0; data_bytes[k] = 0;
+        k = k + 1;
+    }
+
+    # text, rodata, rel.ro, data - each on its own 16 KiB page, mirroring the linker's
+    # per-kind segment layout.
+    val base: u64 = 0x100000000;
+    val pg:   u64 = 0x4000;
+    var segs: [4]of.LoadSegment;
+    segs[0].vaddr = base;          segs[0].filesz = 16; segs[0].memsz = 16;
+    segs[0].flags = of.SEG_FLAG_READ | of.SEG_FLAG_EXECUTE; segs[0].bytes = ?text_bytes[0];
+    segs[1].vaddr = base + pg;     segs[1].filesz = 16; segs[1].memsz = 16;
+    segs[1].flags = of.SEG_FLAG_READ; segs[1].bytes = ?ro_bytes[0];
+    segs[2].vaddr = base + pg * 2; segs[2].filesz = 16; segs[2].memsz = 16;
+    segs[2].flags = of.SEG_FLAG_READ; segs[2].bytes = ?relro_bytes[0];
+    segs[3].vaddr = base + pg * 3; segs[3].filesz = 16; segs[3].memsz = 16;
+    segs[3].flags = of.SEG_FLAG_READ | of.SEG_FLAG_WRITE; segs[3].bytes = ?data_bytes[0];
+
+    # one absolute pointer to slide, in the rel.ro segment only.
+    var brs: [2]of.BaseReloc;
+    brs[0].seg_index = 2; brs[0].seg_offset = 0; brs[0].target = base;
+    brs[1].seg_index = 1; brs[1].seg_offset = 0; brs[1].target = base;
+
+    var dyn: of.DynamicInfo;
+    dyn.interp = intern.STR_NIL;
+    dyn.base_relocs = ?brs[0];
+    dyn.base_reloc_count = 1;
+    dyn.pie = true;
+
+    val tf_r: R.Result[fs.TempFile, str] = fs.temp_create(?alloc, "macho_relro_seg");
+    if (R.is_err[fs.TempFile, str](tf_r)) { ret 1; }
+    var tf: fs.TempFile = R.unwrap_ok[fs.TempFile, str](tf_r);
+    val tpath: str = fs.temp_path(?tf);
+
+    val em: R.Result[bool, str] = emit_dyn_exec(?alloc, ?itn, ?isa_vt, ?segs[0], 4, pg, base,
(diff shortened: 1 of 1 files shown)


octalide:
## Verification

**Acceptance — green on a real Mac.** `cd.yml` `workflow_dispatch` on this branch at `343f8cea`:
[run 30127667356](https://github.com/briar-systems/mach/actions/runs/30127667356) — `seed aarch64-darwin` success, `build aarch64-darwin` **success** (cross-built seed execs, `dep pull`, three-stage native self-host converged `b == c`, `./c test .` = **778 passed, 0 failed**, `mach.lang.target.of.macho 5 ok`).

An earlier dispatch of the fix commit alone ([run 30127349305](https://github.com/briar-systems/mach/actions/runs/30127349305)) was likewise green.

**Nothing else moved.** Building the *pre-fix* source tree with the pre-fix and post-fix compilers gives byte-identical output on every non-Mach-O target:

| target | result |
| --- | --- |
| x86_64-linux (ELF) | byte-identical |
| aarch64-linux (ELF) | byte-identical |
| x86_64-windows (COFF) | byte-identical |

**Self-host.** x86_64-linux three-generation release fixpoint: `b == c` byte-identical (`a != b` expected against the 4.1.0 seed).

**Suites.** `778 passed, 0 failed, 778 total` through the from-source compiler on linux, and the same on aarch64-darwin. The new checks are folded into the existing `of.macho:dyn_exec` case, so the count matches the v4.2.0 baseline exactly.

**The regression test bites.** With the old permission-derived rule restored and everything else unchanged, the suite reports `777 passed, 1 failed` at `of.macho:dyn_exec`; with the fix, `778 / 0`.

## Found, not fixed

The COFF executable writer names PE sections the same way (`segment_name` keys on permissions), so a windows image built since #2126 carries two `.rdata` sections. PE tolerates duplicate section names and the windows lanes are green, so this is cosmetic rather than fatal — reporting it rather than widening this PR.

octalide:
aarch64-linux three-generation release fixpoint (cross-built with the fixed compiler, self-hosted under `qemu-aarch64`): `b == c` byte-identical. Both ELF ISAs converge.

octalide:
## Review round 2 — all eight addressed

The two CONFIRMED findings and F1/F3 turned out to share one root: the writer emitted **one `LC_SEGMENT_64` per linker load segment**, so any two load segments that map the same way wrote the same 16-byte name twice, and every predicate keyed on load segments while the emitter keyed on commands. Fixed structurally — the writers now emit **one command per run of identically-mapped load segments**, carrying one `section_64` each.

### F6 — CONFIRMED, fixed

The fixture now has five segments in the linker's kind order (text, data, rodata, rel.ro, bss) and puts a base relocation **in the writable `__DATA` segment as well**. `__DATA` is asserted on name, vaddr, `initprot`, `maxprot`, and the *absence* of `SG_READ_ONLY`.

Mutation run: deleting the `vm_prot_for(...) != VM_PROT_READ` guard from `seg_is_data_const` (the successor to `&& prot == VM_PROT_READ`) now gives **777 passed, 1 failed**.

### F2 — CONFIRMED, fixed

`emit_exec`'s read-only segments merge into a single `__RODATA` command spanning both, so the name is written once. New case `ts_exec_merges_read_only` asserts exactly one `__RODATA`, its vaddr, its span, and its protection. Mutation (make the static writer emit per-segment again): **777 / 1**.

### F1 — you were right; the domains now coincide

`__TEXT` is still its own command — it maps the mach header below the linker's bytes, so it can share none — but it now goes through the *same* `seg_vm_prot` / `seg_is_data_const` rules as every other segment, and the guard counts the same commands the emitter flags. A read-only, rebased first segment is flagged like any other instead of counting toward a rejection it can never trigger.

### F3 — settled in favour of merging, error kept only for the inexpressible case

You were right that the error was reachable from a legitimate link. Two read-only segments the rebase stream both writes into are **adjacent** in the linker's kind order, so they now share one `__DATA_CONST` carrying a section each — the pre-#2126 object and the hand-written `.quad _sym` in `__TEXT,__const` both link. The format constrains the image to one such segment, not to one such input.

The error survives only for two rebased read-only segments *separated* by a differently-mapped one, which has no single command to share and no structural fix. `ts_dyn_exec_split_data_const` pins it.

### F8 — fixed

The diagnostic now names both segments (index + vaddr) and one offending relocation offset, e.g.

```
macho: read-only load segments 1 (vaddr 4294983680) and 3 (vaddr 42941032192, e.g. the pointer at +0) both need base relocations, but Mach-O maps a single read-only __DATA_CONST and they are not adjacent
```

and the test asserts on its content, not just that an error occurred.

### F7 — fixed

Segment and sections are now written from one `gname`, so they cannot diverge, and `t_sections_match_segname` walks **every** section of **every** segment and compares the 16-byte `segname` field against the enclosing command. Mutation (write `"__DATA_CONST"` into the section's `segname` while the segment stays `__RODATA`): **777 / 1**.

The rebase stream is checked the same way: `t_rebase_vaddr` walks the opcodes, resolves each through the *emitted* segment table, and asserts the absolute address the linker recorded — so a wrong command index or an offset measured from a member instead of its command is caught. Mutation (measure from the member): **777 / 1**.

### F5 — fixed

`macho_kind_from` takes `is_rodata`, so a `__RODATA` section re-reads as `SK_RODATA` instead of falling through to mutable `SK_DATA`. Asserted in `ts_exec_merges_read_only`.

### F4 — confirmed, and the narrowing is deliberate

You are right that `maxprot` narrows from `rw-` to `r--` for un-rebased read-only segments. That is intended and now says so in `seg_vm_prot`'s doc: the segment holds nothing the loader slides and nothing in the image may remap constants, so capping it is the point rather than a side effect. `__LINKEDIT` has always shipped `r--`/`r--`. **If you would rather keep `maxprot` open, say so and I will widen it — it is one term.**

## Verification

| check | result |
| --- | --- |
| **darwin dispatch** [run 30132773689](https://github.com/briar-systems/mach/actions/runs/30132773689) | `build aarch64-darwin` **success** — seed execs, `dep pull`, native `b == c`, `./c test .` **778 / 0** |
| PR CI | 16 / 16 SUCCESS |
| real darwin image vs. the already-green `343f8cea` | **byte-identical** — the merging is inert for the compiler's own layout |
| ELF x86_64 / aarch64, COFF x86_64 (pre-fix tree, pre- vs post-fix compiler) | **byte-identical** |
| x86_64-linux 3-gen release fixpoint | `b == c` byte-identical |
| aarch64-linux 3-gen release fixpoint (qemu) | `b == c` byte-identical |
| suites | **778 / 0** on linux and on darwin, `of.macho` 5 ok |

Mutation coverage, all four **777 / 1**: drop the read-only guard; make a section's `segname` diverge from its segment; stop merging in the static writer; measure rebase offsets from the member instead of its command.

Issue #2179, opened 2026-07-24
ci: darwin lane never gates dev -> main merges — a platform-fatal regression reached a tag through 17/17 green CI


octalide:
The darwin lane was intended to gate `dev` → `main` merges — the release path only — so that the slow, metered macOS runners are not paid for on every merge into `dev`. **That gate does not exist in the workflows.**

`.github/workflows/ci.yml` triggers on `pull_request: branches: [dev, main]` and contains no darwin or macOS job at all. The only on-Mac path in the repo is the manual `workflow_dispatch` in `cd.yml`, which nothing in the merge path invokes.

## What this cost

#2172. The aarch64-darwin seed was SIGKILLed at exec by a Mach-O regression introduced in #2126. It sat on `dev` through the entire 4.2.0 cycle, passed the release PR (#2169) **17/17 green**, and reached a pushed `v4.2.0` tag. The first thing to notice was CD's publish step, by which point the tag was public and the release had to be superseded by 4.2.1. A release-PR-only darwin gate would have caught it at #2169 and cost one macOS job.

## Fix

Add a darwin job to `ci.yml` that runs **only for PRs whose base is `main`** — i.e. the `dev` → `main` release merge — via a job-level condition on `github.base_ref`. Merges into `dev` must remain unaffected and must not schedule a macOS runner.

The job should do what `cd.yml`'s dispatch darwin lane already does: cross-build the stage-0 seed on linux, then on `macos-15` (Apple Silicon, aarch64-darwin native) exec the seed, `dep pull`, run the three-stage native self-host to `b == c`, and run the suite. Factor it out of `cd.yml` rather than duplicating the steps if that is clean — a copy that drifts from the CD lane is worse than no gate, because it would report green for a path CD does not actually take.

## Bar

- A PR targeting `dev` schedules **no** macOS runner (verify on a real PR, not by reading the YAML).
- A PR targeting `main` runs the darwin lane and fails if the darwin binary does not exec and self-host.
- Reverting the #2175 fix on a test branch targeting `main` makes the gate go red — the gate must demonstrably catch the exact regression that motivated it, not merely run.

Related: #2172, #2175, and the release supersession recorded in the 4.2.1 CHANGELOG entry.

octalide:
Closed by #2180 (merged to `dev` at `f36b6201`).

Verified on real runs, not read off the YAML:
[...]

PR #2181, opened 2026-07-24
test(2179): revert #2175 to prove the darwin release gate catches it — DO NOT MERGE

octalide:
Throwaway PR for #2179 acceptance verification only. Reverts the #2175 Mach-O fix (1845166f) — the exact regression that let #2172 reach a pushed v4.2.0 tag — on a PR targeting main, to prove the new darwin release gate goes red on it.

Will be closed and the branch deleted once the run is confirmed red. Do not merge.


octalide:
Rebuilding on the fix/2179 branch — this base lacked the darwin gate itself (still unmerged), so no darwin job was scheduled here. Recreating.

Issue #2177, opened 2026-07-24
ct: generic instantiated inside a variadic pack $each body is never re-validated — discloses a secret (9th path)


octalide:
## Security: confidentiality leak — the 9th proven escape path in the generic-secrecy family (#2157 / #2168)

Found by the max-effort review of PR #2170. **Pre-existing — this leaks on `origin/dev` today**, and is independent of anything #2170 changes (an ablated build with #2170's concreteness gate removed still accepts it and still prints the secret).

### The leak (prints the secret)

```
var g: ^u32 = 99;
fun show[U](s: U) u32 { print.printlnf("leaked = {}", s); ret 0; }
fun sink(va: ...) u32 { var acc: u32 = 0; $each a in va { acc = acc + show[^u32](g) + a; } ret acc; }
// sink(1u32) in main
```

Both the `dev` compiler and the #2170 branch compiler report **0 errors**; the linked binary prints `leaked = 99`.

### Control proving it is the pack body specifically

The identical `show[^u32](g)` call written **outside** a pack body is correctly rejected with the pack-secrecy diagnostic. Only the in-pack spelling escapes.

### Root cause

`reinfer_pack_each_body` (`src/lang/fe/sema.mach`) sets `collect_insts = false`. `record_instance` (`src/lang/fe/sema/context.mach:302`) early-returns on that flag, so a generic instantiated from inside a pack `$each` body is never enqueued and its body is never re-inferred under the instance substitution. The pack body is a blind spot for the entire per-instance constant-time re-validation, not just for this one sink.

### Fix

The pack-body re-inference must record instances like every other path. If `collect_insts = false` exists to prevent double-recording during the lowering-time re-inference, that is an ordering problem to solve properly, not a reason to skip validation — the current behavior trades a duplicate-work concern against a silent secret disclosure.

Note this sits in the same function as #2174 (`reinfer_pack_each_body` types deferred bodies against template parameter types rather than instance types). The two are natural companions; fix them together if the delta is shared.

### Bar

Build and run the reproducer above and confirm rejection; enumerate the analogous surface (every generic-body operation reachable only from inside a pack `$each`) and probe each by building and running, not by reasoning. This family has produced nine proven leaks and **every one was found by executing a probe after a previous fix was called airtight**. Add permanent test cases.

Part of epic #1643. The constant-time guarantee is documented as an incomplete preview in 4.2.0; this is one of the reasons why.

octalide:
[...]

Issue #2176, opened 2026-07-24
COFF: windows images carry two .rdata sections — PE sections named by permission, not kind


octalide:
Found while root-causing #2172. The COFF exec writer's `segment_name` names PE sections by **permission**, the same permission-derived naming that made the Mach-O writer emit two `__DATA_CONST` segments and one fatal `SG_READ_ONLY` too many.

Since #2126 (`feat/1865-reloc-addend`, commit `a450f3eb`) deleted `coalesce_kind` so object round-trips preserve `SK_RELRO` separately from `SK_RODATA`, the link sees two read-only merged kinds. On Windows both are therefore named `.rdata`, so **every windows image built since #2126 carries two `.rdata` sections.**

## Severity: cosmetic, not fatal

PE tolerates duplicate section names — the loader keys on the section table entry, not the name — and every windows CI lane is green (build, release, test, int). This is not the Windows analogue of #2172; there is no known functional breakage. It is untidy output that misrepresents the image to any tool that groups sections by name (debuggers, `dumpbin`, PE inspectors), and it is the same latent design smell: a name derived from permissions rather than from what the section *is*.

## Fix

Name PE sections from the section kind rather than from the permission bits, so `SK_RELRO` and `SK_RODATA` land in distinct, correctly-named sections. #2175 does the analogous thing for Mach-O — read there for the shape, but note the Mach-O change was forced by a hard kernel constraint that has no PE equivalent, so this one is free to be a straightforward naming fix.

## Bar

Windows image builds and runs; a test asserting no duplicate section names in the emitted PE; ELF and Mach-O output byte-identical (this should be COFF-local).

Not blocking any release. Filed from #2172's investigation so it is not lost.

Issue #2167, opened 2026-07-24
sema: deep secrecy-placement over-rejects differing-shape aggregates lacking field offsets


octalide:
### Summary
The deep secrecy-placement comparison (`generics.secrecy_structure_equal_deep` / `sse_deep_walk`, added in #2165) is **sound but conservatively imprecise** for differing-shape aggregates: when two record/union variants have different field counts, it can only prove `^` placement identical when **both sides are entirely secret** (`deep_all_secret`). A differing-shape pair where a public field lies BEYOND the secret overlap is **rejected** even though it is byte-safe.

### Example (rejected, but actually safe)
```
rec A { x: ^u32; }
rec B { x: ^u32; y: u32; }   // public `y` is beyond A's single secret field
uni U { a: A; b: B; }
```
Every byte `A` overlays in `B` is secret in both; the only public byte (`B.y`) lies past `A`'s extent, so no public alias can read a secret. This is safe, but the check fails closed because sema has no field OFFSETS to prove the non-overlap.

### Why it is deferred
`type.FieldEntry` carries `{name, ty}` with **no offset**; the deep walk cannot compute byte ranges, so it cannot distinguish "public field overlaps a secret" (unsafe) from "public field is beyond the secret extent" (safe). It fails closed — the safe direction. The common cases (matched shapes, all-secret differing shapes, fully-public) are all handled precisely; only this beyond-the-overlap differing-shape case is over-rejected.

### Proper fix
Plumb field offsets / layout into sema so the placement comparison can reason about byte ranges instead of field-index alignment. Then `sse_deep_walk`'s differing-count arm can admit a pair whose public bytes never overlap a secret byte, rather than requiring all-secret.

### Scope / priority
- **Non-security** — this is an over-rejection (false positive), not a leak. No secret escapes.
- Low priority; the workaround is to give the aggregates matching shapes.
- Depends on layout information being available at the sema secrecy-check point.

Follow-up to #2165 (PR #2166).

PR #2182, opened 2026-07-24
release: mach 4.2.1

octalide:
mach 4.2.1 — the release that actually delivers 4.2.0.

`v4.2.0` was tagged at `d7697778` but **never published**: its CD run failed `build aarch64-darwin` — the cross-built seed was SIGKILLed at exec — so `publish release` skipped and no artifacts and no release object ever shipped. The tag stays where it is. This release supersedes it; the payload is otherwise identical, so a reader coming from 4.1.0 gets the whole 4.2.0 feature set here. Ledger: #2172.

The one code delta over 4.2.0 is the darwin fix (#2172, PR #2175), already merged to `dev` at `9ed2822c`.

Verification and the CHANGELOG entry follow in a comment once the entry lands.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,62 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [4.2.1] - 2026-07-24
+
+The release that delivers 4.2.0. Its entire payload — auto-vectorization, loop
+rotation, fallthrough elision, bounded recursion peeling, loop-carried phi
+coalescing, in-memory linking, real Mach-O DWARF, parallel `-g` codegen, and the
+constant-time preview — reaches users for the first time here, so the
+`## [4.2.0]` **Added**, **Fixed**, and **Changed** sections below apply in full
+to this release. On top of them sits one fix: a mach-built aarch64-darwin
+executable execs again. Built with mach 4.1.0.
+
+(v4.2.0 was tagged but never published: its CD run caught the cross-built
+aarch64-darwin seed being SIGKILLed at exec, so `publish release` skipped and no
+artifact and no release object ever left the workflow — the full ledger is
+#2172. This release supersedes the tag; the payload is otherwise identical.
+Upgrading from 4.1.0 lands the whole 4.2.0 feature set at once.)
+
+(**The constant-time support in this release is an experimental preview and its
+guarantee is not complete.** The flow typing, the codegen taint contract, the
+translation validator, and the timing harness are all in — but proven
+secret-disclosure paths remain open. A generic instantiated inside a
+variadic-pack `$each` body is never constant-time re-validated, and the program
+prints the secret with zero diagnostics (#2177). `$fields` reflection projection
+inside a generic erases a secret field's secrecy (#2168). Both are joined by
+#2158, #2159, and #2167. The `^` / `#[oblivious]` surface has not been audited
+and is not sound today. **Do not build production cryptography on this
+version.** Epic #1643 stays open; #2177 extends the holes enumerated under
+4.2.0's Added below.)
+
+### Fixed
+- of: **a mach-built aarch64-darwin executable execs again.** The Mach-O writer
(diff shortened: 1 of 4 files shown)


octalide:
## Verification — from-source compiler throughout, never the PATH seed (still 4.1.0)

All bars re-run against the **locked** `mach-std` pin, not its default branch (see the correction note at the bottom).

| bar | result |
| --- | --- |
| **darwin, `cd.yml` dispatch on `dev`** | [run 30133095704](https://github.com/briar-systems/mach/actions/runs/30133095704) — `seed aarch64-darwin` success, **`build aarch64-darwin` success**. The cross-built seed execs, `dep pull` runs, the native three-stage self-host converges `b == c`, `./c test .` passes. This is the whole point of the release and it is proven on a real Mac before the cut. |
| **3-gen fixpoint, release, branch tip** | A ≠ B (expected — the 4.1.0 seed predates 4.2.0's features), **B == C byte-identical** `b92dd86ecd5e5a4f3f07c93d0a9fae452723fd3146724de2d526b8dbaf7ae881` |
| **3-gen fixpoint, debug, branch tip** | A ≠ B, **B == C byte-identical** `b789bbbd39bd9f2b75226fb4b49658812113bb00a47f33e288054ceb27487815` |
| **compiler suite** | **778 passed / 0 failed** — v4.2.0 baseline 778/0, **delta 0** |
| **mach-std @ pinned `eff1e8e`** | **611 passed / 0 failed**, 53 suites — v4.2.0 baseline 611/0, **delta 0** |
| **pin provenance** | `eff1e8e40686d9993047ef027bd8d7e6a7e2f5be` is exactly upstream `mach-std` `origin/main` HEAD, verified from a fresh clone; `mach.lock` is byte-identical at `v4.2.0` and at this branch, so the pin did not move |
| **int linux** | 72/72 PASS, 0 FAIL, both profiles |
| **int linux-riscv64 (qemu 11.0.2)** | 57/57 PASS, 0 FAIL, including `regression/1852-rv64-self-host [release]` — green first run, no flake, so no `v4.2.0` baseline comparison was needed |
| **version string** | the built compiler reports `mach 4.2.1`; `mach.toml` is the only version string in the repo, so CD's `verify` assertion `v$version == tag` is satisfied |

The `v4.2.1` tag is deliberately **not** created or pushed here — the owner handles the tag and the CD publish.

## What changed over 4.2.0

Two commits: the version bump, and the CHANGELOG + `doc/language/secrecy.md` entry. **No compiler source moved on this branch** — the darwin fix arrived on `dev` through #2175 and is inherited.

The CHANGELOG entry supersedes the dead tag rather than restating it: `## [4.2.0]` stays in place and 4.2.1's lede says plainly that its Added / Fixed / Changed sections apply in full, since that release never reached a user. It documents the segment-command fix, and calls out one user-visible behavior change under `### Changed` — `maxprot` narrowing to `r--` for un-rebased read-only segments, so an `mprotect` asking for write over a page holding a `val` constant now fails on darwin where it used to succeed. Anyone doing self-patching or const-cache-warming needs to be able to find that.

The constant-time preview caveat now names **#2177** alongside #2168 / #2158 / #2159 / #2167 — a confirmed, independently reproduced secret disclosure that reproduces on this release's code. `doc/language/secrecy.md` gained the same entry, so the docs and the CHANGELOG do not disagree about what is broken.

## Correction worth recording

My first pass reported **mach-std 663 / 0** and I nearly wrote it into the release. It was wrong: **mach-std's default branch is `dev`**, so a bare `git clone` lands on `982680c4`, not on the pin. @release-420 caught it. Re-measured at `eff1e8e` the number is 611 / 0.

The part worth keeping: the **three-generation fixpoint is blind to this mistake**. All four stage binaries came out byte-identical across the two std trees — the compiler imports none of the thirteen `dev`-only modules — so the strongest bar in the suite cannot detect a wrong dependency tree. Only the mach-std suite count exposes it, which is why that bar needs an exact number rather than "green". Two sessions hit this in one day.

## Found, not fixed — not in scope here

#2176 (windows images carry two `.rdata` sections; PE sections named by permission, not kind) is the COFF sibling of this root cause, filed off #2175 and left open. PE tolerates duplicate section names and every windows lane is green, so it is cosmetic rather than fatal, and it is not mentioned in the 4.2.1 entry. Flagging it rather than widening a release PR.

PR #2183, opened 2026-07-24
test(2179): revert #2175 to prove the darwin release gate catches it — DO NOT MERGE

octalide:
Throwaway PR for #2179 acceptance verification only. Branched from fix/2179-darwin-release-gate (so the darwin gate itself is present) with one extra commit reverting the #2175 Mach-O fix (1845166f) — the exact regression that let #2172 reach a pushed v4.2.0 tag — to prove the darwin release gate goes red on it when targeting main.

Will be closed and the branch deleted once the run is confirmed red. Do not merge.

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -13,13 +13,12 @@ permissions:
 # to its own self-host fixpoint, then publishes the archives + SHA256SUMS.
 # correctness is CI's job — every change reaches main through a PR it gates.
 #
-# darwin has no prior release to self-seed from, so seed-darwin cross-builds the
-# stage-0 seed on linux first; collapse that bootstrap once a darwin archive ships.
-# the darwin archive builds on macos-15 (apple silicon), aarch64-darwin natively.
+# the darwin lane (seed cross-build + native fixpoint + self-test) lives in
+# darwin-lane.yml, shared with ci.yml's dev -> main release gate; see that file.
 #
-# workflow_dispatch runs the darwin lane (seed + native fixpoint + self-test) on a
-# branch with no tag, so on-Mac validation needs no release; the tag-verify and
-# publish steps are gated to tag pushes and stay skipped on dispatch.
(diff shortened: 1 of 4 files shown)


octalide:
Correction: reverting only 1845166f (as the issue named) does NOT reproduce the #2172 SIGKILL — CI went green (run https://github.com/briar-systems/mach/actions/runs/30133576096). Root-caused: the actual exec-SIGKILL fix is 90855ee8 ("mark only the rebased read-only segment SG_READ_ONLY"); 1845166f is a separate follow-on fix (duplicate LC_SEGMENT command merging), not the exec fix. Force-pushed a revert of the full #2175 merge (9ed2822c, mainline 1) instead, which reverts both. Re-running.

octalide:
Closing: this acceptance test is misconstructed and its green result does not demonstrate anything.

The branch **contains** the #2175 Mach-O fix rather than reverting it. The diff against `main` shows `seg_is_rebased`, `rebased_ro_seg_count` and the `rebased` parameter being **added** (the same +214/-14 as PR #2175) because `main` is still at 4.2.0 and predates the fix — so it renders as a revert while actually being the fix. The darwin gate passed because it built a correct tree.

Also closing because an open PR targeting `main` is a hazard while the 4.2.1 release is being cut.

The correct construction is simpler: branch off `main` (which **already** lacks the fix), add only the darwin-lane workflow files, and PR that to `main`. The gate must go **red**. See #2179.

octalide:
Correcting my own close comment: the conclusion stands, the stated reason does not.

I said the diff renders backwards because `main` predates the fix. That was true when I started looking and false by the time I posted — #2182 (release 4.2.1) merged to `main` in between, so `main` now carries the fix.

The conclusion is unchanged, and provable without reference to any diff base: **`6010fff4` contains `seg_is_rebased` three times.** The fix is present on the branch; it was never reverted. So the darwin lane built a correct tree and its green result demonstrates nothing about whether the gate catches #2172.

Acceptance point 3 of #2179 is still unmet.

Issue #2223, opened 2026-07-25
resolve: a module-qualified member silently falls through to a file-local of the same name — wrong value, no diagnostic


octalide:
A module-qualified member reference **silently resolves to a file-local declaration of the same name**. No diagnostic, no ambiguity error — the wrong value.

Confirmed independently against the shipped **4.2.2** compiler, not just a work branch.

## Repro

```mach
# src/other.mach
pub val X: i64 = 1;
```
```mach
# src/main.mach
use repro.other;
val X: i64 = 2;            # file-local, same leaf name

#[symbol("main")]
fun main(argc: i64, argv: **u8) i64 {
    ret other.X;           # returns 2
}
```

Exit code is **2**. It must be 1.

`doc/language/use.md:19` states plainly that `module.member` reaches the member, so this is a defect, not a documented shadowing rule.

## Why it is worse than it looks

The resolver **can** reject a bad qualified member — a missing name gives `error: no symbol OPC_JAL exported by mach.lang.target.isa.riscv64.inst`. So the lookup does consult the module. It falls through to the local binding **only when a local of that name happens to exist.**

That is the shape that produces a wrong-but-plausible program: it needs a name collision to trigger, and when it triggers it is silent. Code that was correct stays correct until someone adds an unrelated file-local with a colliding leaf name, at which point an unrelated function starts reading a different value.

## How it was found, and what it produced

While adding a riscv64 machine-opcode surface (#2193): `riscv64/encode.mach` has a file-local `val JAL: u32 = 0x6F` (the 7-bit major-opcode **field value**), and the new `riscv64/inst.mach` has `pub val JAL: MachOp = 59` (the **instruction**). Both are the names the RISC-V manual uses for different concepts.

Every `inst.JAL` / `inst.JALR` / `inst.LUI` / `inst.AUIPC` in the encoder silently became the major opcode. Because `0x6F` = 111 and `0x67` = 103 are themselves valid indices into the opcode table, the printer emitted **fluent nonsense**: `jal zero, .3` rendered as `fmv.w.x zero, .3`, and `jalr zero, 0(ra)` as `fcvt.s.l zero, 0(ra)`. Right operands, right shape, wrong instruction. No crash, no diagnostic.

This is the #2187 bug class — an artifact confidently disagreeing with reality — arriving **through the compiler** rather than through a second printer. It was caught only because someone was reading rendered assembly against what they knew the encoder emits.

## Fix

[...]

Issue #2214, opened 2026-07-25
backend: three additive fields on isa.Inst/isa.Operand to cover aarch64 and riscv64


octalide:
Child of #2212. Additive only, **no behaviour change**, no byte movement — this unblocks aarch64 and riscv64 to use the shared operand model instead of each inventing one.

Derived from the #2193 and #2194 workers each mapping every form their encoder emits against `isa.Inst`. Both independently found the operand model fits 1:1 and only a handful of fields are missing.

## The three fields

**1. `isa.Inst.src3: Operand`** — arity 4 and second immediates.

- aarch64 `msub Rd, Rn, Rm, Ra` needs four operands, and every integer remainder in the corpus is an MSUB.
- `ubfm Rd, Rn, #immr, #imms` carries two immediates, and the bitfield moves are *every* immediate shift and *every* sub-word extend on aarch64 (1356 `uxtb` + 48 `sxtw` in one 31-module corpus).
- Shift modifiers (`add Rd, Rn, Rm, lsl #sh`, `movk Rd, #imm16, lsl #N`) then ride `src3` as an immediate — the same thing x64 already does for the `CMPPS` imm8 predicate on `src2`.

**2. `isa.Inst.vec_lane: u8`** — NEON lane arrangement.

`Operand.size` is a byte width: 16 says "128-bit" but cannot distinguish `.4s` from `.8h` from `.16b`, and the arrangement rides *every* operand of a 3-same form. **Reuse `mir.MirInstr.vec_lane`'s existing encoding** (`mir.vec_lane_kind` / `vec_lane_elem_bytes`) rather than defining a parallel one.

This also closes a known bug class: #2037 was a vector lane arrangement **silently dropped** when regalloc rebuilt a `MirInstr`. Giving the machine instruction the same field the MIR instruction has — instead of a parallel encoding — removes the mismatch that made that drop possible. x64 ignores the field; its packed ops are typed by opcode (`paddd` vs `paddw`).

**3. `isa.Operand.sym_mod: u8`** — relocation modifier on a symbol operand. **Needed by both targets, which is the strongest signal it belongs in the shared type.**

- aarch64: `adrp x16, sym` + `ldr x0, [x16, :lo12:sym]`, plus `:got:` / `:got_lo12:`. Dropping the modifier renders a `[x16]` that assembles to the wrong instruction — load-bearing, not cosmetic.
[...]

Issue #2213, opened 2026-07-25
backend: split IsaVTable by target family (register machine vs whole-module emitter)


octalide:
Child of #2212. Independent of #2193 / #2194 — different file, can proceed in parallel.

## The problem

One flat \`IsaVTable\` serves two fundamentally different kinds of target, so half its slots are dead for any given one:

- \`emit_module\` is **nil for every register machine** (x86_64, aarch64, riscv64, mos6502).
- \`select\` / \`encode\` / \`emit_asm\` are nil or unused for **SPIR-V**, which forks to a whole-module emitter right after MIR lowering — no isel, no regalloc, no frame, no encode.

Mutually-exclusive slots in one record is precisely the half-baked-per-target shape #2212 exists to remove. It also means nothing tells you which slots a new target actually owes: the contract is "fill in whatever applies", which is how targets diverge.

Compounding it, every slot is a type-erased \`*u8\` function pointer, so the compiler cannot check that a target provides a coherent set — a target can leave a required slot nil and fail at runtime instead of at build time.

## Direction

Two contracts, one per family:

- **Register machine** — isel → regalloc → frame → encode. Owes: instruction selection, encoding, register classes and reserved/scratch registers, frame and stack-pointer registers, relocation application, DWARF/CodeView register maps. Optionally an asm printer and (later) an asm parser.
- **Whole-module emitter** — MIR → module bytes. Owes far less: no register classes, no frame registers, no relocation application in the same sense.

The dispatch that already exists (\`emits_whole_module\`, added by #2159) is the correct fork and should become the type-level distinction rather than a predicate over a shared record.

**Do not** attempt to make SPIR-V conform to the register-machine contract. Its shape is right; the vtable is what is wrong.

## Bar

- A target cannot compile while leaving a slot its family requires unset — the failure moves from runtime to build time. Say explicitly what enforces that, given the type-erased pointers.
- No behavior change: all five targets emit **byte-identical** output before and after. This is a restructuring, not a rewrite.
- Suites and the three-generation fixpoint unchanged.
- Adding a hypothetical sixth target should require touching strictly fewer places than today. Demonstrate that concretely rather than asserting it — e.g. by listing what a new register machine must now provide versus what it had to before.

## Coordination

#2193 (riscv64) and #2194 (aarch64) are live and touch their ISA directories plus the shared sink spine in \`be/codegen/encode.mach\`. This issue touches \`src/lang/target/isa.mach\`. Expect to rebase around them; do not edit their files.

Issue #2212, opened 2026-07-25
backend: complete instruction, effect and target ownership contracts


octalide:
Mandatory Mach v5 release gate.

Start from a current source census. The target-family split, existing asm parsing,
printer coupling and byte-accounting are already partly or fully implemented.
Inventory the remaining parallel AArch64 instruction representation, encoding
decisions, effect descriptions, registry ownership and nominal register IDs.

Acceptance:

- Register-machine instructions have one authoritative structural/effect description
  consumed consistently by selection, allocation checks, encoding, asm and printing.
  Whole-module emitters retain their appropriate separate contract through briar-systems/mach#2963.
- Generalize form tables where they eliminate duplicated encoding decisions. Keep
  algorithmic relaxation and variable-length encoding as algorithms. No target-wide
  rewrite merely to satisfy a slogan about one IR.
- Existing construction invariants remain enforced. Unknown or incomplete forms
  fail without publishing partial accepted output. Missing callback/capability
  contracts are discovered at registration or the earliest typed boundary.
- VRegId and PRegId are non-interchangeable. Registry descriptors have explicit
  ownership and lifetime, and resolved targets cannot outlive their owner. Choose
  immutable ownership rather than a redundant version scheme where sufficient.
- Inline assembly/atomic memory effects remain correct under briar-systems/mach#3110 inlining and
  optimization. Ordinary read/write/control/ordering effects are modeled, not
  accidentally preserved only because the compiler cannot see the instruction.
- Native and independent encoding/assembly controls prove the actual remaining
  changes. Preserve bytes for pure representation/lifetime changes and document
  intentional generated-code differences. Relevant memory-order controls use the
  required native hosts rather than asserting a litmus execution proves all orders.

The synthesized host-target leak was fixed on dev at c86dac09. Remaining briar-systems/mach#3116 work is registry/descriptor ownership and borrow lifetime. Transfer briar-systems/mach#3114 virtual/physical register identities here, with negative interchange controls. Section identity belongs to briar-systems/mach#3113. Carry briar-systems/mach#2118 remaining encoding/form-table and AArch64 representation work without restarting completed target families.

## Tracking

Parent: [mach#3112](https://github.com/briar-systems/mach/issues/3112).

Required sub-issues:

- [ ] [mach#3127](https://github.com/briar-systems/mach/issues/3127) link: accept canonical RISC-V extension strings and reject what the target lacks
[...]

PR #2233, opened 2026-07-25
fix(emit-asm): render the riscv64 .s from the encoder's instruction stream

### LINKED ISSUE #2193: asm(riscv64): render --emit-asm from the encoder sink
Child of #2192. Wire riscv64's encoder to the asm sink introduced for x64 in #2187, so `--emit-asm` on rv64 renders from the encoder's own instruction stream rather than a parallel MIR-level printer.

**Harder than x64.** rv64 funnels through a single `emit_word(st, word)`, but the word is already packed by `pack_r`/`pack_i`/… before it arrives, so the sink receives an encoded word with no structure. Faithful text needs either a small rv64 decoder (~300-400 lines) or mechanical call-site edits that carry structure to the sink. Pick deliberately and say why.

**Until this lands, rv64 `--emit-asm` fails loudly** (from #2187) rather than emitting a misleading artifact. This issue removes that refusal.

**Bar:** every instruction in the `.s` corresponds to what the encoder actually emitted, proven by structural comparison against a disassembly of the `.o`; the sink's byte accounting matches the encoder's buffer length; emitted object bytes byte-identical pre/post; mutation-test the totality guard.

octalide:
Closes #2193. The riscv64 slice of #2192 / #2212, generalizing the shape #2197 proved on x86-64. Also lands the `Operand.sym_mod` slice of #2214 (aarch64, merged in #2226, carries a target-local one and can migrate onto it).

Merged with `dev` after #2226 landed; verification below is re-run on the merged tree.

## The finding: riscv64 had no machine-instruction surface

`riscv64.ADD` / `MOV` / `JMP` / `RET` are instruction-**selection** tags, deliberately width- and form-independent. `riscv64.MOV` is not an instruction — `mv` is `addi rd, rs, 0`. `riscv64.RET` is not one either — `ret` is `jalr zero, 0(ra)`. A single `ADD` tag becomes `add`, `addw`, `addi` or `addiw` depending on width and operand shape.

So the closest thing riscv64 had to a name for an emitted instruction was the `(major opcode, funct3, funct7)` triple passed to a packer and discarded one layer above the bytes. That is the missing half a shared assembler needs, not scaffolding for a printer, and it is the concrete thing #2213 can write the register-machine contract from.

`isa/riscv64/inst.mach` names them: the RV64I/M/A/FD forms the encoder can actually produce.

## The structure was never lost, only discarded early

`pack_r(opcode, funct3, funct7, rd, rs1, rs2)` receives every field of the word as an argument. Capturing at the packer rather than at `emit_word` means **no RV64 decoder exists anywhere**: `emit_r` / `emit_i` / `emit_s` / `emit_b` / `emit_u` / `emit_j` classify the same arguments that produce the bytes into a concrete instruction and notify. 119 call sites, a mechanical rename.

riscv64 now notifies with the same `isa.Inst` x86-64 does. One representation, three consumers — nothing for a second one to disagree with.

## What did not fit, and what it cost

| x64 assumption | riscv64 need | resolution |
|---|---|---|
| 3 operands | R-type `rd, rs1, rs2` is the widest form | fits |
| composite class-tagged `reg` | float bank must not alias the integer one | fits — `REG_CLASS_ID_XMM`, self-describing |
| `OPK_MEM` base+index+scale+disp | base+disp only | fits, strict subset |
| `OPK_LABEL` block ids | same | fits |
| **no relocation modifier** | `%pcrel_hi` / `%pcrel_lo` on the two halves of every global reference | **`isa.Operand.sym_mod`** (#2214) |
| renders at emit time | branch relaxation rewrites instructions *after* layout | **buffered notifications in the shared sink** |

The two things that stayed riscv64-local are rendering syntax, not representation: a store prints its transferred register before its address, and memory prints `disp(base)` rather than `[base + disp]`.

## Relaxation: the hazard, handled

An over-range conditional becomes an inverted guard plus a `jal`; an over-range `jal` becomes an `auipc`/`jalr` trampoline — **after** the function is laid out. Render at emit time and the `.s` says `blt` where the object holds `bge` + `jal`: #2187 recurring inside its own fix. So notifications buffer, the relax pass edits them exactly as it edits the text, and rendering happens once the layout is final. `insert_text` shifts note offsets alongside the block / fixup / symbol / relocation tables it already shifts — next to the tables they must stay consistent with, not in ISA-private memory a future shift path could forget.

Forced all three paths with a generated 1 MB function — guard, trampoline, and the guard widening from `.+8` to `.+12` on the second pass:

```
  bge a2, a0, .+12                 |    bge  a2, a0, 0x107af4
  auipc t0, %pcrel_hi(.2)          |    auipc t0, 0xffef8
  jalr zero, %pcrel_lo(.2)(t0)     |    jr   0x538(t0)
```

## Before / after

`sum_to`, riscv64 — printed vs the `.o` from the same run:

| before (`.s`) | after (`.s`) | `llvm-objdump` |
|---|---|---|
| `cbr.lt_s a2, a0, .2, .3` | `blt a2, a0, .2` / `jal zero, .3` | `blt` / `j` |
| `add a2, a2, 1` | `addi a2, a2, 1` | `addi` |
| `mv a0, a1` | `addi a0, a1, 0` | `mv` (alias of `addi`) |
| `j .1` (does not exist) | *(absent)* | *(elided)* |
| *(no epilogue)* | `ld ra, 8(sp)` / `ld s0, 0(sp)` / `addi sp, sp, 16` / `jalr zero, 0(ra)` | matches |

**No pseudo-instructions.** The text names what was encoded — `addi a0, zero, 5`, not `li a0, 5` — because a pseudo spelling hides which real instruction, or how many, the encoder chose. That is the property this artifact exists to show.

## Object-vs-asm cross-check (the acceptance criterion)

Printed `.s` vs `llvm-objdump` of the `.o` the same run produced — per-module mnemonic multiset and instruction count, with objdump's pseudo-mnemonics normalized to their base forms.

| corpus | before | after |
|---|---|---|
| riscv64, control-flow + full std | 26/26 diverged, 49,380 / 102,503 | **0 diverged, 102,503 / 102,503** |
| riscv64, `int/surface` + `int/regression`, **debug** | — | **903 modules, 0 diverged, 2,995,196 / 2,995,196** |
| riscv64, same, **release** | — | **903 modules, 0 diverged, 3,938,366 / 3,938,366** |
| riscv64, `vectorize-emit` (own `out` layout, checked separately) | — | 17 modules, 0 diverged, 32,045 / 32,045 |
| riscv64, `freestanding-riscv64` flat image (emits no `.o`) | — | 17 / 17, verified against the wrapped image |
| x86-64 regression check (spine touched) | — | **889 modules, 0 diverged, 1,985,637 / 1,985,637** |
| aarch64 regression check (#2226 runs under the same spine) | — | **872 modules, 0 diverged, 1,858,315 / 1,858,315** |

The checker was validated against x86-64's known-good `0 diverged, 69,976 / 69,976` from #2197 *before* any change, so the instrument was calibrated first.

## Byte identity

Printer-only, proven rather than argued. Baseline is a compiler built from `origin/dev` at the merge point.

| check | result |
|---|---|
| pre/post, riscv64 debug + release | 903 + 903 objects identical, 0 differ |
| pre/post, x86-64 debug + release | 889 + 889 objects identical, 0 differ |
| pre/post, aarch64 debug + release | 872 + 872 objects identical, 0 differ |
| with vs without `--emit-asm`, riscv64 debug + release | 903 + 903 objects identical, 0 differ |

## The guard, and what mutation-testing changed about it

The mutation test **found a real weakness, and the guard is stronger for it.** Dropping exactly one notification did *not* fail the build under byte accounting: `sink_unaccounted` only ever moves `accounted` forward to the buffer length, so a missing notification mid-sequence is absorbed by the next one, and only a miss at the *end* of a MIR opcode is caught.

RISC-V instructions are a fixed four bytes and the encoder notifies once per instruction, so the check can be exact: **emitted bytes must equal four times the instructions notified.** No blind spot — a dropped notification stays one short forever.

Mutation, on the final code:

| state | result |
|---|---|
| one notification dropped, guard **on** | build fails: `--emit-asm: prologue emitted instructions the riscv64 assembly printer did not render` |
| one notification dropped, guard **off** | ships quietly — cross-check catches it: **1 diverged, 102,502 / 102,503** |
| guard restored | 0 diverged, 102,503 / 102,503 |

## Suites and fixpoint

- mach: **814 / 0** — `dev`'s 812 plus the two printer tests deleted in #2197, restored
- mach-std: **611 / 0** at pin `eff1e8e`, confirmed by `git rev-parse HEAD` against `mach.lock`
- three-generation fixpoint seeded by a `dev`-current compiler: **A == B == C**, byte-identical (`30c57f5…`)
- seeded by the 4.2.2 PATH seed, A ≠ B — reproduced identically on **clean `dev`**, so it is `dev`'s codegen having moved past the seed, not this branch

## The `OPC_` prefix is deliberate — do not revert it as scaffolding

It separates the encoding FIELD values from the instructions, whose names overlap them: the RISC-V manual calls both the `JAL` opcode field value and the `jal` instruction by the same name, and after this PR both live in the riscv64 target.

It was *introduced* because the resolver defect in #2223 (found on this branch) made the collision silently miscompile. **That defect is fixed and merged, and the prefix stays anyway** — reviewed and ruled on that basis. The compiler bug made the ambiguity dangerous; the human ambiguity was there first and remains, and #2214 and #2118 will both add more names to the same space. 274 mechanical occurrences inside one file is a cheap price for making a collision that already bit once structurally impossible.

So: it stands on readability, not on necessity, and it is not dead scaffolding left over from a bug that no longer exists.

## Follow-up filed

**#2250 — the byte-accounting guard has a blind spot.** `sink_unaccounted` only moves `accounted` forward to the buffer length, so a dropped notification mid-sequence is absorbed by the next one and only an end-of-opcode miss is caught. Found by mutation-testing this PR's guard; the exact `bytes == 4 × notifications` check here closes it for RISC-V but **does not transfer** — x86-64 is variable-width and aarch64 has multi-word paths.

The artifacts from #2197, #2226 and this PR are not suspect: each was independently verified by an object-vs-asm cross-check at 0 diverged. What the blind spot costs is future protection. The guard-off measurement here (**1 diverged, 102,502 / 102,503**) is the demonstration that the cross-check catches what the guard misses — an argument for keeping both rather than treating either as sufficient.

## Note for #2118

The `classify_*` functions are a **hand-written inverse of a format table that does not exist yet**: they map `(major opcode, funct3, funct7, rs2)` to an opcode, the reverse of what #2118 would build. RV64I/M is a dense 3-key lookup and the A extension is funct5 × width with ordering falling out as flags — both transcribe straight across. **F/D does not:** `funct7` encodes operation *and* precision in bit 0, `funct3` is the rounding mode for arithmetic but the relation for compares and the sign-injection variant for `fsgnj`, and `rs2` is a width *selector* rather than a register in the `fcvt` family. A format table wants those three as explicit per-form fields, not as a uniform key. Budget for it.

### DIFF
--- a/doc/cli.md
+++ b/doc/cli.md
@@ -63,7 +63,7 @@ A verbosity flag (`-v`/`-vv`) and `--quiet` together is a parse error.
 | `--lib <name>`   | artifact name    | narrow the build to one `static`/`shared` `[artifact.<name>]` (mutually exclusive with `--bin`) |
 | `-o <path>`      | path             | override the artifact path, rooted at the project root (build/run/test) |
 | `--all-targets`  | —                | build every declared `[target.*]`, not just the default (mutually exclusive with `-o`, which names one path) |
-| `--emit-asm`     | —                | emit per-module assembly text (`.s`) — each line is one machine instruction the encoder emitted for that module, so the `.s` corresponds to the `.o` instruction for instruction. Supported on x86-64; riscv64 and aarch64 fail naming their tracking issue rather than emit a misleading file. Emission is opt-in, controlled only by this flag |
+| `--emit-asm`     | —                | emit per-module assembly text (`.s`) — each line is one machine instruction the encoder emitted for that module, so the `.s` corresponds to the `.o` instruction for instruction. Supported on x86-64 and riscv64; aarch64 fails naming its tracking issue rather than emit a misleading file. Emission is opt-in, controlled only by this flag |
 | `--emit-ir`      | —                | emit per-module SSA IR text (`.ir`), the final post-pipeline IR the object is built from (so it varies with the profile's `opt`) — emission is opt-in, controlled only by this flag |
 | `--verify-ir`    | —                | run the IR verifier after each optimisation pass |
 
--- a/src/lang/be/codegen/encode.mach
+++ b/src/lang/be/codegen/encode.mach
@@ -47,6 +47,7 @@ use mach.lang.be.codegen.mir;
 use mach.lang.intern;
 use mach.lang.source;
 use mach.lang.target.abi;
+use mach.lang.target.isa;
 use mach.lang.target.of;
 use tos: mach.lang.target.os;
 use target: mach.lang.target.resolved;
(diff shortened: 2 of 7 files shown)


octalide:
Post-merge dispositions, recorded here so they are not re-litigated from the diff alone.

**`OPC_` stays.** Reviewed and kept on the readability ground, not the one that introduced it. The resolver defect (#2223) that made the `JAL` field-value / `jal` instruction collision silently miscompile is fixed and merged; the prefix remains because the two concepts genuinely share a name in the manual and now share a namespace in this target, and #2214 and #2118 will both add more names to it. **Not dead scaffolding — do not revert it as such.** The source comment above the group states the concept distinction; this is the record of the decision.

**#2250 filed** for the byte-accounting blind spot this PR's mutation test exposed. Stated precisely, because it is easy to over-read: the artifacts from #2197, #2226 and this PR are **not** suspect — each was independently verified by an object-vs-asm cross-check at 0 diverged. What the blind spot costs is *future* protection, which is the guard's actual job. The exact `bytes == 4 × notifications` check here closes it for RISC-V and does **not** transfer; whoever takes #2250 needs to work out what is achievable per target and state what each invariant fails to catch rather than assuming the fixed-width answer generalizes.

**Both mechanisms earn their place.** The guard-off measurement (1 diverged, 102,502 / 102,503) is the demonstration that the cross-check catches what the guard misses. Neither is sufficient alone.

**For whoever takes #2118:** read this PR before designing. The `classify_*` functions are a hand-written inverse of the format table that increment would build, and the F/D extension is the part that does not transcribe cleanly — details in the body.

Issue #2225, opened 2026-07-25
ct: a generic union instantiated at a secret is never re-checked at the instance (10th path)


octalide:
Split out of #2168 / PR #2170, where three review rounds established this is a distinct problem from the reflection-projection leak that issue is about.

## The leak

```mach
uni U[T] { a: T; b: u32; }
var u: U[^u32];  u.a = 5;  if (u.b == 5) { ... }  ret tbl[u.b];
```

Compiles clean. The non-generic twin is a hard error (*"union variants must agree on secrecy"*).

**Root cause:** `check_uni_secrecy` runs only at the union **declaration** (`infer_nominal_uni`), where it compares `T` — a `TYPE_GENERIC_PARAM`, classified a public leaf — against `u32` and agrees. The CT re-validation worklist records only generic **function calls** (`record_instance` is reached solely from `infer_generic_call`), so a generic **nominal type** instantiated at a secret never enters it.

## Why this needs its own design pass

An attempt inside PR #2170 produced a deep type-graph walk that a max-effort review found bypassable four independent ways, plus a use-after-free. Recording them so the next attempt starts from the constraints rather than rediscovering them:

1. **The offending instance need not be spelled.** `rec Box[T] { u: U[T]; }` at `Box[^u32]` materializes `U[^u32]` inside its own substituted field table and never names it. A sweep over source annotations cannot see it.
2. **An AST sweep reads un-substituted template types.** A `U[T]` written inside a generic *function body* materializes `U[^u32]` during re-validation, after `resolve_all_types` has already cached the template resolution — so the sweep sees `{T, u32}` and agrees.
3. **Anonymous unions cannot be classified at all** — see the companion issue. `generic_nominal_of` on a `MODULE_NIL` nominal yields `TYPE_ERROR`, so any guard keyed on "is this instance a union" fails open. This affects mach-std's `Result`.
4. **Function-typed fields hide instances.** `rec Box[T] { f: fun(*U[T]) u32; }` at `Box[^u32]` materializes `fun(*U[^u32]) u32`; a walk that does not descend `TYPE_FUN` misses it. The sibling `secret_deep_walk` *does* descend it, precisely because a reinterpret can round-trip a whole signature.
5. **Borrow hazard.** `type.get` returns a pointer into a growable array; `ensure_fields` → `build_instance_table` → `substitute` interns types and can reallocate it. Any walk must re-fetch after materializing, not borrow across.
[...]

PR #2238, opened 2026-07-25
fix(sema): apply the mixed-secrecy union rule to anonymous unions

### LINKED ISSUE #2224: ct: anonymous unions escape the mixed-secrecy rule entirely — verified leak on shipped 4.2.2 (incl. mach-std Result)
## Security: confidentiality leak, verified on shipped 4.2.2

An **anonymous** `uni { ... }` is never subjected to the mixed-secrecy union rule. The named twin is a hard error; the inline form compiles and leaks.

## Verified

```mach
#[symbol("main")]
fun main(argc: i64, argv: **u8) i64 {
    var v: uni { a: ^u32; b: u32; };
    v.a = 1337;
    val leaked: u32 = v.b;
    ret leaked::i64;
}
```

Built and run with the **released 4.2.2** compiler: exit **57** = `1337 & 0xFF`. The secret written through `a` is read back public through `b`.

The named equivalent is correctly rejected:

```
error: union variants must agree on secrecy: overlapping fields are part secret ^
       and part public, which would alias the same storage at two classifications
```

## Root cause

`check_uni_secrecy` is called from exactly one place — `infer_nominal_uni` (`fe/sema/infer.mach:172`), i.e. only for a named `uni` **declaration**. An inline anonymous union goes through `resolve_type_anon` (`infer.mach:2353`), which calls `build_field_table` and nothing else. Nobody applies the rule.

## This also exposes mach-std's `Result[T, E]`

`dep/mach-std/src/types/result.mach:12` is `pub rec Result[T, E] { tag: bool; value: uni { ok: T; err: E; } }` — an **anonymous** union. So `Result[^u32, u32]` places `ok: ^u32` and `err: u32` in the same storage with no diagnostic, and the secret is readable through the public variant:

```mach
var r: R.Result[^u32, u32];
r.value.ok = key;
if (r.value.err == 0) { ... }   # branches on the secret through the public overlay
```

That is the most-used generic type in the standard library. Confirm this specific case as part of the fix rather than assuming it follows from the general one.

## Fix

Apply the mixed-secrecy rule wherever a union **type** is formed, not only at a named declaration. The rule exists to stop secret bytes being read through a public overlay; which syntactic form declared the union is irrelevant to that.

Check whether the same asymmetry exists for anonymous `rec` — if any rule fires only at `infer_nominal_rec`, it has the same hole.

## Bar

Build and run; do not reason. This family has produced ten proven leaks and every one was found by executing a probe after a previous fix was called airtight.

- The reproducer above is rejected, and the `Result[^u32, u32]` case with it.
- A legal anonymous union (all-public, or all-secret) still compiles — test both.
- Mutation-test the new gate: remove it, show a test fails with its count, restore.
- Compiler output byte-identical unless the sweep finds an existing victim in mach or mach-std, in which case that is the correct outcome and must be reported per victim.

Part of epic #1643. Related: #2168 (the generic-instance half of the same rule), #2177, #2191.

octalide:
Closes #2224.

`check_uni_secrecy` fired from exactly one place — `infer_nominal_uni` — so only a **named** `uni` declaration was subject to the mixed-secrecy rule. An inline `uni { ... }` resolves through `resolve_type_anon`, which builds a field table and nothing else, so the rule never reached it.

Reproduced on a from-source `dev` build (not the 4.2.2 seed):

```mach
var v: uni { a: ^u32; b: u32; };
v.a = 1337;
val leaked: u32 = v.b;   # exit 57 = 1337 & 0xFF
```

Six distinct anonymous shapes leaked on `dev`, not one: inline, as a record field, deep through a record variant, behind a pointer, in an array, and as a function parameter.

## Placement

The rule runs as one pass over the module's unions after `infer_all_decls`, rather than at the site the issue names. Both earlier placements were built and measured:

- **`resolve_type_anon`** — runs once per *reference* to a shape, so `[2]uni {...}` reported one site twice and `*uni {...}` three times.
- **during type resolution or decl inference** — the deep secrecy comparison reads the variants' field tables, which do not exist yet. It fails closed and rejects a legal all-public `uni { a: S; b: P }`.

## Commits

1. **the leak fix** — the rule reaches anonymous unions.
2. **tests** for the anonymous rule, including the over-rejection controls.
3. **declaration-order independence** — a pre-existing false positive found while measuring the placement. On `dev`, `uni U { a: S; b: P; }` written *above* two entirely public records is rejected, while the identical union written *below* them compiles. Both forms now run in the one pass with every field table complete. This changes what compiles; the newly-accepted set is characterised exactly in the verification comment.

## Not covered here

`Result[^u32, u32]` is **not** reachable from the anonymous side — its variants `T` and `E` agree at the declaration, and the violation exists only after substitution, which does not re-resolve the AST. That is #2225's half, whose check keys on the instance's template nominal being a `uni` and so covers anonymous and named generic unions at one site.

Probing it surfaced #2239: an anonymous `uni` inside a generic record loses union layout in every instance, which currently **masks** #2225's hole.

## Verification

Reproduction, rejections, over-rejection controls, mutation tests, suites (mach 806/0, mach-std 611/0 @ `eff1e8e`), byte-identity and the three-generation fixpoint are in the [verification comment](https://github.com/briar-systems/mach/pull/2238#issuecomment-5076688699).

### DIFF
--- a/src/lang/driver/tests.mach
+++ b/src/lang/driver/tests.mach
@@ -4511,6 +4511,29 @@ test "mach.lang.driver:secret_anon_uni_field_secrecy" {
     ret bad;
 }
 
+# secret flow typing: the union rule is DECLARATION-ORDER INDEPENDENT (#2224). It reads
+# the variants' field tables, so running it during decl inference made a variant naming a
+# record declared LATER unresolvable, and the comparison - which must fail closed - then
+# rejected an entirely public union purely for being written above its variants. Each
+# program below is asserted in both orders and must give the same verdict either way:
+# public variants clean, a genuinely mixed pair rejected, named and anonymous alike
+test "mach.lang.driver:secret_uni_secrecy_order_independent" {
+    var bad: i32 = 0;
+
+    # all-public variants: clean whether the union is written above or below them
+    if (!ut_sema_clean("uni U { a: S; b: P; }\nrec S { k: u32; }\nrec P { k: u32; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    if (!ut_sema_clean("rec S { k: u32; }\nrec P { k: u32; }\nuni U { a: S; b: P; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    if (!ut_sema_clean("fun f() u32 { var v: uni { a: S; b: P; }; ret v.b.k; }\nrec S { k: u32; }\nrec P { k: u32; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    if (!ut_sema_clean("rec S { k: u32; }\nrec P { k: u32; }\nfun f() u32 { var v: uni { a: S; b: P; }; ret v.b.k; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+
+    # a genuinely mixed pair: rejected in both orders, never order-dependently missed
+    if (!ut_sema_rejects("uni U { a: S; b: P; }\nrec S { k: ^u32; }\nrec P { k: u32; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    if (!ut_sema_rejects("rec S { k: ^u32; }\nrec P { k: u32; }\nuni U { a: S; b: P; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    if (!ut_sema_rejects("fun f() u32 { var v: uni { a: S; b: P; }; ret v.b.k; }\nrec S { k: ^u32; }\nrec P { k: u32; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    if (!ut_sema_rejects("rec S { k: ^u32; }\nrec P { k: u32; }\nfun f() u32 { var v: uni { a: S; b: P; }; ret v.b.k; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    ret bad;
+}
+
 # constant-time (#1646): floating point is variable-latency and gated by default, so
 # a secret operand of an FP add / sub / mul / compare is a sema error, while the same
 # ops on public floats are clean - the gate keys on secrecy, not shape
--- a/src/lang/driver/tests.mach
+++ b/src/lang/driver/tests.mach
@@ -4492,6 +4492,25 @@ test "mach.lang.driver:secret_uni_field_secrecy" {
     ret bad;
 }
 
+# secret flow typing: the same rule reaches an ANONYMOUS `uni { ... }`, which has no
+# declaration to hang the check on and so escaped it entirely - a part-secret part-public
+# inline union leaked the secret through the public variant (#2224). Covered inline, as a
+# record field, deep through a record variant, and under a pointer / array spine; the
+# uniformly-secret and uniformly-public inline unions, and an anonymous `rec` with a
+# secret field (which has no overlapping storage), must still compile
+test "mach.lang.driver:secret_anon_uni_field_secrecy" {
+    var bad: i32 = 0;
+    if (!ut_sema_rejects("fun f() u32 { var v: uni { a: ^u32; b: u32; }; v.a = 1; ret v.b; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    if (!ut_sema_rejects("rec Box { v: uni { a: ^u32; b: u32; }; }\nfun main() i32 { ret 0; }\n"))                           { bad = 1; }
+    if (!ut_sema_rejects("rec S { k: ^u32; }\nrec P { k: u32; }\nrec Box { v: uni { a: S; b: P; }; }\nfun main() i32 { ret 0; }\n")) { bad = 1; }
+    if (!ut_sema_rejects("fun f(p: *uni { a: ^u32; b: u32; }) u32 { ret p.b; }\nfun main() i32 { ret 0; }\n"))               { bad = 1; }
+    if (!ut_sema_rejects("fun f() u32 { var v: [2]uni { a: ^u32; b: u32; }; ret v[0].b; }\nfun main() i32 { ret 0; }\n"))    { bad = 1; }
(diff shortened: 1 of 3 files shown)


octalide:
## Verification

Everything below was **built and run**, never reasoned. The PATH seed is 4.2.2; every verdict is from a from-source compiler built on this branch, and every baseline verdict is from a from-source compiler built on unmodified `dev` (`d4b7b19`). `dep/mach-std` pinned at `eff1e8e`, `git rev-parse HEAD` checked against `mach.lock`.

### The leak, and its rejection

| | dev | this branch |
|---|---|---|
| `uni { a: ^u32; b: u32; }`, write `a`, read `b` | **compiles, exit 57** (`1337 & 0xFF`) | rejected |
| same union as a `rec` field | **compiles, exit 57** | rejected |
| deep: `uni { a: S; b: P; }`, `S { k: ^u32 }`, `P { k: u32 }` | **compiles, exit 57** | rejected |
| same, with `S`/`P` declared *after* | **compiles, exit 57** | rejected |
| behind a pointer: `*uni { a: ^u32; b: u32; }` | **compiles, exit 57** | rejected |
| in an array: `[2]uni { a: ^u32; b: u32; }` | **compiles, exit 57** | rejected |
| as a function parameter | **compiles, exit 57** | rejected |

Six distinct shapes leaked on `dev`, not one. All now carry the same diagnostic the named twin has always carried.

### Over-rejection controls — every one still compiles

`uni { a: u32; b: u32; }` · `uni { a: ^u32; b: ^u32; }` · `uni { a: S; b: P; }` with both records public · both records all-secret · anonymous `rec { a: ^u32; b: u32; }` (fields do not overlap, so there is no aliasing channel) · a mixed-secrecy union inside an **untaken** comptime-`$if` arm, named and anonymous (a discarded branch is not compiled, and stays unchecked — verified identical to `dev`).

### `Result[^u32, u32]` — not covered here, and why

Still accepted. It is **not** reachable from the anonymous side: the anon uni in `Result[T,E]` has variants `T` and `E`, which *agree* at the declaration because both are abstract public leaves. The violation exists only after substitution, and substitution does not re-resolve the AST — `generics.substitute` interns a `TYPE_INSTANCE` and `build_instance_table` builds its table, so no anonymous-side hook can see it. Confirmed empirically: `Result[^u32,u32]` and a local `rec Pair[T,E] { v: uni { a: T; b: E; } }` are both still accepted with this fix in.

This is #2225's half, and per the ruling on that issue its check keys on the instance's **template nominal being a `uni`**, which covers anonymous and named generic unions at one site.

While probing it I found something that changes the record, filed as **#2239**: an anonymous `uni` inside a generic record loses union layout in every instance. Measured `(?r.value.err) - (?r.value.ok)` on `dev`: stdlib `Result[u32,u32]` = **4 bytes**, a local `rec Local[T,E]` of the same shape = **4 bytes**, the non-generic twin = **0** (correct), a named generic `uni Named[u32,u32]` = **0** (correct). Store/load agrees — write `err=2`, then `ok=1`, read `err` → `2`, no aliasing. So `Result`'s variants occupy *distinct* storage today. The type system genuinely accepts the pun, but no byte leaks — **the layout bug is masking #2225's hole, and fixing it without fixing the rule converts a masked hole into a live leak.**

### Declaration-order independence (commit 3)

Found while measuring where to place the anonymous check. On `dev`:

```mach
uni U { a: S; b: P; }     # rejected: "variants must agree on secrecy"
rec S { k: u32; }
rec P { k: u32; }
```

Both variants are entirely public. Written *below* `S` and `P` the identical union compiles. `check_uni_secrecy` ran inside `infer_all_decls`, so a forward-referenced variant had no field table and `secret_deep_walk` — which must fail closed — reported a mismatch.

**Programs this newly accepts**, exactly: a named `uni` at least one of whose variants names an aggregate declared **later in the same module**, directly or through a pointer/array spine, and whose variants genuinely agree on secrecy. Measured cases that flip `dev` reject → accept: `uni U { a: S; b: P; }` above two public records; `uni U { a: S; b: u32; }` above the record; `uni U { a: *S; b: *P; }` above them. Cases that do **not** flip: all-primitive variants (no field table needed, accepted on `dev` already); anonymous forward references (no check existed on `dev`); anything in an untaken `$if` arm. A genuinely mixed pair is still rejected in **both** orders — asserted both ways in the test.

Nothing in `mach` or `mach-std` was in the newly-accepted set — both build unchanged, and a program `dev` *rejects* cannot be in a tree that builds.

### Mutation tests

| mutation | result |
|---|---|
| anonymous-union loop disabled | **804 passed, 2 failed, 806** — `secret_anon_uni_field_secrecy`, `secret_uni_secrecy_order_independent` |
| named check restored to `infer_nominal_uni` (the `dev` placement) | **805 passed, 1 failed, 806** — `secret_uni_secrecy_order_independent` |
| restored | **806 passed, 0 failed, 806** |

Each gate is independently load-bearing for a distinct test.

### Suites

| suite | result |
|---|---|
| `mach` on `dev` (baseline) | **804 / 0** |
| `mach` on this branch | **806 / 0** (+2 new tests) |
| `mach-std` @ `eff1e8e` | **611 / 0** |

`mach-std`'s default branch is `dev`, so a bare clone is not the pin and reports 663; `git rev-parse HEAD` was checked against `mach.lock` before and after the run.

### Byte-identity

Unmodified `dev` source compiled by the `dev` compiler and by this branch's compiler:

| profile | `dev` compiler | this branch's compiler |
|---|---|---|
| debug | `c36c53aa…` | `c36c53aa…` |
| release | `d6224952…` | `d6224952…` |

Identical, and all **204** object files diff clean. No victim in `mach` or `mach-std` — the rule rejects nothing either tree contains.

### Three-generation fixpoint

| generation | this branch | unmodified `dev` |
|---|---|---|
| A (seed 4.2.2-built) | `7441f107…` | `c16ffafc…` |
| B (A-built) | `f7a79946…` | `c36c53aa…` |
| C (B-built) | `f7a79946…` | `c36c53aa…` |

**B == C** on both. A ≠ B is the installed 4.2.2 seed lagging current `dev`, reproduced identically on unmodified `dev`, so it is not a property of this branch.

PR #2249, opened 2026-07-25
fix(sema): re-validate instances a deferred `$each` body instantiates (#2177, #2174)

### LINKED ISSUE #2174: sema: reinfer_pack_each_body types deferred bodies against template parameters, not instance types
Found while fixing #2168. `reinfer_pack_each_body` types a deferred `$each` body against the **template** parameter types — it does not carry `ctx.subst`, so the substitution that maps a generic parameter to its instance type is not applied during the re-inference.

## Impact: diagnostic quality, not a soundness hole

After #2170 sema is authoritative for the constant-time gates, so this does not leak a secret. What it does is produce misleading diagnostics: a probe against the pre-fix blind spot reported `incompatible operand types 'U' and 'i64'` — naming the raw template parameter `U` rather than the concrete instance type the user actually wrote. A user reading that message has no way to connect it to their call site.

## Fix

Carry `ctx.subst` into `reinfer_pack_each_body` so the deferred body is typed against instance types, matching how the rest of per-instance re-inference works after #2157 and #2168.

## Bar

A test asserting the diagnostic names the concrete instance type, not the template parameter. Confirm the change is diagnostic-only: no accepted program becomes rejected and no rejected program becomes accepted.

Serialize behind #2170 — it touches the same sema re-inference area.

### LINKED ISSUE #2177: ct: generic instantiated inside a variadic pack $each body is never re-validated — discloses a secret (9th path)

octalide:
Closes #2177
Closes #2174

## The gap

A pack element's type is fixed only at monomorphization, so a pack `$each` body is type-checked **exactly once** — by `reinfer_pack_each_body`, during **lowering**, after the sema drain has run and freed its worklist. That episode set `collect_insts = false`, and `infer_generic_call` is the only enqueue point `drain_revalidations` consumes. So a generic whose only call site sat inside a pack body was looked at once with recording disabled: never enqueued, never re-inferred under its concrete type arguments.

## Shape chosen, and why

Recording there needs either a second drain in lowering or pack-body typing hoisted into sema. **The drain becomes a property of an inference episode, not of the module sema pass**: whichever pass type-checks a body drains the instances that body reaches. `sema()` already obeyed that invariant; `reinfer_pack_each_body` now does too, through the same `revalidate_one`.

Hoisting pack-body typing into sema was rejected on structure, not effort: pack monomorphization keys on the call-site element type-list and lives in lowering, and lowering must still type each element as it emits it. Sema-side validation typing would be a **second enumeration of the same instances**, and two enumerators that must agree is precisely the shape that produced this blind spot.

Supporting pieces the shape requires:

- the worklist is owned across one module's lowering (`LowerContext.ct_insts`), so its `drained` cursor validates each distinct instance exactly once regardless of element or pack-instance count;
- `drain_revalidations` splits into a deps-building wrapper and `drain_instances`, so the deferred-body drain reuses the module-spanning snapshot the unroll already built;
- `revalidate_one` **saves and restores** the origin context's comptime resolvers rather than clearing them — the drain now runs during lowering, where that context carries lowering's own installation.

#2174 falls out of the same episode: it carries the enclosing instance's substitution, so a deferred body naming a generic parameter is typed against the concrete instance type.

## Reproduction and rejection

Built from source (the 4.2.2 PATH seed was never the test compiler). On `origin/dev` @ `d4b7b191` the issue's reproducer builds with **zero diagnostics**, runs, prints `leaked = 99`. On this branch it is rejected with `error: a secret value cannot be passed to a variadic \`...\` argument` — the same diagnostic the out-of-pack twin has always produced.

**Two further escapes** through the same hole were found by probing and are closed here:

| shape reached only from a pack `$each` body | pre | post |
|---|---|---|
| secret handed to a variadic pack (the issue) | ran, `leaked = 99` | rejected |
| aggregate `==` at a record instance (#2127) | ran, `r = 1` | rejected |
| `#[oblivious]` generic returning `^u8` through a public `u8` | ran, `r = 2` | rejected |

Each is paired with its out-of-pack control, which was and stays rejected.

## Enumerated surface

`==`/`!=` and passing to a pack are the only operations a generic body can perform on a bare `T` (arithmetic on a template parameter is refused at the template pass), so the reachable surface is small and was probed exhaustively by building and running. Full results, including the shapes that did **not** leak, are in the worker report on the issue.

## Verification

- compiler suite **806 / 0** (804 baseline + the 2 tests added)
- mach-std **611 / 0** at pin `eff1e8e` (`git rev-parse HEAD` verified against `mach.lock`)
- three-generation fixpoint **B == C == D**, byte-identical
- pre-fix and post-fix compilers emit a **byte-identical** artifact from the same secret-free sources: analysis only, no codegen effect
- every new assertion confirmed to fail without the fix, per-assertion, via a bitmask build

## Out of scope, escalated

`PackInstance` carries no generic type arguments, so a **pack-tailed generic** loses its substitution entirely. `fun f[T](t: T, va: ...)` fails at lowering with `call argument typing: ...`, and `fun f[T](va: ...)` silently drops the substitution — a `T`-typed local in its pack body is not seen as secret at a `^` instantiation. Closing that means keying pack instances on (type-args × element-types), which lands in `me/lower/expr.mach` (owned by #2223). Filed separately rather than papered over.

### DIFF
--- a/doc/language/secrecy.md
+++ b/doc/language/secrecy.md
@@ -171,9 +171,10 @@ run it with `bash int/ct/ct.sh`.
 
 What does not hold — the known open holes:
 
-- **a generic instantiated inside a variadic-pack `$each` body is never
-  re-validated**, so it computes on a secret and prints it with no diagnostic.
-  Proven, security-blocking (briar-systems/mach#2177).
+- **a pack-tailed generic loses its type arguments at monomorphization**, so its
+  `$each` body is typed against the raw parameter and a value that is secret at
+  the instance is not seen as one — it reaches a variadic pack with no
+  diagnostic. Proven, security-blocking (briar-systems/mach#2251).
 - **`$fields` reflection projection inside a generic erases a secret field's
   secrecy**, which discloses the secret. Proven, security-blocking
   (briar-systems/mach#2168).
--- a/src/lang/driver/tests.mach
+++ b/src/lang/driver/tests.mach
@@ -6185,3 +6185,194 @@ test "mach.lang.build.engine.execute_warm:sema_phase_diagnostic_fails_build" {
     session.dnit(?s);
     ret bad;
 }
+
+# the error diagnostics `src` leaves on the sink after sema + lower, or -1 when a
+# pass returned a hard failure
+#
+# The deferred-`$each` gates report through the sink and let the pass return ok -
+# the engine's own `run_phase` gate is what fails the build - so a pack-body
+# rejection is invisible to `ut_lower_rejects` (which demands an Err) and needs the
+# sink read the engine performs. A pass that returns a hard failure carrying no
+# located diagnostic counts as one error rather than zero, or a lowering that dies
+# with a raw message would read here as a clean compile. Both passes run against one
+# scaffolded project torn down here
+fun ut_ct_lower_errs(src: str) i32 {
+    var alloc: A.Allocator;
+    if (O.is_some[str](page.make(?alloc))) { ret -1; }
+    val sr: R.Result[session.Session, str] = session.init(?alloc);
+    if (R.is_err[session.Session, str](sr)) { ret -1; }
+    var s: session.Session = R.unwrap_ok[session.Session, str](sr);
+
+    var names: [1]str;
+    names[0] = "main.mach";
+    var texts: [1]str;
+    texts[0] = src;
+    val pr: R.Result[project.Project, outcome.Fail] = ut_build(?s, ?alloc, ?names[0], ?texts[0], 1);
+    if (R.is_err[project.Project, outcome.Fail](pr)) { session.dnit(?s); ret -1; }
+    var p: project.Project = R.unwrap_ok[project.Project, outcome.Fail](pr);
+
+    var errs: i32 = -1;
+    val se: R.Result[bool, outcome.Fail] = passes.run_sema_pass(?p);
+    if (R.is_ok[bool, outcome.Fail](se)) {
+        val le: R.Result[bool, outcome.Fail] = passes.run_lower_pass(?p);
+        errs = ut_count_errors(?s.diags)::i32;
+        if (R.is_err[bool, outcome.Fail](le) && errs == 0) { errs = 1; }
+    }
+
+    project.dnit_project(?p);
+    session.dnit(?s);
+    ret errs;
+}
+
+# 0 when sema + lower leave an error diagnostic containing `needle`, 1 otherwise.
+# the needle is load-bearing: a bare "was rejected" also passes on an unrelated
+# earlier error, which is exactly how a probe in this family reported a fix that
+# was not there
+fun ut_ct_lower_diag(src: str, needle: str) i32 {
+    var alloc: A.Allocator;
+    if (O.is_some[str](page.make(?alloc))) { ret 1; }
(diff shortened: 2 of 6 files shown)


octalide:
Rebased onto `dev` @ `dbdc0e67` (merge, not rebase). The only conflict was in `driver/tests.mach`, where both sides appended — resolved by taking dev's file and re-appending this branch's block, so the addition stays one contiguous hunk. `fe/sema.mach` and `me/lower/context.mach` auto-merged with #2224's `check_uni_secrecy_all` and #2223's `declaring_comptime_ctx`; both are orthogonal to this change, and the merged `SemaContext` / `LowerContext` field blocks were re-read to rule out a silent same-slot collision.

Re-verified on the merged tree:

- compiler suite **826 / 0** (824 dev baseline + the 2 tests here)
- mach-std **611 / 0**, `dep/mach-std` at `eff1e8e` per `mach.lock`
- three-generation fixpoint **B == C == D**, byte-identical
- byte-identity re-measured against a compiler built from *this* tree with only the four fix files reverted to `dev`: both emit `fb4113e6…` from the same sources — **identical**, so the change is analysis-only
- the issue's reproducer still rejected, the public and `:^` controls still clean

Issue #2205, opened 2026-07-25
codegen: no SROA -- small aggregates never reach registers


octalide:
Part of #2199.

## What

Add scalar replacement of aggregates, so a small `rec`/`uni` whose address never
escapes is decomposed into individual SSA values and kept in registers instead of
being built and re-read through stack memory.

Today it never is. `Result[i64, str]` is a 24-byte record (`{i8, {i64, ptr}}`)
that should live in two registers. Here is what one iteration of a loop that
calls a `Result`-returning function and unwraps it lowers to on 4.2.1 — note that
`step_checked`, `ok`, `err`, `is_err` and `unwrap_ok` are **all correctly
inlined**, so this is not an inlining failure:

```
block bb2:
  %9  = alloca {i8, {i64, ptr}} 1        # (1) slot for the returned Result
...
block bb11:                              # err path
  %49 = alloca {i8, {i64, ptr}} 1        # (2)
  memzero %49
  store 0: i8, ...                       # write tag
  store @.Lc...: ptr, ...                # write payload
block bb13:                              # ok path
  %61 = alloca {i8, {i64, ptr}} 1        # (3)
  memzero %61
  store 1: i8, ...
  store %44: i64, ...
block bb15:
  %72 = phi {i8, {i64, ptr}} [bb12 %49, bb14 %61]
  store %72, %9                          # spill the phi into (1)
block bb21:
  %87 = alloca {i8, {i64, ptr}} 1        # (4) — to read the tag
  store %9, %87
  %90 = load i8 ...                      # is_err
block bb16:
  %73 = alloca {i8, {i64, ptr}} 1        # (5) — to read it again
  store %9, %73
  %76 = load i8 ...                      # tag, second time
  %84 = load i64 ...                     # unwrap_ok
```

**Five `alloca`s per iteration**, three of them with a redundant `memzero` that is
immediately overwritten by explicit field stores, and the value round-trips
through memory three separate times before its payload is read. Every one of
those allocas is a fresh slot for a value that never escapes and is dead by the
end of the iteration.

The kernel this comes from measures **20.6x `cc -O3`** — the worst of 19 in the
benchmark, and the C side is doing exactly the same work via the idiomatic
`int` status plus out-parameter.

Three separable wins, in dependency order:

1. **SROA / mem2reg for non-escaping aggregates.** Decompose the record into
   per-field SSA values; the phi becomes two scalar phis and the tag test becomes
   a register compare.
[...]

PR #2265, opened 2026-07-25
fix(me): mem2reg aggregate promotion (#2246) and dce phi cycles (#2247)

### LINKED ISSUE #2246: me: mem2reg promotes an aggregate-typed slot and substitutes a null it cannot represent
Found while landing #2205 (SROA). Not a bug on `dev` today, but a live trap the
moment another pass hands mem2reg the shape.

## What

`mem2reg` classifies an alloca as promotable purely from escape analysis — its
result must only be the pointer of a load or a store. It never asks whether the
slot's *value* fits an SSA register. An aggregate slot whose every use happens to
be a whole-aggregate copy (a `store %src, %slot`, its address never reaching a
gep or a memzero) therefore passes the test, and `alloca_ty` is taken from that
store's value type — the aggregate.

Promotion then does two things the IR has no representation for:

- it places phis of aggregate type at the iterated dominance frontier of the
  store blocks, and
- for a path with no prior store it substitutes `value.const_null(alloca_ty)`.

An aggregate flows by address in this IR, so a `const_null` stamped with an
aggregate type is not address-producing — exactly what `VC_AGG_ADDRESS` forbids,
and what the back end would read as a null pointer. The phi it feeds is usually
dead, but a dead phi cycle survives DCE (see the companion issue), so it reaches
the back end.

## Reproduction

Not reachable from lowered IR on `dev`: an aggregate local always carries a
memzero or a field gep, both of which escape it out of the promotable set. It
became reachable the moment #2205 created per-field slots — a field whose type
was a `uni` became a slot whose only use was a whole-slot copy, and
`--verify-ir` on `std/data/toml` reported

```
error: aggregate representation: aggregate-typed operand is not address-producing
```

against

```
%2167 = phi uni{ptr, i64, f64, i8, {ptr, i64, i64}, {ptr, ptr, i64, i64}}
        [bb64 null: uni{...}, bb190 %2167, bb40 %2167, bb58 %1989]
```

#2205 closed its own exposure by only splitting an aggregate whose whole tree
splits to scalars, so it never leaves an aggregate slot behind. The mem2reg hole
itself is untouched.

## Why not just guard it

The obvious guard — drop a candidate whose `alloca_ty` is an aggregate — is
**not** inert: it changes six modules' release output on `dev`
(`fe/sema/infer`, `target/abi/{aapcs64,mos6502,spirv,sysv,win64}`), because
mem2reg is today deleting real dead aggregate copies through exactly this path.
So the fix has to keep the promotions whose IDF is empty (no phi, no undefined
path) and drop only the rest, or express an undefined aggregate value in a form
the representation invariant admits.

## Acceptance

- An aggregate-typed slot never yields a phi with a `const_null` incoming, and
  `--verify-ir` stays clean on a module built to produce one.
- The dead-aggregate-copy removal `dev` relies on today is preserved (the six
  modules above stay byte-identical).
- A regression test builds the shape directly (a whole-aggregate copy into a slot
  on one arm of a diamond, no loads) and pins the outcome.

### LINKED ISSUE #2247: me: dce's use-count sweep cannot remove a dead phi cycle
Found while landing #2205 (SROA).

## What

`dce`'s value sweep is use-count driven: an instruction is dead when its use
count reaches zero, and removing it decrements its operands' counts. That is
exact for a DAG and wrong for a cycle — a set of phis that only feed each other
keeps every member's count at one, so none is ever seeded and the whole cycle
survives to the back end.

mem2reg produces exactly such cycles routinely. It places phis at the iterated
dominance frontier of a slot's store blocks with no liveness pruning, so a slot
stored on one arm of a diamond inside a loop gets a phi at the merge and another
at the loop header, each naming the other, and neither naming anything else.

## Cost

Measured on the `result_chain` kernel from #2199 while #2205 was in an earlier
shape that promoted more slots: six dead loop-carried phis survived DCE, spilled
(the loop ran out of registers) and cost **six loads and six stores per
iteration** for values nothing reads. The final #2205 design sidesteps the shape
rather than fixing DCE, so the limitation is untouched and still reachable —
any pass that promotes a conditionally-stored slot inside a loop hits it.

It is also what keeps the aggregate-typed dead phi of the companion mem2reg
issue alive all the way to code generation.

## Fix

Replace the use-count sweep with a mark-and-sweep from roots — the side-effecting
instructions (stores, memzeros, calls, asm, volatile loads) and the terminators
— marking transitively through operands. A phi reached from no root is dead
regardless of who else names it, which is exactly the cycle case.

This changes emitted code at every optimisation level (`dce` is in the always-on
sequence), so it is not additive with the current debug output and needs a
deliberate re-baseline rather than a quiet fix.

## Acceptance

- A function with a phi cycle that no root reaches comes out with the cycle gone.
- The three-generation self-host fixpoint still holds on both profiles.
- The suite and int goldens are re-blessed together with the change, not
  piecemeal.

octalide:
Two middle-end passes surfaced by SROA (#2205), fixed independently.

Closes #2246
Closes #2247

---

### Reproduction

The promotion path is live on `dev`. A probe build of the compiler (instrumented
`mem2reg`, reporting every promoted candidate whose `alloca_ty` is an aggregate)
finds **11 sites** in a release self-build, and from source:

```mach
fun diamond(c: i64, a: Agg, b: Agg) i64 {
    var s: Agg = a;
    if (c > 0) { s = b; }
    ret c + 1;                       # s is never read
}
```

promotes an aggregate slot and **places an aggregate-typed phi** at the merge
(`stores=2 phis=1`). Both arms store, so both incomings are address-producing and
`--verify-ir` stays quiet — which is exactly the issue's "not a bug on `dev` today,
but a live trap". The `const_null` half needs the arm-only store the issue names, and
that shape is only constructible directly in IR (an aggregate local from source always
carries a memzero, and the inliner hoists a cloned callee slot into the entry block, so
the store always dominates). Built directly — an `alloca St` in the entry, a
whole-aggregate `store %p1` on one arm of a diamond, no loads — `dev`'s pass yields
**1 phi and exactly 1 VC_AGG_ADDRESS violation** (measured by asserting those counts
against the reverted pass; the test passes with `!= 1` flipped in).

### Why the obvious guard is not inert — accounted per module

Dropping every aggregate-`alloca_ty` candidate reproduces the issue's report exactly:
`fe/sema/infer`, `target/abi/{aapcs64, mos6502, spirv, sysv, win64}` — six objects,
release. The probe explains all six. **Every one of the 11 promotions has
`loads=0 stores=1 phis=0`**: it is the home slot of an **unused by-value aggregate
parameter**, e.g. `sysv.classify_arg`'s `agg: abi.AggLayout` ("lp64d-only; unused
under SysV"). The lowerer emits `%s = alloca Agg; store %pN -> %s`; mem2reg promotes
it and `strip_promoted` deletes both, so the dead copy vanishes.

With the naive guard the slot survives mem2reg, **SROA then splits it field-wise**,
and the resulting leaf loads on the parameter cannot be removed — `is_pure(OP_LOAD)`
is `false`, so DCE never drops a load. Minimal case, reduced from `classify_arg`:

```
# dev
fn @ignores(%p0: i64, %p1: {i8, [2]{i8, i32, i8}}): i64 {
  block bb0:  %5 = add i64 %p0, 1 ;  ret %5

# naive guard: +17 instructions, six dead leaf loads
  block bb0:  %9  = gep ptr %p1, 0, 0 ;  %10 = load i8 %9
              ... 6 gep/load pairs over the aggregate's leaves ...
              %5 = add i64 %p0, 1 ;  ret %5
```

### Fix

The pass promotes a slot into SSA **values**, and an aggregate has no SSA value form
in this IR — it flows by address. So:

- an aggregate slot that is **read back** is not promotable at all. Substituting the
  load with the value a store placed in the slot hands the reader the *source's*
  address — an alias, not the copy the store performed. (Not just conservative: that
  is a miscompile the pass would otherwise be free to commit.)
- what survives is an aggregate slot **nothing reads**, where promotion means deleting
  unobservable memory traffic and no phi is needed — so `insert_phis` places none, and
  no undefined incoming can arise.

### Result

The 11 promotions `dev` relies on all keep. **Objects byte-identical on all five
targets, both profiles** (208/208 each), so the six modules stay untouched:

| target | release | debug |
|---|---|---|
| linux-x86_64 | 0 differ | 0 differ |
| linux-arm64 | 0 differ | 0 differ |
| linux-riscv64 | 0 differ | 0 differ |
| windows-x86_64 | 0 differ | 0 differ |
| darwin-aarch64 | 0 differ | 0 differ |

---

### Reproduction

```mach
fun kernel(n: i64) i64 {
    var acc: i64 = 0;  var dead: i64 = 0;  var i: i64 = 0;
    for (i < n) {
        if (i % 2 == 0) { dead = 7; }
        acc = acc + i;  i = i + 1;
    }
    ret acc;
}
```

`dev`, `--emit-ir` (release) — `%32` and `%31` name only each other, and survive:

```
block bb1:  %32 = phi i64 [bb0 0, bb6 %31]        # loop header
block bb6:  %31 = phi i64 [bb4 7, bb5 %32]        # merge
```

and in the emitted asm the cycle costs a callee-saved register (`r12`, spilled in the
prologue), plus `mov r12, rsi` / `mov r12, 7` / `mov rsi, r12` per iteration.

### Fix

Mark-and-sweep from the roots over the same removability policy: the block-listed
instructions DCE may not remove are the roots, marking walks their value operands
transitively, and a removable instruction no root reaches is dead. Ghost pool entries
stay outside both sets, so they still pin nothing and are still reported as surviving.

### Measured win

A register-pressure kernel with eight live accumulators and six conditionally-stored
dead slots (the `result_chain` cost model the issue cites), release, x86_64:

| | dev | branch |
|---|---|---|
| loop-body instructions | 119 | 30 |
| stack loads / iteration | 27 | 4 |
| stack stores / iteration | 27 | 3 |
| retired instructions | 19.27 G | 11.33 G (−41%) |
| cycles | 4.55 G | 2.08 G (−54%) |
| wall | 1.34 s | 0.61 s (**2.2x**) |

Output checksum identical on both. The dead cycles were not just extra phis: they held
enough registers to spill all eight live accumulators, which is why the win exceeds the
issue's "six loads and six stores".

Over the compiler's own source, total object bytes (same input, both compilers):

| target | release | debug |
|---|---|---|
| linux-x86_64 | −8.4% | −1.2% |
| linux-arm64 | −17.1% | −2.4% |
| linux-riscv64 | −12.6% | −1.2% |
| windows-x86_64 | −9.1% | −1.3% |
| darwin-aarch64 | −20.9% | −3.2% |

`.text` alone on linux-x86_64 release: 11,541,349 → 10,076,350 (−12.7%).

### Output delta — per target, both profiles

`dce` is in the always-on sequence, so this is **not** additive with the current debug
output, as the issue anticipated. Same source, two compilers:

| target | release | debug |
|---|---|---|
| linux-x86_64 | 107 / 208 objects | 86 / 208 |
| linux-arm64 | 107 / 208 | 86 / 208 |
| linux-riscv64 | 107 / 208 | 86 / 208 |
| windows-x86_64 | 108 / 208 | 87 / 208 |
| darwin-aarch64 | 107 / 208 | 86 / 208 |

Every change is a removal of a dead phi cycle (and the pure computation only that
cycle kept alive): every target shrinks, on both profiles, and no golden needed
re-blessing — the int observables are behavioural and all 79 linux cases plus the
full riscv64 leg pass unchanged.

`-g` additivity holds: `.text` byte-identical with and without `-g` across all 208
objects, on both profiles.

---

## Verification

- **Fixpoint** — three generations, both profiles: `B == C == D` on release and on
  debug (the seed-built generation A differs, as expected, since the 4.2.2 seed lacks
  both fixes).
- **Suites** — `dev` baseline read fresh: mach **856 / 0**, mach-std **611 / 0** at
  pin `eff1e8e` (`git rev-parse HEAD` checked against `mach.lock`). Branch: mach
  **859 / 0** (+3 new tests), mach-std **611 / 0**.
- **int** — `linux` 79 / 79 pass, `linux-riscv64` all pass under qemu, no re-bless.
  The `linux-arm64` leg is `native` in `targets.conf` and cannot execute on this
  x86_64 host; CI covers it.
- **`--verify-ir`** — clean over a full compiler self-build on both profiles, and over
  the mach-std suite.

### Mutations

| test | mutation | result |
|---|---|---|
| `dce.run:dead_phi_cycle_removed` | restore `dev`'s use-count `value_dce` | **4 passed, 1 failed** / 5 |
| `mem2reg.run:aggregate_slot_takes_no_phi` | drop the `insert_phis` aggregate skip | **5 passed, 1 failed** / 6 |
| `mem2reg.run:read_aggregate_slot_stays_in_memory` | drop the read-back rejection | **5 passed, 1 failed** / 6 |
| both mem2reg tests | restore `dev`'s whole `mem2reg` | **4 passed, 2 failed** / 6 |

With the fixes in place: 5 / 5 and 6 / 6.

---

## Note for a follow-up (no change made here)

`sroa.mach` is untouched, as scoped. Its design note records that it "builds no phis"
because an earlier snapshot-at-edges shape left six dead loop-carried phis that DCE
could not remove — the cost measured above. That cost is now gone, so the shared-slot
partition and its two extra rules (a read-only-and-local phi, self-contained incoming
slots) are no longer forced by *this* constraint and a per-incoming design may be back
on the table. Whether it is actually simpler is a SROA design question I did not
attempt; worth a follow-up issue rather than a change here.

### DIFF
--- a/src/lang/me/pass/dce.mach
+++ b/src/lang/me/pass/dce.mach
@@ -2,15 +2,21 @@
 #
 # Two cooperating sweeps run on each non-extern function:
 #
-#   1. value DCE - an instruction whose result is unused and which is removable
-#      once unused (a pure value op, or an OP_PHI, which is not `is_pure` but has
-#      no side effects) is dead. A per-instruction use count drives a worklist:
-#      removing one instruction decrements its operands' counts, which may expose
-#      further dead instructions. Volatile loads and every store/call survive
-#      even when unused, since they may carry side effects. Block lists, not the
-#      instruction pool, are the source of truth for liveness: a pool entry no
-#      longer named by any block - a ghost left by a pass that unlinked it -
-#      contributes no uses and is never seeded, so it cannot pin live values.
+#   1. value DCE - a mark-and-sweep from the roots. The roots are the block-listed
+#      instructions DCE may never remove (the terminators and every side-effecting
+#      opcode - stores, memzeros, calls, allocas, asm, loads, debug values - plus
+#      any volatile op); marking walks their value operands transitively. An
+#      instruction that is removable-when-unused (a pure value op, or an OP_PHI,
+#      which is not `is_pure` but has no side effects) and that no root reaches is
+#      dead. Reachability, not a use count, is the criterion: a use count is exact
+#      on a DAG but wrong on a cycle, where a set of phis naming only each other
+#      keeps every member's count at one and none is ever seeded (#2247). mem2reg
+#      produces such cycles routinely - a slot conditionally stored inside a loop
+#      gets a phi at the merge and another at the loop header, each naming the
+#      other and nothing else. Block lists, not the instruction pool, are the
+#      source of truth for liveness: a pool entry no longer named by any block - a
+#      ghost left by a pass that unlinked it - is neither a root nor a removal
+#      candidate, so it neither pins live values nor is reported dead.
 #   2. reachability DCE - blocks unreachable from the entry block (BlockId 0)
 #      along terminator edges are removed. The block array compacts, every
 #      BlockId reference (terminator targets, predecessor lists, phi incomings)
@@ -94,12 +100,17 @@ fun dce_marked(ctx: ptr, iid: id.InstructionId) bool {
 
 # value DCE for one function
 #
-# Builds a per-instruction use count, seeds a worklist with the unused
-# removable-when-unused instructions, and drains it: each removal decrements its
-# operands' counts and may expose further dead instructions. Block lists are the
-# source of truth for liveness - a pool entry no longer named by any block (a
-# ghost left by an earlier pass) contributes no uses and is never seeded.
-# Survivors are compacted out of their block lists in a final sweep.
+# One sweep over the block lists partitions the listed instructions: a root (an
+# instruction DCE may not remove) is marked live and pushed, a removable one is
+# flagged dead until proven otherwise. Draining the worklist marks each live
+# instruction's value operands live, clearing their dead flags, so what survives
+# is exactly the set the roots reach. A phi cycle no root reaches stays flagged
+# and is removed whole - the case a use count cannot see, since its members keep
+# each other's counts at one. Block lists are the source of truth for liveness: a
+# pool entry no longer named by any block (a ghost left by an earlier pass) is
+# neither seeded as a root nor flagged dead, so it pins nothing and is reported
+# as surviving, exactly as before. The dead instructions are compacted out of
+# their block lists in a final sweep.
 # ---
 # m:   the owning module; its allocator backs the scratch arrays
 # fn:  the function to process
@@ -108,22 +119,22 @@ fun value_dce(m: *ir.Module, fn: *ir.Function) R.Result[bool, str] {
     val ilen: u32 = fn.instr_len;
     if (ilen == 0) { ret R.ok[bool, str](true); }
 
-    val res_use_count: R.Result[*u32, str] = A.allocate[u32](m.alloc, ilen::usize);
-    if (R.is_err[*u32, str](res_use_count)) {
-        ret R.err[bool, str](R.unwrap_err[*u32, str](res_use_count));
+    val res_live: R.Result[*bool, str] = A.allocate[bool](m.alloc, ilen::usize);
+    if (R.is_err[*bool, str](res_live)) {
+        ret R.err[bool, str](R.unwrap_err[*bool, str](res_live));
     }
-    val use_count: *u32 = R.unwrap_ok[*u32, str](res_use_count);
+    val live: *bool = R.unwrap_ok[*bool, str](res_live);
 
     val res_dead: R.Result[*bool, str] = A.allocate[bool](m.alloc, ilen::usize);
     if (R.is_err[*bool, str](res_dead)) {
-        A.deallocate[u32](m.alloc, use_count, ilen::usize);
+        A.deallocate[bool](m.alloc, live, ilen::usize);
         ret R.err[bool, str](R.unwrap_err[*bool, str](res_dead));
     }
     val dead: *bool = R.unwrap_ok[*bool, str](res_dead);
 
     val res_work: R.Result[*id.InstructionId, str] = A.allocate[id.InstructionId](m.alloc, ilen::usize);
     if (R.is_err[*id.InstructionId, str](res_work)) {
-        A.deallocate[u32](m.alloc, use_count, ilen::usize);
+        A.deallocate[bool](m.alloc, live, ilen::usize);
         A.deallocate[bool](m.alloc, dead, ilen::usize);
         ret R.err[bool, str](R.unwrap_err[*id.InstructionId, str](res_work));
     }
@@ -131,14 +142,12 @@ fun value_dce(m: *ir.Module, fn: *ir.Function) R.Result[bool, str] {
 
     var i: u32 = 0;
     for (i < ilen) {
-        use_count[i] = 0;
-        dead[i]      = false;
+        live[i] = false;
+        dead[i] = false;
         i = i + 1;
     }
 
-    count_uses(fn, use_count);
-
-    # seed: every block-listed instruction that is removable-when-unused and unused.
+    # mark: seed the roots from the block lists, then close over their operands.
     var wlen: u32 = 0;
     i = 0;
     for (i < fn.block_count) {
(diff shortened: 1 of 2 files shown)


octalide:
Post-merge re-verification, against the `dev` tip this landed on (`362fcc39`, 22
commits past the branch point) rather than the branch point the body reports. Same
conclusions, numbers within noise.

**Fixpoint** — three generations past the seed-built A, both profiles: release
`B == C == D`, debug `B == C == D`.

**Suites** — `dev` baseline re-read at `362fcc39`: mach **865 / 0**, mach-std
**611 / 0** at pin `eff1e8e` (`git rev-parse HEAD` checked). Merged branch: mach
**868 / 0** (+3), mach-std **611 / 0**.

**int** — `linux` and `linux-riscv64` (qemu) all cases pass, no re-bless.

**`--verify-ir`** — clean over a full self-build on both profiles.

**`-g` additivity** — `.text` byte-identical with and without `-g`, 208/208 objects,
both profiles.

**#2246 inertness re-proved at the new tip** — the mem2reg change alone, same source,
two compilers: **0 differing objects** on all five targets, both profiles. The
aggregate-promotion census is unchanged at the new tip: **11 sites, all
`loads=0 stores=1 phis=0`**.

**#2247 delta re-measured at the new tip:**

| target | release objects differing | debug | release bytes | debug bytes |
|---|---|---|---|---|
| linux-x86_64 | 107 / 208 | 86 / 208 | −8.36% | −1.16% |
| linux-arm64 | 107 / 208 | 86 / 208 | −16.98% | −2.42% |
| linux-riscv64 | 107 / 208 | 86 / 208 | −12.51% | −1.23% |
| windows-x86_64 | 108 / 208 | 87 / 208 | −9.04% | −1.27% |
| darwin-aarch64 | 107 / 208 | 86 / 208 | −20.81% | −3.22% |

Kernel benchmark, best of five: wall 1.307 s → 0.592 s (**2.21x**), cycles
4.42 G → 2.01 G (−54.7%), retired instructions 19.27 G → 11.33 G (−41.2%); loop body
119 → 30 instructions and 27+27 → 4+3 stack loads/stores per iteration. Output
checksum identical.

Issue #2245, opened 2026-07-25
target(spirv): freestanding-spirv is advertised but cannot produce any artifact


octalide:
Surfaced by #2213 (PR #2242) while building a byte-identity harness. Pre-existing.

`mach info targets` advertises **`freestanding-spirv`**, but it cannot produce an artifact by either route:

- **Executable link fails.** `emit_spirv_image` adds no symbols, so `_start` never resolves.
- **`--emit obj` fails** unless the manifest explicitly pins `of = "spv"`, because `os = freestanding` defaults to `raw`, and the raw format refuses object output.

So the only way to get a `.spv` out is a manifest that overrides the object format the target itself selected.

## Why it matters

Advertising a target that cannot build anything is worse than not listing it — someone selects it, gets a link error about `_start` on a GPU target that has no notion of an entry symbol, and has no way to tell whether they misconfigured something or the target is simply not wired.

It also interacts with **#2121** (`mach build --target spirv` writing a `.spv` end to end), which is the feature this blocks, and with **#2213**, which just made SPIR-V a first-class family rather than a register machine with nil slots.

## Fix

Decide and implement one of:

1. **Make it work** — the whole-module emitter path should select `spv` as its object format without a manifest override, and `--emit obj` should be the supported output since a linked executable is meaningless here.
2. **Stop advertising it** until #2121 lands, so `mach info targets` lists only targets that can build.

Option 1 is the better outcome and is largely #2121's ground; option 2 is acceptable as an interim if that is far off. **Do not leave it advertised and broken.**

## Bar

- Whichever option: `mach info targets` and what the toolchain can actually do must agree.
- If made to work: a `.spv` produced from a default manifest with no `of` override, and validated by an external SPIR-V validator rather than only by mach's own reader.
- Other targets byte-identical.

Issue #2211, opened 2026-07-25
codegen: loops are not rotated -- top-tested with an unconditional branch back


octalide:
Part of #2199.

Loops in the benchmark corpus are emitted top-tested with an unconditional branch
back from the body, so every iteration executes two branches where a rotated,
bottom-tested loop executes one.

## Repro

```mach
pub var u8_a: [262144]u8;
pub val N_BYTES: i64 = 262144;

fun count() i64 {
    var n: i64 = 0;
    var i: i64 = 0;
    for (i < N_BYTES) {
        if (u8_a[i] > 127) { n = n + 1; }
        i = i + 1;
    }
    ret n;
}
```

`mach build . --profile release --emit-ir`:

```
block bb1:                                       # header: tests every iteration
  %27 = phi i64 [bb0 0, bb6 %26]
  %28 = phi i64 [bb0 0, bb6 %20]
  %6  = load i64 @..N_BYTES: ptr
  %7  = cmp_lt_s i8 %28, %6
  cbr %7: i8, bb2, bb3
block bb2:                                       # body
  ...
                                                 # ...and the body ends `br bb1`
```

The same shape reaches the binary. From a disassembled 4.2.1 release build of an
element-wise f32 loop, the body ends:

```asm
add    rcx,0x1
jmp    <header>        # unconditional branch back, then a conditional test
```

## Expected

The loop is rotated: the condition is evaluated once before entry and then tested
at the bottom, so the steady-state iteration takes a single conditional backward
branch and no unconditional jump.

## Actual

Header-tested with an unconditional `jmp` back from the body. Two branches per
iteration, and the header block is re-entered every time — which also keeps the
invariant `load` of the bound (see #2204) inside the loop rather than in a
preheader.

## Prior art

#1937 (`codegen: loop rotation, bottom-test loops`) covered this and is closed,
and 4.2.0's changelog lists loop rotation among the shipped optimizer work. This
issue is **not** a request to revisit that one — it is a report that the
transform does not reach the loop shape above on 4.2.1. Worth establishing early
which of these it is, because the fixes differ:

- the pass runs but declines this shape (multiple phis in the header? a
  non-constant bound? the `for (cond)` lowering producing a header the matcher
  does not recognize?), or
- the pass runs and the rotated form is undone by a later block-layout or
  cleanup pass, or
- the pass does not run in this pipeline configuration.

Checking the IR immediately after the rotation pass versus at emission should
separate the first from the second.

## Environment

[...]

PR #2298, opened 2026-07-25
assess(me): LICM register pressure (#2255) — do not merge, #2246 already closed it

### LINKED ISSUE #2255: me: LICM has no register-pressure model — hoisting costs ~2.8% on the compiler's own build
Split out of #2204 (PR #2221, merged) with its measurements. Part of #2199.

LICM ships with a real tradeoff, measured rather than estimated:

| | array workload | compiler self-build |
|---|---|---|
| instructions | 1.345 G → 0.798 G (**−40.7%**) | 260.15 G → 263.53 G (+1.30%) |
| cycles | 256.4 M → 162.0 M (−36.8%) | 78.87 G → 80.70 G (+2.32%) |
| wall | 75.99 → 48.01 ms (**1.58x**) | 25.39 → 26.11 s (+2.8%) |

Residual static footprint: **+2,801 instructions, 90% of them frame-slot references**. The middle end hoists without any estimate of register demand, so a hoist that lengthens a live range can force a spill that costs more than the redundant computation it removed.

## Two hypotheses already tested and eliminated — do not repeat them

**Copy coalescing does not pay for it.** Measured across #2222 landing: +2.95% → +2.67%, static spill growth essentially unchanged (+14,263 → +14,205 frame-slot refs). Coalescing removes reg-reg *copies*; it does not reduce the register *pressure* longer live ranges create.

**Trip-count gating cannot help.** #2139 already exposes everything needed (`is_counted`, `iv.init`, `iv.step`, `counted.bound`, `bound_inclusive`) — a compile-time trip count is ~15 field reads. Instrumented over the 206-module build, hoist sites are **86.3% uncounted, 11.9% counted with a runtime bound, 1.8% counted with a constant trip**. Exactly **two** hoists in the entire build sit in a loop provably tripping ≤ 1, so a short-trip gate reaches 0.12% of hoists.

The stronger "hoist only from counted loops" variant was **built and measured**, not predicted: it destroys 72% of the array win (1.58x → 1.16x) and removes **none** of the compiler cost — retired instructions identical to three significant figures. The array benchmark's hoists are 87.2% uncounted against the compiler's 86.3%; countedness does not separate the workloads.

**Conclusion: the cost is concentrated in counted, hot loops — exactly the ones any trip-count gate would keep.** An earlier framing calling this "the classic short-trip-loop tradeoff" was an inference and is wrong; the static attribution was right.

## What would actually close it

1. **A pressure-aware hoisting cost model.** The middle end would need to estimate register demand, which it has no basis for today. The larger and more general fix.
2. **Middle-end loop rotation**, so the preheader is guarded by the loop's own entry test rather than running on loop entry. Note #2211 rotates in the back end, after the middle end has already placed the preheader.

## Urgency

This will **drift upward**. The baseline frame-slot count rose **16.5%** when SROA (#2205) landed — more promoted aggregates, more pressure — so the same static footprint costs more cycles now than it did before. Each future promotion-style optimization makes this worse.

## Bar

- Compiler self-build regression measurably reduced, with the array win preserved — both measured on a clean pair emitted by the same pipeline differing only in the change. (#2204's original +0.24% figure was wrong precisely because the two sides came from different pipelines; the true baseline was +2.95%.)
- Debug builds byte-identical; three-generation fixpoint `A == B == C`.
- Whatever heuristic lands is mutation-tested, and its decisions are attributable — no unexplained tuning constants.

octalide:
**Do not merge.** This is the assessment #2255 asked for, and the answer is that neither
of its two candidate closers should be built — because **the regression #2255 documents no
longer exists.** It was removed by #2246 (PR #2265), not by anything here. The pressure
model on this branch is complete, tested and mutation-pinned, and on current `dev` it is a
**measured regression**. It is pushed as the record behind that claim.

Recommendation: close #2255 as resolved by #2246, and close this PR.

## The finding

Every figure is `instructions:u` (this host has `perf_event_paranoid=2`, so perf narrowed
the counter to user space and kernel work was never counted). Wall and cycles are omitted:
the fleet was loaded and a same-binary cycle comparison swung **1.044x**, so they carry no
signal. The instruction-count floor, measured today by running the same binary against
itself six times under load average 20.6, is a spread of **0.00003%**.

Both sides of every pair are emitted by pipelines differing only in the change under test.

### LICM's cost on the compiler's own build, bisected across `dev`

| tree | no LICM | LICM | delta |
|---|---|---|---|
| `c382092c` (this branch's base, and what #2255 measured) | 260.863 G | 264.248 G | **+1.298%** |
| `362fcc39` (#2245 merged) | 264.045 G | 267.446 G | +1.288% |
| **`f308adbd`** (#2246: *sweep dead values by reachability from roots, not by use count*) | 224.773 G | 223.082 G | **−0.753%** |
| `origin/dev` today | 206.610 G | 205.129 G | **−0.716%** |

`f308adbd` is a single commit and it does two things at once: it cuts the compiler's own
retired instructions by **14.9%** (264.0 G → 224.8 G), and it inverts LICM's sign.

The mechanism is the one #2255 named, arriving from the other side. DCE swept by *use
count*, which cannot remove a dead cycle — a value used only by other dead values. Those
survived into the back end and occupied registers in exactly the loops LICM hoists into, so
each hoist tipped a loop into spilling. Sweeping by reachability from roots removes them,
loop pressure falls, and the same hoists now fit. **#2255's diagnosis was right; the
pressure was not LICM's to fix.** Removing it at the source is the better fix, and it
already landed.

### What this branch does to current `dev`

| | array workload | compiler self-build |
|---|---|---|
| no LICM | 19.163 G | 204.413 G |
| `dev` today | 11.737 G | 202.946 G (**−0.72%**) |
| this branch | 11.737 G (+0.0007%) | 204.422 G (**+0.73%**) |

It preserves the array win to seven digits and gives back the entire compiler win. On
current `dev` there is no version of "decline a hoist under pressure" that helps, because
the hoists it declines are now profitable.

## The other two hypotheses, assessed

**Middle-end loop rotation — assessed and rejected before building.** Rotation guards the
preheader with the loop's own entry test, so hoisted work is skipped when the loop trips
zero times. It does not change register pressure at all: a hoisted value is live across the
loop body either way, and the peak inside the loop is identical. Since #2255's residual is
90% frame-slot traffic, rotation cannot touch it. It would also make the problem worse — a
guarded preheader is reached only when the loop runs, which *relaxes* the speculation rules
and admits more hoists. The branch-count win it would otherwise buy is already banked by
#2211 (PR #2220) in the back end.

**A pressure-aware cost model — built, measured, and it does not pay.** Three revisions:

| revision | array win kept | compiler, on `c382092c` |
|---|---|---|
| budget alone | 37% | (static −54%) |
| budget + address exemption | 75% | (static −54%) |
| budget + address chains, width casts cut | **95.5%** | **−0.03%** (from +1.30%) |

On its own base it met the bar. It does not survive the base moving.

## What the model does, for the record

- **Demand** — a backward liveness solve over the function, re-read per candidate with that
  value *and every value already lifted* forced live throughout. Measuring it that way
  rather than adding one per hoist is what keeps it honest: a hoisted value was already
  live over part of the loop, so its cost at the busiest point is zero as often as one.
- **Supply** — the register file the target describes: a bank's declared size less what
  `MachineModel` withholds from value allocation. x86_64: 16 − {RSP, RBP} − {R8, R9, R10} = 11.
  No tuned constants. Where either is unavailable — a whole-module emitter with no register
  machine, or a bank the model does not declare — there is no budget and safety alone decides.
- Over budget, only what costs more than the reload replacing it is still lifted: an address
  computation and the arithmetic feeding it, with the chain **cut where it crosses a width
  cast** (a width cast re-presents an index the loop already holds and cannot be coalesced
  with its source, so lifting it makes the loop carry the datum twice).

## A correction to #2255's framing, which stands regardless

**The static spill attribution does not locate the dynamic cost.** On the base #2255 was
filed against, declining exactly the width-cast hoists removed **100%** of the retired-
instruction penalty while giving back only **14.5%** of the +2,263 frame-slot references
(603,806 → 606,069 → 605,740). `objdump` counts are immune to everything about machine load,
so this half is not in question. A handful of hot spills carried the cost and the static
count could not see which.

The supporting bisection of LICM's hoist population by opcode class, on that base, each
variant built and interleaved:

| variant declined | compiler self-build | array win kept |
|---|---|---|
| nothing (shipped) | +1.30% | 100% |
| loads | +1.29% | 55% |
| bitcasts | +1.34% | — |
| compares | +1.30% | — |
| ALU ops | +1.32% | — |
| **width casts** (`zext`/`sext`/`trunc`, 17.8% of sites) | **−0.05%** | **100%** |

## Verification

- **Suites** — mach **920 / 0** (dev baseline **914 / 0**, plus the 6 added); test-name
  inventory against dev: **0 dropped**. mach-std **611 / 0** at pin `eff1e8e`, verified
  against `mach.lock`. Both re-run through the branch's own self-hosted compiler.
- **int** — green on `linux` and `linux-riscv64`.
- **Debug byte-identical** — 208 objects and the binary, `dev` vs this branch.
- **Three-generation fixpoint** — `B == C` byte-identical.
- **Mutation testing** — seven rules disabled in turn, all seven detected:

| mutation | result |
|---|---|
| budget never binds | 5 of 15 fail |
| register file ignored, raw bank size | 2 fail |
| address exemption removed | 1 fail |
| width-cast rule removed | 1 fail |
| address chain not followed | 1 fail |
| lifted values not held live | 1 fail |
| liveness never solved | 1 fail |

- **Pass compile-time cost** — +0.79% (264.247 G → 266.328 G to compile the tree), after a
  lazy build and popcount bank counting cut it from +2.34%.

## What the next person needs

Nothing here, if #2255 closes. If a future promotion-style pass reopens the pressure
question, this branch has the whole apparatus: the model, six IR-level tests, seven
mutations, and the measurement harness — `licmdelta.sh` reports LICM's on/off delta at any
commit in two builds and is what produced the bisection table above.

### DIFF
--- a/src/lang/me/transform/licm.mach
+++ b/src/lang/me/transform/licm.mach
@@ -2600,3 +2600,155 @@ test "mach.lang.me.transform.licm:width_cast_is_cut_from_an_address_chain" {
     intern.dnit(?itn);
     ret code;
 }
+
+# demand counts a value that merely PASSES THROUGH the loop, not only one the body
+# names. The twenty values below are computed before the loop and consumed after it,
+# so nothing inside the loop mentions them -- yet the allocator must hold every one
+# across every iteration, and the invariant multiply in the body has no register to
+# move into. Reading pressure from the body's own instructions alone would see an
+# empty loop and lift it.
+test "mach.lang.me.transform.licm:demand_counts_values_passing_through" {
+    var alloc: A.Allocator;
+    page.make(?alloc);
+    var tgt: resolved.Target;
+    if (mk_target(?tgt) != 0) { ret 1; }
+
+    val rm: R.Result[ir.Module, str] = ir.init(?alloc, intern.STR_NIL);
+    if (R.is_err[ir.Module, str](rm)) { ret 1; }
+    var m: ir.Module = R.unwrap_ok[ir.Module, str](rm);
+
+    val rt: R.Result[ir_type.IrTypeId, str] = ir_type.intern_int(?m.types, 64);
+    if (R.is_err[ir_type.IrTypeId, str](rt)) { ir.dnit(?m); ret 1; }
+    val i64t: ir_type.IrTypeId = R.unwrap_ok[ir_type.IrTypeId, str](rt);
+
+    val rf: R.Result[u32, str] = ir.function_add(?m, intern.STR_NIL, ir_type.IRT_NIL, 0);
+    if (R.is_err[u32, str](rf)) { ir.dnit(?m); ret 1; }
+    val fn: *ir.Function = ?m.functions[R.unwrap_ok[u32, str](rf)];
+    var bld: builder.Builder = builder.init(?m);
+    builder.set_function(?bld, fn);
+
+    var b: [4]id.BlockId;
+    var i: u32 = 0;
+    for (i < 4) {
+        val rb: R.Result[id.BlockId, str] = builder.new_block(?bld);
+        if (R.is_err[id.BlockId, str](rb)) { ir.dnit(?m); ret 1; }
+        b[i] = R.unwrap_ok[id.BlockId, str](rb);
+        i = i + 1;
+    }
+
+    val n: u32 = 20;
+    val rv: R.Result[*value.Value, str] = A.allocate[value.Value](?alloc, n::usize);
+    if (R.is_err[*value.Value, str](rv)) { ir.dnit(?m); ret 1; }
+    val vs: *value.Value = R.unwrap_ok[*value.Value, str](rv);
+
+    # the preheader computes twenty values and enters the loop.
+    builder.set_block(?bld, b[0]);
+    i = 0;
+    for (i < n) {
+        val ra: R.Result[value.Value, str] = builder.emit_add(?bld, value.param(0, i64t), value.const_int(i::u64, i64t));
+        if (R.is_err[value.Value, str](ra)) { A.deallocate[value.Value](?alloc, vs, n::usize); ir.dnit(?m); ret 1; }
+        vs[i] = R.unwrap_ok[value.Value, str](ra);
+        i = i + 1;
+    }
+    if (R.is_err[bool, str](builder.emit_br(?bld, b[1]))) { A.deallocate[value.Value](?alloc, vs, n::usize); ir.dnit(?m); ret 1; }
+
+    builder.set_block(?bld, b[1]);
+    val rphi: R.Result[value.Value, str] = builder.emit_phi(?bld, i64t);
+    if (R.is_err[value.Value, str](rphi)) { A.deallocate[value.Value](?alloc, vs, n::usize); ir.dnit(?m); ret 1; }
+    val iv: value.Value = R.unwrap_ok[value.Value, str](rphi);
+    val rc: R.Result[value.Value, str] = builder.emit_cmp_lt_s(?bld, iv, value.param(0, i64t));
+    if (R.is_err[value.Value, str](rc)) { A.deallocate[value.Value](?alloc, vs, n::usize); ir.dnit(?m); ret 1; }
+    if (R.is_err[bool, str](builder.emit_cbr(?bld, R.unwrap_ok[value.Value, str](rc), b[2], b[3]))) { A.deallocate[value.Value](?alloc, vs, n::usize); ir.dnit(?m); ret 1; }
+
+    # the body names none of the twenty; only the invariant multiply is a candidate.
+    builder.set_block(?bld, b[2]);
+    val rp: R.Result[value.Value, str] = builder.emit_mul(?bld, value.param(0, i64t), value.param(0, i64t));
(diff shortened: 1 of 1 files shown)


octalide:
Closing unmerged, on the author's recommendation and with its agreement.

The work is complete and green — mach 920/0, mach-std 611/0 at pin, int green, debug byte-identical across 208 objects, fixpoint `B == C`, seven mutations all detected. It should not ship.

#2255's premise was dissolved by #2246 (see the bisection on the issue): sweeping dead values by reachability rather than use count cut the compiler's retired instructions **14.9%** and flipped LICM from **+1.30%** to **−0.72%**. Against that baseline this pressure model is a **+0.73% regression** on the compiler self-build while preserving the array win to seven digits — it declines exactly the hoists that are now profitable.

Recording it here rather than deleting the branch, because the apparatus is sound and the measurement is the deliverable.

Issue #2313, opened 2026-07-25
codegen: integer reductions do not vectorize (no reassociation licence needed)


octalide:
Part of #2199.

**Integer reductions do not vectorize.** Integer addition is associative, so no reassociation licence is needed and no fast-math gate applies — but a plain `s = s + a[i]` declines at every width tested.

## Measured

Same module, same build (`--profile release`, `bba44b7b`), packed-instruction counts from the disassembly:

| loop | body | packed instrs | verdict |
| --- | --- | --- | --- |
| `bare` | `ya[i] = a * xa[i] + ya[i]` (elementwise) | 10 | vectorized |
| `red_i32` | `s = s + ia[i]::i64` | **0** | scalar |
| `red_f32` | `s = s + xa[i]` | 0 | scalar |
| `cond_count` | `if (ia[i] > ib[i]) { n = n + ia[i]::i64; }` | 0 | scalar |
| `cond_branchless` | `n = n + (ia[i] > ib[i])::i64 * ia[i]::i64` | 0 | scalar |

`red_f32` declining is **correct** under default flags — #2160 gated float-reduction vectorization behind fast-math, which is right, and `cc -O3` will not reassociate an FP reduction either. This issue is not about that row; it is included to show the discriminator is the reduction, not the element type.

`cond_branchless` is the one that rules out the obvious alternative explanation. It contains **no control flow at all** — a compare producing 0/1, a multiply, an add — and still declines. So this is not a missing if-conversion; the reduction itself is what is refused.

## Why it costs up to 22x

| kernel | mach packed | cc -O3 packed | vs `cc -O3` |
| --- | --- | --- | --- |
| `u8_count` | 0 | 43 | **22.47x** |
| `i32_filter_sum` | 0 | 19 | **14.36x** |
| `i32_madd` | 0 | 1 | 1.11x |

`u8_count` is `if (u8_a[i] > 127) { n = n + 1; }` over a byte array — GCC turns it into packed compares accumulated into a vector counter, 43 packed instructions. Mach emits a scalar loop with a taken branch, and pays both the lane count and the mispredicts. `i32_filter_sum` is the same shape at 32-bit width with a data-dependent predicate.

`i32_madd` is listed for honesty: it is already at 1.11x after #2206's module-val fold, so vectorizing it is worth very little. The win here is concentrated in the two byte/branch kernels.

## Bar

[...]

Issue #2237, opened 2026-07-25
codegen(x64): every scalar float op routes allocated XMM operands through the fixed scratch pair


octalide:
Part of #2199. Split out of #2203, which reported these copies as an allocator
gap; they are not — they are emitted by the x86-64 encoder, below the allocator.

## What

`encode_float_arith` (and the float `MOV` / conversion / comparison paths beside
it) route **every** scalar float operand through the fixed XMM scratch pair,
unconditionally, even when the allocator already placed both operands and the
destination in allocatable XMM registers. From `src/lang/target/isa/x64/encode.mach`:

> load `lhs` -> FLOAT_SCRATCH0 and `rhs` -> FLOAT_SCRATCH1, run the SSE op
> `SCRATCH0 = SCRATCH0 op SCRATCH1`, then store SCRATCH0 -> `dst`. operating in
> fixed scratch registers (which the allocator never assigns) keeps every float
> value out of the XMM allocator, exactly as cmach does.

The second sentence is no longer true of the rest of the backend: float values
*are* allocated to XMM registers (xmm0-xmm13 below). Only the operations are not.

## Evidence

`dst[i] = alpha*x[i] + y[i]` over module-scope `f32` arrays, release, 4.2.2. The
MIR reaching the encoder is a clean three-address `FMUL v2, v0, v1` with all three
operands in allocated XMM registers; the encoder turns each op into four
instructions:

```asm
movss xmm14, dword [rdx]   # load x[i] -> scratch
movss xmm1,  xmm14         # scratch -> allocated
movss xmm14, xmm0          # alpha -> scratch      (emit_float_load lhs)
movss xmm15, xmm1          # x[i]  -> scratch2     (emit_float_load rhs)
mulss xmm14, xmm15         # the work
movss xmm2,  xmm14         # scratch -> allocated  (emit_float_store dst)
```

`mulss xmm2, xmm0, xmm1` needs `movaps xmm2, xmm0 ; mulss xmm2, xmm1` — two
instructions where the encoder emits six. Nine of the loop's 23 instructions are
these scratch shuttles. The same pattern covers `FADD` / `FSUB` / `FDIV`, the
float `MOV`, the conversions and `FCMP`.

## Why #2203 cannot fix it

The allocator sees `FMUL [v2, v0, v1]` — three vregs, no copies. There is nothing
to coalesce: the copies do not exist until the encoder emits them, and the scratch
registers it emits them against are held out of the allocatable bank on purpose.
[...]

PR #2322, opened 2026-07-25
feat(me): vectorize affine loop indices

### LINKED ISSUE #2312: codegen: the vectorizer requires a zero-based induction-variable index, declining every affine offset
Part of #2199. Found by profiling `f32_matmul`, the residual the epic predicted would need "at least one more child once someone profiles those two".

**The vectorizer only accepts an index that is the loop's own zero-based induction variable.** Any affine offset declines — including one hoisted into its own induction variable.

## Measured

A single module, one loop body (`ya[?] = a * xa[?] + ya[?]`), five index shapes, built `--profile release` at `bba44b7b` and counted from the disassembly:

| loop | index shape | packed instrs | verdict |
| --- | --- | --- | --- |
| `bare` | `ya[i]`, `i` from 0 | **10** | vectorized |
| `affine` | `ya[base + j]` | 0 | scalar |
| `affine_iv` | `t = base; ... ya[t]; t = t + 1` | 0 | scalar |
| `affine_load` | bare store, `xa[base + i]` load | 0 | scalar |
| `affine_store` | bare load, `ya[base + i]` store | 0 | scalar |

`base` is a runtime value, so this is not about constant folding. `affine_iv` is the sharp one: the index there **is** a bare variable incremented by one per iteration — it merely starts at a runtime offset rather than zero — and it still declines. Neither the load side nor the store side is privileged; either one being offset is enough.

## Why it costs 42x

`f32_matmul`'s inner loop is exactly this shape:

```mach
d.mat_out[i * d.MAT_N + j] = d.mat_out[i * d.MAT_N + j] + aik * d.mat_b[k * d.MAT_N + j];
```

The source is already in the interchanged `i-k-j` order on both sides — GCC is not doing anything clever with the loop nest, it is vectorizing the elementwise inner loop as written. Mach declines it purely on the index shape, and the result is 22 instructions per element against C's 8 instructions per 8 elements:

```
mach (inner loop, 22 instrs)          cc -O3 (inner loop, 8 instrs, 2x unrolled)
  mov/add/lea    x2  addressing         movaps (%rcx,%rax,1),%xmm0
  movss  (%rax),%xmm14                  mulps  %xmm3,%xmm0
  movss  %xmm14,%xmm1                   addps  (%rdx,%rax,1),%xmm0
  ... 8 movss reg-reg copies ...        movaps %xmm0,(%rdx,%rax,1)
  mulss  %xmm15,%xmm14                  movaps (%rsi,%rax,1),%xmm1
  addss  %xmm15,%xmm14                  mulps  %xmm2,%xmm1
  movss  %xmm14,(%rax)                  addps  %xmm1,%xmm0
                                        movaps %xmm0,(%rdx,%rax,1)
```

Two useful FP operations out of 22 instructions. `f32_matmul` measures **42.36x `cc -O3`** and is the single worst kernel in the suite — it has not moved since the epic recorded 41.7x on 4.2.1, through every child that has landed since.

## Note on what else is in that 42x

Vectorization is the largest factor but not the only one. The same inner loop also shows 8 register-to-register `movss` copies (#2237, PR #2266 queued) and per-access address recomputation (`mov`/`add`/`lea` rather than a pointer increment). Fixing the index rule alone should be worth roughly the lane count; the rest is separately attributable.

## Bar

- `affine`, `affine_iv`, `affine_load` and `affine_store` all emit packed instructions.
- `f32_matmul` re-measured against `cc -O3` — the number, not "it vectorizes now".
- The `int/surface/vectorize-emit` case gains a per-verdict row for the affine shape, so a future regression is a golden failure rather than a silent decline (#2207's lesson).
- Checksums stay bit-identical: this is an integer-index change, not a reassociation.

octalide:
Closes #2312

The vectorizer accepted an index only when it *was* the loop's induction-variable phi, so every affine offset declined — `a[base+i]`, either access side alone, and even a cursor `t = base; ...; t = t + 1` that is a bare variable incremented by one. `f32_matmul`'s inner loop is exactly that shape and was the worst kernel in the suite.

## The rule

`analysis/dependence.mach` decomposes an index into `off + stride*iv`. Leaves: the induction variable, any loop-invariant value, and a secondary add-recurrence. Interior nodes: add, subtract, and multiply by a constant. Only a **unit stride** is admitted — any other leaves a vector access's lanes non-adjacent. Nothing in the decomposition emits, so an offset that cannot fold to a single value declines rather than growing arithmetic the guard would have to materialize.

Widening the index widens what must be proven, so three things move with it:

- **Same-base pairs.** Two accesses to the same base at different offsets are a nonzero cross-iteration distance the runtime guard cannot discharge — the ranges are the same storage at a shift, not two objects that might overlap. `DEPENDENT` once either writes. Equal offsets remain the distance-0 self dependence that always vectorized. `a[i] = a[i-1] + 1` still declines, now for this reason rather than for the index shape.
- **The alias guard.** Each range endpoint shifts by that access's own offset. A zero offset emits no arithmetic, so an unshifted loop's guard is unchanged.
- **Recurrences.** Every add-recurrence is stepped by the lane count, and the remainder re-enters from the vector loop's final value *per phi* rather than from the induction variable's alone.

## `f32_matmul`

Retired instructions (`instructions:u`), best of 3, C at baseline x86-64:

| | cc -O3 | mach before | mach after |
|---|---|---|---|
| `f32_matmul` | 203,503,663 | 3,420,561,569 (**16.81x**) | 763,013,734 (**3.75x**) |

**4.48x fewer instructions.** Wall clock on the same runs: **42.57x → 3.54x** vs `cc -O3`. Array-loops geomean 2.98x → 2.57x on instructions, 3.22x → 2.52x on wall; whole suite 26.15 G → 23.49 G retired instructions.

Every other kernel's delta is exactly 1.00x, and that is not a measurement claim: of the benchmark's objects only `kernels.o` and `data.o` differ at all, and within them only `k_f32_matmul` (0 → 10 packed) and `data.init` (0 → 2 packed).

## Verification

- **Checksums** `=` on every kernel, before and after.
- **Byte-identity**: compiling the compiler's own 212 objects with the baseline and with this branch, **205 are byte-identical**. The 7 that differ do so only by newly-vectorized loops — each is a `dst[off+i] = src[i]` copy (`dup_bytes`, `write_fixed16`, `seg_dup`, …), packed count up, nothing else moved.
- **Fixpoint** A == B == C.
- **Suites** through the from-source HEAD compiler: mach 911/0, mach-std 611/0 at pin `eff1e8e`; `int` all green on `linux` and `linux-riscv64`. (`linux-arm64` is a `native` leg and cannot execute here; its structural `vectorize-emit` golden is blessed and passes.)
- **Mutation**: 7 of 11 guards fail a case when mutated. The 4 that survive now record at their own site why they are justified rather than dead, per the convention 793f3869 set.

## Goldens

`vectorize-emit` gains a verdict per shape (offset inline and hoisted, either access side, pointer parameter and module-scope array, the matmul row, an affine reduction → `simd`; same-base shifted store at runtime and constant distance, doubled stride, reversed stride, a cursor stepping by two → `scalar`). `vectorize` pairs each vectorizing shape with a `#[scalar]` twin over 5 offsets × 4 trip counts, including offsets no lane count divides.

## Not in scope

**#2313 is mis-diagnosed and is not fixed here.** Integer reductions already vectorize — the shipped golden asserts `sum_i64 simd`, and an `i32` accumulator over an `i32` array emits 6 packed instructions on `dev`. What `u8_count`, `i32_filter_sum` and `i32_madd` actually need is a **widening** accumulator (`s: i64 = s + a[i]::i64` over a narrower array) plus, for the two branchy ones, if-conversion and vector compares. Details on the issue.

### DIFF
--- a/int/surface/vectorize-emit/expect.linux-arm64.txt
+++ b/int/surface/vectorize-emit/expect.linux-arm64.txt
@@ -1,5 +1,17 @@
 arch=aarch64
 add_i32k simd
+aff_const_dep scalar
+aff_cursor simd
+aff_cursor2 scalar
+aff_glob simd
+aff_load simd
+aff_matmul_row simd
+aff_ptr simd
+aff_reduce simd
+aff_reverse scalar
+aff_shift_dep scalar
+aff_store simd
+aff_stride2 scalar
 axpy_f32 simd
 carried_i32 scalar
 cond_i32 scalar
--- a/int/surface/vectorize-emit/expect.linux-arm64.txt
+++ b/int/surface/vectorize-emit/expect.linux-arm64.txt
@@ -3,6 +3,7 @@ add_i32k simd
 aff_const_dep scalar
 aff_cursor simd
 aff_cursor2 scalar
+aff_cursor_k simd
 aff_glob simd
 aff_load simd
 aff_matmul_row simd
--- a/int/surface/vectorize-emit/expect.linux-riscv64.txt
+++ b/int/surface/vectorize-emit/expect.linux-riscv64.txt
@@ -1,5 +1,17 @@
 arch=riscv64
 add_i32k scalar
+aff_const_dep scalar
+aff_cursor scalar
+aff_cursor2 scalar
+aff_glob scalar
+aff_load scalar
+aff_matmul_row scalar
+aff_ptr scalar
+aff_reduce scalar
+aff_reverse scalar
+aff_shift_dep scalar
+aff_store scalar
+aff_stride2 scalar
 axpy_f32 scalar
 carried_i32 scalar
 cond_i32 scalar
--- a/int/surface/vectorize-emit/expect.linux-riscv64.txt
+++ b/int/surface/vectorize-emit/expect.linux-riscv64.txt
@@ -3,6 +3,7 @@ add_i32k scalar
 aff_const_dep scalar
 aff_cursor scalar
 aff_cursor2 scalar
+aff_cursor_k scalar
 aff_glob scalar
 aff_load scalar
 aff_matmul_row scalar
--- a/int/surface/vectorize-emit/expect.linux.txt
+++ b/int/surface/vectorize-emit/expect.linux.txt
@@ -1,5 +1,17 @@
 arch=x86_64
 add_i32k simd
+aff_const_dep scalar
+aff_cursor simd
+aff_cursor2 scalar
+aff_glob simd
+aff_load simd
+aff_matmul_row simd
+aff_ptr simd
+aff_reduce simd
+aff_reverse scalar
+aff_shift_dep scalar
+aff_store simd
+aff_stride2 scalar
 axpy_f32 simd
 carried_i32 scalar
 cond_i32 scalar
--- a/int/surface/vectorize-emit/expect.linux.txt
+++ b/int/surface/vectorize-emit/expect.linux.txt
@@ -3,6 +3,7 @@ add_i32k simd
 aff_const_dep scalar
 aff_cursor simd
 aff_cursor2 scalar
+aff_cursor_k simd
 aff_glob simd
 aff_load simd
 aff_matmul_row simd
--- a/int/surface/vectorize-emit/src/main.mach
+++ b/int/surface/vectorize-emit/src/main.mach
@@ -48,6 +48,36 @@ pub fun add_i32k(a: *i32, b: *i32, k: i32, n: i64) { var i: i64=0; for(i<n){ a[i
 # both gaps in one loop: a module-scope array scaled by an invariant scalar
 pub fun glob_mul_f32k(k: f32) { var i: i64=0; for(i<512){ gf_a[i]=gf_b[i]*k; i=i+1; } }
 
+# affine indices (#2312): an index that is the induction variable shifted by a
+# loop-invariant offset, through a pointer parameter and a module-scope array, on the
+# store side alone and the load side alone, and with the offset hoisted into a cursor
+# of its own — a second variable incremented by one that merely starts at the offset
+pub fun aff_ptr(a: *f32, b: *f32, o: i64, n: i64) { var i: i64=0; for(i<n){ a[o+i]=b[o+i]+b[o+i]; i=i+1; } }
+pub fun aff_glob(o: i64, n: i64) { var i: i64=0; for(i<n){ gi_a[o+i]=gi_b[o+i]+gi_c[o+i]; i=i+1; } }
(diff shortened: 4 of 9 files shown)


octalide:
**Re-measured from one tree, after #2266 landed. The isolated win is 3.21x, not the 4.48x this PR's body claims.**

The original numbers were taken against `26565c8e`, which predates #2266 (x86-64 two-address scalar float form, #2237). Eight of the twenty-two instructions in `f32_matmul`'s inner loop were scratch routing that #2266 removes on its own, so the earlier before/after straddled two changes and credited part of #2266's win here.

Re-run with both sides built from **one tree** — current `dev` (`9f24932c`), and the same tree with only the three files this PR touched restored to their pre-#2312 blobs (`dependence.mach`, `vectorize.mach`, `versioning.mach`). Everything else, #2266 included, is byte-identical between the two. Measurement tree verified green first: mach **948/0**, mach-std **611/0** at pin `eff1e8e`, `int` all green on `linux`; fixpoint **B == C**.

Retired instructions (`instructions:u`), best of 3, C at baseline x86-64:

| | cc -O3 | dev −#2312 | dev | ratio |
|---|---|---|---|---|
| `f32_matmul` | 202,503,645 | 2,428,043,704 (**11.99x**) | 755,513,599 (**3.73x**) | **3.21x fewer** |

Array loops geomean **2.58x → 2.30x**, whole suite **2.52x → 2.37x**, whole-binary total 23.32 G → 21.64 G retired. Wall clock on the same trees, for context only — this host's task-clock has sd 9.97% and max/min 1.42, so nothing rests on it: **15.60x → 3.45x**, checksums `=`.

How the two stack, all on the current tree's terms:

| tree | `f32_matmul` vs `cc -O3` |
|---|---|
| before both | 16.81x |
| +#2266 only | 11.99x |
| +#2266 +#2312 | **3.73x** |

They compose cleanly — operand routing and index recognition are orthogonal — but only the 11.99x → 3.73x step is this PR's.

Unchanged by the re-measurement: every other kernel's delta is exactly 1.00x, and that is a code fact rather than a measurement one. Of the benchmark's objects only `kernels.o` and `data.o` differ between the two trees, and within them only `k_f32_matmul` (0 → 10 packed) and `data.init` (0 → 2 packed).

Two of the baseline's own rows moved a lot between the old and new trees — `f64_dot` 4.26x → 2.49x, `f32_norm` 4.00x → 2.25x, `rec_particle` 4.01x → 2.51x. That is #2266, correctly showing up on both sides here and attributed to neither.

PR #2324, opened 2026-07-25
perf(sroa): carry a Result's union payload into registers

### LINKED ISSUE #2315: codegen: a Result built by an inlined constructor still round-trips through the stack
Part of #2199. Follows #2205, which is closed — SROA landed, but this shape still round-trips through the stack.

**A `Result[T, E]` built by an inlined constructor is materialized in a stack slot and read straight back, every iteration.** `result_chain` measures **19.49x `cc -O3`** and has not moved since the epic recorded 20.6x on 4.2.1.

## Measured

`k_result_chain`, built `--profile release` at `bba44b7b`: **95 instructions** against C's **19** for the whole kernel. The loop body, with the aggregate traffic marked:

```
lea    -0x48(%rbp),%rdx        <- stack slot for the Result
movq   $0x0,(%rdx)             <- zero 16 bytes
movq   $0x0,0x8(%rdx)
lea    (%rdx),%rsi
movb   $0x1,(%rsi)             <- store the discriminant
lea    0x8(%rdx),%rsi
lea    (%rsi),%rdi
mov    %rcx,(%rdi)             <- store the payload
jmp    ...
mov    (%rdx),%rcx             <- read the whole thing back
```

Both arms do this — the error arm builds its own 16-byte temp at `-0x30(%rbp)` with the same zero / store-tag / store-payload / reload sequence. The call *is* inlined (there is no `call` in the loop), so this is not an inlining failure: the constructor's storage simply never leaves memory.

C's side is an `int` status plus an out-parameter, which is the ordinary way C propagates an error — a compare and an add, no memory round trip.

## Why it is worth more than its row

`Result[T, E]` is how essentially every fallible stdlib call reports, so this tax is not confined to one benchmark row — it is paid by ordinary Mach code that touches the standard library at all. The epic said as much when it put #2205 in the "value representation" layer and predicted it was the change that would move `result_chain`. It landed and the row did not move; that is worth knowing before anyone plans further work on the assumption that it did.

`rec_particle` (**8.10x**, 81 instructions against C's 25) is plausibly the same defect seen through a plain `rec` at a 24-byte stride rather than a discriminated one, but I have not separated the two — that is a question for whoever takes this, not a claim.

## Bar

- The `Result` temp does not reach memory in `k_result_chain`; the discriminant and payload live in registers across the branch.
- `result_chain` re-measured against `cc -O3` — the number.
- Whether `rec_particle` shares the cause, answered from its disassembly either way.
- Checksums bit-identical.

octalide:
Closes #2315.

## Mechanism

SROA (#2205) never saw `Result` at all. `splittable_at` requires the *whole* type
tree to reduce to scalars; `Result[T, E]` lowers to `{i8, uni{T, E}}`, and the
union field is neither a scalar nor recursively splittable (`agg_field_count`
returns 0 for `IRT_UNION` by design — union members overlap). So the candidate was
refused at the type test, before escape analysis ever ran.

That was the only blocker. A probe with the identical control-flow shape over a
plain two-field `rec` splits today and reaches registers with no allocas at all —
so this is not an inlining, escape-analysis, ABI-classification, or mem2reg gap.

## Fix

A union field becomes **one** slot rather than per-member slots, whenever every
member is a scalar covering the whole cell. Then each store writes all of it and
each load reads all of it, so the bits one member reads are exactly the bits
another wrote — the reinterpretation a union exists to express is preserved
exactly. The slot's storage type is picked from the members themselves (first
integer, else first pointer, else member 0), so at least one member always
accesses it with no cast, and a member of another type is a plain `bitcast`.

A member narrower than the cell is refused: a partial store would have to preserve
the bytes it does not write, which one whole-cell store cannot express.

`#2246`'s rule is untouched. The cell's slot is a **scalar**, so mem2reg's
aggregate-read rejection does not apply to it and does not need to — no aggregate
load is ever substituted with a stored address.

Measurements to follow on this PR before it is marked ready.

### DIFF
--- a/src/lang/me/transform/sroa.mach
+++ b/src/lang/me/transform/sroa.mach
@@ -13,6 +13,14 @@
 # of field slots, so the address merge becomes a merge of the slots' contents, and
 # the mem2reg run that follows turns those slots into the scalar phis.
 #
+# A union field is the one member that does not decompose: its members share storage,
+# so per-member slots would break the reinterpretation a union exists to express. It
+# becomes ONE slot instead, whenever every member is a scalar covering the whole cell
+# (`union_leaf_ty`) -- then every store writes all of it and every load reads all of
+# it, so the bits one member reads are exactly the bits another wrote, and a member
+# access naming a different type than the cell's is a plain bit reinterpret. This is
+# what carries `Result[T, E]`, whose payload is an anonymous `uni`, into registers.
+#
 # ESCAPE CONDITION (the whole legality argument). A candidate is an OP_ALLOCA of a
 # struct / array type (never a union: its fields overlap) with a constant count of
 # one, or an OP_PHI of that same type, whose whole type tree splits to scalars.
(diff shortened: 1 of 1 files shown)


octalide:
Stopped by priority, not by a defect: the owner closed the performance program. Findings, measurements and the reasoning about which routes are open are recorded on #2315.

Issue #2320, opened 2026-07-25
build: test bodies are compiled into every release object and discarded only at link


octalide:
Found during #2294 (PR #2300), which disproved the assumption that a test-only change cannot move objects.

**`mach build` compiles `test` block bodies into the per-module object, on every build, at every profile.** They are discarded only at link.

Proof, not inference: `strings` on a plain **release** build of `of/elf.o` finds `elf_unres` — a literal used only inside a `test` block. Measured across six targets × both profiles, a change touching only test bodies moved **84 of 2556 files**: exactly the six touched objects plus the twelve binaries linking them. The binaries differ by relayout rather than content — `strings` on the release binary finds no test name, so the symbols really are dropped at link.

## Why this is worth a decision

It may be intended. `mach test` needs those bodies, and compiling them unconditionally keeps one code path instead of two. But nobody has stated that, and it has three costs that nobody has measured:

- **Every release object carries dead code and dead rodata** through codegen, optimization and object emission, to be thrown away by the linker.
- **Build time** is spent optimizing and encoding functions that cannot be called in the artifact.
- **It silently invalidates a common assumption.** "This change is test-side only, so objects are byte-identical" is false here — and that assumption was in a verification bar I wrote. Anyone reasoning about object identity after touching a test will get a surprise and, worse, may bless it as noise.

The third is the immediate one: the whole byte-identity discipline this repo leans on assumes that touching a test does not move a release object.

## The question to answer

Should `test` bodies be excluded from a non-`test` build?

**Establish first whether they can be.** `test` blocks may reference module-private items and may be the only reference keeping something alive; dropping them could change what the optimizer considers reachable, which is a *different* object delta rather than none. Find out before proposing a fix.

Then decide with numbers, not principle:

- What fraction of a release object is test-derived? Per module and in total.
- What does the compiler's own build time change by, if they are skipped?
- Does skipping them change any *non-test* output? It must not.

If the answer is "intended and cheap", say so **in the code** where someone auditing object identity will find it, and the issue closes with that. That is a perfectly good outcome — the defect here is the silence, not necessarily the behaviour.

## Bar

- The per-module and total test-derived fraction of a release build, measured.
- If skipping: **non-test output byte-identical**, one target one profile, widened only on divergence; build-time delta reported both ways; `mach test` still finds and runs every test.
- If keeping: a comment where it will actually be read, plus a note in whatever documents the byte-identity discipline, so the next person's bar does not carry the false premise mine did.
[...]

PR #2325, opened 2026-07-25
fix(sema): reject a generic type named without its type arguments

### LINKED ISSUE #2311: sema: a generic type named without its arguments is silently accepted, and $size_of lays out the template
Found while fixing #2178. **Pre-existing, not introduced by that fix** — verified
identical on `dda66e3d` (before) and `418373f2` (after).

## A generic type named without its arguments is accepted everywhere

There is no arity check on the **zero-argument** spelling of a generic type. `Box`
where `Box[T]` was declared is silently accepted in every type position:

```mach
rec Box[T] { v: T; w: T; }

fun f() u32 { var b: Box; ret 0; }          # accepted, 0 diagnostics
fun g() u64 { ret $size_of(Box); }          # accepted, 0 diagnostics
fun h() u64 { ret $offset_of(Box, v); }     # accepted, 0 diagnostics
pub def Alias: Box;                         # accepted, 0 diagnostics
```

The non-zero-argument spellings *are* checked — `Box[u32, u32]` reports
`wrong number of type arguments: expected 1, found 2`. Only the bare name escapes.

## It is not inert: the template gets laid out, and the numbers are wrong

The layout intrinsics answer against the **template**, with `T` unbound and
apparently sized as a pointer. Compiled and run:

```
size_of(Box)      = 16
size_of(Box[u8])  = 2
size_of(Box[u16]) = 4
size_of(Box[u64]) = 16

offset_of(Box,w)     = 8
offset_of(Box[u8],w) = 1
```

`$size_of(Box)` is a compile-time constant that is **wrong for every
instantiation except the pointer-width one**, produced with no diagnostic. A
`val BUF: [$size_of(Box)]u8` or an allocation sized this way is silently the wrong
size. That is the same fail-open shape as #2178 — a plausible answer where a
diagnostic belongs.

`$fields(Box)` likewise unrolls the template's own fields with `T` unbound
(`fields(Box) count=2` for the record above), so a reflection walk runs against a
type that has no layout.

## And the unbound parameter escapes into concrete signatures

```mach
rec Box[T] { v: T; }
fun f(b: Box) u32 { ret b.v; }
```

```
error: type mismatch: expected u32, found T
 --> ./src/main.mach:3:21
  |
3 | fun f(b: Box) u32 { ret b.v; }
  |                     ^^^^^^^^
  |
  = note: mach has no implicit widening; cast the value with `value::Type`
```

`T` is a generic parameter of `Box`, and `f` is not generic — yet `T` appears as
the type of a field read in `f`'s body, and the diagnostic offers to fix it with a
cast. The message names a type the user cannot write at that point and cannot
satisfy. The real error is one line earlier, on `b: Box`.

## Where it belongs

`resolve_type_named` (`src/lang/fe/sema/infer.mach`) branches on
`named.args_len > 0` to `instantiate_named`, and otherwise falls to
`type_for_symbol`, which interns the nominal without ever asking whether the
declaration is generic. `generics.check_arity` exists and is already called from
`instantiate_named` and from `infer_generic_call`; the zero-argument path never
consults it.

## Related

Same family as the bare-template defect #2257 addressed on the lowering side (a
template reached with nothing concrete substituted in). This is its front-end
twin: the template is *named* with nothing substituted, and sema accepts it.

## Bar

- `var b: Box;`, `$size_of(Box)`, `$offset_of(Box, f)`, `def A: Box;`, and
  `fun f(b: Box)` are each rejected, naming the arity: the message should say the
  type needs its type arguments, not report a downstream mismatch on `T`.
- `$fields(Box)` is rejected for the same reason rather than unrolling a template.
- Every legal spelling still compiles: `Box[u32]`, `Box[T]` inside a generic,
  nested and multi-argument instances (`int/regression/2178-intrinsic-type-operand`
  covers these and must stay green).
- A test asserting the *diagnostic*, not merely that the build fails — the
  "expected u32, found T" message above is a build failure for the wrong reason.

octalide:
Closes #2311

## The defect

`resolve_type_named` forks to `instantiate_named` only when `named.args_len > 0`;
the fall-through reached `type_for_symbol`, which interned the nominal without
ever asking whether the declaration was parameterized. So a bare `Box` where
`Box[T]` was declared was accepted in every type position, with zero diagnostics.

It was not inert. The layout intrinsics answered against the **template**, with
`T` unbound and sized as a pointer:

```
size_of(Box)      = 16      offset_of(Box, w)     = 8
size_of(Box[u8])  = 2       offset_of(Box[u8], w) = 1
```

`$size_of(Box)` was a compile-time constant wrong for every instantiation but the
pointer-width one, produced silently.

## The fix

`type_for_symbol` is the single map from a resolved type-name symbol to the type
it names, and **both** callers reach it only with the argument-free spelling
(`resolve_type_named` forks first; a `$type_of` comparand cannot spell an
instance). The arity rule lands there: a declaration that declares type
parameters does not name a type when spelled bare.

`generics.declared_arity` is the count-only counterpart to the existing
`check_arity` for the case with no argument list to check against; both now share
a `declaring_ast` helper, so "which module declares this" has one definition.

Only the symbol kinds whose `decl` **is** the type declaration are asked
(`SYM_REC`, `SYM_UNI`, `SYM_IMPORTED`). A `SYM_GENERIC` carries its *owner's*
decl, whose arity is not the parameter's own — asking it would reject every
generic body in the language.

## The diagnostic

```
error: generic type `Box` named without type arguments: expected 1, found 0
 --> ./src/main.mach:8:22
  |
8 | fun f() u32 { var b: Box; ret 0; }
  |                      ^^^
  |
  = note: a generic declaration is a template, not a type; name it with its type arguments
```

The `T`-leaking signature now points at the real cause instead of a type the user
cannot write:

| | |
|---|---|
| **before** | `error: type mismatch: expected u32, found T` at `ret b.v` (col 21), with a "cast the value" note |
| **after** | the message above at `b: Box` (col 10) — one diagnostic, on the signature |

## Surface

**Newly rejected** (each was silently accepted before, each verified to carry the
diagnostic above): local `var b: Box`, parameter `b: Box`, pointee `*Box`, return
type `fun f() Box`, field of a non-generic record, `def Alias: Box`,
`$size_of` / `$align_of` / `$offset_of` / `$fields`, an array length
`[$size_of(Box)]u8`, the `$type_of(x) == Box` comparand, a generic `uni`, and a
two-parameter generic (`expected 2, found 0`).

**Already correct, and still accepted** — verified compiling clean on both the
base and fixed compilers:

- every instance spelling: `Box[u32]`, `Box[T]` inside a generic, nested
  `Box[In[T]]`, multi-argument `Pair[A, B]`, cross-module `O.Option[u32]`
- a declaration with **no** type parameters named bare in every position above
  (local, parameter, pointee, `def` target, all four intrinsics, `$type_of`
  comparand) — `rec Plain` and `uni V`
- a generic **parameter** named bare inside its owner: `var x: T`, `*T`, `[2]T`,
  `$size_of(T)`, `$align_of(T)`, `$fields(T)`
- a `def` alias naming an instance (`def BA: Box[u32]`) and used bare

`int/regression/2178-intrinsic-type-operand` uses only argument-bearing spellings
and is unaffected.

## Verification

- **Objects byte-identical, linux-x86_64 release.** The full compiler + mach-std
  corpus compiled with the base compiler and with the fixed compiler produced
  byte-identical objects (`diff -r`) *and* a byte-identical linked binary
  (`cmp`). The base compiler was itself confirmed byte-identical to one built
  from pristine `dev` HEAD `b9748c03`.
- That same run is the corpus check: the fixed compiler builds the compiler and
  mach-std clean, so neither uses a bare generic name anywhere.
- **Suites through the from-source HEAD compiler:** `914 passed / 0 failed` on
  pristine `dev` HEAD → `915 passed / 0 failed` on this branch. Delta is exactly
  the one new test.
- mach-std materialized with `git archive` from pin `eff1e8e`.
- **Mutations** — see the comment below.
- No fixpoint run: the change is a front-end rejection and codegen did not move
  (objects byte-identical).

## Pre-existing, out of scope

A type-name diagnostic in an intrinsic operand or a signature is reported more
than once (repeated type-resolution passes). This is unchanged by the fix — on
`dev`, `$size_of(Box[u32, u32])` already reported `wrong number of type
arguments` twice, and two wrong-arity sites in a signature already reported three
times. Surfacing it rather than absorbing it; not touched here.

### DIFF
--- a/src/lang/driver/tests.mach
+++ b/src/lang/driver/tests.mach
@@ -552,7 +552,14 @@ fun ut_vec_sema(s: *session.Session, alloc: *A.Allocator, src: str) R.Result[pro
     names[0] = "main.mach";
     var texts: [1]str;
     texts[0] = src;
-    val pr: R.Result[project.Project, outcome.Fail] = ut_build(s, alloc, ?names[0], ?texts[0], 1);
+    ret ut_vec_sema_n(s, alloc, ?names[0], ?texts[0], 1);
+}
+
+# ut_vec_sema over an `n`-file project, for a rule that only fires ACROSS modules:
+# a reference whose origin is another module resolves to SYM_IMPORTED and reads its
+# declaration from that module's AST, a path a single-file fixture never enters
+fun ut_vec_sema_n(s: *session.Session, alloc: *A.Allocator, names: *str, texts: *str, n: u32) R.Result[project.Project, outcome.Fail] {
+    val pr: R.Result[project.Project, outcome.Fail] = ut_build(s, alloc, names, texts, n);
     if (R.is_err[project.Project, outcome.Fail](pr)) { ret pr; }
     var p: project.Project = R.unwrap_ok[project.Project, outcome.Fail](pr);
     val se: R.Result[bool, outcome.Fail] = passes.run_sema_pass(?p);
@@ -576,6 +583,56 @@ fun ut_vec_diag(src: str, needle: str) i32 {
     session.dnit(?s);
     if (hit) { ret 0; }
     ret 1;
+}
+
+# ut_vec_diag over a two-module project: `main` is `main.mach`, `lib` is `lib.mach`,
+# reachable from main as `uniontest.lib`
+fun ut_vec_diag2(main: str, lib: str, needle: str) i32 {
+    var alloc: A.Allocator;
+    if (O.is_some[str](page.make(?alloc))) { ret 1; }
+    val sr: R.Result[session.Session, str] = session.init(?alloc);
+    if (R.is_err[session.Session, str](sr)) { ret 1; }
+    var s: session.Session = R.unwrap_ok[session.Session, str](sr);
+
+    var names: [2]str;
+    names[0] = "main.mach";
+    names[1] = "lib.mach";
+    var texts: [2]str;
+    texts[0] = main;
+    texts[1] = lib;
+
+    val pr: R.Result[project.Project, outcome.Fail] = ut_vec_sema_n(?s, ?alloc, ?names[0], ?texts[0], 2);
+    if (R.is_err[project.Project, outcome.Fail](pr)) { session.dnit(?s); ret 1; }
+    var p: project.Project = R.unwrap_ok[project.Project, outcome.Fail](pr);
+    val hit: bool = ut_diag_has(?p.s.diags, needle);
+    project.dnit_project(?p);
+    session.dnit(?s);
+    if (hit) { ret 0; }
+    ret 1;
+}
+
+# whether the two-module project type-checks with zero error diagnostics
+fun ut_sema_clean2(main: str, lib: str) bool {
+    var alloc: A.Allocator;
+    if (O.is_some[str](page.make(?alloc))) { ret false; }
+    val sr: R.Result[session.Session, str] = session.init(?alloc);
+    if (R.is_err[session.Session, str](sr)) { ret false; }
+    var s: session.Session = R.unwrap_ok[session.Session, str](sr);
+
+    var names: [2]str;
+    names[0] = "main.mach";
+    names[1] = "lib.mach";
+    var texts: [2]str;
+    texts[0] = main;
+    texts[1] = lib;
+
+    val pr: R.Result[project.Project, outcome.Fail] = ut_vec_sema_n(?s, ?alloc, ?names[0], ?texts[0], 2);
+    if (R.is_err[project.Project, outcome.Fail](pr)) { session.dnit(?s); ret false; }
+    var p: project.Project = R.unwrap_ok[project.Project, outcome.Fail](pr);
+    val clean: bool = ut_count_errors(?s.diags) == 0;
+    project.dnit_project(?p);
+    session.dnit(?s);
+    ret clean;
 }
 
 # 0 when compiling `src` through sema + lower + codegen reaches a DEFINED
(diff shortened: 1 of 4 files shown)


octalide:
## Mutation results

Every mutation was applied to the guard, the compiler rebuilt from the mutated
source, and the resulting compiler run against the suite. Baseline is
**915 passed / 0 failed**.

| # | mutation | result |
|---|---|---|
| M1 | guard disabled (`if (false)`) | **914 / 1** — `generic_named_without_type_arguments`, exit 14 |
| M2 | `SYM_IMPORTED` dropped from the kind gate | **914 / 1** — same test, exit **22**, the cross-module case |
| M3 | `SYM_GENERIC` added to the kind gate | mutated compiler **cannot compile any generic body** — `var x: T` inside `fun f[T]` reports the new error; 181 errors on a 4-line program, and it cannot compile the compiler itself |
| M4 | `arity > 0` → `arity >= 0` | mutated compiler **cannot compile the compiler** — 169,396 errors, every non-generic nominal reference rejected |

M2 is why the cross-module coverage was added in `64fac177`: with only single-file
fixtures, dropping `SYM_IMPORTED` passed the whole suite. That arm was
unfalsifiable before.

M3 and M4 kill the compiler outright rather than one test, which is the honest
result for those two edges — they are not narrow guards but the rule's domain.

## Verification, final tree

- **Same source, base vs fixed compiler, linux-x86_64 release: objects
  byte-identical (`diff -r`) and the linked binary byte-identical (`cmp`)**, over
  the full compiler + mach-std corpus. The base compiler was confirmed
  byte-identical to one built from pristine `dev` HEAD `b9748c03`.
- **Branch vs `dev`, same compiler, release:** exactly three objects differ —
  `check.o`, `generics.o`, `infer.o`, the three files edited. `tests.mach` is not
  in a `mach build` object set at all (only `mach test` compiles it), so the
  test-only commits contribute **zero** object delta. No `#2320` interaction.
- `mach test` through the from-source HEAD compiler: `914 / 0` on pristine `dev`
  HEAD → **`915 / 0`** here. Delta is exactly the one new test.
- `int/run.sh --target linux`: **all cases passed**, both profiles, including
  `regression/2178-intrinsic-type-operand`.
- mach-std at pin `eff1e8e` (materialized with `git archive`): **`611 / 0`**.
- No fixpoint run: codegen did not move (objects byte-identical).

## Flagged: two documented degradations in `declared_arity`

Both read as 0 (no arity rule applies) rather than reporting:

- an origin AST that cannot be reached — the reference fails at its own leaf
  instead (`type_for_symbol` reports the unresolvable import), so a missing
  dependency never masquerades as an arity error;
- a declaration that cannot carry type parameters (`def`, `val`) — genuinely has
  no parameters.

Both mirror the contract `check_arity` already documents and enforces on the
argument-bearing path; the first is `check_arity`'s existing `ret true` verbatim.
Raising them here rather than leaving them silent.

Issue #2294, opened 2026-07-25
test: 16 hand-built IsaVTable sites in of/elf.mach read uninitialized fields — the #2215 census does not cover them


octalide:
Found during #2244 (PR #2282), partially fixed there.

`src/lang/target/of/elf.mach`'s tests hand-build vtables as `var isa_vt: isa.IsaVTable;` — a declaration with no initializer — at **16 sites**. Those sites leave the vtable's fields uninitialized, so the test reads whatever the stack happened to hold.

This is not theoretical. `attributes_for` is called unconditionally on every exec emit, so those 16 sites were **reading a garbage function pointer and comparing it to nil**. Whether the comparison said "nil" was a property of the stack, not of the program.

#2282 initialized `.reloc` explicitly at all 16. **`machine`, `emitter` and `model` are still uninitialized at those same sites** — pre-existing, and deliberately left alone so the seam work stayed scoped.

## Why this is worth closing properly rather than initializing three more fields

A test that reads uninitialized memory can pass for reasons unrelated to what it asserts, and can fail on an unrelated change that happens to move the stack. That is worse than a broken test: it is a test that will eventually accuse the wrong commit.

The structural angle matters more. #2215 established that a record can only be hand-built from a declaration with no initializer, and turned that into a **source census** (`src/lang/target/isa/tests.mach`) that forbids the form outside a permitted constructor set — the failure exit code *is* the site count. The point of that design is that it forbids hand-building **including for fields that do not exist yet**, which is exactly the trap here: the seam fields were added later and 16 existing sites silently acquired garbage.

So the question is not "initialize three more fields." It is **why these 16 sites are exempt from the census**, and whether that exemption is right. #2244 judged that the ELF writer's tests legitimately hand-build `IsaVTable`s and that a textual permit-list would be dominated by test sites. That judgment may well be correct — but it means the census has a hole precisely where the next field addition will land, and that should be a deliberate, documented decision rather than a gap.

## Bar

- Establish whether the 16 sites genuinely need a hand-built vtable, or whether a test constructor (`isa.machine_isa` with test values, or a purpose-built `isa.test_vtable`) serves them. Prefer removing the exemption over documenting it.
- If a constructor works: all 16 converted, the census extended to cover `of/elf.mach`, and the census shown to fire on a reintroduced hand-built site — **demonstrate it, then revert**.
[...]

Issue #2329, opened 2026-07-25
dev is red: a bare `isa.IsaVTable` in mir/context.mach fails the construction census


octalide:
`dev` fails its own suite. Reproduced on a clean checkout at `538a9669` with nothing else applied:

```
mach.lang.target.isa.blank:every_construction_site_is_constructed
  ./src/lang/target/isa/tests.mach:450  (exit 1)
922 passed, 1 failed, 923 total
```

Failing on every self-test lane (release x86_64-linux, release aarch64-linux, release x86_64-windows, test aarch64-linux), and therefore on **every open PR's merge build** — so other work is currently reading a red CI that is not its own.

## The site

```
src/lang/be/codegen/mir/context.mach:1293
test "mach.lang.be.codegen.mir.context.vector_scaffolding"

    var arch: isa.IsaVTable;
    arch.pointer_width = 8;
```

## Why it is a race, not a regression

Neither change was wrong on its own:

- **#2289** (PR #2318, shared layout policy) added that bare `IsaVTable` in a test helper.
- **#2300** (PR #2294 census expansion) widened the census to scan the whole of `src` rather than two chosen subtrees, and added `IsaVTable` to the censused set.

They landed without seeing each other, and there is no textual conflict between them — the collision is semantic, in a *count*. This is the third instance of that shape today.

## Fix

Convert the site rather than raise the budget, matching what #2300 did to 55 other sites. The helper only wants `pointer_width = 8`, and `isa.machine_isa` is what the sibling test helpers in `ctvalidate.mach` and `of/macho.mach` already use.

## Separately: the assertion cannot say what it found

Worth fixing while this is fresh, because it is what made the cause hard to locate:

- `if (unpermitted != 0) { ret unpermitted::i32; }` returns `1` for a single unpermitted site, and the count-mismatch path also returns `1` — the two failure modes are **indistinguishable from the exit code**.
- The docstring promises "the count rises by one **and the site is named**", but no site name reaches the test output. Locating this one required re-implementing the census externally to get the path and line.

A diagnostic that cannot distinguish the thing it is reached for invites exactly the wrong first guess.

Issue #2326, opened 2026-07-25
ci: nothing runs after a merge — two PRs that are each green can leave dev red


octalide:
`dev` went red on 2026-07-25 at `538a9669` and **nothing reported it**. It was found by a worker whose own verification started failing, not by CI.

## The gap

CI runs on **pull requests**. GitHub Actions builds `refs/pull/N/merge` — the PR merged into `dev` *as of that moment* — which catches a collision between the PR and the `dev` that existed when the run fired.

**Nothing runs after a merge lands on `dev`.** So two PRs that each merge cleanly, are each green against the `dev` they saw, and conflict only *in combination* produce a red `dev` with no signal at all.

## The instance

- **PR #2300** widened the source census to walk all of `src` rather than two chosen subtrees.
- **PR #2318** added a hand-built censused record in a newly-scanned file — `src/lang/be/codegen/mir/context.mach`, `var arch: isa.IsaVTable;` in the `mir.context.vector_scaffolding` test.

Neither branch could see the other: each branched before the sibling merged, and each merge-CI ran against a `dev` that lacked it. Both green. Combined: `922 passed, 1 failed`.

## Why this is not a one-off

This is the **third** instance of the clean-auto-merge family in a single day, and the first to reach `dev`:

1. **Bitflag collision** — two branches each taking "the next free value" for an `FN_FLAG_*` bit. Git appends both lines, no conflict marker, both constants end up equal.
2. **Duplicate import line** (PR #2287) — `dev` and the branch each added the same `use` line; git appended both; mach rejects the duplicate binding. Green locally on the branch, red only on `refs/pull/N/merge`.
3. **This one** — semantic rather than textual, and invisible to *both* PRs' CI.

The pattern generalizes: **a clean textual merge is not a semantic merge**, and the current signal only ever compares a branch against a past `dev`.

## What to consider — decide with evidence, do not assume the first option

- **Post-merge CI on `dev`.** The direct fix. Cost is one full matrix run per merge; at today's rate that is substantial, and it detects rather than prevents — `dev` is already red by the time it fires.
- **Require branches to be current before merge** (up-to-date-with-base). Prevents rather than detects, but serializes merges and forces a rebase-and-re-run per PR, which at this repo's concurrency may be worse than the disease.
- **A cheap post-merge smoke lane** — build plus `mach test` on one target, not the full matrix. Would have caught this exact failure in well under a minute.
[...]

Issue #2244, opened 2026-07-25
link: apply_reloc is patched onto the vtable by string-matching the arch name after registration


octalide:
Surfaced by #2213 (PR #2242), reported rather than fixed.

`be.linker.register_reloc_hooks` walks the ISA registry and attaches relocation appliers by **comparing the target's name string**:

```
if (str_equals(vt.name, "x86_64")) { vt.apply_reloc = ...; }
```

So a capability the target owns is bolted on **after** registration by a lookup in a different subsystem, keyed on a string.

## Why it is worth closing

- **It defeats the completeness guarantee #2213 just built.** That PR made the family contracts constructor-arity-checked, so omitting a required slot is a sema arity error. `apply_reloc` sits outside that: a target can register fine, pass every check, and simply not receive a hook — failing at link time instead of build time.
- **It is silent when it misses.** Rename a target, add a target whose name does not match a branch, and the hook is never attached. Nothing says so.
- **It inverts ownership.** Every other machine capability is provided *by* the target. This one is provided *to* it, by the linker, which has to know the set of arch names in advance — exactly the coupling the retargeting work exists to remove.

Note #2213 kept it on `IsaVTable` as an object-format seam, which is defensible: mos6502 is a register machine with no packer, so it is not universally owed. That is an argument about *which* record it belongs on, not about the string-matching.

## Fix

Have the target provide its relocation applier the way it provides everything else. If some targets genuinely have none, model that as an explicit optional capability — the pattern #2213 introduced with `with_asm_printer` / `with_dwarf_regs` — rather than as an absent string match.

## Bar

- All five targets emit **byte-identical** output; this is a wiring change.
- Adding a hypothetical sixth target requires no edit to `be/linker.mach`. Demonstrate that concretely.
- Suites; three-generation fixpoint `A == B == C`; mach-std **611 / 0** at pin `eff1e8e` (a bare clone lands on `dev` and reports 663 — not the pin).

PR #2331, opened 2026-07-25
fix(test): construct the scaffolding vtable instead of declaring it bare — dev is red

octalide:
**`dev` is red.** `mach.lang.target.isa.blank:every_construction_site_is_constructed` fails at `538a9669` — 922 / 1. This is a one-line fix to make it green.

## Cause

The census finds exactly one unpermitted site, and it is mine:

```
src/lang/be/codegen/mir/context.mach:1293  var arch: isa.IsaVTable;
```

Added by `c17c5cec` (#2318) to give the vector-scaffolding test the resolved target the shared layout policy needs.

**Neither change was wrong when it merged.** #2294 added the census; #2318 added a bare declaration. Each was green on its own branch, and the pair is red. There was no textual conflict, so nothing flagged it — a semantic merge collision.

## The census is right on the substance

Not just bookkeeping. `var` does not zero, so every field of that vtable except `pointer_width` held the frame's residue. Nothing read them, which is the only reason it worked — and "nothing reads it today" is precisely the property that stops being true later.

It now goes through `machine_isa`, the same family constructor `ctvalidate`'s `t_isa` was moved to for this exact reason, so a field added to `IsaVTable` later is cleared here with no visit to this site.

## Verification

- **923 / 0** on this branch (`dev` is 922 / 1).
- Byte-identity linux-x86_64 release: the change is confined to a `test` block, so no shipped code moves. Reported below rather than assumed — a test-only change *can* move objects (#2320).

### DIFF
--- a/src/lang/be/codegen/mir/context.mach
+++ b/src/lang/be/codegen/mir/context.mach
@@ -1290,8 +1290,11 @@ test "mach.lang.be.codegen.mir.context.vector_scaffolding" {
     classes[0].kind = isa.RC_GENERAL;
     classes[1].kind = isa.RC_FLOAT;
     classes[1].bit_width = 128;
-    var arch: isa.IsaVTable;
-    arch.pointer_width = 8;
+    # through the family constructor, not a bare declaration: `var` does not zero, so
+    # the fields this test does not care about would otherwise hold the frame's residue
+    # (#2294). Only `pointer_width` is read here - the layout queries `op_width_of`
+    # routes to need a resolved target.
+    var arch: isa.IsaVTable = isa.machine_isa(isa.ARCH_UNKNOWN, "test", 0, 8, nil, 0, nil, nil, nil);
     var tgt: target.Target;
     tgt.isa                 = ?arch;
     tgt.model.reg_classes   = ?classes[0];


octalide:
Byte-identity verified rather than assumed: both compilers built the identical `dev` source tree at `--profile release` and `diff -r` over the whole output is **identical**. The change is confined to a `test` block and moves no shipped code — but per #2320 that is a check, not an inference.

octalide:
Closing as a duplicate — `dev` was already repaired by #2330, merged as `c327004a`. Verified green independently: **923 passed, 0 failed** from a source build of merged `dev`.

Our messages crossed; you and `iss-2237` converged on the same one-line conversion within minutes of each other, and its patch was already verified in both directions (923/0 on clean `dev`, 926/0 on its own branch) when I dispatched you. That is my coordination failure, not duplicated effort on your part — I should have checked who had already reached it before assigning.

Both fixes were the same call and the same reasoning: **convert rather than permit**, matching the 55 sites #2300 converted and `ctvalidate`'s `t_isa`.

Your analysis stands on the record regardless, and the substantive half is worth keeping: `var` does not zero, so every field of that vtable except `pointer_width` held the frame's residue. Nothing read them, which is the only reason it worked — and *"nothing reads it today"* is exactly the property that stops holding later. That is the same mechanism that put garbage in 16 `of/elf.mach` sites when `IsaVTable` gained a seam in #2244.

The structural gap you identified — that our green-on-branch signal cannot see this class at all — is filed as #2326, and the collision itself as #2329.

Issue #2257, opened 2026-07-25
comptime: $type_of over a T-typed local in a generic folds against the template parameter — silently wrong arm


octalide:
Found during #2254 (PR #2256), deliberately not widened into. **Pre-existing on `dev` and unchanged by that fix.**

A `$type_of` gate over a `T`-typed local inside a **plain generic** — no variadic pack involved — folds against the **template parameter** rather than the instance type, and silently selects the fallback arm.

## Reproducer

```mach
fun pick[T](seed: T) u32 {
    var out: T = seed;
    var r: u32 = 0;
    $if ($type_of(out) == u32) { r = r + 1; }
    $or                        { r = r + 100; }
    ret r;
}
```

`pick[u32](x)` returns **100**. It must return 1.

## Root cause

Distinct from #2254's. Lowering's `resolve_type_comparison_operand` reads `expr_type` **raw**, while `revalidate_one`'s per-instance types are derived into a **scratch array lowering never sees**. So the comptime fold evaluates against the template parameter.

Note this is why #2254's `$if ($type_of(out) == u32)` case was the *still-live* form of that miscompile: #2252 turned the aggregate truncation into a rejection, but the comptime fold kept mis-selecting with no diagnostic.

## Severity: silent wrong code, not a secrecy hole

No gate is bypassed and nothing leaks — `$type_of` is a comptime type comparison, so the failure is that the *wrong arm* is compiled. But it is silent: no diagnostic, no crash, a plausible answer.

That makes it worse than it looks in one specific way. `$if ($type_of(x) == T)` is the documented idiom for per-element dispatch inside `$each` bodies — `doc/language/*` and the skill both present it as the sanctioned pattern, and the stdlib's `vformat` is built on it. Any generic using that idiom over a `T`-typed **local** silently takes the wrong branch.

## Fix

The comptime fold must read the instance type when one exists. Per #2249's design, the CT re-validation drain is a property of an **inference episode** — whichever pass types a body drains the instances it reaches. #2254 extended that to a pack-tailed function's whole body by **re-deriving** from `type_resolved_ty` (a pristine input written once) rather than substituting on the shared, in-place-rewritten `decl_type`. The same reasoning applies here: a read leaf cannot know how many substitutions a derived slot has absorbed.

[...]

Issue #2178, opened 2026-07-24
sema: $fields(Name[args]) inside a generic silently drops the entire $each body with no diagnostic


octalide:
Found by the max-effort review of PR #2170. **Pre-existing on `dev`.**

### A whole `$each` body vanishes with zero diagnostics

```
rec Box[T] { v: T; }
fun leak[T](b: Box[T]) u32 { var acc: u32 = 0; $each f in $fields(Box[T]) { acc = b.[f]; } ret acc; }
// main: var s: ^u32 = 1234; var b: Box[^u32]; b.v = s; print.u64(leak[^u32](b)::u64);
```

Builds clean — **0 diagnostics** — runs, and prints `0`. The `$each` body was neither type-checked (the `^u32` → `u32` assignment is never reported) nor emitted. The body silently does not exist.

### Root cause

`intrinsic_operand_type` (`src/lang/fe/sema/infer.mach:1296`) resolves its type operand through `symbol_for_expr`, which returns none for a `Name[args]` index-expression spelling like `Box[T]`. The function then returns `TYPE_ERROR`, and its comment asserts *"resolve already reported the unresolved name"* — **which is untrue here**: both `Box` and `T` resolve fine. Downstream, `field_seq_owner` and `infer_stmt_comptime_each` both `ret ok` on `TYPE_ERROR` without reporting, so the loop is dropped in silence.

By contrast `resolve_type_ref` routes a `Name[args]` spelling through `instantiate_named` (`infer.mach:2088`) and handles it correctly. This leaf cannot resolve a generic-args spelling at all, despite being documented as a peer type-source leaf.

The concrete twin `$fields(Box[u32])` *does* report (`unresolved identifier u32`), so only the in-generic spelling is silent.

### Why it matters beyond the dropped code

A silently dropped body is a fail-open: any constant-time gate, type error, or diagnostic that body would have produced simply never fires. The example above hides an unchecked secret projection. Two independent defects would each have caught it, and neither ran.

### Fix

Resolve a `Name[args]` type spelling at this leaf the way `resolve_type_ref` does, via `instantiate_named`. Where a type operand genuinely cannot be resolved, **report** rather than returning a `TYPE_ERROR` that callers absorb — and correct the false comment. Audit `field_seq_owner` / `infer_stmt_comptime_each` for any other path that treats `TYPE_ERROR` as "nothing to do" without a diagnostic having been filed.

### Bar

[...]

Issue #2144, opened 2026-07-23
sema silently poisons unbound identifier to TYPE_ERROR, hiding resolver gaps until lower


octalide:
## Problem

`infer_ident` (src/lang/fe/sema/infer.mach) poisons an operand whose `expr_resolved` slot is `SYMBOL_NIL` to `TYPE_ERROR` **silently** — no diagnostic — on the assumption that the name resolver already reported the failure. When the resolver is *also* silent (e.g. a missing `bind_expr` arm never visits the operand, as in #2138), nothing surfaces in sema at all; the error only appears later in `lower` (symbol_reference/symbol_lvalue in src/lang/me/lower/expr.mach report "unresolved identifier").

This is a defensive fallback masking a structural gap: the sema-only unit harness (`ut_sema_clean`) never runs lowering, so a resolver hole passes the in-memory sema tests and only breaks a full `mach build`. #2138's `:^` strip cast hid this way through the entire 4.1.0 release.

## Fix direction

1. **Remove the silent poison.** A `SYMBOL_NIL` reaching sema is an invariant breach — the resolver must have bound the symbol *or* emitted a diagnostic. Sema should assert/emit rather than quietly coerce to `TYPE_ERROR`, so any future resolver gap surfaces immediately in-memory instead of leaking to `lower`.
2. **Close the harness fidelity gap** so sema-level unit tests can catch resolve/lower divergences (or document that end-to-end int cases are required for any new expr kind's resolution).

## Context
Surfaced by #2138 (declassify `:^` resolution, fixed in #2143 with an end-to-end case but the underlying fallback remains). Frontend territory (#1644).

## Sequencing
Touches src/lang/fe/sema/infer.mach — serialized behind #2137 (ct foundation, in review) and #2136 (required-check), which are on the same file. Dispatch once sema is quiescent.

PR #2343, opened 2026-07-25
fix(sema): resolve each type node once per pass, so a diagnostic fires once

### LINKED ISSUE #2334: fe: a type-name diagnostic in an intrinsic operand or signature is reported two or three times
Surfaced during #2311 (PR #2325), confirmed pre-existing and deliberately not touched there.

A type-name diagnostic in an intrinsic operand or a signature is **reported more than once**, because type resolution runs over the same AST types repeatedly and each pass re-reports.

Measured on `dev`, before #2311's change and unaffected by it:

- `$size_of(Box[u32, u32])` reports `wrong number of type arguments` **twice**.
- Two wrong-arity sites in one signature report **three** times.

#2311's new "generic type named without type arguments" diagnostic inherits the same shape, so the count grows with each rule added on this path rather than being specific to any one of them.

## Why it is worth fixing rather than tolerating

A diagnostic printed two or three times reads as *two or three distinct problems*. The reader's first move is to look for the other sites, find none, and lose confidence in the error count — which is exactly the currency a compiler's diagnostics trade in. It also makes `N errors` meaningless as a signal, and this repo's tests increasingly assert on diagnostic text and counts (#2297's tests pin text precisely because two causes shared one message).

It is also a growth problem: every new rule on the type-resolution path acquires the duplication automatically, so the cost rises as the front end gets stricter — and it has been getting stricter all week.

## What to establish first

**Why does type resolution visit the same AST type more than once?** #2144's investigation found `resolve_all_types` walks every AstType in index order, *including ones the resolver deliberately never visits* — discarded `$if` arms and the orphan nodes `probe_type_payload` leaves behind for every bracket call. That is a plausible source of a second visit, but it is a hypothesis, not the answer. Instrument and count before changing anything.

Then decide where the fix belongs:

- **Suppress at the sink** — deduplicate identical (message, span) pairs. Cheapest, and wrong if two genuinely distinct problems can share a message and a span.
- **Resolve each AstType once** — memoize or mark visited. Correct if the repeat visits are redundant, and a possible compile-time win; wrong if some passes legitimately need to re-resolve after substitution.
- **Report at one designated pass** — only the pass that owns the decision emits. Most principled, most invasive.

Do not pick by taste. The answer follows from *why* the repeat visits exist.

## Bar

- The instrumented count of repeat visits, per site class, before any change.
- Each of the three reproducers above emits **exactly one** diagnostic, verified by running the compiler and counting, not by reading the code.
- **No diagnostic is lost.** Build the compiler and mach-std with the fix and diff the full diagnostic output against `dev` for a corpus of deliberately-broken inputs — the failure mode of any dedup is swallowing a real second error.
- Two genuinely distinct errors at the same span still both report; construct that case explicitly.
- **Objects byte-identical, linux-x86_64 release** — this is diagnostics only. Note a test-only change can still move objects (#2320), so measure rather than assume.
- Report any compile-time delta; if the fix is memoization it may be a small win, and that is worth recording either way.
- Mutation-test whatever guard lands; suites; mach-std 611 / 0 at pin `eff1e8e`.

octalide:
Closes #2334

Instrumented first, per the issue's bar. The stated hypothesis turned out to be
wrong, and one of the three candidate fixes is unsafe.

## The hypothesis was not the cause

#2144's finding — `resolve_all_types` walks nodes the resolver never visited
(discarded `$if` arms, `probe_type_payload` orphans) — describes **extra distinct
nodes**, not repeat visits of one node, and cannot duplicate a diagnostic. The
measured causes are two others.

**A — the index walk and structural recursion overlap.** `resolve_all_types`
visits every AstType *by index*, and every composite resolver (`resolve_type_ptr`,
`_array`, `_secret`, `_fun`, `instantiate_named`, `def_underlying`) recurses into
children that are themselves indexed nodes. Traced on the two-site reproducer:

```
RTR mod=0 tid=4 phase=1 emitted=1     # Box[u32,u32], reached by its own index
RTR mod=0 tid=4 phase=1 emitted=1     # the SAME node, reached as *Box's pointee
RTR mod=0 tid=9 phase=1 emitted=1     # Box[u32,u32] in return position — once
```

2 + 1 = 3. That is why `*Box[...]` reported twice and a bare `Box[...]` once.

**B — body inference re-resolves.** `intrinsic_operand_type` resolves its operand
again at body-inference time, after `resolve_all_types` already did:

```
RTR mod=0 tid=5 phase=1 emitted=1
RTR mod=0 tid=5 phase=5 emitted=1
```

Both are the fact already documented at `src/lang/fe/sema/context.mach:148`.
#2225 solved it **for its own rule only**, with a private `UniReports` set — the
instance-level workaround for the class-level defect.

## Instrumented counts — full compiler + mach-std build

| | |
|---|---|
| `resolve_type_ref` calls | **309,354** |
| distinct (module, tid, substitution) | 191,001 |
| **redundant repeats** | **118,353 — 38.3%** |
| …no substitution active | 94,944 |
| …under a substitution | 23,409 |

Repeat-visit phases: type-resolution 45,303; body inference 46,469;
re-validation 26,581. First-visit phases: type-resolution 180,568;
re-validation 10,433.

**Body inference makes 46,469 resolutions and produces zero first visits** —
every one is a repeat.

## Why the memo is keyed the way it is

Under a substitution the same node resolves **exactly once per distinct
substitution**:

```
RTR mod=0 tid=13 phase=6 subst=84696351     # Box[u8]
RTR mod=0 tid=13 phase=6 subst=101473970    # Box[u64]
```

Those are not redundant — they are the same question with different answers. A
memo keyed on the node alone would collapse them and answer with the **template's**
types, which is the #2178 / #2257 defect class. So the memo is **nil on every
episode**, and keyed on the node only where that is sound: within one module sema
pass `subst` is nil throughout (`init_context` sets it so; the re-validation drain
builds *new* contexts rather than mutating this one).

It also records whether a resolution **reported**, not merely that it happened.
`intrinsic_operand_type` asks "did this operand fail silently?" by watching the
sink across its own call; a suppressed repeat files nothing, so without the
recorded outcome that guard fired its `internal:` message on every ordinary bad
operand. That interaction is the only one in the tree — `reported_since` has
exactly one sema caller.

**Not sink dedup**, and the corpus shows why concretely: one span can carry four
genuinely distinct problems (below). A sink comparing spans would destroy them,
and one comparing messages would be a coin flip whenever two instances produce
the same text. The memo never reads the sink at all.

**Not one designated reporting pass**: body inference legitimately needs the
answer — an operand inside a generic body must resolve under the episode's
substitution, which is what #2178 fixed. It re-resolved only because it had no way
to ask "already done".

## Result — counted by running, not by reading

| reproducer | sites | dev | this branch |
|---|---|---|---|
| `$size_of(Box[u32, u32])` | 1 | 2 | **1** |
| two wrong-arity sites in one signature | 2 | 3 | **2** |
| `$size_of(Box)` (#2311's rule) | 1 | 2 | **1** |
| `$fields(Box)` | 1 | 2 | **1** |
| bare generic, six sites | 6 | 10 | **6** |

## No diagnostic lost

A 20-program corpus of deliberately-broken inputs (mismatches, unknown names,
arity over and under, missing fields, secrecy leaks, mixed-secrecy unions,
`$fields` misuse, bad array lengths, type-as-value, nested composites, `def`
targets, in-generic-body errors), full diagnostic text and span compared:

```
DISTINCT (file, message, span) in dev but not here [lost]:  none
DISTINCT (file, message, span) here but not in dev [new]:   none

dev:         46 emitted, 30 distinct
this branch: 30 emitted, 30 distinct
spans still emitting the same diagnostic twice: none
```

Identical distinct sets; only the duplicate multiplicity is gone.

## Two genuinely distinct errors at the same span

The corpus produced one naturally, and I then built the strongest form — **four**
distinct diagnostics at one identical span, the template's own check plus one per
instance:

```
error: type mismatch: expected u32, found T      main.mach:3:43
error: type mismatch: expected u32, found ^u32   main.mach:3:43
error: type mismatch: expected u32, found ^u64   main.mach:3:43
error: type mismatch: expected u32, found u64    main.mach:3:43
```

Byte-identical on dev and on this branch. Pinned as a test.

## Verification

- **Objects byte-identical, linux-x86_64 release**, same source compiled by the
  dev-HEAD compiler and by this one — `diff -r` clean and the linked binary
  `cmp`-identical. This is diagnostics only, and it shows.
- `mach test` through the from-source HEAD compiler: `925 / 0` on dev
  `c446a879` → **`926 / 0`** here. Delta is exactly the one new test.
- `int/run.sh --target linux`: **all cases passed**, both profiles.
- mach-std at pin `eff1e8e` (`git archive`): **`611 / 0`**.
- **Compile time: ~10% faster** on a full release build — seven alternating
  pairs, mean 16.46s → 14.79s, faster in six of seven. Machine was shared and
  loaded (fix range 13.6–16.5s, dev 15.6–17.0s), so the direction is solid and
  the magnitude is approximate.
- Mutations: see the comment below.
- No fixpoint run: codegen did not move.

### DIFF
--- a/src/lang/driver/tests.mach
+++ b/src/lang/driver/tests.mach
@@ -6134,6 +6134,58 @@ test "mach.lang.driver:generic_named_without_type_arguments" {
     ret bad;
 }
 
+# a type-name diagnostic fires ONCE PER SITE, not once per resolution (#2334)
+#
+# `resolve_all_types` walks every AstType by index while every composite resolver
+# recurses into children that are themselves indexed nodes, and `intrinsic_operand_type`
+# resolved its operand again during body inference. Each resolution re-emitted the
+# node's diagnostics, so one bad annotation read as two or three separate problems and
+# `N errors` stopped meaning anything. These are exact COUNTS: a presence assertion
+# cannot see a duplicate, which is why the defect survived every diagnostic test the
+# suite already had
+test "mach.lang.driver:type_diagnostic_fires_once_per_site" {
+    var bad: i32 = 0;
+
+    # one site, through the intrinsic operand -- the leaf that resolved twice. 2 on dev.
+    if (ut_sema_errs("rec Box[T] { v: T; w: T; }\nfun f() u64 { ret $size_of(Box[u32, u32]); }\n#[symbol(\"main\")]\nfun main(argc: i64, argv: **u8) i32 { f(); ret 0; }\n") != 1) { bad = 1; }
+    if (ut_sema_errs("rec Box[T] { v: T; w: T; }\nfun f() u64 { ret $size_of(Box); }\n#[symbol(\"main\")]\nfun main(argc: i64, argv: **u8) i32 { f(); ret 0; }\n") != 1) { bad = 2; }
+    if (ut_sema_errs("rec Box[T] { v: T; w: T; }\nfun f() u64 { var n: u64 = 0; $each g in $fields(Box) { n = n + 1; } ret n; }\n#[symbol(\"main\")]\nfun main(argc: i64, argv: **u8) i32 { f(); ret 0; }\n") != 1) { bad = 3; }
+
+    # a node nested in a composite: the index walk reaches it, and so does the pointer
+    # that encloses it. TWO sites here, and the second is not behind a composite -- 3 on
+    # dev, which is the count that proves the two causes are independent.
+    if (ut_sema_errs("rec Box[T] { v: T; w: T; }\nfun ptr(p: *Box[u32, u32]) u32 { ret 0; }\nfun ret_ty() Box[u32, u32] { var b: Box[u32]; ret b; }\n#[symbol(\"main\")]\nfun main(argc: i64, argv: **u8) i32 { ret 0; }\n") != 2) { bad = 4; }
+
+    # every composite that recurses: pointer, array, `^`, and a `fun` parameter list.
+    if (ut_sema_errs("rec Box[T] { v: T; }\nfun f(p: *Box) u32 { ret 0; }\n#[symbol(\"main\")]\nfun main(argc: i64, argv: **u8) i32 { ret 0; }\n") != 1) { bad = 5; }
+    if (ut_sema_errs("rec Box[T] { v: T; }\nfun f() u32 { var a: [4]Box; ret 0; }\n#[symbol(\"main\")]\nfun main(argc: i64, argv: **u8) i32 { f(); ret 0; }\n") != 1) { bad = 6; }
+    if (ut_sema_errs("rec Box[T] { v: T; }\nfun f(x: ^Box) u32 { ret 0; }\n#[symbol(\"main\")]\nfun main(argc: i64, argv: **u8) i32 { ret 0; }\n") != 1) { bad = 7; }
+    if (ut_sema_errs("rec Box[T] { v: T; }\nfun f(cb: fun(*Box) u32) u32 { ret 0; }\n#[symbol(\"main\")]\nfun main(argc: i64, argv: **u8) i32 { ret 0; }\n") != 1) { bad = 8; }
+
+    # six distinct sites, one diagnostic each. 10 on dev, and the shape a reader would
+    # have to hand-audit to discover the count was fiction.
+    if (ut_sema_errs("rec Box[T] { v: T; w: T; }\nrec Outer { b: Box; }\ndef Alias: Box;\nfun a(p: *Box) u64 { ret $size_of(Box) + $align_of(Box) + $offset_of(Box, w); }\n#[symbol(\"main\")]\nfun main(argc: i64, argv: **u8) i32 { ret 0; }\n") != 6) { bad = 9; }
+
+    # THE PRESERVATION EDGE, and the reason this is a memo rather than a dedup at the
+    # diagnostic sink. One span, FOUR genuinely distinct problems: the template's own
+    # check plus one per instance. They differ only by the substitution in force, which
+    # a sink that compares spans -- or messages that happened to coincide -- would
+    # destroy. The memo keys on the AST node and never reads the sink, so it cannot.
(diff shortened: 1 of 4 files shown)


octalide:
## Mutation results

Each mutation applied to the guard, the compiler rebuilt from the mutated source,
and the resulting compiler run against the suite. Baseline **926 / 0**.

| # | mutation | result |
|---|---|---|
| M1 | memo disabled (`memo_state` always `NONE`) | **925 / 1** — `type_diagnostic_fires_once_per_site`, exit 9 |
| M2 | memo forgets whether the resolution reported (always `QUIET`) | **925 / 1** — same test, exit **13**, the `internal:`-guard assertion |
| M3 | the re-validation episode shares the main pass's memo | mutated compiler **cannot compile the compiler** — 1,812 errors |
| M4 | only the intrinsic-operand half bypasses the memo | **925 / 1**, and the two causes separate cleanly (below) |

**M3 is the one that matters most.** Handing an episode the main pass's memo is
precisely "key on the node, ignore the substitution" — the design the issue
proposed as option 2. It serves the template's answer to every instance and
1,812 errors fall out of a compiler that otherwise builds clean. The "nil on
every episode" property is load-bearing, not incidental.

**M4 separates the two causes**, which is what shows the test covers both rather
than one masking the other:

| reproducer | dev | this branch | M4 |
|---|---|---|---|
| `$size_of(Box[u32, u32])` — 1 site, intrinsic operand | 2 | **1** | **2** (regressed) |
| `*Box[u32,u32]` + return type — 2 sites, composite recursion | 3 | **2** | **2** (still fixed) |

Reverting the body-inference half alone regresses the intrinsic case and leaves
the composite case fixed, so cause A and cause B are independently pinned.

M2 is the interaction I would have missed by reading rather than running: with
reported-ness dropped, every ordinary bad intrinsic operand gains a spurious
`internal: could not intern the type named by this comptime intrinsic argument`,
because the #2178 guard watches the sink across its own call and a suppressed
repeat files nothing. That is the one place in the tree where suppressing a
duplicate is observable beyond the duplicate itself — `reported_since` has
exactly one sema caller, which is why the blast radius is a single guard.

Issue #2277, opened 2026-07-25
driver: ProjectConfig.simd and .vectorize are written and never read


octalide:
Flagged during #2160 (PR #2268), deliberately not widened into.

`ProjectConfig.simd` and `ProjectConfig.vectorize` (`src/lang/driver/project.mach`, `src/lang/driver/config.mach`) are **written and never read**. The effective values come from the resolved profile; these two fields are a second, inert copy of the same settings.

#2160 added a third profile key (`float_reassoc`) and correctly **declined to mirror it** into `ProjectConfig` rather than add a third dead field. That is the right call, and it leaves the structure visibly wrong: two of the three SIMD-related profile keys have a shadow field, the newest does not, and none of the shadows do anything.

## Why this is worth fixing rather than tolerating

A config field that is written and never read is indistinguishable, from the outside, from one that works. Anyone adding a target or a driver path will reasonably wire a new setting into `ProjectConfig` by symmetry with `simd`, and it will silently have no effect. That is the same failure shape as a defensive fallback: it makes a broken wiring look like a working one.

## What to establish first

Find out **why** they exist. A field that is written by the manifest reader and read by nothing is either (a) a leftover from before profile resolution owned these levers, or (b) an intended driver-level override that was never finished. Those want opposite fixes — deletion in the first case, wiring in the second.

Read the profile-resolution path (`manifest.mach` and the driver's profile plumbing) and say which it is, with evidence, before changing anything. Do not assume deletion is right because it is smaller.

## Bar

- State which of (a) or (b) applies and what the evidence is.
- Sweep for the general case, not just these two: any other `ProjectConfig` field (or peer config record) that is written and never read. Report the full list; a one-off fix here leaves the same trap elsewhere.
- If deletion: objects **byte-identical** on every target, both profiles — dead fields cost nothing at runtime, so anything else means they were not dead.
- If wiring: a runnable case showing the driver-level setting taking effect, and a golden pinning it.
- Consider whether the "written and never read" property can be made **structurally impossible** for config records rather than found by inspection. #2215's source-census test (`src/lang/target/isa/tests.mach`) is the working precedent for turning a convention into a build failure.
- Suites; mach-std 611 / 0 at pin `eff1e8e` (a bare clone lands on `dev` and reports 663 — not the pin); three-generation fixpoint `B == C`.

Issue #2273, opened 2026-07-25
spirv: a non-variadic call misdiagnoses as a variadic ABI problem


octalide:
Found during #2245/#2121 (PR #2264), deliberately not fixed there.

On a SPIR-V target, `ret add(a, b)` — an ordinary non-variadic call — fails with:

```
mir.lower_ir: ABI vtable exposes no variadic save model
```

from `be/codegen/mir/abi.mach`. The SPIR-V ABI erases `arg_passing` / `ret_passing` / `va_model`, and `require_abi` fires **before** the emitter can say anything about calls.

## Why it is worth filing separately

Implementing calls on SPIR-V is **#2120**'s job. But a plain non-variadic call reporting a *variadic* problem is a **misdiagnosis independent of that feature** — the user gets pointed at variadics they never wrote, in a message from a pass that is not the one refusing them.

It is also the same family as the two nil-guard bugs #2242 found: `mir.require_abi` and `ctvalidate.run` both guarded `tgt.arch` **before** the whole-module fork, so each would have rejected SPIR-V outright once that field went nil. Shared pre-fork machinery asking register-machine questions of a registerless target is the recurring shape, and this is another instance of it.

## Fix

Either the whole-module path should not reach `require_abi`'s variadic check at all, or the emitter should refuse unsupported constructs **first**, with its own loud message — it already does exactly that for unsupported instructions (`instruction is not yet supported by the SPIR-V target`). Match that.

Audit the rest of the pre-fork path for the same shape while you are there; #2242 fixed two, this is a third, and there may be more.

## Bar

- `ret add(a, b)` on a SPIR-V target reports a message naming **calls**, not variadics.
- Machine targets byte-identical.
- Suites; three-generation fixpoint; mach-std **611 / 0** at pin `eff1e8e` (a bare clone lands on `dev` and reports 663 — not the pin).

Related: #2120 (which implements calls and closes the underlying gap), #2242.

PR #2354, opened 2026-07-25
refactor(target): remove the dead FP argument-register slot and close the debug-hook wiring gap

### LINKED ISSUE #2295: target: dwarf.register_debug_hooks is the last capability wired onto a vtable after registration
Surfaced by #2244 (PR #2282), which removed the stronger instance of this shape and flagged this one as the last survivor.

`dwarf.register_debug_hooks` wires a capability onto an already-registered ISA vtable **after the fact**. #2282 eliminated the equivalent for relocations — `be.linker.register_reloc_hooks`, which attached each arch's relocation packer by string-matching the vtable's name — by making the seam a **required positional parameter** of `isa.machine_isa`, so an arch that omits it is a build-time arity error and an arch that has none writes an explicit `nil` a reader can see.

This is the weaker instance of the same shape, and #2244 was right not to expand into it. Two things genuinely distinguish it: it keys on the **model's own registered name** rather than on a table of names a foreign subsystem holds, and its hook genuinely lives in the back end rather than in the target. Its doc comments were updated in #2282 to stop citing the now-deleted `register_reloc_hooks` and to state plainly that it is the last one.

## The question

Is after-the-fact wiring correct here, or is it the same defect with better manners?

The failure mode to check for is the one #2282 closed: can an ISA register successfully, pass every check, and end up with **no debug hook** — with nothing pointing at the gap until something fails far away? If yes, the fix is the same shape (make it a declared parameter, absence expressed in the type). If no — if the keying genuinely cannot miss — then this is fine and the answer is to say so, in the code, so the next person auditing this pattern does not re-open it.

Either outcome is a good result. Do not assume the #2282 shape must be replicated; establish whether the hazard exists.

## Bar

- Enumerate what happens today for each registered ISA, including one that legitimately has no debug hook. Does any path silently get nothing?
- If it can miss: same treatment as the reloc seam — declared, arity-checked, absence explicit. Demonstrate the failure firing with a temporary stub target and then remove it, as #2282 did.
- If it cannot miss: record why, next to the code, with the reasoning that makes it verifiable rather than a claim.
- Objects byte-identical on every target, both profiles — this is a wiring change, not a codegen change.
- Mutation-test whatever enforcement lands; state the mutation and the `N-1 / 1` result.
- Suites; mach-std 611 / 0 at pin `eff1e8e` (materialize with `git archive` from the pin and diff byte-for-byte — a bare clone lands on `dev` and reports 663); three-generation fixpoint `B == C`.

### LINKED ISSUE #2341: abi: AbiVTable.fp_arg_regs is written by all six conventions and read by nothing
Found during #2273 (PR #2290), scoped out because removing it touches every ABI module.

`AbiVTable.fp_arg_regs` is **written by all six conventions and read by nothing.**

## Why this is the same family as #2273, not just tidiness

#2273 was a vtable slot demanded ahead of the fork that needed it: `lower_call` read `abi_va_model` on **every** call and `require_abi` demanded `gp_arg_regs` on every call and parameter list, though both are variadic-only facts. The consequence was a non-variadic call reporting a variadic ABI problem — a wrong diagnosis pointing at the wrong subsystem, which is how it stayed unfixed through two prior sightings (#2242 found the shape twice).

`fp_arg_regs` is the end state of the same drift: a slot that outlived its consumer entirely. Nothing misreports today because nothing reads it — but a field written by six implementations and read by none is indistinguishable, from the outside, from one that works. The next person adding a convention will populate it by symmetry and get silence. That is exactly what happened with `ProjectConfig.vectorize` in #2277: it was added **five days after** the live path had moved, mirroring an already-dead field.

## Establish before deleting

**Why does it exist?** Either it is a leftover from before the register file owned this, or it is an intended surface a consumer never grew. Those want opposite fixes. #2277's investigation traced its equivalent through git history to the exact commit that orphaned it, and that is the standard here.

Note the neighbouring fact from #2273: `gp_arg_regs` **does** have a consumer, but only a variadic one — the win64 tail-float duplication, which is the register file's only reader. So the question is not "are the register-file slots dead" but specifically "did `fp_arg_regs` ever have a reader, and what removed it".

**Sweep for the class while you are there.** #2277 found nine more dead fields beyond the two it was filed for, including a deep-allocated array rebuilt every build and touched again only to free it. Check every `AbiVTable` field and its peers for the same shape, and report the full list rather than fixing one.

## Bar

- The (a)-leftover vs (b)-unfinished determination, with evidence from history rather than inference.
- The full sweep result across `AbiVTable` and peer records.
- If deleting: **objects byte-identical, one target where identity is genuinely a claim** — a dead field costs nothing at runtime, so any delta means it was not dead and you must explain it. Note a test-only change can still move objects (#2320), so measure rather than assume.
- All six conventions still classify correctly, demonstrated by building and running on the three linkable ISAs — not by reading the classifier.
- Mutation-test anything added; suites; mach-std 611 / 0 at pin `eff1e8e`; fixpoint only if codegen moves.

## Related

Two soft-degrade paths were flagged in #2290 and deliberately left in place, on unreachable-with-a-wrong-answer paths: `cfg_refusal` falling back to a bare reason if interning the function-qualified message fails, and `reg_value` falling back to the machine word when a consumer names no width. Both are documented at their sites. They are not this issue, but they are the same "is this a fallback or a contract" question and whoever takes this may want to rule on them at the same time.

octalide:
Closes #2341
Closes #2295

Two ends of one shape: a vtable slot whose consumer is missing.

---

## #2341 - `AbiVTable.fp_arg_regs`: leftover, or unfinished?

**Neither. It never had a reader, in either compiler.**

| commit | date | what |
| --- | --- | --- |
| `5f882a32` | 2025-11-24 | C bootstrap declares `sysv64_get_fp_arg_regs` in `mir/abi/sysv64.h` |
| `602d69ad` | 2025-11-24 | defines it in `mir/abi/sysv64.c` - no caller |
| `1cb7a70e` | 2025-12-01 | "Remove unused MIR components" deletes it, still uncalled |
| `06cb9b84` | 2026-05-31 | the self-hosted rebuild reintroduces it as an `AbiVTable` slot |

`git grep get_fp_arg_regs` at every commit the pickaxe reports returns exactly the
declaration and the definition, never a call. `git grep fp_arg_regs <commit> -- src/`
outside `target/abi/` is empty at every self-hosted commit that touches it. So it is not a
slot whose consumer was removed - it is one **mirrored from `gp_arg_regs` by symmetry at
birth**, the `ProjectConfig.vectorize` shape from #2277, and it survived the one commit
that was explicitly deleting unused MIR components.

There is no consumer to grow either. `arg_passing` / `ret_passing` hand back a concrete
register or stack slot per value, so lowering never enumerates an FP bank. `gp_arg_regs`
survives only because the win64 variadic tail-float duplication must name the GP register
at an FP argument's index (#2273) - a need with no FP twin.

Removed: the field, the six vtable writes, the four orphaned `fp_param_regs` accessors,
the mos6502 zero stub, and the three self-test assertions that only exercised them. The
record now states why no FP twin exists, so the next convention does not add one back by
symmetry.

## #2295 - `dwarf.register_debug_hooks`: can after-the-fact wiring miss?

**Yes - and it was demonstrated, not argued.** Neutering the registrar and building a
compiler:

- registration: silent
- `mach info targets`: clean, every target listed
- plain build: succeeds, correct output
- `-g` build: `error: codegen: debug producer registered no produce hook`

So it is not silent the way the reloc seam was - there is a purpose-built guard at
dispatch - but it is **late and anonymous**: `-g` only, at codegen, naming neither the
model nor the missing registrar. The identity list (target layer) and the be-layer
registrars are hardcoded in two files in two layers, joined only by the order
`driver.setup_registry` calls them.

**The #2244 treatment does not transfer.** That seam worked because `isa.machine_isa` is
one constructor every arch calls and an arch can name its own packer. Here the identity
must be registrable *without* the back end - `mach info`, `mach doc` and `mach init`
compose registries from `target.register_all` alone, and a format naming an unregistered
model has to fail `select_of` loudly there too - while the producer can only be named from
`be`. That is the same import direction that erases `produce` to `*u8`. Threading it as a
required parameter would push a be-layer pointer through the target layer's generic
bring-up and make ~40 call sites write `nil` meaning "no back end here" rather than "no
producer exists", institutionalizing the ambiguity rather than removing it.

**So the pairing is enforced at the join**, the one point where both halves are in scope:
`of.debug_hooks_complete` walks the registry after the registrars run, and
`setup_registry` refuses any registry holding an unwired identity, naming the model. Doc
comments that #2282 left describing this as the unguarded survivor now say what guards it.

**Stub-target demonstration.** A second identity (`"codeview"`) added to
`debug_register_identities` with no registrar, then removed:

```
error: debug model 'codeview' is registered but no producer wired its hooks
       (its be-layer registrar is missing from driver.setup_registry)
```

fired on *every* build, `-g` or not, at bring-up - against pre-fix behaviour where the
same gap was invisible until a `-g` codegen.

## Verification

**Objects byte-identical** - `2580` artifacts, **6 declared targets x 2 profiles**
(`linux-x86_64`, `linux-arm64`, `linux-riscv64`, `windows-x86_64`, `darwin-x86_64`,
`darwin-aarch64`; release + debug), built from one fixed source tree by the pre-change
compiler and by the post-change compiler. `sha256sum` sets equal. Covers sysv64 / aapcs64
/ lp64 / win64 and elf / coff / macho. The fixed tree is `dev` at `5f31576`; the eight
commits `dev` has taken since touch only `.github/`, `dist/`, `doc/` and `int/`, no
`src/`, so the measurement still stands against current `dev` (the compiler binary is
bit-identical before and after merging it).

**Suites** - `948 -> 950`, `0 failed`. The delta is exactly the two tests added; no test
was removed, only assertions on the deleted accessors.

**Mutation tests**

| mutation | result |
| --- | --- |
| `setup_registry` drops the `debug_hooks_complete` call | **949 / 1** (`driver:setup_registry_refuses_an_unwired_debug_identity`) |
| `debug_hooks_complete` never reports a gap | **948 / 2** (that test + `of.debug_hooks_complete:refuses_unwired_identity`) |

The first mutation initially survived, against a state-asserting test; the test was
rewritten to plant an unwired identity and drive `setup_registry` through it.

**int, three linkable ISAs** - `linux` and `linux-riscv64` green as configured;
`linux-arm64` green with its leg switched to `qemu` run-mode locally only (this host is
x86_64 - `int/targets.conf` is unmodified in the branch, and `run.sh` is explicit that the
aarch64 leg must stay `native` in CI).

**mach-std** - `611 passed, 0 failed` at pin `eff1e8e`, materialized with
`git archive eff1e8e | tar -x` rather than a clone.

**Fixpoint** - `gen2 == gen3` (`680ad94f...`). `gen1 != gen2` is the pre-existing
seed-vs-HEAD divergence: unmodified `dev` reproduces the same shape (`dgen1 != dgen2`,
`dgen2 == dgen3`) from the same seed.

## Sweep - the rest of the class

Fourteen more write-only fields and one dead record across `AbiVTable`'s peers, reported
rather than fixed. See the PR thread.

### DIFF
--- a/src/lang/be/codegen/dwarf.mach
+++ b/src/lang/be/codegen/dwarf.mach
@@ -2906,12 +2906,16 @@ var dwarf_debug_vtable: of.DebugVTable;
 # The target layer owns the `DebugRegistry` slot but cannot name this be-layer
 # producer, so the wiring lands here. It is the last capability attached to a
 # registered vtable after the fact: the ISA relocation packers were the other, and
-# they became a contract the arch declares at construction (#2244). Unlike them
-# this one is keyed on the model's own registered name rather than on a table of
-# names a subsystem holds, so nothing here has to know the set in advance. Sets the DWARF vtable's identity and its erased `produce`
-# hook, then registers it under "dwarf" - the name elf / mach-o's `default_debug`
-# resolves against. Idempotent: `debug_register` keeps the first write-once entry.
-# Must run at startup (from `driver.setup_registry`) before any `select`.
+# they became a contract the arch declares at construction (#2244). That treatment
+# is not available here - the identity must be registrable without the back end
+# present - so the pairing is enforced at the join instead: `driver.setup_registry`
+# runs `of.debug_hooks_complete` right after this call and refuses a registry
+# holding any identity a registrar never wired (#2295).
+#
+# Sets the DWARF vtable's identity and its erased `produce` hook, then registers it
+# under "dwarf" - the name elf / mach-o's `default_debug` resolves against.
+# Idempotent: `debug_register` keeps the first write-once entry. Must run at startup
+# (from `driver.setup_registry`) before any `select`.
 # ---
 # reg: the debug registry to install the DWARF producer into
 # ret: ok(true) once registered
--- a/src/lang/driver.mach
+++ b/src/lang/driver.mach
@@ -45,6 +45,7 @@ use mach.lang.query;
 use mach.lang.readout;
 use mach.lang.session;
 use mach.lang.target;
+use mach.lang.target.of;
 
 use mach.lang.driver.project;
 use dq: mach.lang.driver.query;
@@ -86,13 +87,20 @@ fwd dcfg.resolve_artifact_path;
 # substrates then wires the one remaining be-layer hook onto the registered
 # dimensions (the DWARF debug producer), so every backend-build path starts from a
 # fully composed registry
+#
+# This is also the only place both halves of the debug dimension are in scope, so
+# it is where the identity/producer pairing is checked: `debug_hooks_complete`
+# refuses a registry holding a model identity whose registrar was never added,
+# instead of letting the gap surface at `-g` codegen far away (#2295).
 # ---
 # reg: the target registry to populate before any select
 # ret: ok(true) once registered and wired, or the first error
 pub fun setup_registry(reg: *target.TargetRegistry) R.Result[bool, str] {
     val rr: R.Result[bool, str] = target.register_all(reg);
     if (R.is_err[bool, str](rr)) { ret rr; }
-    ret dwarf.register_debug_hooks(?reg.debug);
+    val dr: R.Result[bool, str] = dwarf.register_debug_hooks(?reg.debug);
+    if (R.is_err[bool, str](dr)) { ret dr; }
+    ret of.debug_hooks_complete(?reg.debug);
 }
 
 # load the project's root manifest, select `target_name`, and
--- a/src/lang/driver/tests.mach
+++ b/src/lang/driver/tests.mach
@@ -9181,3 +9181,38 @@ test "mach.lang.driver:shift_executes_on_a_6502_core_mos6502" {
     ret bad;
 }
 
+
+# `setup_registry` refuses a debug identity whose producer no registrar wired
+#
+# The identity list (target layer) and the be-layer registrars are joined only by
+# the order this function calls them, so a model added to one and not the other
+# registers cleanly, resolves cleanly through `select_of`, and surfaces only when a
+# `-g` build reaches codegen. Standing in for that model: a hook-less entry planted
+# in the registry before bring-up, which `dwarf.register_debug_hooks` does not
+# claim. `setup_registry` must refuse it and name it, and must still accept the
+# registry it composes itself (#2295)
+test "mach.lang.driver:setup_registry_refuses_an_unwired_debug_identity" {
+    var alloc: A.Allocator;
+    if (O.is_some[str](page.make(?alloc))) { ret 1; }
+    val sr: R.Result[session.Session, str] = session.init(?alloc);
+    if (R.is_err[session.Session, str](sr)) { ret 1; }
+    var s: session.Session = R.unwrap_ok[session.Session, str](sr);
+
+    # the registry the driver composes on its own comes up wired.
+    if (R.is_err[bool, str](driver.setup_registry(?s.registry))) { session.dnit(?s); ret 1; }
+    if (of.debug_registered_count(?s.registry.debug) == 0)       { session.dnit(?s); ret 1; }
+    session.dnit(?s);
+
+    # a model whose registrar was never added: bring-up must refuse, naming it.
+    val sr2: R.Result[session.Session, str] = session.init(?alloc);
+    if (R.is_err[session.Session, str](sr2)) { ret 1; }
+    var s2: session.Session = R.unwrap_ok[session.Session, str](sr2);
+    var orphan: of.DebugVTable;
+    orphan.id = of.DBG_CODEVIEW; orphan.name = "codeview"; orphan.produce = nil;
+    of.debug_register(?s2.registry.debug, ?orphan);
+    val res: R.Result[bool, str] = driver.setup_registry(?s2.registry);
+    if (!R.is_err[bool, str](res))                                  { session.dnit(?s2); ret 1; }
+    if (!ut_str_contains(R.unwrap_err[bool, str](res), "codeview")) { session.dnit(?s2); ret 1; }
+    session.dnit(?s2);
+    ret 0;
+}
--- a/src/lang/target.mach
+++ b/src/lang/target.mach
@@ -140,7 +140,9 @@ pub fun host_abi_id() u32 {
 # debug: debug-info producer vtable registry; populated post-hoc by the be-layer
 #        `dwarf.register_debug_hooks` (its producers live in the back end). the last
 #        dimension wired after registration rather than declared at it - the ISA
-#        relocation packers stopped being (#2244)
+#        relocation packers stopped being (#2244). it stays post-hoc because the
+#        target layer must register a model's identity without the back end present,
+#        so `of.debug_hooks_complete` enforces the pairing at bring-up instead (#2295)
 pub rec TargetRegistry {
     isa:   isa.IsaRegistry;
     os:    os.OsRegistry;
--- a/src/lang/target/abi.mach
+++ b/src/lang/target/abi.mach
@@ -263,9 +263,9 @@ pub def ArgPassingFn: fun(i32, u64, u64, bool, u8, bool, bool, bool, i32, i32, u
 pub def RetPassingFn: fun(u64, u64, bool, u8, bool, bool, u8, u8, AggLayout) ParamSlot;
 
 # fill a caller-owned `out` buffer with a register bank in canonical order and
-# return the count written. the erased `gp_arg_regs` / `fp_arg_regs` /
-# `callee_saved` pointers are all cast back to this. the caller must size `out` to
-# hold at least the bank's maximum count
+# return the count written. the erased `gp_arg_regs` / `callee_saved` pointers are
+# both cast back to this. the caller must size `out` to hold at least the bank's
+# maximum count
 # ---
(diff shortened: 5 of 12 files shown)


octalide:
## Sweep: the rest of the class across `AbiVTable` and its peers

Every field of every runtime-dispatch / machine-description record in `src/lang/target/`
(`AbiVTable`, `IsaVTable`, `OfVTable`, `DebugVTable`, `OsVTable`, `resolved.Target`,
`MachineModel`, `RegMachine`, `ModuleEmitter`, `RelocSeam`, `RegClass`, `Register`,
`RegisterFile`, `ParamSlot`, `ParamPiece`, `SigLayout`, `VaModel`, `AggLayout`,
`AggField`, and the four registries), classified by assignment-position vs read-position
occurrences across `src/`, then confirmed by hand.

**Fourteen dead fields beyond the one this PR removes, plus one dead record.**

| record | field | writes | reads |
| --- | --- | ---: | ---: |
| `AbiVTable` | `fp_arg_regs` | 6 | 0 |
| `IsaVTable` | `reg_classes` | 1 | 0 |
| `IsaVTable` | `reg_class_count` | 1 | 0 |
| `RegMachine` | `fp_scratch_reg` | 1 | 0 |
| `RegClass` | `bit_width` | 14 | 0 |
| `MachineModel` | `index_width` | 5 | 0 |
| `MachineModel` | `scratch_regs` | 5 | 0 |
| `MachineModel` | `scratch_len` | 5 | 0 |
| `MachineModel` | `addr_forms` | 5 | 0 |
| `MachineModel` | `addr_scales` | 5 | 0 |
| `MachineModel` | `addr_disp_bits` | 5 | 0 |
| `MachineModel` | `stack_grows_down` | 5 | 0 |
| `MachineModel` | `fp_relative_frame` | 5 | 0 |
| `resolved.Target` | `of_id` | 2 | 0 |
| `isa.RegisterFile` | *(whole record)* | - | - |

They are not one problem. Four distinct groups:

### 1. `RegMachine.fp_scratch_reg` - the same defect as `fp_arg_regs`, one layer down

The FP twin of a live GP slot. `RegMachine.scratch_reg` / `scratch_reg2` are read by
`be.codegen.legalize` (`mach.stage_lo = tgt.arch.scratch_reg`) and by `regalloc`
(`reserve_reg`, spill staging). `fp_scratch_reg` is written once, from a **required
positional parameter of `isa.reg_machine`**, and read nowhere.

Sharpest of the group, because the concept is live while the field is dead: the arm64 and
riscv64 encoders both name *"the vtable's `fp_scratch_reg`"* in their comments
(`isa/arm64/encode.mach:226`, `isa/riscv64/encode.mach:203`) and then use their own
constants. The value each ISA passes into `reg_machine` and the register its encoder
actually reserves can drift apart with nothing to catch it - the arity check #2244 relies
on guards *forgetting* a slot, not *lying* about one, and here nothing reads the slot to
disagree with.

### 2. `IsaVTable.reg_classes` + `reg_class_count`, and `RegClass.bit_width`

A whole descriptor superseded in place. Live readers reach the register classes through
`MachineModel.reg_classes` (`tgt.model.reg_classes`, ~8 sites in `target.mach`,
`mir/context.mach`, `regalloc.mach`); the `IsaVTable` pair is written by the constructor
at `isa.mach:901-902` and read by nobody.

Three doc comments still direct readers to `target.arch.reg_classes`
(`be/codegen/mir.mach:751`, `mir/context.mach:274` and `:300`) - a path that does not
exist: `Target.arch` is a `*RegMachine`, which has no such field.

`RegClass.bit_width` is set by all five ISA register modules and three sites in
`mir/context.mach` and read nowhere, while its siblings `kind` and `len` are read.

### 3. The eight `MachineModel` fields - dead behind a **stale** excuse

These are the deliberate-looking ones, and the excuse no longer holds. The record's
docstring says:

> this definition is purely additive: the stages still read the legacy constants and
> vtable fields until the rewiring lands (issue #1600 group B)

**#1600 is closed** (delivered v2.6.0), and its close-out explicitly *defers or declines*
the group-B remainder - the per-stage `StageVTable` seam and model-parametric MIR are
gated on "a target that cannot run the model-driven stages", and composable type spelling
was declined outright. So eight fields are populated by five ISAs on every build, pointing
at a closed epic for when they will be read.

This is the shape #2277 warns about at its worst: not merely dead, but dead *with a
comment that reads like a plan*, so the next ISA fills them in confidently.

### 4. `resolved.Target.of_id` and the `RegisterFile` record

`of_id` is derived at `target.mach:316` (`t.of_id = t.of.id`) and zeroed at
`driver/project.mach:452`, and read nowhere - while its four siblings `os_id`, `arch_id`,
`abi_id`, `debug_id` all have readers. A single dead member of a five-member symmetric
block: exactly the shape that gets re-added by symmetry.

`isa.RegisterFile` is declared at `isa.mach:91` and **used nowhere**. Its only trace is
four ABI docstrings claiming the register-file accessors return their count "so the call
composes with `RegisterFile`" (`sysv.mach:332`, `win64.mach:308`, `aapcs64.mach:341`,
`lp64.mach:405`). A type declared for a consumer that never existed, with four comments
asserting the composition as though it happens.

---

Not filing these unilaterally. Suggested split if wanted: one issue for group 1 (the
`fp_scratch_reg` drift hazard - the only one with a correctness edge), one for group 2
(supersession plus the three misdirecting comments), one for group 3 (**re-decide**: read
them or drop them, and either way fix the reference to a closed epic), one for group 4.

Issue #1858, opened 2026-07-03
incremental: test-mode main-neutralization mutates query-owned IR in place


octalide:
Part of #1164. Surfaced by #1618's back-half binding (PR #1850).

## What

The test build path neutralizes a module's `main` by mutating the lowered `ir.Module` in place. Since #1618, that IR is owned by the query engine (Q_LOWER's cached value), so the mutation writes through into the cache. On the CLI's single-shot session this is safe — the session dies before any reuse — but a long-lived session interleaving test and normal builds of the same module would serve neutralized IR to a normal build (or un-neutralized IR to a test build), whichever computed first. Same latency class as #1586: harmless today, wrong the moment a persistent or warm-session consumer exists. Confirmed by inspection during the #1618 integration; not observable through the CLI.

## Fix

The invariant to establish: query-owned values are immutable after compute; consumers never write through a cached handle.

Candidates:
1. Neutralize at the consumption site on a clone of the module's IR, keeping query-owned values untouched. Simple, but clones a whole module to tweak one symbol.
2. Make neutralization a derived query keyed on (module, test-mode), so both variants cache independently and query-owned IR is immutable by construction. Preferred — test-mode is an input to the computation and the query key is where inputs belong.

## Acceptance

- No consumer mutates query-owned IR (test path included); the neutralized variant is produced without writing through the Q_LOWER handle.
- A long-lived session interleaving test and non-test builds of the same module yields correct IR for both (driver test in the existing `last_changed`-probe style).
- Byte-identical self-host fixpoint; unit suite and int harness green.

PR #2359, opened 2026-07-25
feat(build): parallelize per-module optimization across cores

### LINKED ISSUE #2100: build: parallelize per-module lowering/optimization across cores
Part of #1825. Follow-up split out of #1826 (parallel codegen).

## Problem

#1826 parallelized the per-module **codegen** phase across a worker pool (codegen 17.7s -> 3.3s on a 16-core self-build), but left **lowering + optimization** serial. Lowering is now the dominant back-half phase: ~6.7s of the ~11.7s parallel-codegen wall on the release self-build. Parallelizing it is the next win (and would push the overall speedup well past the current ~2.15x).

## Why it was deferred

Lowering is where **monomorphization** happens: a generic/comptime instance from another module is monomorphized *into* this module's IR. That path writes shared session state mid-pass:

- it **interns mangled instance names into the shared session interner**, and
- it populates **monomorphization caches** off the session,

with no per-module firewall. #1826's codegen isolation (per-worker child interner + arena) does not cover the lowering writers, and the query engine's Q_LOWER compute is a serial memoizing evaluator besides.

## Direction

Isolate the monomorphization writers per worker the way #1826 isolated codegen: a per-worker child interner over a frozen base for mangled names, and per-worker (or synchronized) monomorph caches, then fold results back deterministically in topo order. The Q_LOWER query binding (serial evaluator) needs the same prestage-into-engine treatment #1826 used for Q_CODEGEN. Determinism bar is identical: byte-identical to the serial build across core counts, all targets, both profiles.

octalide:
Closes #2100.

## The measurement that shaped this

Instrumented the `lower` phase before designing (temporary timers around
`lower.lower_module` / `oblivious.check_module` / `pipeline.run`, reverted).
214-module release self-build, 16 cores, 11.4 s wall:

| | release | debug |
|---|---:|---:|
| monomorphize + lower | 2.32 s | 2.51 s |
| oblivious check | 3 ms | 4 ms |
| optimize (`pipeline.run`) | **5.21 s** | 0.72 s |
| `lower` phase total | 7.5 s | 3.2 s |

The phase is **69 % optimizer**. The issue's premise - that the work is
monomorphization, needing a child interner for mangled names and per-worker
monomorph caches - describes the smaller half; the larger half needs none of
that.

## What is and isn't parallelizable

**Monomorphization is not**, and not for want of a cache. An instance body is
re-derived into the **declaring** module's `expr_type` / `decl_type` arrays
under the instance's substitution (`sema.mach`: *"Instances are emitted one at
a time, so one shared array suffices"*), and installs resolvers and opens
comptime frames on that module's shared `ComptimeCtx`. Two workers
monomorphizing from one origin produce wrong types, not merely nondeterministic
ones. `generics.substitute` also interns into the shared `s.types`, whose ids
reach the IR as `Function.ret_sem` / `DbgVar.ty_sem` for the `-g` producer.
Parallelizing that half means forking the whole front-end mutable typing state
per worker - tracked as a follow-up with the numbers.

**Optimization is**, completely. `pipeline.run` reads the target read-only,
touches the interner only through `intern.lookup` on the scalarize
require-error path (the one `intern.intern` in the pass/transform tree is a
`licm.mach` test helper), and allocates every byte through `Module.alloc`.

## The change

`Module.alloc` becomes per-module. `ir.ModuleHome` is the private page-backed
arena one lowered module owns every byte of; it travels with the module in the
Q_LOWER handle, so `q_lower_finalize` releases the artifact with one
`arena.dnit` rather than a field-by-field walk. That makes the module the sole
owner of its free set and removes the last shared mutable state from
optimization - so byte-identity across `--jobs` is structural, not argued.

The pass then runs in two stages:

1. `lower_prepare` settles every Q_LOWER staleness verdict up front. The
   topo-ordered lowered-surface refresh moves out of `run_lower_one`; safe
   because `Q_LOWERED_SURFACE` depends only on `Q_RESOLVE`, and it is what
   fixes each verdict for the rest of the pass.
2. Stale modules are monomorphized serially, then optimized across a pool sized
   by `--jobs`, and each finished module is staged for `q_lower_compute` to
   adopt - so Q_LOWER stays the value's system of record and no cached value is
   mutated in place (#1858). Warm-valid modules never reach the prelude.

Also fixed: an `IrAsm` payload's `binds` array is module-owned storage
(`ir.function_dnit` frees it through `m.alloc`) but lowering allocated it from
`s.alloc` - identical while the two were the same allocator, wrong once they
are not.

## Verification

Re-verified after merging current `dev` (see the verification comment below for
the full record): determinism is byte-identical across `--jobs 1/2/4/16` in both
profiles, the three-generation fixpoint holds in both profiles, `mach test` is
1001/1001 through the from-source compiler, the warm/incremental determinism
check passes, and the `int` suite passes on linux. Wall-clock numbers are
deliberately **not** included: this machine had peer builds running throughout,
so any figure would be noise. They are deferred to a serialized benchmarking run
before merge.


octalide:
## Revived on current `dev` — verification record

Merged `dev` twice (it advanced mid-verification, picking up #2432), so this
branch now sits on `26155157`. Everything below was re-run from scratch against
the final merge `c10c7a1e`, all in a single pinned process.

### Merge

One conflict, in `q_lower_compute`. `dev` had added a `test_mode` parameter to
`lower.lower_module` (#2320 — test declarations are lowered only into the
test-mode variant) on the very block this branch relocated into `lower_raw` /
`lower_optimize`. Resolved on the side of `dev`'s structure: its hunk was dropped
as already-relocated, and `test_mode` was threaded through `lower_raw` to both
callers — `q_lower_compute` unpacks it from the back-half key, and
`run_lower_parallel` now reads `test_build(p)` once so its staged IR is
necessarily the variant the query will adopt it under. The auto-merged
`lower_module` signature correctly carries both sides' new parameters (`malloc`
from here, `test_mode` from `dev`).

Everything else auto-merged, including `dev`'s lowered-surface fingerprint
rewrite, the `write_build_config` platform-tag/base additions from the #2426
target restructure, and the mach-std pin bump to 0.21.0. The net diff against
`dev` is exactly this branch's 9 files, and removes nothing `dev` added.

### Constant-block distinctness

Checked explicitly, because parallel branches each taking "the next" enum value
auto-merge to the same value with no conflict marker. **This branch introduces no
module-level constants, enum values, bitflags, or opcodes at all** — every
numeric literal it adds is a local loop counter. The collision hazard is
vacuously satisfied rather than merely unobserved.

### Determinism (the core bar)

The parallel pool is confirmed *engaged*, so the byte-identity below is not
vacuous. A temporary probe in `run_lower_parallel` (built, observed, reverted —
not committed) reports:

```
PLWPROBE workers=16 stale=215 test_mode=0
PLWPROBE workers=4  stale=215 test_mode=0
PLWPROBE workers=1  stale=215 test_mode=0
PLWPROBE workers=16 stale=220 test_mode=1   # under `mach test`
```

The 215 → 220 stale-module delta under `mach test` also confirms the merged
`test_mode` threading is live: test IR reaches lowering only in the test variant.

Byte-identical output across worker counts, each from a clean `out/`:

| profile | `--jobs` | sha256 |
|---|---|---|
| debug | 1, 2, 4, 16 | `5efcca73…0773` (all four identical) |
| release | 1, 16 | `0522c3e1…0f00e` (both identical) |

### Standard bars

| bar | result |
|---|---|
| three-generation fixpoint, debug | `a == b == c` = `5efcca73…0773` |
| three-generation fixpoint, release | `ra == rb == rc` = `0522c3e1…0f00e` |
| `mach test` via from-source compiler | 1001 passed, 0 failed |
| `mach test` via release compiler | 1001 passed, 0 failed |
| warm/incremental determinism (`int/lib/check-determinism.sh`) | pass |
| `int` suite, linux target | all cases passed |

The suite-count delta is accounted for exactly: current `dev` measured with the
same compiler is 1000, this branch is 1001, and the branch adds exactly one test
(`test(ir): pin that a ModuleHome owns its module's whole free set`).

### Performance: deliberately not measured

No wall-clock or speedup figures here. Two peer workers were building
concurrently on this machine throughout, so any number produced now would be
noise rather than a measurement. **Benchmarking is deferred to a serialized run
before merge.**

octalide:
## Re-verified on dev `3c1f0c3c` (head `a9428230`)

Merged current `dev` again — it had advanced past the previous run with #2434
(x86-64 indirect JMP encoding, `src/lang/target/isa/x64/encode.mach` only). No
conflict; no overlap with this branch's area. The net diff against `dev` is still
exactly this branch's 9 files.

All bars re-run from scratch in a single pinned process against `a9428230`:

| bar | result |
|---|---|
| three-generation fixpoint, debug | `a == b == c` = `5e969cfe…a514` |
| three-generation fixpoint, release | `ra == rb == rc` = `f23c7d77…c465` |
| final-binary determinism, debug `--jobs 1/2/4/16` | all four identical |
| final-binary determinism, release `--jobs 1/16` | both identical |
| `mach test` via from-source compiler | 1005 passed, 0 failed |
| `mach test` via release compiler | 1005 passed, 0 failed |
| warm/incremental determinism | pass |
| `int` suite, linux | all cases passed |

Suite delta still exactly +1: current `dev` measured with the same compiler is
1004, this branch is 1005, and the branch adds one test.

## Object-level determinism: no nondeterminism found, on this branch or on `dev`

A report reached me that on current `dev` a default-jobs `mach build` produces
byte-**different** individual `.o` files run-to-run while the final binary stays
identical, and that `--jobs 1` restores object-level identity. **That does not
reproduce here.** Measured with the same method on both sides, three clean builds
per configuration, output binary written to an identical path every run:

| configuration | objects differing | final binary |
|---|---|---|
| `dev`, default jobs, debug | 0 / 215 | identical |
| `dev`, default jobs, release | 0 / 215 | identical |
| branch, default jobs, debug | 0 / 215 | identical |
| branch, default jobs, release | 0 / 215 | identical |
| branch, `--jobs 16` | 0 / 215 | identical |
| branch, `--jobs 1` | 0 / 215 | identical |

Object-level determinism is complete on both branches, at every job count and in
both profiles. **This branch changes the determinism status in neither direction.**

Worth recording how the contrary result can arise, because I produced it myself
first and it is convincing until checked. Two apparatus faults, either alone
enough to report "all 215 objects differ":

1. Objects are nested (`obj/std/memory.o`, `obj/mach/lang/...`), so comparing
   them by `basename` looks up a path that does not exist in the second tree.
   Every comparison then "fails" and every object reads as differing.
2. Passing a different `-o` path to the two runs makes the builds genuinely
   different builds.

The tell is that the failure was **uniform** — 215/215, including at `--jobs 1`
where a single thread does all the work and nondeterminism has nowhere to come
from. Genuine nondeterminism is selective. The corrected harness compares by
relative path and self-checks by diffing a tree against itself (must report 0
differences) before any result is trusted.

octalide:
## Serialized benchmark (coordinator run, promised in the revival record)

Method: seed-built compilers from this branch (a9428230) and dev (485129c5), each building its own full compiler tree cold (`out/` deleted between runs), alternating dev/branch repeats in one pinned process, fleet builds held during the window. 16-core host.

| config | dev (median of 3) | branch (median of 3) | delta |
|---|---:|---:|---|
| debug, `--jobs 16` | 4.83 s | 4.29 s | **−11 %** |
| debug, `--jobs 1` | 8.05 s | 7.97 s | parity |
| release, `--jobs 16` | 15.45 s | 8.89 s | **−42 % (1.74×)** |
| release, `--jobs 1` | 28.44 s | 28.74 s | parity |

Noise floor: within-config spread ≤ 0.2 s (debug j16), ≤ 0.8 s (release j16); both deltas are far outside it. Serial (`--jobs 1`) parity in both profiles confirms no regression when the pool is off.

Context: dev's existing parallel codegen already buys 28.4 → 15.5 s in release; this branch's parallel optimization extends the total parallel speedup to ~3.2× over fully-serial. Release gains dominate because the optimize phase is the largest remaining serial chunk there.

Issue #2231, opened 2026-07-25
perf: inline atomics through parsed asm — remove the call and stack-slot staging


octalide:
Blocked on #2229 (parse inline asm into `isa.Inst`). Child of #2212, and part of the Mach-vs-C gap (#2199).

`std.sync.atomic` is implemented as stdlib functions with per-arch `asm` bodies selected by `$if $mach.build.arch` — real instruction sequences, carefully done (x86 TSO reasoning, aarch64 LDAR/STLR pairing, rv64 `.aqrl` AMOs). mach has **no runtime builtins** and that is worth keeping.

The cost is that an opaque block cannot be inlined. So a `lock xadd` — one instruction — is a **function call plus stack-slot staging**: the bodies stage `{ptr}` through a register because that is how asm operands bind. Every atomic operation in the compiler's own worker pool, every `mutex` acquire, every `thread` handoff pays that.

## Two wins, and the second is the important one

**Cost.** Once the block is `isa.Inst`, the inliner can inline it and regalloc can allocate across it. The instruction becomes the instruction.

**Correctness of the guarantee.** Today the memory-ordering behaviour of these functions is an **accident of asm opacity** — it holds because the optimizer cannot reason about the block, not because anything promises it. The day the optimizer learns anything about asm, it evaporates silently. A modeled effect set makes it a contract.

Note the current design is simultaneously *too weak* (guarantees rest on an accident) and *too strong* (an opaque block blocks all code motion across it, including motion that is perfectly legal around a relaxed atomic).

## Also worth doing here

Everything is **sequential consistency only** — no relaxed/acquire/release. So aarch64 eats a trailing `DMB ISH` on every read-modify-write that an acquire or release would not need. Orderings should be a **comptime parameter** (mach already has them, and a runtime ordering is meaningless — comptime rejects it by construction). mach has no enums, so `Order` is a `def` over an integer with `val` constants, exactly like `PRIM_CLASS_*`.

## Bar

- `std.sync.atomic` keeps its **exact source surface** — no new language surface, no builtins, callers unchanged.
- Measured: instruction count and runtime for an atomic increment before and after, plus a contended-counter benchmark.
- Correctness under real concurrency, not just single-threaded: the compiler's own parallel codegen (`--jobs`) is the natural stress case, and output must stay byte-identical at every core count.
- Litmus tests per ISA (message-passing, store-buffering, IRIW) if orderings are added — that is the real price of making the memory model a specification.
[...]

PR #2367, opened 2026-07-25
fix(asm): one `{name}` effect model across the three targets

### LINKED ISSUE #2258: asm: the three targets' {name} effect models disagree — unifying them makes x86-64 more precise
Surfaced by #2232 (PR #2253), which unified inline-asm parsing onto one spine and deliberately left this divergence stated rather than silently changed.

When an inline-asm `{name}` binding is **unresolved** — pre-frame, so the displacement is genuinely unknown — the three targets model the enclosing statement differently:

- **riscv64 and aarch64** model the statement around the unresolved binding.
- **x86-64** treats the whole statement as unmodeled, so its effects stay worst-case within the GP bank.

## Why it was not unified in #2253

It is **load-bearing, not cosmetic**. x86-64's worst-case treatment is exactly what makes every atomic and syscall block report all-16-GP-clobbered, which is in turn why regalloc's stack-slot staging is what it is. Changing it **moves register allocation** — real emitted-code change, not a refactor — and #2253's acceptance bar was byte-identity across 2472 objects. Riding a behaviour change in on a unification PR would have destroyed the one property that made that PR reviewable.

So it stayed a **stated per-target fact**. This issue is where it gets decided deliberately.

## Why it is worth doing

Unifying **upward** — giving x86-64 the modelled treatment the RISC targets already have — makes it *more precise*, and precision here is what unlocks the rest of the asm-awareness work: an over-broad clobber set is why regalloc stages through stack slots rather than allocating across a block, which is a direct cost on every `std.sync.atomic` call (see #2231).

It is also a correctness-shaped question, not just a performance one. Three models for one concept is the drift shape epic #2212 exists to remove, and the divergence is currently invisible unless someone reads all three grammars.

## Constraints

- **Conservative by default still applies.** More precise must never mean optimistic: an instruction or binding the model cannot account for keeps worst-case effects. **Opaque-and-slow is safe; modeled-and-wrong is a miscompile**, and today's conservatism is why the atomics' ordering guarantees hold at all.
- The derived clobber set must be **proven equivalent-or-narrower** against the current one across the whole corpus before it ships — #2243 and #2253 both used a shadow module running the old derivation beside the new one, armed to panic on mismatch, and that is the bar.
- Expect emitted output to change on x86-64. **Byte-identity is not the bar here**; a measured delta with an explanation is.

## Bar

- One effect model, with any remaining per-target behaviour justified in the code rather than emergent.
- Shadow-module equivalence showing the new set is never *wider* than the old on any block in mach, mach-std or `int/`.
- Measured: register-allocation and code-size delta on x86-64, plus a runtime measurement on an atomics-heavy workload since that is the motivating cost.
- Suites; three-generation fixpoint `A == B == C`; mach-std **611 / 0** at pin `eff1e8e` (a bare clone lands on `dev` and reports 663 — not the pin).

Part of #2212. Related: #2231 (inline atomics), #2230 (`asm` inside `#[oblivious]`).

octalide:
Closes #2258.

## What diverged

The three grammars gave three answers to one question. For an **unresolved** `{name}` — parsed pre-frame, where the displacement genuinely is unknown:

| | operand kind | statement | effect | emit-side guard |
|---|---|---|---|---|
| riscv64 | `AOP_BIND`, accepted where an address is (`rv_is_addr`) | `AI_INST` | modelled | `rv_emit` RVP_LOAD/STORE only |
| aarch64 | `AOP_BIND`, accepted where an address is (`a64_is_addr`) | `AI_INST` | modelled | `a64_emit_ldst` only |
| x86-64 | `AOP_BIND` anywhere | `AI_OPAQUE` | **all 16 GP** | the spine's `AI_OPAQUE` refusal |

The root cause, once written down: x86-64's model was a function of the **parse phase**, not of the instruction stream. Parse the same body *with* a laid-out frame and x86-64 already derived the narrow set — the binding became `[rbp + off]`, and `mark_written` credits a memory destination with nothing. `clobbers`' own doc says the set "is a property of the parsed instruction stream rather than a second reading of it"; on x86-64 that was false.

## The one model

A `{name}` denotes a frame slot. A frame slot is memory. Memory is not a register. So a binding contributes nothing to the write set on any target, and the statement around it folds to the same set resolved or unresolved. That sentence now lives once, in `asm.mach`'s header, and `AI_OPAQUE` — whose sole producer was x86-64's escape — is gone.

The emit-side half was spelled three times as well (the spine's `AI_OPAQUE` refusal plus two per-target `mem.kind != AOP_MEM` checks, all three with the same message). It is now **one scan over every operand of every statement** in the shared driver.

## Finding: riscv64's guard had a hole

`rv_amo_addr` → `rv_is_addr` accepts a binding as an AMO address, and `rv_emit`'s `RVP_AMO` / `RVP_AMO_LR` arms re-checked nothing — they read `item.ops[N].reg::u32` directly. `AOP_BIND` carries `reg = -1`, so `amoadd.d.aqrl a0, a2, {slot}` reached `emit_r` with a negative register index and encoded a **different instruction**. aarch64 had no such hole (`a64_bare_addr` refuses a binding at parse); x86-64 had none (`AI_OPAQUE` caught it). The shared scan closes it by construction, and `riscv64.encode:inline_asm_atomics` now pins it.

Not reachable from `encode_block` (which always supplies bindings), but reachable from `rv_asm_text` — the encoder-test entry point, which passes `pl = nil` on all three targets.

## Conservative by default: unchanged

An unrecognized mnemonic still refuses the block. A `.byte` payload still writes every register in every bank. A body the parser cannot read still folds to `all_gp | all_fp`. What went away is one *category* judgement ("a statement I can't fully resolve is opaque") replaced by a *per-operand fact* ("this operand is memory").

## Shadow-module equivalence — 0 wider

Two instrumented compilers (the old derivation and the new), dumping `(isa, body, gp, fp)` for every `asm.clobbers` call, over **mach + mach-std + `int/`**, all six targets, both profiles:

```
distinct (isa, body) derivations: 109   total calls: 15570
  identical: 77   narrower: 32   WIDER: 0
unchanged by isa:  aarch64 40/40   riscv64 27/27   x86_64 10
```

Every one of the 32 narrowings is x86-64 and every one carries a `{name}`. Nothing anywhere got wider.

## Measured

**Objects.** riscv64 and aarch64 **byte-identical** across the whole corpus, *both* profiles (0/428 objects differ per target/profile). x86-64 moved in exactly 5 artifacts — the four modules holding an x86-64 `asm` block with a `{name}`, plus the linked binary:

| object (release) | `.text` | |
|---|---|---|
| `std/sync/atomic.o` | 1883 → 1578 | −305 |
| `std/system/os/linux/shared.o` | 15263 → 14753 | −510 |
| `std/system/os/linux/x86_64.o` | 570 → 453 | −117 |
| `std/system/panic.o` | 151 → 111 | −40 |
| **total** | | **−972** |
| `bin/mach` exec segment | 0xa3c5ae → 0xa3c1ee | −960 |

Per function, only the functions that *contain* inline asm moved. Every other function in those objects is byte-identical; nothing else in 1284 objects moved at all.

**The case where the old model was needlessly conservative** — `std.sync.atomic.load`, x86-64 release, 97 → 53 bytes:

```asm
;; before                              ;; after
push  %rbp                             push  %rbp
mov   %rsp,%rbp                        mov   %rsp,%rbp
sub   $0x60,%rsp                       sub   $0x30,%rsp
mov   %rbx,-0x38(%rbp)                 mov   %rbx,-0x28(%rbp)
mov   %r12,-0x40(%rbp)
mov   %r13,-0x48(%rbp)
mov   %r14,-0x50(%rbp)
mov   %r15,-0x58(%rbp)
lea   -0x10(%rbp),%rax                 lea   -0x10(%rbp),%rax
mov   %rdi,(%rax)                      mov   %rdi,(%rax)
lea   -0x20(%rbp),%r8                  lea   -0x20(%rbp),%rbx
mov   %r8,-0x30(%rbp)      ;; stage
mov   -0x30(%rbp),%r8      ;; reload
movq  $0x0,(%r8)                       movq  $0x0,(%rbx)
mov   -0x10(%rbp),%rcx                 mov   -0x10(%rbp),%rcx    ;; --- asm block
mov   (%rcx),%rax                      mov   (%rcx),%rax
mov   %rax,-0x20(%rbp)                 mov   %rax,-0x20(%rbp)    ;; --- asm block
mov   -0x30(%rbp),%r8      ;; reload
mov   (%r8),%rax                       mov   (%rbx),%rax
mov   -0x38(%rbp),%rbx                 mov   -0x28(%rbp),%rbx
mov   -0x40(%rbp),%r12
mov   -0x48(%rbp),%r13
mov   -0x50(%rbp),%r14
mov   -0x58(%rbp),%r15
mov   %rbp,%rsp                        mov   %rbp,%rsp
pop   %rbp                             pop   %rbp
ret                                    ret
```

The body writes RAX and RCX. The old set was all 16, so the prologue preserved **all five** callee-saved GP registers and the address had to be **staged through a stack slot** across the barrier — the exact cost #2231 names. The new set is `{RAX, RCX}`: one callee-saved register saved, and it is saved because the allocator now *uses* it to hold the address across the block.

**Runtime**, atomics-heavy workload (3M iterations × `store` / `load` / `fetch_add` / `fetch_sub` / `exchange` / `cas`, six values live across each barrier), x86-64 release, 15 interleaved runs:

```
base:  min 69.2 ms   median 71.1 ms
new:   min 49.9 ms   median 50.9 ms
       -27.9% (min)   -28.5% (median)
```

Both binaries print identical results.

## Mutation tests

Baseline **949 / 0**.

| mutation | result | broke |
|---|---|---|
| **M1** drop the shared emit-side bind scan | **947 / 2** | `riscv64.encode:inline_asm_atomics`, `x64.encode:asm_unresolved_binding_refused_at_emit` |
| **M2** drop the implicit-write fold (the optimistic direction — a real miscompile) | **946 / 3** | all three ISAs' `asm_clobbers` tests |
| **M3** restore the old worst-case bind rule, in the shared fold | **946 / 3** | all three ISAs' `asm_clobbers` tests |

M1 shows one shared check now covers two ISAs that each had their own. M2 matters *because* of this change: under the old model a missing `implicit` fact was masked on every bind-bearing x86-64 body by the all-GP fold — it no longer is. M3 shows the new x86-64 assertions discriminate rather than pass vacuously.

## Suites

- mach **949 / 0** through the from-source HEAD compiler (948 on `dev` + the one test added).
- mach-std **611 / 0** at pin `eff1e8e` (materialized by `git archive` + `diff -r` against the submodule; a bare clone lands on `dev` and reports 663).
- `int/`: **linux** all pass, **linux-riscv64** all pass (qemu). The **linux-arm64** leg is `native` in `targets.conf` and this x86_64 host cannot execute it (`Exec format error`) — CI runs it on `ubuntu-24.04-arm`. Covered here instead by byte-identical aarch64 objects across the corpus and by the shadow's 40/40 unchanged aarch64 derivations; `int/surface/freestanding`'s aarch64 and riscv64 binaries are byte-identical, and only its x86-64 binary moved.
- Three-generation fixpoint **B == C** (`e80aadf5…`). A ≠ B is expected: the seed predates this codegen change.

## `std.sync.atomic` specifically

Its memory-ordering behaviour is untouched — no instruction, ordering suffix or fence moved; the emitted asm bodies are the same instructions, and only the *surrounding* register allocation changed. Its objects do move, which is the point of the issue, and the delta is the table above.

https://claude.ai/code/session_01Ntqq8dS4TDqoYPsx3JpVy4

### DIFF
--- a/src/lang/target/isa/arm64/encode.mach
+++ b/src/lang/target/isa/arm64/encode.mach
@@ -4401,8 +4401,6 @@ fun a64_emit_branch(st: *encode.EncodeState, c: *iasm.Cursor, l: *iasm.Labels, i
 fun a64_emit_ldst(st: *encode.EncodeState, c: *iasm.Cursor, item: *iasm.Item) R.Result[bool, str] {
     val rt: *iasm.Operand = ?item.ops[0];
     val mem: *iasm.Operand = ?item.ops[1];
-    if (mem.kind != iasm.AOP_MEM) { ret R.err[bool, str]("internal compiler error: inline-asm statement reached the encoder unresolved"); }
-
     val is_load: bool = item.code == A64M_LDR || item.code == A64M_LDRB || item.code == A64M_LDRH;
     var w: u8 = 4;
     if (a64_is64(rt))                                          { w = 8; }
--- a/src/lang/target/isa/asm.mach
+++ b/src/lang/target/isa/asm.mach
@@ -47,6 +47,16 @@
 # Precision is earned per mnemonic, on that mnemonic's own table row, and never
 # credited by category - `opaque-and-slow` is safe, `modeled-and-wrong` is a
 # miscompile.
+#
+# A `{name}` binding is the one operand every grammar reads the same way, so it is
+# stated here rather than three times (#2258): it denotes a frame slot, a frame slot
+# is memory, and memory is not a register - so it contributes nothing to the write
+# set and the statement around it is modeled whether or not the frame is laid out
+# yet. That is what makes the effect model a property of the parsed instruction
+# stream rather than of the parse phase: the same body folds to the same set before
+# register allocation and at emit time. The emitter never sees an unresolved binding,
+# because it always parses with a binding context, and the driver CHECKS that rather
+# than assuming it.
 
 use std.text.parse;
 use std.types.bool.bool;
@@ -189,14 +199,12 @@ pub rec Mnemonic {
 }
 
 # the parse outcomes one statement can have
+#
+# There is no "parsed but unmodeled" outcome: a statement the grammar accepts is one
+# the effect model describes, and one it does not accept refuses the block.
 pub val AI_INST:   u8 = 0;  # a modeled machine instruction
 pub val AI_LABEL:  u8 = 1;  # a numeric local-label definition
 pub val AI_BYTES:  u8 = 2;  # a raw byte directive's payload
-# a statement whose mnemonic is known but whose operands are not resolvable - a
-# `{name}` binding parsed before the frame is laid out, on a target whose effect
-# model does not attempt to model one. the instruction is not modeled, so its
-# effects fall back to everything the target's inline grammar can reach
-pub val AI_OPAQUE: u8 = 3;
 
 # what a parsed instruction still needs the encoder to resolve
 pub val AR_NONE:  u8 = 0;  # nothing: the instruction is self-contained
@@ -865,9 +873,10 @@ fun unknown_message(c: *Cursor, lo: usize, hi: usize) str {
 # target's - `[rbp + off]`, `off(s0)`, `[x29, #off]` - so the ISA's lexer calls this
 # and builds its own memory operand.
 #
-# Returns none when the parse carries no binding context: before the frame is laid
-# out a binding's displacement genuinely is unknown, and the ISA decides whether its
-# effect model can still describe the statement.
+# Returns ok(false) when the parse carries no binding context: before the frame is
+# laid out a binding's displacement genuinely is unknown. The operand stays a binding
+# and the statement is still modeled - what the displacement turns out to be cannot
+# change which registers the statement writes, since a slot is memory either way.
 # ---
 # c:   the cursor
 # lo:  body offset of the `{`
@@ -912,10 +921,13 @@ fun bind_name_eq(interner: *intern.Interner, name: intern.StrId, body: str, star
 
 # the registers one parsed item writes, folded into the running masks
 #
-# The whole effect model. A modeled instruction writes exactly the registers its
-# mnemonic's facts name; an item the parser could not model writes worst-case. An
-# unreadable byte stream reaches every bank, while an unresolved statement is bounded
-# by the general-purpose bank - no target's inline grammar names a vector register.
+# The whole effect model, and the same one on every target. An instruction writes
+# exactly the registers its mnemonic's facts name among its own operands, plus the
+# ones its form writes regardless. A raw byte payload is not an instruction the
+# parser read at all, so it reaches every register in every bank.
+#
+# A `{name}` binding names a frame slot, never a register, so a statement carrying
+# one folds to the same set resolved or unresolved (#2258).
 # ---
 # g:    the target's grammar
 # item: the parsed statement
@@ -926,10 +938,6 @@ pub fun fold_clobbers(g: *Grammar, item: *Item, gp: *u32, fp: *u32) {
     if (item.kind == AI_BYTES) {
         @gp = @gp | g.all_gp;
         @fp = @fp | g.all_fp;
-        ret;
-    }
-    if (item.kind == AI_OPAQUE) {
-        @gp = @gp | g.all_gp;
         ret;
     }
     @gp = @gp | item.implicit;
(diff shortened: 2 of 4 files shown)


octalide:
Rebased onto `dev` at `4721a8f9` (15 commits). The four modules this PR touches — `isa/asm.mach`, the three ISA `encode.mach` — were **untouched** by those commits, so the shadow-equivalence and mutation results in the description stand verbatim. Re-verified everything that could move:

- **Objects, rebased base:** same 5 x86-64 artifacts move (`std/sync/atomic.o`, `std/system/os/linux/shared.o`, `std/system/os/linux/x86_64.o`, `std/system/panic.o`, `bin/mach`). **linux-arm64 and linux-riscv64 byte-identical: 0 of 214 objects differ, both profiles.**
- **Suites:** mach **951 / 0** (baseline `4721a8f9` is 950 — the one test this PR adds); mach-std **611 / 0** at pin `eff1e8e`. Both through the from-source HEAD compiler.
- **`int/`:** linux all pass, linux-riscv64 all pass (qemu).
- **Fixpoint:** `B == C` at `b62b010c…`.

Issue #2365, opened 2026-07-25
isa: RegMachine.fp_scratch_reg is declared and never read, while two encoders cite it and hardcode their own


octalide:
Found by the sweep in #2341 (PR #2354). **This one has a correctness edge — it is not just dead surface.**

`RegMachine.fp_scratch_reg` is the FP twin of the live `scratch_reg` / `scratch_reg2`. It is written from a **required positional parameter** of `isa.reg_machine` and **read nowhere**.

## Why it is worse than the fields removed in #2354

The arm64 and riscv64 encoders both **cite it in comments** — *"the vtable's `fp_scratch_reg`"* — and then **hardcode their own register**.

So there are two facts that must agree, one declared and one hardcoded, with **nothing checking that they do**. Today they happen to match. If a target's declared value is ever changed, the encoder keeps using the hardcoded one and the comment continues to assert the declared one is authoritative. The register the compiler *says* it reserved and the register it *actually* reserved would diverge silently, and a clobbered FP scratch is a wrong-answer bug rather than a crash.

**#2244's arity check does not cover this.** That mechanism guards against *forgetting* a slot — an arch that omits the seam is a build-time error. It does not guard against **lying** about one: `fp_scratch_reg` is dutifully supplied by every target and simply not consulted.

## Decide the direction with evidence

Two coherent ends, and the sweep does not settle which:

- **Make it live** — have the encoders read the declared value instead of hardcoding. Correct if the field is the intended contract; the comments already assert it is. Objects should be byte-identical if the declared and hardcoded values currently agree — and if they do **not**, that divergence is the finding and it is a live bug.
- **Delete it** — if per-encoder choice is genuinely the right model, remove the field, remove the parameter, and **fix the comments that claim otherwise**. A comment pointing at a field nobody reads is how this survives another year.

Establish which by checking whether the declared and hardcoded values agree on every target *today*. That answer decides the direction and is cheap to get.

## Bar

- The per-target comparison of declared versus hardcoded, reported. This is the deliverable even if nothing changes.
- **Objects byte-identical on one target where identity is genuinely a claim** — unless the values diverge, in which case say exactly where.
- Whichever direction: no comment left asserting a relationship that does not hold.
- Mutation-test whatever enforcement lands; suites; mach-std 611 / 0 at pin `eff1e8e`.

Issue #2274, opened 2026-07-25
fe: the `-0.0` float literal evaluates to +0.0 (sign of zero dropped)


octalide:
Found while implementing #2160, which needed a negative zero as an exact IEEE identity element. Out of that issue's scope to fix (`fe/sema` is #2225's ground), so filed rather than patched; #2160 works around it by spelling the constant as a bit pattern.

## What

Mach cannot express negative zero as a literal. Both the source-level literal and a
runtime negation of `+0.0` produce **positive** zero:

```mach
val NEG_ZERO_BITS: u64 = 0x8000000000000000;

pub fun probe() {
    val a: f64 = -0.0;                    # expected 0x8000000000000000
    var x: f64 = 0.0;
    val c: f64 = -x;                      # expected 0x8000000000000000
    val d: f64 = NEG_ZERO_BITS:~f64;      # the workaround
    p.printlnf("lit={} neg={} bitpat={}", a:~i64, c:~i64, d:~i64);
}
```

Observed on 4.2.2 (release, x86_64-linux):

```
lit(-0.0)=0 runtime_neg(0.0)=0 bitpat=-9223372036854775808
```

Only the bit-pattern reinterpret produces the correct value. IEEE-754 unary negation is
a sign-bit flip and is exact for every operand, zeros included, so `-0.0` must be
`0x8000000000000000` and `-x` for `x == +0.0` must be negative zero.

## Why it matters

`-0.0` is not a curiosity — it is the exact additive identity under round-to-nearest
(`x + (-0.0) == x` for every `x`, where `x + (+0.0)` turns `-0.0` into `+0.0`). Any code
that seeds an accumulator, writes an identity element, or round-trips IEEE data through
Mach source silently loses the sign. The middle end already treats the two zeros as
distinct — `me/pass/cse.mach` value-numbers float constants by bit pattern precisely so
`x + 0.0` and `x + (-0.0)` do not merge, and `me/pass/algebraic.mach` declines to fold
`x + 0.0` for the same reason — so the defect is upstream of them, in the front end.

## Where

Not diagnosed beyond the observation. Two candidates, and the `-x` result suggests both
may be in play:

- the comptime/constant-folding path for a negated float literal (`fe/comptime.mach` /
  `me/pass/constfold.mach`), which likely computes `0 - 0.0` and gets `+0.0`
- the scalar float-negate lowering, if the negation is not folded

A fix should assert the bit pattern, not numeric equality: `-0.0 == 0.0` is *true* in
[...]

Issue #2239, opened 2026-07-25
layout: an anonymous uni inside a generic record loses union layout — variants do not overlap (mach-std Result affected)


octalide:
Found during #2224. **Pre-existing on `dev` and on released 4.2.2.** This is a layout/lowering bug, not sema.

An anonymous `uni` inside a **generic** record does not get union layout in its instances: its variants occupy **distinct storage** instead of overlapping.

## Measured

Offset between the two variants, `(?r.value.err) - (?r.value.ok)`, on baseline `dev`:

| shape | offset | correct? |
|---|---|---|
| stdlib `Result[u32, u32]` | **4 bytes** | **no** |
| local `rec Local[T, E] { v: uni { a: T; b: E; } }` | **4 bytes** | **no** |
| non-generic twin | 0 bytes | yes |
| named generic `uni Named[u32, u32]` | 0 bytes | yes |

Store/load agrees with the offsets: write `err = 2`, then `ok = 1`, read `err` back → **2**. No aliasing. The variants are simply two separate fields.

So the defect is specific to an **anonymous** union nested inside a **generic** aggregate — the named generic form and the non-generic form are both correct.

## Two consequences

**1. Union semantics are violated.** Writing one variant and reading another is the entire point of a `uni`. Any code relying on that through a generic gets a value that was never written there.

**2. `Result[T, E]` is 4 bytes larger than it should be, on every instantiation.** That is the most-used generic in the standard library, so this is a size regression across essentially every mach program, plus the cache cost of carrying it.

## Coupling with #2225 — read this before fixing either

#2224 established that the type system genuinely accepts a mixed-secrecy pun through `Result[^u32, u32]` — the rule that should reject it does not fire for generic union instances (that hole is **#2225**).

This layout bug is currently **masking** that hole: the secret and public variants do not overlap, so no byte actually leaks today.

**Fixing this layout bug without fixing #2225 converts a masked hole into a live leak.** Whoever takes this must confirm #2225 has landed, or land them together. Say explicitly in the PR which it was.

## Bar

- Root-cause it in instance layout — likely `build_instance_table` / `substitute_aggregate` treating the anonymous nominal as a record when materialising a generic instance. Establish the mechanism before changing anything.
- After the fix, all four shapes in the table above report offset 0, and the store/load probe shows genuine aliasing.
- **Report the size delta** for `Result` and for any stdlib type that shrinks — objects will change, so byte-identity is not the bar here; an explained delta is.
[...]

PR #2388, opened 2026-07-25
refactor(isa): remove 14 dead descriptor fields, one dead record, and the comments that misdirect

### LINKED ISSUE #2366: isa: 14 dead descriptor fields and one dead record, three of them behind comments that misdirect
The remainder of the sweep from #2341 (PR #2354), which removed `AbiVTable.fp_arg_regs` and wired `dwarf.register_debug_hooks` at the join. **14 dead fields and 1 dead record**, in three groups that want different treatment. `RegMachine.fp_scratch_reg` is filed separately as #2365 because it carries a correctness edge; this issue is the rest.

## Group A — a descriptor superseded in place

`IsaVTable.reg_classes`, `IsaVTable.reg_class_count`, `RegClass.bit_width` — superseded by `MachineModel.reg_classes`.

Three comments still direct readers to **`target.arch.reg_classes`, a path that does not exist**: `Target.arch` is a `*RegMachine`. So the stale surface is actively misdirecting, not merely inert.

## Group B — eight fields behind a plan that already closed

`MachineModel.index_width`, `scratch_regs`, `scratch_len`, `addr_forms`, `addr_scales`, `addr_disp_bits`, `stack_grows_down`, `fp_relative_frame` — populated by **all five ISAs**, read by **none**.

They sit behind a docstring deferring to **#1600 group B**. **#1600 is closed** (v2.6.0), and its close-out defers or declines that remainder.

This is the worst version of the #2277 shape: **dead, with a comment that reads like a plan.** A reader finds a deferral, assumes work is pending, and leaves it — indefinitely. #2277's `ProjectConfig.vectorize` was added *five days after* the live path moved, by mirroring an already-dead field; a field with an open-sounding deferral attached is even more likely to be mirrored.

**This group wants a re-decision, not a mechanical delete.** Either the fields describe something a future addressing-mode or frame-layout consumer genuinely needs — in which case say so against a *live* reference — or they do not and they go. Either way **the closed-epic reference must go**, because it is what makes the current state look deliberate.

## Group C — an asymmetric member and an unused record

- `resolved.Target.of_id` — one dead member of a five-member symmetric block whose **four siblings are all read**. Asymmetry inside an otherwise-uniform block is exactly what invites a reader to populate it by pattern.
- `isa.RegisterFile` — declared and used nowhere, while **four ABI docstrings assert the accessors "compose with" it**.

## Suggested order

Group B first — it is the largest and the only one carrying a false plan. Then A, then C. Each is independently landable.

## Bar

- **Objects byte-identical, one target where identity is genuinely a claim.** A dead field costs nothing at runtime, so any delta means it was not dead and you must say where. Note a `test` block inside a shipped module *can* move a release object (#2320), though an orphan test module cannot.
- **No comment left pointing at a deleted or nonexistent path.** Group A's three misdirecting comments and Group C's four docstrings are as much the defect as the fields.
- For Group B specifically: the re-decision stated, with the closed **#1600** reference removed regardless of which way it goes.
- All five ISAs still register and classify correctly, demonstrated by building and running on the three linkable ones.
- Suites; mach-std 611 / 0 at pin `eff1e8e` via `git archive`; fixpoint only if codegen moves.

octalide:
Closes #2366 — the remainder of the #2341 sweep. `RegMachine.fp_scratch_reg` is **not** touched; that was #2365 / PR #2382, which is now merged and fully preserved here.

## Group B is a re-decision, and the evidence changed what I would have done

Eight `MachineModel` fields, populated by **all five ISAs** and read by **none**, behind a docstring deferring to **#1600 group B — closed in v2.6.0**.

The tempting defence is "these describe five real machines, keep them for a future addressing-mode consumer." I checked whether they actually do, and two of them do not:

- **`stack_grows_down` is `true` on all five**, including a stackless SPIR-V where the value is meaningless. It distinguishes nothing.
- **`addr_disp_bits` means two different things.** x86-64 (32), rv64 (12) and 6502 (8) record the ISA's immediate field width — a machine fact. aarch64 records **32**, and its own comment gives it away: *"the shared displacement is a signed 32-bit (the MIR operand's `disp` field)"*. aarch64 hardware encodes a **scaled imm12** or an **unscaled imm9** and has no 32-bit displacement form at all. A model-driven consumer folding against this would have formed addresses aarch64 cannot encode.
- **`scratch_regs` is superseded**: each encoder names its own scratch registers directly (aarch64 IP0/IP1 in `arm64/encode.mach`) rather than consulting the model.

So the decision is **delete**, and the reason is stronger than "unused": **unread machine description is unverified machine description.** Nothing reads these, so nothing can catch a wrong value — and one had already drifted into meaning something else entirely. A future consumer would have had to re-verify every value anyway.

Removing `addr_forms` orphaned the `AddrForm` flag set it was the only reader of, so that goes too — otherwise this change would have created exactly the dead surface it exists to remove.

The docstring now states the rule rather than a plan: **every field here is read**, and such a description is re-added *with* its consumer, in the same change.

## Group A — a descriptor superseded in place

`IsaVTable.reg_classes` / `reg_class_count` and `RegClass.bit_width`, all superseded by `MachineModel.reg_classes`. The vtable pair was fed by a parameter chain through `machine_isa` / `emitter_isa` / `isa_common` that existed **only** to populate them, so the chain goes with them — 17 call sites updated.

The three comments pointing at **`target.arch.reg_classes`, a path that does not exist** (`Target.arch` is a `*RegMachine`) now name `tgt.model.reg_classes`. A fourth comment in `riscv64/register.mach` described "the two register-class bit_widths above" and the removed addressing/frame facts; it now describes what remains.

## Group C

- `resolved.Target.of_id` — the one dead member of a five-member block whose four siblings are read.
- `isa.RegisterFile` — declared and used nowhere, while four ABI docstrings asserted the accessors *"compose with"* it. Those now say the caller holds the `(regs, count)` pair, which is what actually happens.

## Verification

- **Objects byte-identical: 146/146.** Built the pinned std (`eff1e8e`, via `git archive`) for `linux-x86_64` and `linux-arm64` with a baseline compiler and with this one; every object matches. A dead field costs nothing at runtime, and nothing moved.
- **mach-std at pin: 611 passed / 0 failed.**
- **Compiler suite: 952 / 5 / 957 — identical to pristine `dev`**, same five pre-existing failures (`printer:signed_zero`, four `generic_layout_runtime` aliasing cases). This change removes **zero** test blocks: 955 at my base, 955 after; `dev`'s 957 is +2 from #2383.
- **All five ISAs still register**: `mach info` output identical to baseline.
- **Built and run on the three linkable ISAs**: `int` corpus fully green on `linux` and `linux-riscv64` (qemu). `linux-arm64` is `native` run-mode in `targets.conf` — an ARM runner in CI — so it cannot execute on this x86_64 host; I built and ran a case under explicit `qemu-aarch64` instead and diffed it against the golden.

https://claude.ai/code/session_01Ntqq8dS4TDqoYPsx3JpVy4

### DIFF
--- a/src/lang/be/codegen/ctvalidate.mach
+++ b/src/lang/be/codegen/ctvalidate.mach
@@ -585,8 +585,8 @@ fun t_target(trust_mul: bool, trust_var_shift: bool) target.Target {
 # emitter: the whole-module payload, or nil for a register machine
 # ret:     the vtable
 fun t_isa(name: str, emitter: *isa.ModuleEmitter) isa.IsaVTable {
-    if (emitter != nil) { ret isa.emitter_isa(isa.ARCH_UNKNOWN, name, 8, nil, 0, nil, emitter); }
-    ret isa.machine_isa(isa.ARCH_UNKNOWN, name, 0, 8, nil, 0, nil, nil, nil);
+    if (emitter != nil) { ret isa.emitter_isa(isa.ARCH_UNKNOWN, name, 8, nil, emitter); }
+    ret isa.machine_isa(isa.ARCH_UNKNOWN, name, 0, 8, nil, nil, nil);
 }
 
 # a secret laundered through a phi join (out-of-SSA predecessor moves into one result vreg)
--- a/src/lang/be/codegen/mir.mach
+++ b/src/lang/be/codegen/mir.mach
@@ -748,7 +748,7 @@ pub rec MirBlock {
 # a virtual register and its register-allocation state
 # ---
 # id:         the vreg's own index into MirFunction.vregs
-# class:      index into target.arch.reg_classes selecting the register bank
+# class:      index into `tgt.model.reg_classes` selecting the register bank
 # spill_slot: stack-frame slot index when spilled (-1 while unspilled)
 # assigned:   physical register id assigned by regalloc (MIR_VREG_NIL until then)
 # vector:     true when the vreg holds a 128-bit SIMD vector; distinguishes a vector
--- a/src/lang/be/codegen/mir/abi.mach
+++ b/src/lang/be/codegen/mir/abi.mach
@@ -1865,7 +1865,7 @@ fun probe_ret(size: u64, align: u64, is_float: bool, fp_eightbytes: u8,
 test "mach.lang.be.codegen.mir.abi.require_abi:variadic_facts_are_not_call_preconditions" {
     # a whole-module ISA: this is a convention with no register machine behind it,
     # which is exactly the shape the two variadic facts are absent from.
-    var vt: isa.IsaVTable = isa.emitter_isa(isa.ARCH_SPIRV, "probe", 8, nil, 0, nil, nil);
+    var vt: isa.IsaVTable = isa.emitter_isa(isa.ARCH_SPIRV, "probe", 8, nil, nil);
     var av: abi.AbiVTable;
     av.arg_passing = probe_arg::*u8;
     av.ret_passing = probe_ret::*u8;
--- a/src/lang/be/codegen/mir/context.mach
+++ b/src/lang/be/codegen/mir/context.mach
@@ -271,7 +271,7 @@ pub fun drain_pending_dbg(ctx: *LowerCtx, mi: *mir.MirInstr) R.Result[bool, str]
 # Grows the function's vreg table on demand.
 # ---
 # ctx:       the lowering context
-# reg_class: register-class index into target.arch.reg_classes
+# reg_class: register-class index into `tgt.model.reg_classes`
 # ret:       the new vreg id, or an allocation error
 pub fun new_vreg(ctx: *LowerCtx, reg_class: u32) R.Result[u32, str] {
     val f: *mir.MirFunction = ctx.f;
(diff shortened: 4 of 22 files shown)


octalide:
Correcting one line in the PR body above, because it implies `dev` is red and it is not.

I reported *"Compiler suite: 952 / 5 / 957 — identical to pristine `dev`, same five pre-existing failures."* The comparison was sound — I ran my branch and pristine `dev` under the same apparatus and got identical counts, which is what established that this change adds no failures and removes no tests. **But the failures are not pre-existing on `dev`. They are an artifact of my apparatus**, and `dev` is clean.

I invoked the suite as `mach test .` with the **PATH seed, `mach 4.2.2`**, rather than a from-source compiler. Same tree (`99217fe0`), two compilers:

| compiler | result |
|---|---|
| PATH seed `mach 4.2.2` | 952 / 5 / 957 |
| from-source, built from that same tree | **957 / 0 / 957** |

The five are **regression tests for compiler bugs that `dev` has fixed and the 4.2.2 seed still has**, so the seed miscompiles the test bodies and they fail. Each names its own bug:

- `me.ir.printer.print_value:signed_zero` — #2274, `-0.0` folding to `+0.0`
- `me.lower.generic_layout_runtime.{anon_in_generic,named_generic,named_in_generic}:aliases` and `two_instantiations:each_aliases` — #2239, an anonymous `uni` in a generic record not getting union layout

They are doing exactly what they were written to do. My branch under a from-source compiler is **957 / 0 / 957**.

Nothing about the merged change is affected — the byte-identical 146/146, the mach-std 611/0 at pin, and the cross-ISA runs all used from-source compilers. Only this one line was wrong.

PR #2397, opened 2026-07-26
feat(target): add bmos — ReturnInfinity's BareMetal exokernel — as an x86-64 target OS

### LINKED ISSUE #2396: target: add bmos — ReturnInfinity's BareMetal exokernel — as an x86-64 target OS
BareMetal is ReturnInfinity's x86-64 exokernel. It is a real OS with its own load and entry contract, not the generic no-OS case `freestanding` already covers, so it wants its own OS value — named `bmos` to keep the two from being confused.

## The contract, from the kernel source

- **Load address** — fixed `0xFFFF800000000000` (`app_start` in `src/sysvar.asm`, `ProgramLocation` in BareMetal-Monitor, and the origin in BareMetal-Demo's `c.ld`). The loader performs no relocation.
- **Image format** — a container-free flat binary (`OUTPUT_FORMAT(binary)`; the asm demos use nasm's default flat output). The monitor reads the file's bytes straight to the load address.
- **Entry** — the first byte of the image: `call [ProgramLocation]`. A program therefore enters at its base.
- **Exit** — by `ret`, back to the monitor's command loop. There is no exit syscall.

The kernel API is not syscalls: a table of absolute function pointers at `0x100010..0x100040`, called indirectly under a custom register convention. That is a library-side fact and no part of it belongs in the compiler.

## Compiler-side scope

- `OS_BMOS` id, `"bmos"` name, and the `$mach.os.bmos` comptime tag.
- A `bmos` `OsVTable`: `raw` flat image, `_start`, base `0xFFFF800000000000`, no PIE, no dynamic loader.
- x86-64 only, refused loudly on any other ISA. This needs a joint (os, isa) capability fact — the object format cannot carry it, because `raw` is isa-agnostic and would otherwise advertise `bmos` on every architecture with an encoder.

The stdlib arm (a bmos `std.system` / API bindings in mach-std) is deliberately out of scope here.

octalide:
Closes #2396.

Adds `bmos` as a target operating system: [BareMetal](https://github.com/ReturnInfinity/BareMetal), Return Infinity's x86-64 exokernel. It is a real OS with its own load and entry contract, not the generic no-OS case `freestanding` covers, so it gets its own id, name, and vtable.

## The contract, and where it came from

Every parameter below was read out of the kernel and its demo programs, not assumed:

| Fact | Value | Source |
|---|---|---|
| Load address | `0xFFFF800000000000`, fixed | `app_start` in `BareMetal/src/sysvar.asm`; `ProgramLocation` in `BareMetal-Monitor/src/monitor.asm`; `. = 0xFFFF800000000000` in `BareMetal-Demo/src/c.ld` |
| Image format | container-free flat binary | `OUTPUT_FORMAT(binary)` in `c.ld`; the asm demos use nasm's default flat output; the monitor `b_nvs_read`s the file straight to the load address |
| Entry | first byte of the image | `call [ProgramLocation]` in the monitor's `exec` and its `init.app` autorun path |
| Exit | `ret`, back to the monitor | `hello.asm` ends in `ret`; `crt0.c`'s `_start` restores the stack and `ret`s |

The loader relocates nothing, so an image is never position-independent, and a flat image has no entry record — which is exactly the `raw` writer's existing enter-at-base contract, so the writer needed no change.

## What's here

- **`OS_BMOS`, `"bmos"`, `$mach.os.bmos`** — id, name, comptime tag.
- **`src/lang/target/os/bmos.mach`** — the vtable: `raw`, `_start`, base `0xFFFF800000000000`, page 1, no PIE, no dynamic loader, no platform library, `sysv64` on x86-64.
- **A joint (os, isa) capability fact.** `bmos` must be x86-64 only, and the object format cannot carry that: `raw` is isa-agnostic by construction, so without it `bmos` would be advertised on every architecture with an encoder. `OsVTable` gains `supported_isas` / `support_isa` / `supports_isa`, mirroring the object format's `cover_isa` / `covers_isa`, and `tuple_capability` gates on it with `TUPLE_OS_NO_ISA`. linux, darwin and windows declare the architectures they run on; freestanding declares no restriction, since there is no OS there to be ported.

The kernel API is a table of absolute function pointers at `0x100010..0x100040` under a custom register convention. That is a library-side fact; nothing in the compiler encodes it. The stdlib arm is a separate change.

## Verification

- **A real artifact.** A Mach hello-world built for `x86_64-bmos` is a 110-byte flat image: entry at offset 0, `call QWORD PTR ds:0x100018` encoded as `ff 14 25 18 00 10 00` — byte-identical to what nasm emits for `call [b_output]` in the shipped `hello.app` — the string data RIP-addressed in the same image, and `ret` at the end. It **runs**: under a Linux harness that maps the API table at `0x100000` and calls the image's first byte, it prints `Hello, world!` and returns; a second fixture returns 45 in RAX.
- **Byte-identity.** Every existing target's output is unchanged (see the comment below for the per-target hashes).
- **Suites** pass, including a mutation-tested guard on the new (os, isa) gate.
- **Not verified: booting.** The image was not run on real BareMetal or under its qemu setup — the reference clones are read-only here. What is proven is the image's shape, its bytes, and its behaviour against a faithful stub of the API table.


octalide:
## Verification

**Byte-identity — 1290/1290 outputs match.** Both compilers (base = `cfa71dc8`, new = this branch) built the *same* source tree — a detached worktree at `cfa71dc8` — for all six declared targets (`linux-x86_64`, `linux-arm64`, `linux-riscv64`, `windows-x86_64`, `darwin-x86_64`, `darwin-aarch64`). Every object and linked image is bit-for-bit identical.

**Suites: 964 passed, 0 failed** (dev is 961; the three added are `os.supports_isa`, `os.os_id_for`, and `target.select:x86_64_bmos_flat_image`).

**Three-generation fixpoint: B == C.**

```
mach-new (A, seed-built)  88b14693...
gen-b    (A builds src)   12dd130e...
gen-c    (B builds src)   12dd130e...   <- B == C
```

**int, linux leg: all cases pass**, including the new `surface/bmos` case (debug + release).

### Mutation tests on the new guard

Four mutations, each reverted after:

| Mutation | Result |
|---|---|
| drop `supports_isa` from `tuple_capability` | `select:x86_64_bmos_flat_image` FAILS |
| bmos declares `supported_isa_any = true` | `select:x86_64_bmos_flat_image` FAILS |
| `support_isa` never advances the count | 177 tests FAIL |
| the unrestricted-os branch answers `false` | 14 tests FAIL (every freestanding tuple) |

### The artifact

A Mach hello-world for `x86_64-bmos`, 110 bytes, disassembled at its load address:

```
ffff800000000000:  55                    push   rbp
ffff800000000020:  48 8d 0d 39 00 00 00  lea    rcx,[rip+0x39]  # 0xffff800000000060 -> "Hello, world!"
ffff80000000003d:  ff 14 25 18 00 10 00  call   QWORD PTR ds:0x100018
ffff80000000005c:  c3                    ret
ffff800000000060:  "Hello, world!"
```

Those seven call bytes are byte-identical to what nasm emits for `call [b_output]` in BareMetal-OS's shipped `hello.app`. No header, entry at offset 0, exit by `ret`.

**It runs.** A Linux harness that maps the API table at `0x100000` (with `b_output` pointing at a stub) and calls the image's first byte prints `Hello, world!` and returns cleanly; a second fixture returns 45 in RAX. The load address itself cannot be reproduced in user space — it is a kernel-half canonical address — so this exercises the image's code and its API-call convention, not its placement.

### Not verified

- **Booting.** The image was never run on real BareMetal or under its qemu setup; the reference clones are read-only in this environment.
- **The stack-alignment and bss consequences.** BareMetal does not guarantee a 16-byte-aligned RSP at entry (its own `crt0.c` aligns it before calling C, because SSE faults otherwise) and does not zero `.bss`. Both are documented in `doc/manifest.md` and in the vtable's header; neither is *handled*, because both belong to a bmos startup shim in mach-std.
- **Multi-function entry placement.** `raw`'s enter-at-base contract (pre-existing, shared with `freestanding`) means `_start` must be the first function emitted; declaring it first in the entry module does that today, and any other layout is refused at link with `raw: flat image must enter at its base address`. That refusal is loud, but it is a positional rule rather than a guarantee the compiler enforces by construction.

Issue #2352, opened 2026-07-25
asm(aarch64): no msr/mrs mnemonic, so PSTATE.DIT cannot be set from Mach source


octalide:
Found while implementing `octalide/mach-std#304` (`crypto.ct`), the last listed child of the constant-time epic #1643. **It is the one thing standing between that epic and a real hardware DIT mode.**

## The gap

`msr DIT, #imm` is the **only** way to write `PSTATE.DIT` on aarch64, and it is EL0-executable — so userspace can legitimately set it. But the mnemonic is not reachable from Mach source:

```
asm aarch64 { msr DIT, #1 }
  → error: unsupported aarch64 inline-asm mnemonic 'msr'
```

`A64_MNEMONICS` in `src/lang/target/isa/arm64/encode.mach` has 42 rows and contains neither `msr` nor `mrs`.

Verified by building, not by reading the table.

## The other two targets are not gaps, and that matters for scope

- **x86_64** — Intel DOITM is `IA32_UARCH_MISC_CTL` bit 0, an MSR. **Ring-0 only; no userspace program can set it on any OS.** `wrmsr` is also absent from the x86_64 table, but that is moot: an x86_64 `ct.begin` is architecturally a no-op, not a missing feature.
- **riscv64** — Zkt is an always-on architectural guarantee with **no mode bit**. Nothing to set.

So this issue is **aarch64-only**, and closing it closes the whole DIT question.

## Scope

A mnemonic table row plus encoder support for the system-register move forms. `msr DIT, #imm` encodes as `0xD503415F` for `#1`. Add `mrs` alongside it — reading `DIT` back is how a caller verifies the mode actually engaged, and a setter with no getter is half an API.

**Do not accept a `.byte` workaround.** The worker who found this explicitly declined to reach the encoding through the parser's raw-byte path, and was right to: hand-assembled opcodes in the stdlib to route around a missing mnemonic is the workaround shape this repo rejects, and it would bypass the assembler's effect model entirely. The compiler now parses inline asm into structured instructions (`isa/asm.mach`), and a system-register write has real effects the model should know about.

That last point is the actual design question: **what does the effect model say about `msr`?** It writes processor state that no register or memory operand names. Conservative-by-default is the existing contract — an unrecognized effect means reads and writes everything, touches memory, is a barrier — so the safe answer is available. But state it deliberately rather than inheriting it, and note that #2258 is currently unifying the three targets' `{name}` effect models, so coordinate if that is still in flight.

[...]

PR #2399, opened 2026-07-26
feat(asm): add the x86-64 indirect `call` forms, and give `.byte` its own bound

### LINKED ISSUE #2398: asm(x86-64): no indirect `call`, so every BareMetal API call must be raw `.byte`s
Found while implementing the `bmos` target (#2396 / PR #2397). **This is the blocker for the mach-std bmos arm.**

## What

BareMetal exposes its entire API as a table of fixed absolute addresses you `call` through — `b_output` at `0x100018`, `b_input` at `0x100010`, and so on. In nasm that is `call [b_output]`, assembling to `ff 14 25 18 00 10 00` (disp32-absolute indirect call). BareMetal own C shim spells it `asm volatile ("call *0x00100018")`.

mach cannot express it. In `isa/asm.mach`, x86-64 `call` is `X64P_BRANCH` only — a direct relative call to a label. There is **no memory-indirect and no register-indirect form**:

```
call [0x100018]     # no
call rax            # no
```

So today every single BareMetal API call has to be emitted as raw bytes.

## And the raw-byte workaround is itself capped

`.byte` accepts at most **4 values per statement** — `MAX_OPS = 4` in `isa/asm.mach` — while `BYTES_MAX` is `STMT_MAX/2`. The 7-byte call therefore needs **two** `.byte` lines:

```mach
asm x86_64 {
    .byte 0xff, 0x14, 0x25, 0x18
    .byte 0x00, 0x10, 0x00
}
```

That works and is what PR #2397 verified against — the emitted bytes are identical to nasm shipped `hello.app`. But an entire stdlib OS backend written as split raw-byte literals is unreviewable and unmaintainable, and the bytes encode the target address opaquely, so nothing can check the address is one of the seven legal API slots.

## Scope

Two separable pieces; the first is the one that unblocks bmos.

1. **An absolute-memory indirect call**, `call [imm]`, encoding ModRM `/2` with mod=00 rm=100 and a SIB of `0x25` — exactly the `ff 14 25 <disp32>` form. This is the only shape BareMetal needs.
2. **A register-indirect call**, `call rax` (`ff /2`, mod=11). Not needed for bmos, but it is the other half of the same opcode and leaving it out invites a second visit.

Independently, **raise or remove `MAX_OPS = 4` for `.byte`** so a byte sequence up to `BYTES_MAX` fits one statement. The cap looks like an operand-count limit applied to a directive that is not an instruction; check whether it is load-bearing for anything else before changing it.

## Bar

- `call [0x100018]` assembles to `ff 14 25 18 00 10 00`, asserted **byte-exact**, not merely "assembles".
- Round-trip through a disassembler that understands the format, per the #2352 discipline — and confirm the disassembler reads the format before trusting a count from it.
- The effect model: an indirect `call` clobbers per the ABI exactly as a direct one does. It must not be modelled as opaque (see #2258).
- A mutation on the SIB byte or the ModRM must fail the test.
- Existing targets byte-identical.

Related: #2396 (bmos), #2258 (the asm `{name}` effect model), #2352 (the aarch64 mnemonic-table precedent).

octalide:
Closes #2398.

x86-64 `call` in inline asm was branch-only, so an ABI whose entry points are fixed addresses rather than symbols — BareMetal's kernel call table at `0x100010..0x100040` (#2396) — was reachable only as raw `.byte`s. That is unreviewable at stdlib scale, and it is *opaque to the effect model*: a raw byte payload makes the block clobber every register in every bank, because the parser cannot read it.

## The three shapes

`call` becomes its own operand pattern, forking on the operand's shape:

```mach
asm x86_64 {
    call some_symbol     # direct:   E8 rel32, relocated — unchanged
    call rax             # register: FF /2, mod=11
    call [0x100018]      # absolute: ff 14 25 <disp32>
    call [rax + 8]       # computed
}
```

The encoder already had both indirect paths behind `FLAG_INDIRECT`; only the grammar could not reach them.

## The absolute form needed a real addressing mode

x86-64 has **two** register-less memory forms, and `isa.Operand.base` is where they are told apart. The shared model's `-1` already means rip-relative (`mod=00 rm=101`) — that is what lowering produces for every symbol reference — so it cannot also mean absolute. `x64.MEM_BASE_ABS` names the second one, which takes the SIB escape (`mod=00 rm=100`, SIB `base=101 index=100`), giving `ff 14 25 <disp32>`. Its displacement is sign-extended to 64 bits, so an address outside signed 32-bit range is refused rather than truncated to a different address.

## The effect model is unchanged, and deliberately not opaque

An indirect `call` **reads** its target and writes nothing. What the callee destroys is the asm block's own call barrier — exactly as for a direct call. `writes: AW_NONE, implicit: 0`, same row as before (#2258).

## `.byte`'s cap: load-bearing elsewhere, so it did not move

`MAX_OPS = 4` is documented as, and is, a real machine fact about **instructions** — "the widest form any supported target spells: aarch64's post-index `ldp Xa, Xb, [Xn], #imm`". Raising it would weaken a genuine refusal. The conflation was that a *directive* was being routed through the *operand* splitter at all: `.byte` now splits on its own, bounded by the `BYTES_MAX` (256) that was already derived and already the array's size. An instruction naming five operands is still refused.

Two behaviours change beyond the cap, both flagged: a value outside a byte's range was **silently truncated** (`.byte 0x1ff` emitted `0xFF`) and is now refused; and a bare `.byte` was already an unknown-mnemonic error before the payload parser ran, so an unreachable empty-payload guard came out rather than going in.

## Verification

See the comment below for the full numbers: byte-exact assertions against nasm, a disassembler round-trip *after* confirming the disassembler reads the format, seven mutations, 1296/1296 byte-identical outputs across six targets, and a three-generation fixpoint.

### DIFF
--- a/doc/language/asm.md
+++ b/doc/language/asm.md
@@ -42,6 +42,47 @@ pub fun add_via_asm(a: i64, b: i64) i64 {
 }
 ```
 
+## Calls (x86-64)
+
+`call` takes three shapes, and which one a statement means is read off the
+operand:
+
+```mach
+asm x86_64 {
+    call some_symbol     # direct: E8 rel32, relocated against the symbol
+    call rax             # indirect through a register: FF /2, mod=11
+    call [0x100018]      # indirect through an absolute address: ff 14 25 <disp32>
+    call [rax + 8]       # indirect through a computed address
+}
+```
+
+The absolute form exists for a fixed-address ABI — one whose entry points are
+addresses rather than symbols, like BareMetal's kernel call table at
+`0x100010..0x100040`. Its displacement is sign-extended to 64 bits, so an address
+outside signed 32-bit range is refused rather than silently truncated. `call
+[symbol]` is refused too: the rip-relative form would mean "call the pointer
+*stored* at the symbol", which is not what the `call symbol` beside it means.
+
+An indirect call clobbers exactly as a direct one does — the callee's caller-saved
+registers, which the surrounding block's barrier already covers. The register or
+memory holding the target is **read**, not written.
+
+## Raw bytes
+
+`.byte` emits its values verbatim, for an encoding the ISA's mnemonic table does
+not name:
+
+```mach
+asm x86_64 {
+    .byte 0x0f, 0x01, 0xd0    # xgetbv
+}
+```
+
+One directive carries a whole sequence (up to 256 values), not four. A value that
+does not fit in a byte is refused. Raw bytes are an instruction stream the parser
+cannot read, so the block's clobber set becomes **every register in every bank** —
+which is why a real mnemonic is always preferable where one exists.
+
 ## What the compiler infers
 
 - **Operand direction.** Position within an instruction determines whether
--- a/src/lang/target/isa/asm.mach
+++ b/src/lang/target/isa/asm.mach
@@ -119,6 +119,7 @@ pub val AOF_HAS_INDEX: u8 = 0x04;  # a memory operand carries an index register
 pub val AOF_WRITEBACK: u8 = 0x08;  # the access writes its updated base register back
 pub val AOF_STACK_PTR: u8 = 0x10;  # the register names the stack pointer, not the zero register
 pub val AOF_WIDE:      u8 = 0x20;  # the index register is the wide (64-bit) form
+pub val AOF_MEM_ABS:   u8 = 0x40;  # a memory operand addresses an absolute constant (`imm` set, no base, no index)
 
 # a parsed inline-asm operand
 #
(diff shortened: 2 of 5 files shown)


octalide:
## Verification

### Byte-exact, against an independent assembler

The same six statements, assembled by nasm and by mach, compared as bytes:

```
nasm: ff142518001000 ffd0 41ffd3 ff10 ff5008 ff54d810
mach: ff142518001000 ffd0 41ffd3 ff10 ff5008 ff54d810
IDENTICAL
```

`call [0x100018]` `call rax` `call r11` `call [rax]` `call [rax + 8]` `call [rax + rbx*8 + 16]`. The REX.B on `r11` and the SIB escape on the absolute form are both asserted, not assumed. The unit test asserts these same bytes through `iasm.run` — the block path a real body takes — so CI holds the line.

### The disassembler reads the format

Per the #2352 discipline, confirmed *before* being trusted: objdump decodes BareMetal-OS's **nasm-built** `sys/hello.app` as

```
   c:	ff 14 25 18 00 10 00 	call   QWORD PTR ds:0x100018
```

so it understands this encoding on a byte stream mach did not produce. It then decodes mach's output identically.

### Seven mutations, all caught

| Mutation | Result |
|---|---|
| SIB base field `5` → `4` | `indirect_forms_byte_exact` FAILS |
| ModRM rm field `4` → `5` | `indirect_forms_byte_exact` FAILS |
| absolute routed to the rip-relative encoding | `indirect_forms_byte_exact` FAILS |
| the 64-bit-register guard removed | `refused_forms` FAILS |
| the signed-32-bit address guard removed | `refused_forms` FAILS |
| `.byte` bound `BYTES_MAX` → `4` | `sequence_is_not_an_operand_list` FAILS |
| the byte-range check removed | `sequence_is_not_an_operand_list` FAILS |

### Everything else

- **Byte-identity: 1296/1296.** Both compilers built the same source (a detached worktree at `4ec26513`) for all six declared targets; every object and image is bit-for-bit identical.
- **Suites: 968 passed, 0 failed** (dev is 964; four added).
- **Three-generation fixpoint: B == C** (`21f75f79`).
- **int, linux leg: all cases pass.**
- **End to end**: the bmos hello-world, rewritten from two `.byte` lines to one `call [0x100018]`, still prints `Hello, world!` under the Linux API-table shim from #2397.

### Not verified / deliberately out of scope

- **`jmp` has the same gap and I did not close it.** `encode_jmp` handles `FLAG_INDIRECT` only for a *register* operand and otherwise falls through to `E9 rel32` — so an indirect JMP with a memory operand would silently encode as a direct relative jump. It is unreachable today (nothing constructs that combination) and outside this issue's scope, but it is a silent-wrong-code fallthrough sitting one function above `encode_call`. Filed separately rather than fixed here, since a fix is untestable without also adding `jmp [mem]` to the grammar.
- **`call [symbol]` is refused, not implemented.** The rip-relative indirect form would relocate correctly as it happens (`x86_reloc_offset` returns `inst_size - 4`, right for `ff 15 <disp32>`), but its meaning — call the pointer *stored* at the symbol — differs from the `call symbol` beside it, and nothing in the language produces it. Shipping an untested relocation shape to support a form nobody asked for is worse than a named refusal.
- **Merge hazard**: this takes `AOF_MEM_ABS = 0x40` in the shared flag space (whose header already warns about exactly this, #2137) and `X64P_CALL = 5` / `X64F_INDIRECT = 0x0200` in the x86-64 grammar's private space. Neither exists on `origin/dev` now; re-read both constant blocks after merging.

PR #2422, opened 2026-07-30
fix(git): ignore root dependency symlinks

Angluca:
/dep/ only dir, `ln -sf ../dep/ dep` is file can't ignore
I ofen use `ln -sf` generat dep, So ...


octalide:
Unanchored `dep` pattern also ignores nested paths such as `src/dep`. Using `/dep` instead, which should work just fine.

Issue #2426, opened 2026-08-03
target: pull bmos out of the hardcoded OS set — freestanding + platform tag


octalide:
## Problem

`bmos` is modeled as a first-class hardcoded OS (`OS_BMOS`), a peer of linux/darwin/windows — but it is not that kind of thing. BareMetal has no hosted platform runtime: no syscalls (its API is an absolute function-pointer table reached via inline asm), no dynamic linker, no libdirs. Modeling it as an OS forces the compiler to carry an identity it barely uses, and splits BMOS knowledge across two independently-hardcoded places — the compiler OS *and* the mach-std bmos arm — that can, and do, drift out of sync. That drift is #2408: the compiler target and the mach-std bmos arm do not compose; no configuration builds with std linked.

## Root cause

Comparing the two `OsVTable`s, `bmos` differs from generic `freestanding` in exactly one load parameter: **`base_addr`** (`0xFFFF800000000000` vs `0`). Format is already `raw` (freestanding's default), entry is `_start`, entry-at-base is a property of the `raw` writer, and `page_size` / `stack_probe` / `native_abi` are identical. The x86_64-only and raw-only "tightenings" are not compiler knowledge — they belong to the platform library, which already exists in mach-std (`runtime/bmos.mach`, `runtime/bmos/x86_64.mach`, `system/bmos.mach`) and, per `os/bmos.mach:31-34`, is already reached through inline asm with "nothing in the compiler encod[ing] it."

So BMOS-in-the-compiler is `freestanding` + one number.

## The change

Split BMOS's identity in two:

- **`os = freestanding`** — what the compiler needs for codegen/link, parameterized by a per-target `base` override.
- **`platform = "bmos"`** — a user-defined string tag the manifest declares, exposed to comptime as `$mach.build.platform`, that libraries key on to select their backend. Open string, no compiler enum.

### Compiler — delete bmos as an OS
- Delete `src/lang/target/os/bmos.mach`.
- `target/os.mach`: drop `OS_BMOS`, the `bmos` arms in `os_id_for` / `os_name_for`, the two test asserts, and the docstring mentions.
- `target.mach`: drop `use …os.bmos`, the `bmos.register_bmos(?reg.os)` call, and the bmos select tests.
- `fe/comptime.mach`: drop the `bmos` name→id arm.

### Compiler — generalize freestanding + add the platform tag
[...]

PR #2427, opened 2026-08-03
feat(target): freestanding platform tag + base override (#2426 PR A)

octalide:
Part of #2426.

## Scope

The **additive, seed-safe** half of #2426: it lets a bespoke platform (bmos) be expressed as `os = freestanding` + a manifest platform tag + a base override, **without removing the existing bmos OS yet**. `OS_BMOS`, `src/lang/target/os/bmos.mach`, and the `$mach.os.bmos` comptime selector are all untouched, so mach-std keeps composing (its dead `$if` bmos arm still resolves `$mach.os.bmos`). The bmos-OS removal + the mach-std re-gate to `$mach.build.platform == "bmos"` are follow-up PRs.

## Three additive changes

1. **Manifest schema.** `[target.X]` gains two optional keys: `platform = "<tag>"` (an open string, STR_NIL when unset) and `base = <addr>` (u64 load-address override, 0/unset defers to the OS default). Both parse, validate as known keys, and thread onto `TargetDef`; the native host-synthesis path initializes them too.
2. **`$mach.build.platform`.** A new comptime selector returning the resolved platform string, folding an unset tag to `""` (interned, so `== ""` holds). It is an OPEN string — no enum, no `PLATFORM_*` ids. Checked before the `$mach.build.<name>` define lookup so a manifest define cannot shadow it.
3. **Base override → linker.** The resolved `Target` gains a per-target `base`; `target.select_of` initializes it to 0, the driver overwrites it from the selected target, and the linker's `os_base_addr` prefers a nonzero override over the OS vtable's `base_addr`. This is exactly the mechanism a bmos project will use (`os=freestanding, base=0xFFFF800000000000, platform=bmos`).

The build fingerprint keys on both new fields so two builds differing only in `platform`/`base` do not alias in the incremental cache.

## Design notes

- **`resolved.Target` holds a *borrowed* OS vtable pointer** (`os: *OsVTable`), it does not copy `base_addr`. So the override cannot mutate the shared vtable — it is a new per-target `base: u64` field on the resolved target, read by the linker with the vtable as fallback.
- **`base` is a single `u64` where 0 == "unset / use OS default"**, matching the issue wording. Freestanding's vtable base is already 0, so an unset base on freestanding links at 0; a nonzero manifest base wins. No separate "is-set" bool.
- The x86_64-only test path (`target.mach` `select:x86_64_freestanding_raw`) reads `t.os.base_addr` directly; it constructs its own segments and is not the production image-base path, so it is left as-is. The authoritative image base is `os_base_addr` in the linker.

## Tests

- `manifest.parse:platform_and_base` — a `[target.X] platform="bmos" base=0xFFFF800000000000` surfaces both; a plain target leaves platform STR_NIL and base 0.
- `comptime.resolve_mach_path:build_platform` — a set tag folds verbatim; unset folds to interned `""`.
- `linker.os_base_addr:manifest_base_override` — freestanding base 0 defers to the vtable; a nonzero override wins.

## Verification

- Clean self-host build (seed → HEAD): exit 0.
- `mach test` through the freshly-built HEAD compiler: **996 → 999 passed, 0 failed** (+1 each in manifest / comptime / linker).
- Self-host fixpoint: HEAD-built-HEAD (B) == B-built-HEAD (C), byte-identical; B passes 999/0. Change does not perturb the compiler's own output (its manifest sets no platform/base).

### DIFF
--- a/src/lang/be/linker.mach
+++ b/src/lang/be/linker.mach
@@ -3739,14 +3739,16 @@ fun image_name(modules: *of.ObjectImage, module_count: u32) intern.StrId {
     ret modules[0].name;
 }
 
-# the base virtual address for the active OS's executables, read
-# from the OS vtable (`tgt.os.base_addr`). the loader parameters belong to the OS
-# layer, so the linker never branches on the numeric os id. `link` validates
-# `tgt.os != nil` before any vaddr assignment, so the vtable is always present
+# the base virtual address for the active target's executables. a per-target
+# manifest `base` override (nonzero `tgt.base`) wins; otherwise the OS vtable's
+# `base_addr` applies. the loader parameters belong to the OS layer, so the linker
+# never branches on the numeric os id. `link` validates `tgt.os != nil` before any
+# vaddr assignment, so the vtable is always present
 # ---
-# tgt: active target - supplies the OS base address
-# ret: the base virtual address for the OS's executables
+# tgt: active target - supplies the base override and the OS base address
+# ret: the base virtual address for the target's executables
 fun os_base_addr(tgt: *target.Target) u64 {
+    if (tgt.base != 0) { ret tgt.base; }
     ret tgt.os.base_addr;
 }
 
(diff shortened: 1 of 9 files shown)


octalide:
Reviewed (coordinator): additive-only, all 17 CI lanes green (build/incremental/int/release/test across x86_64/aarch64/windows/riscv64; darwin gate skipped), self-host fixpoint byte-identical (B==C), +3 targeted unit tests at the parse / comptime-fold / linker-base layers. Design decisions verified against the diff: per-target `base` u64 on resolved.Target (vtable is borrowed, cannot mutate); `base==0` means "use OS default" (sound — no target wants an explicit base 0 over a nonzero default, and freestanding defaults to 0); `$mach.build.platform` folds an open string with unset -> interned "" before the define lookup; platform+base added to the build fingerprint. Merging via --admin (own-PR review requirement).

PR #2429, opened 2026-08-03
fix(link): place the flat-image entry symbol at the image base

### LINKED ISSUE #2409: linker(raw): flat image entry symbol is displaced by dependency functions — needs symbol placement control
## Summary

When building flat binary (`raw`) object format targets (such as BareMetal OS exokernel applications, microkernels, or bootloaders), execution starts at byte offset `0` of the binary image (`base address`).

Mach's flat binary image writer (`src/lang/target/of/raw.mach`) enforces that the entry symbol (`_start`) must sit at offset `0` (`entry == base`).

However, because Mach's dependency resolver and backend linker order functions in `.text` based on dependency graph traversal and module path sorting (e.g., dependencies like `std.*` before project modules, or earlier module names alphabetically), importing functions from an external dependency or earlier-sorted module places those functions at offset `0` of `.text`. This displaces `_start` to an offset > 0 and fails with:
```
error: raw: flat image must enter at its base address
```

While `std.runtime` works around this for standard executables by acting as the top-level entry module, freestanding and bare-metal developers lack a direct mechanism to specify symbol ordering or absolute placement.

## Proposed Solution

Introduce explicit symbol placement controls in Mach, such as:

1. **`#[location(...)]` / `#[section(...)]` Function Attribute**:
   Allow developers to declare explicit section or offset placement on entry routines, interrupt vector tables, and boot signatures:
   ```mach
   #[symbol("_start")]
   #[location(section = ".text.entry", offset = 0)]
   pub fun start() i64 {
       # Guaranteed to be emitted at offset 0 of .text
   }
   ```
2. **Target Section Ordering in `mach.toml`**:
   Support configurable section ordering for flat binary generation:
   ```toml
   [target.of.raw]
   section_order = [".text.entry", ".text", ".rodata", ".data", ".bss"]
   ```

## Benefits

* Gives freestanding and embedded developers deterministic control over symbol layout without relying on module naming tricks or link-order side effects.
* First-class support for exokernel and bare-metal targets (BareMetal OS, Redox, custom microkernels).

octalide:
Closes #2409

## Problem

A flat (`raw`) image carries no entry record, so execution begins at the image base — the entry symbol (`_start`) must sit at offset 0 of the merged `.text`. `accumulate_sections` concatenates input `.text` sections in link order, so a module linked **before** the entry-defining one displaces `_start` and the raw writer rejects the image:

```
error: raw: flat image must enter at its base address
```

Entry-at-base held only by the runtime-imported-first convention; a layout that imports a function-defining module before the runtime broke a freestanding/bmos build.

## Fix

`reorder_flat_entry_first` (linker): for a flat-image executable (`tgt.of.produces_image` — true only for `raw` — and `LINK_EXE`), restack the merged `.text` placements so the entry symbol's input section leads, then relay the rest in their original stable order. Runs after `accumulate_sections`, before vaddr assignment.

The reorder is expressed **entirely through `placements[].offset`** — the same kind of offset `accumulate_sections` already produces. Every downstream reader (vaddr finalization, symbol carry, byte copy, relocation patching via `placements[].vaddr`) is placement-offset-driven, so symbol addresses and relocations resolve unchanged by construction. No relocation-specific failure mode is introduced.

Gated on `produces_image`, so `elf`/`macho`/`coff`/`spv` are untouched. The entry symbol must still be the first function in its own defining section (the runtime contract, already documented in `std.runtime.bmos`); this removes the cross-module link-order dependency, which is the reported bug.

## Tests / verification

- New linker regression test `link_modules:flat_entry_leads_regardless_of_order`: links a function-defining module **before** the entry-defining one, asserts the flat image begins with the entry module's bytes and the writer accepts entry-at-base. Falsified — neutering the reorder makes exactly this test fail (`999 passed, 1 failed`).
- Existing `raw.emit_exec:entry_must_be_base` rejection test still passes (unchanged).
- `mach test` through the freshly-built HEAD compiler: **1000 passed, 0 failed** (linker suite 12 → 13; +1 test).
- Self-host fixpoint intact: three generations byte-identical (`genA == genB == genC`). The change is inert for the compiler's own ELF build.

### DIFF
--- a/src/lang/be/linker.mach
+++ b/src/lang/be/linker.mach
@@ -559,6 +559,20 @@ fun link_modules(s: *session.Session, tgt: *target.Target, modules: *of.ObjectIm
         ret R.err[bool, str](R.unwrap_err[bool, str](acc_r));
     }
 
+    # a flat image (raw) enters at its base, so the entry symbol's section must lead
+    # the merged .text regardless of link order; restack the text placements so the
+    # entry-defining section is first (#2409). only a flat-image executable needs it;
+    # every other format carries an explicit entry record and is left untouched.
+    if (tgt.of.produces_image && mode == LINK_EXE) {
+        val ei_r: R.Result[intern.StrId, str] = intern.intern(?s.interner, tgt.os.entry_symbol);
+        if (R.is_err[intern.StrId, str](ei_r)) {
+            free_scratch(alloc, placements, sec_total, sec_base, module_count, merged, merged_count, debug_names, debug_count, merged_to_out, ?strm, ?atoms);
+            ret R.err[bool, str](R.unwrap_err[intern.StrId, str](ei_r));
+        }
+        reorder_flat_entry_first(modules, module_count, merged, placements, sec_base,
+                                 ?atoms, R.unwrap_ok[intern.StrId, str](ei_r));
+    }
+
     # assign each merged section a virtual address (an executable or a shared
     # object; a relocatable library carries none). sections are placed in canonical
     # order from the OS base, each kind page-aligned; a position-independent image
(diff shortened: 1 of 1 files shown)


octalide:
Reviewed (coordinator): reorder is placement-offset-only for SK_TEXT, gated on produces_image + LINK_EXE (raw-only; elf/macho/coff/spv untouched); vaddr, symbol carry, byte copy and relocation patching all read the new offsets so they resolve unchanged by construction. Mutation-falsifiable regression (neutering the reorder fails exactly the new test); the existing entry-must-be-base rejection still passes; 1000/0; self-host fixpoint byte-identical A==B==C. All CI green. Merging via --admin.

Issue #2408, opened 2026-07-26
bmos: the compiler target and mach-std bmos arm do not compose — no configuration builds with std linked


octalide:
Found while building a standalone demo project against released **mach 4.3.1** and **mach-std 0.20.0**. Each half was verified in isolation; together they do not work.

## Reproduction

A project with `os = "bmos"` and `[dep.mach-std] ref = "branch/main"` (0.20.0). Three configurations, all failing except the one that links no standard library at all:

| program | result |
|---|---|
| own `_start`, **no std dep** | **builds** — an 8-byte image |
| own `#[symbol("_start")]`, `use std.system.bmos` | `error: raw: flat image must enter at its base address` |
| `#[symbol("main")]` instead, `use std.system.bmos` | `error: undefined entry symbol '_start'; ensure the target startup library is linked` |

The second case is minimal — one function, `_start`, literally first in the file:

```mach
use std.types.size.usize;
use bmos: std.system.bmos;

#[symbol("_start")]
fun start() i64 {
    bmos.output("hi\r\n", 4);
    ret 0;
}
```

## The two halves disagree about who owns `_start`

- **The compiler target** documents the program as the provider. `src/lang/target/os/bmos.mach:75`: *"the `_start` entry symbol (a program marks its entry fn `#[symbol("_start")]`)"*, and `bmos_vtable.entry_symbol = "_start"` with **no platform library** (`libdir_len = 0`).
- **mach-std** provides one too. `src/runtime/bmos/x86_64.mach` defines `#[naked] pub fun _start()`, and its own header says *"`_start` MUST BE THE FIRST FUNCTION EMITTED … this module is the entry module"*, calling `main(0, nil)`.

So with std linked there are two candidate entry points and the writer places the wrong one first; without a program-side `_start` nothing supplies it, because bmos declares no startup library to pull the runtime in.

The int fixtures do not catch this because **none of them uses `std`** — `int/surface/bmos` and `int/regression/2402-raw-bss-fill` both define their own `_start` and import nothing. mach-std's CI does not catch it because it never builds a bmos *binary*; the arm is gated declarations that a hosted build never instantiates.

## What needs deciding

Which side owns the entry point, and then make the other side agree:

1. **mach-std owns it** (matching linux/darwin/windows, where the runtime provides the entry and user code writes `main`). Then bmos needs to declare a startup library so the runtime is linked, and `raw` must place that module first.
[...]

PR #2430, opened 2026-08-03
refactor(target): remove bmos as a hardcoded OS (#2426)

### LINKED ISSUE #2426: target: pull bmos out of the hardcoded OS set — freestanding + platform tag

octalide:
Closes #2426

Compiler-side of #2426 (PR B): remove `bmos` as a hardcoded OS. bmos is no longer a first-class OS peer of linux/darwin/windows — it is `os = freestanding` plus a per-target `base` override, selected by libraries through the platform tag (`$mach.build.platform == "bmos"`). That platform-tag model already lives in-tree; the `OS_BMOS` identity was redundant and only invited drift with the mach-std bmos arm (#2408). bmos now ships as the external `mach-bmos` platform package.

## Changes
- **`build(dep)`**: bump mach-std `11d8f618` (0.20.2) → `a35a0c46` (0.21.0), which removed its bmos arm.
- **`refactor(target)`**: delete the `OS_BMOS` id, its `os_id_for` / `os_name_for` arms, the `$mach.os.bmos` resolver arm in `fe/comptime.mach`, the `register_bmos` registration in `target.mach`, the `x86_64_bmos_flat_image` select test, and the `target/os/bmos.mach` OsVTable file.
- **`test(int)`**: retarget the two bmos integration fixtures (`int/surface/bmos`, `int/regression/2402-raw-bss-fill`) from `os="bmos"` to `os="freestanding"` + `base=0xFFFF800000000000` + `platform="bmos"` — the same entry-at-base raw flat-image contract, now through the platform-tag model. Target name (`bmos-x86_64`) and `built=1` goldens unchanged.
- **`docs`**: purge `os="bmos"` / `$mach.os.bmos` from `doc/manifest.md`, `doc/cli.md`, `doc/language/comptime-mach.md`; rewrite the manifest's bmos section to the freestanding + `platform`/`base` model and document those two (previously undocumented) optional target keys. `CHANGELOG.md` history is left for the 4.5.0 release chore.

The freestanding + `base` placement mechanism bmos relied on is unchanged, and the platform-tag model (`platform` manifest field, `$mach.build.platform`) is untouched. Net src/lang: 5 insertions, 171 deletions.

## Verification (from-source HEAD compiler, x86_64-linux)
- **Build**: from-source build against mach-std `a35a0c4` succeeds.
- **Self-host fixpoint**: `B == C` byte-identical (`sha256 be949be5…`) — removing the dormant OS does not perturb the compiler's own ELF codegen.
- **`mach test` through HEAD**: `1000 → 999` suites, 0 failed. The single-suite drop is exactly the removed `x86_64_bmos_flat_image` select test (suite counts are compiler-generation-independent).
- **`mach info targets`**: no longer lists `bmos`; `freestanding-{x86_64,aarch64,riscv64,spirv,mos6502}` remain.
- **Compose check**: a `[target.bmos]` = `os="freestanding"`, `platform="bmos"`, `base=0xFFFF800000000000` with a `#[symbol("_start")]` entry builds a valid raw flat image (single `0xc3` RET) through the HEAD compiler — the mach-bmos model is unaffected.
- **CI**: green (build/incremental/test/release/int across x86_64/aarch64/windows; darwin gated to main).

## Notes
- End-to-end working bmos link still needs #2409 (already merged to dev); this compiler-side pull-out lands independently.


octalide:
Reviewed (coordinator): clean removal — pin 11d8f618->a35a0c4 (mach-std 0.21.0, bmos arm gone), os/bmos.mach deleted, OS_BMOS + os_id_for/os_name_for arms + the $mach.os.bmos resolver arm + register_bmos + the x86_64_bmos_flat_image select test all removed; only the 3 platform-tag sites kept. int fixtures migrated to freestanding+platform (goldens/target-name preserved). Docs rewritten to the platform-tag model (manifest.md "Platform targets", cli.md, comptime-mach.md). 999/0, self-host fixpoint B==C byte-identical, info targets dropped bmos, and a freestanding+platform=bmos build still composes. All CI green. Merging via --admin.

Issue #2479, opened 2026-08-04
asm: .byte/.word/.long/.quad are x86-only — fixed-width targets have no raw-encoding escape


octalide:
`.byte` (and `.word`/`.long`/`.quad`) are registered only in the x86-64 encoder (src/lang/target/isa/x64/encode.mach:5017) and appear in neither the arm64 nor the riscv64 mnemonic tables. On those targets they are refused with `unknown aarch64/riscv64 inline-asm instruction`.

Verified empirically on 4.4.0 with throwaway freestanding aarch64/riscv64 artifacts, probing each directive.

This is arguably the root problem behind the other encoder gaps. On x86, a missing mnemonic is an inconvenience — `.byte` emits the instruction (that is how bmos-mach does port i/o and rdtsc today, see #2468). On aarch64/riscv64 a missing mnemonic is a hard wall: if it is not in the table, the instruction cannot be emitted at all. Concretely this blocks bmos-mach's timer and trap-handler installation on both new architectures (see the sibling issues), with no workaround available in-language.

Fixed-width ISAs make the fix easy and natural: a `.word <u32>` (and `.quad`) escape that emits a raw instruction word, respecting target endianness. That single addition would unblock every current and future encoder gap on those targets without waiting for mnemonic coverage.

Disposition: mach backlog; suggest prioritizing above the per-mnemonic gaps, since it subsumes them.

PR #2504, opened 2026-08-05
feat(riscv64): add the CSR instruction family to inline asm

### LINKED ISSUE #2481: riscv64: no CSR instructions in the inline-asm mnemonic table
The riscv64 inline-asm mnemonic table has 91 entries and no CSR family: `csrr`, `csrrs`, `csrrw`, `csrw`, and `rdtime`/`rdcycle`/`rdinstret` are all refused as unknown instructions. The CSR file is therefore completely unreachable from Mach, and with #2479 (`.byte`/`.word` are x86-only) there is no raw-encoding fallback.

Verified on 4.4.0 against real freestanding riscv64 builds. The rest of the surface is healthy — `li` with large immediates, `ld`/`sd`/`sb`, `la`, and `ecall` all assemble.

Impact in practice (bmos-mach v1, a kernel in Mach): no stvec means no trap vector, so `ecall` has nowhere to land and the syscall layer cannot exist on riscv64. Timekeeping survives only because QEMU virt's CLINT exposes `mtime` as a plain MMIO load at 0x0200BFF8 — an ordinary 64-bit load, no CSR needed — so the port keeps a real 100 ns-resolution clock. Trap handling has no such escape.

`csrrw`/`csrrs`/`csrrc` plus the immediate forms and the `rdtime`/`rdcycle` pseudo-instructions would cover essentially all systems use.

Disposition: mach backlog, maintainer's call on priority.

octalide:
## Summary

Closes #2481. The riscv64 inline-asm mnemonic table had no Zicsr CSR family — `csrr`/`csrw`/`csrrs`/`csrrw` and the `rdtime`/`rdcycle`/`rdinstret` pseudos were all refused as unknown instructions. `stvec` (the trap vector) and `satp` (the MMU) were therefore unreachable, and with #2479 (`.byte`/`.word` are x86-only) there was no raw-encoding fallback.

Follows the shape of #2500 (aarch64 system registers):

- All six real Zicsr instructions: `csrrw`/`csrrs`/`csrrc` (register source) and `csrrwi`/`csrrsi`/`csrrci` (five-bit immediate source).
- Every standard assembler pseudo that spells one of the six: `csrr`, `csrw`/`csrs`/`csrc`, `csrwi`/`csrsi`/`csrci`, and `rdcycle`/`rdtime`/`rdinstret`.
- A CSR name table (`mstatus`, `mtvec`, `mepc`, `mcause`, `mie`, `mip`, `mhartid`, `satp`, `stvec`, `sepc`, `scause`, `sstatus`, `sie`, `sip`, `time`, `cycle`, `instret`), matched case-insensitively in one place (`rv_name_eq`).
- A numeric CSR address form (`csrr a0, 0xc01`) — required since a name table can never be complete. Unlike aarch64's system registers, RISC-V spells no separate escape syntax for this: a CSR operand is already a plain assembler immediate, so the numeric path reuses the shared operand lexer's existing parse rather than re-deriving it.
- `asm_ct_class` classifies all six real ops as `CT_OP_NONE` (and therefore every pseudo, since each pseudo's row names one of the six as its `code`): which CSR is accessed is fixed at assembly time, never an operand, so it cannot be secret, and the six forms are fixed-latency in the operands they do carry.

**Correctness bar (per #2481's instructions):** every CSR address, and every test's expected instruction word, was read from `llvm-mc -triple=riscv64 -mattr=+zicsr --show-encoding`, decoding `imm[31:20]` of the word independently — never read back from the table or test-computed by the same reasoning that built it. No CSR was omitted for uncertainty; all 17 requested names resolved and cross-checked cleanly against llvm-mc.

## Test plan

- [x] `mach.lang.target.isa.riscv64.encode` grew from 36 → 39 tests: `inline_asm_csr_forms` (all six real forms + all ten pseudos, byte-exact), `inline_asm_csr_names` (all 17 named CSRs read back byte-exact, plus mixed-case and numeric-address equivalence), `inline_asm_csr_refusals` (unknown CSR name, out-of-range numeric address, out-of-range `uimm5`)
- [x] `mach test .` — 1058 passed, 0 failed, run through the freshly self-hosted compiler (not the PATH seed)
- [x] Mutation check: flipped `mstatus`'s address by one bit, rebuilt, confirmed `inline_asm_csr_forms` and `inline_asm_csr_names` both failed; reverted, confirmed clean pass again
- [x] Cross-built the compiler itself for `linux-riscv64` (exercises the shared `shape_of`/`build_inst`/`classify_i` codegen paths this PR touches) — produced a valid riscv64 ELF

https://claude.ai/code/session_01MddQFDuSwQSS9bPMAQPbtg

### DIFF
--- a/doc/language/asm.md
+++ b/doc/language/asm.md
@@ -110,6 +110,44 @@ set becomes **every register in every bank**, and an `#[oblivious]` function may
 contain one at all — which is why a real mnemonic is always preferable where one
 exists.
 
+## Control-and-status registers (riscv64)
+
+The Zicsr extension's six instructions — read-write, read-set and read-clear, each
+taking its source from a register or a five-bit immediate — reach a CSR by name:
+
+```mach
+asm riscv64 {
+    csrrw a0, mstatus, a1   # read mstatus into a0, write a1 into it
+    csrr  a0, mtvec         # csrrs a0, mtvec, x0 - the read-only pseudo
+    csrw  stvec, a1         # csrrw x0, stvec, a1 - install a trap vector
+    rdtime a0               # csrrs a0, time, x0  - the unprivileged counters
+}
+```
+
+The named set covers what freestanding code reaches for — the machine and supervisor
+trap vector / exception-PC / cause registers, the interrupt enable / pending pairs, the
+address-translation root, the hart id an SMP boot path reads to tell cores apart, and
+the three unprivileged counters `rdtime` / `rdcycle` / `rdinstret` name. It is
+deliberately not exhaustive: the privileged spec defines several hundred addresses
+across three privilege levels, so **any** CSR is also reachable by its numeric address,
+exactly as a name resolves to one:
+
+```mach
+asm riscv64 {
+    csrr a0, 0xc01   # the same register as `csrr a0, time`
+}
+```
+
+Unlike aarch64's system registers, RISC-V spells no separate escape syntax for this —
+a CSR operand simply parses as the ordinary integer literal it looks like, bounded to
+the twelve bits a CSR address occupies. `csrrwi` / `csrrsi` / `csrrci` (and their
+`csrwi` / `csrsi` / `csrci` pseudos) take a five-bit unsigned immediate in the same
+position a register would occupy in the non-`i` form.
+
+Access permission is not checked: whether a CSR is readable or writable depends on the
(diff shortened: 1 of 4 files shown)


octalide:
Closing and reopening to nudge a stuck CI webhook trigger.

Issue #2509, opened 2026-08-05
link: emit PE .rsrc resources (icon, version info, manifest)


octalide:
## What

A shipped Windows application needs an icon, version metadata, and a DPI-awareness application manifest. These live in the PE `.rsrc` section, which mach does not emit. Resource compilers (rc.exe, windres) are external tooling we do not depend on; PE emission is fully in-house, so resource emission belongs in the linker behind manifest keys — no `.rc` source surface.

Manifest shape (windows-only keys, inert elsewhere, same convention as `subsystem`):

```toml
[artifact.game]
icon = "assets/game.ico"            # ICO container; every embedded size carried over
manifest = "assets/game.manifest"   # XML application manifest
```

**Amended 2026-08-05** (supersedes this issue's original nested `[artifact.game.windows]` proposal, which was wrong): keys go flat on `[artifact.<name>]` and are optional. `doc/manifest.md` states that per-target artifact variation "is not a per-cell exception table — it is a second artifact stanza, so the condition stays visible", which rules out a nested per-OS sub-table. A windows artifact typically already needs its own stanza anyway, since the `.exe` extension is written literally into `out`, so these keys land naturally on that stanza and the condition stays visible as the manifest intends. Optional is correct despite every current `[artifact.X]` key being required — the alternative forces linux-only projects to write meaningless windows keys — and `[link.X]`'s `library` key is the precedent for an optional key. Document them with `Required: no` in the key table, inert on non-windows cells (consistent with how non-matching `[link.X]` entries are skipped).

Emission:

- `RT_GROUP_ICON` + `RT_ICON` entries decomposed from the ICO container.
- `RT_VERSION`: `VS_VERSIONINFO` fixed block sourced from the artifact's `version` in `mach.toml`; string table from name/description keys where present.
- `RT_MANIFEST`: the given file, byte-preserved.
- Standard three-level resource directory tree (type → id → language), language-neutral entries.
- No `.rsrc` section at all when no keys are set — output stays byte-identical to today.

## Acceptance

- A PE built with `icon` shows it in Explorer; the resource tree structure is asserted in CI via an in-tree reader alongside the existing PE verification tests.
- `RT_VERSION` reflects the `mach.toml` version.
- `RT_MANIFEST` is byte-identical to the input file.
- Absent keys → byte-identical output to current emission.
- `doc/manifest.md` updated.

Note: expect `feat` traffic in `src/lang/of/coff.mach` from the subsystem knob issue; land after it and merge `dev` before marking ready.

Issue #2508, opened 2026-08-05
link: PE subsystem selection for windows executables


octalide:
## What

Every PE executable mach emits today carries the console subsystem, so a graphical application opens a console window behind it on launch. The subsystem field already exists in the PE optional header writer (`src/lang/of/coff.mach`); it is simply not configurable.

Add a per-artifact manifest key with a CLI override:

```toml
[artifact.game]
kind = "bin"
subsystem = "gui"   # windows-only; "console" (default) | "gui"
```

- Maps to `IMAGE_SUBSYSTEM_WINDOWS_GUI` (2) vs `IMAGE_SUBSYSTEM_WINDOWS_CUI` (3) in the optional header.
- On non-windows targets the key is accepted and inert, matching the manifest's existing philosophy for target-filtered keys (`[link.X]` os filters); document the inertness.
- CLI: `--subsystem gui|console`, overriding the manifest with the same precedence the other artifact-level CLI overrides use.

## Acceptance

- Default output unchanged (console, byte-identical to today when the key is absent).
- `subsystem = "gui"` produces a PE whose optional header Subsystem field is 2, asserted through the existing PE verification harness.
- Non-windows targets: key accepted, no emission change.
- `doc/manifest.md` and `doc/cli.md` updated.

Issue #2507, opened 2026-08-05
sema: #[embed] — compile-time file embedding


octalide:
## What

Applications need single-file distribution: assets compiled into the executable rather than shipped as a loose directory. The static-linking route (vendored `.a` archives) already makes the executable self-contained for code; there is no equivalent for data. Mach has no mechanism to source a constant's bytes from a file at compile time.

Add a compile-time embedding primitive as a declaration attribute, consistent with the existing attribute idiom (`#[symbol]`, `#[library]`):

```mach
#[embed("assets/logo.qoi")]
const LOGO: []u8;
```

Design:

- The path resolves relative to the declaring source file. If the manifest layer already defines a project-root path convention, support it identically; otherwise file-relative only.
- The bytes are placed in the read-only data section exactly as any other const byte array; the slice length is the file length. No copy at runtime, no I/O path — embedded data is file-backed mapped memory, paged on demand.
- The embedded file becomes a build input: its content participates in incremental fingerprinting the same way `[step]` inputs do, so editing the asset rebuilds the declaring module and an untouched asset stays a cache hit.
- Two embeds of byte-identical content within one artifact deduplicate to a single data placement.
- Works for every artifact kind and target, freestanding included (bytes land in the flat image).

## Acceptance

- Attribute compiles and round-trips (program observes correct length and bytes vs the source file) on linux/windows/darwin x86_64 and aarch64 where applicable.
- Missing file, directory-instead-of-file, and unreadable file each produce a clean frontend diagnostic naming the declaration and the resolved path.
- Fingerprint tests: asset edit → rebuild reflects new bytes; no-op rebuild → cache hit.
- Dedup test: two identical embeds → one read-only data placement.
- `doc/language/` page for the attribute; `doc/manifest.md` note on fingerprinting.

PR #2515, opened 2026-08-05
docs: end-user application distribution guide

### LINKED ISSUE #2511: docs: end-user application distribution guide
Depends on #2507, #2508, #2509.

## What

There is no end-to-end guide for shipping an application built with mach. The pieces exist or are landing — static linking of vendored archives, `#[embed]` (#2507), PE subsystem selection (#2508), PE resources (#2509) — but nothing collects them. Add `doc/distribution.md` covering, per OS:

- **linux**: the fully-static route and where the dynamic boundary belongs (ubiquitous system libraries like X11 stay dynamic); `tar.gz` layout; why no rpath is needed on the static route.
- **windows**: subsystem selection, icon/version resources, the exe-adjacent DLL fallback for anything shipped dynamically, `zip` layout.
- **darwin**: `.app` bundle layout (`Info.plist`, `Contents/MacOS`, `Contents/Resources`), codesigning and notarization recipe via `rcodesign` from a linux host — positioned explicitly as external tooling the compiler does not own — and the Gatekeeper reality that makes it non-optional for public distribution.
- **assets**: single-file shipping via `#[embed]` vs a loose asset directory, and when each is right.
- **cross-building**: producing all three targets from one linux host (`--all-targets`), including the C cross-toolchain requirement for vendored-archive dependencies.

## Acceptance

- `doc/distribution.md` exists and is linked from the doc index.
- Every command shown was run against the current toolchain as part of authoring the PR.
- The page documents only shipped behavior — it lands after the features it references merge, never ahead of them.

octalide:
Adds `doc/distribution.md`, the end-to-end guide for shipping an application built with mach, linked from both doc indexes (`README.md` and `doc/language/README.md`).

## Status: draft, blocked on dependencies

Per the issue, this page documents **only shipped behavior** and must land after the features it references. Three sections are deliberately **not yet written**:

| Section | Blocked on |
|---|---|
| windows subsystem selection | #2508 |
| windows icon / version / manifest resources | #2509 |
| `#[embed]` single-file asset shipping | #2507 |

I verified against the current `dev` toolchain that none of these have shipped yet — the PE optional header hardcodes `IMAGE_SUBSYSTEM_WINDOWS_CUI` (`src/lang/target/of/coff.mach:3278`) and emits no `.rsrc` (RESOURCE data directory is 0/0) — so documenting them now would describe behavior that does not exist. I am watching those three PRs and will write each section against its actual merged surface, merge `dev`, re-run every command, and mark ready.

## What is written and verified

Every command and transcript on the page was run against a toolchain built from `dev` (4.5.0), not the installed release. Verified end to end:

- **linux static route** — a vendored C `.a` links into a fully static ELF (`statically linked`, `not a dynamic executable`); the sample returns `42` from the vendored C at runtime. No rpath is emitted because there is no loader to direct.
- **linux dynamic boundary** — a `system` X11 entry records `DT_NEEDED [libX11.so.6]` (the SONAME, ABI-versioned) with no `DT_RPATH`/`DT_RUNPATH`.
- **windows DLL fallback** — the PE import directory records bare basenames (`objdump -p` shows `kernel32.dll`, `render.dll`), which is *why* an exe-adjacent DLL resolves. Corollary verified: `render.dll` exists nowhere on the build host, so a PE cross-link needs no DLL present.
- **darwin `.app`** — the bundle executable is placed in-tree by the artifact `out` template (`Demo.app/Contents/MacOS/Demo`), no assembly step; the `Info.plist` parses and the executable bit survives the zip round trip.
- **cross-building** — `--all-targets` produces ELF + PE + Mach-O from one linux host with nothing installed for a pure-Mach project.
- **cross-build verification** — the page checks a cross-built binary by **reading its headers**, not by executing it: `objdump -f`/`-p` for the PE container, machine, `PE32+` magic, subsystem, and import directory, and `llvm-objdump --macho --private-headers` for the Mach-O (`MH_MAGIC_64 X86_64 EXECUTE NOUNDEFS`). Two of the three targets cannot run on the host that built them, so header inspection is both the more honest check and the one that works everywhere. `--runner` is noted as a property of the environment a build is *tested* in, with a pointer to `cli.md`, rather than presented as the way to verify a cross-build.
- **vendored C cross-build** — reproduced both failure modes when the step runs the host `cc` for every cell (`coff: not an x86-64 object`, `macho: not a 64-bit Mach-O object`), and the `MACH_TARGET_OS`-dispatched fix. Also documents the Mach-O leading-underscore symbol prefix, which otherwise fails only on the darwin cell.

The rcodesign recipe is the one thing **not** runnable here — signing needs a Developer ID certificate and notarization an App Store Connect key. Its flags are quoted against rcodesign 0.29's upstream documentation (read from the `.rst` sources, not from memory), and the page marks it explicitly as a recipe rather than a verified transcript. rcodesign is framed throughout as external tooling the compiler does not own.

Closes #2511


octalide:
Filed #2516 from this work — out of scope here, no fix in this PR.

While verifying the cross-building section I found that mach's ELF reader rejects **every clang-emitted x86-64 object**: clang gives `.eh_frame` the architecture-specific type `SHT_X86_64_UNWIND` (`0x70000001`), the reader admits only `SHT_PROGBITS`/`SHT_NOBITS`, so the section is dropped and its `.rela.eh_frame` relocation resolves to no section — `error: linker: relocation names an out-of-range section`. gcc spells the same section `SHT_PROGBITS` and links fine.

This matters to the page because `clang --target=` is the natural single-compiler answer for building a vendored C archive for all three targets. The script this PR documents therefore uses `cc` for the linux cell and clang triples for the cross cells, which is a reasonable shape on its own terms and is fully verified today. Once #2516 lands, a uniform clang script would also work; nothing on the page needs to change either way.

octalide:
Amended the cross-building section: the guide now verifies a cross-built binary by **header inspection** rather than by executing it (af9a1bf0).

Reading the headers is the better check on the merits — two of the three targets cannot run on the host that produced them, so an execution-based check only ever covers the one target you were least worried about, and it needs an emulator that CI may not have. `objdump -f`/`-p` establishes the PE container, machine type, `PE32+` magic, subsystem, and import directory; `llvm-objdump --macho --private-headers` establishes the Mach-O header (`MH_MAGIC_64 X86_64 EXECUTE NOUNDEFS`). `NOUNDEFS` in particular proves the vendored C archive linked with nothing left to bind, which running the program would only show indirectly.

`--runner` is still mentioned, but neutrally — as a property of the environment a build is *tested* in, pointing at `cli.md#mach-run` — not as the way to check a cross-build. The PR body is updated to match.

octalide:
Merged `dev` (picking up #2519) and wrote the **windows console/GUI section** against the shipped surface, plus the **darwin vendored-C limit**. Rebuilt the toolchain from the merged branch (4.6.0) and re-ran every existing transcript on the page against it — all still reproduce exactly.

**Subsystem, verified structurally** (no runtime claim — nobody has watched a `gui` binary decline to open a console, and the page does not say otherwise):

| case | Subsystem field |
|---|---|
| key absent (default) | `00000003 (Windows CUI)` |
| `subsystem = "gui"` | `00000002 (Windows GUI)` |
| `--subsystem gui` | `00000002 (Windows GUI)` |
| `--subsystem console` over a `gui` key | `00000003 (Windows CUI)` |

Inertness verified byte-for-byte, not just asserted: a linux build with `subsystem = "gui"` in the manifest, and the same build with `--subsystem gui` on the command line, both produce an ELF with sha256 `5ec12335ac01f3b2eff73d376393573d4735301f98023473ddc0cfe8f4f7fefd` — identical to the build with no key at all.

The page covers the shipping *decision* and links to `manifest.md#subsystem--the-windows-consolegui-selector` and `cli.md#mach-build` for the reference rules rather than restating them. It does note one consequence worth knowing: a `gui` image starts with no console attached, so stdout goes nowhere and diagnostics belong in a file.

**Darwin limit** — new subsection stating without hedging that a darwin build vendoring C, or touching any Apple framework, must run on macOS: the SDK and frameworks are licensed to Apple hardware and are not redistributable, so this leg is not cross-buildable at all rather than merely inconvenient. It includes a per-leg table (pure Mach: yes; vendored C with no system includes: yes; SDK headers or frameworks: no) and points at `darwin-lane.yml`, where mach itself does exactly this split — cross-building the darwin seed on `ubuntu-latest`, then the real fixpoint on `macos-15`. The page intro now links to it, since it is the constraint most likely to break someone's pipeline plan.

Still blocked on #2507 and #2509. #2516 is not on `dev` yet, so the vendored-C script still uses host `cc` for the linux cell.

octalide:
Merged `dev` through #2531 and **simplified the vendored-C script to one clang triple per cell** (04d23476).

Verified the fix against my original repro before changing anything — `gcc.o`, `clang.o`, `gcc.a`, and `clang.a` all link now, and the clang-archive build runs and returns 42. The script's linux branch was only using host `cc` to route around #2516/#2531; with that gone, nothing varies per cell but the triple:

```sh
case "$MACH_TARGET_OS" in
    linux)   triple="$MACH_TARGET_ISA-unknown-linux-gnu" ;;
    windows) triple="$MACH_TARGET_ISA-pc-windows-msvc"   ;;
    darwin)  triple="$MACH_TARGET_ISA-apple-darwin"      ;;
esac
clang --target="$triple" -c -O2 -Ivendor/qz -o "$obj" vendor/qz/qz.c
```

All three targets build from one linux host with it, and I re-ran every transcript on the page against the rebuilt toolchain (4.6.0 at `dev`) rather than assuming the `elf.mach` rewrite left them alone: static/dynamic ELF, `DT_NEEDED`, PE `objdump -f`/`Magic`/`Subsystem`, the import directory, the Mach-O header, the full subsystem matrix, the plist parse, the tar.gz layout, and the exec-bit zip round-trip. All reproduce exactly.

Still draft. Remaining: #2507 (PR #2518 under review), #2509, and #2535 — the windows vendored-C story stays as-is until #2535 lands, per the agreed disposition.

octalide:
Refreshed against shipped 4.7.0: rebased onto current dev, added the `icon`/`manifest` resource section (#2561), expanded `#[embed]` (#2518), and corrected the archive-selection claim (#2520 semantics). Version-resource derivation claims verified against src/lang/target/of/rsrc.mach.

PR #2517, opened 2026-08-05
feat(link): emit PE .rsrc resources (icon, version info, manifest)

### LINKED ISSUE #2509: link: emit PE .rsrc resources (icon, version info, manifest)

octalide:
Emits a PE `.rsrc` section carrying the application icon, version metadata, and
application manifest, driven by two new manifest keys. No external resource
compiler is involved — resource emission is in-house alongside the rest of PE
emission.

```toml
[artifact.game]
kind = "bin"
icon = "assets/game.ico"          # ICO container; every embedded size carried over
manifest = "assets/game.manifest" # optional XML application manifest
```

- `RT_GROUP_ICON` + `RT_ICON` decomposed from the ICO container.
- `RT_VERSION`: `VS_VERSIONINFO` whose fixed block is sourced from `[project].version`.
- `RT_MANIFEST`: the given file, byte-preserved.
- Standard three-level resource directory tree (type -> id -> language), language-neutral.
- No `.rsrc` section at all when neither key is set, so output stays byte-identical to today.

### Schema shape

The issue's snippet shows a nested `[artifact.game.windows]` table, but its prose
specifies "same convention as `subsystem`" — and #2508's `subsystem` is a flat
artifact key. `doc/manifest.md` also states outright that a per-target artifact
condition "is not a per-cell exception table — it is a second artifact stanza",
and the schema has no nested-override precedent anywhere. These keys are
therefore flat `[artifact.<name>]` keys, windows-only and inert elsewhere,
exactly like `subsystem`. Capability is identical; only the spelling differs from
the issue's snippet. Flagging explicitly for review.

### Status

Draft — work in progress. Depends on #2508 landing first (shared edit sites in
`src/lang/target/of.mach` and `src/lang/manifest.mach`); `dev` will be merged in
and the whole thing re-verified before this is marked ready.

Closes #2509

### DIFF
--- a/doc/manifest.md
+++ b/doc/manifest.md
@@ -352,6 +352,8 @@ reads the selected artifact's name.
 | `targets` | yes | Array of declared target names this artifact builds for; `["*"]` means every declared target. |
 | `link`    | yes | Array of `[link.X]` names this artifact links (see below). `[]` for none. |
 | `need`    | yes | Array of `[step.X]` names this artifact demands directly, for step outputs that are not themselves link inputs. `[]` for none. |
+| `icon`    | no  | Windows only. Path to an `.ico` container, relative to the project root. Every image the container embeds is carried into the executable. |
+| `manifest` | no | Windows only. Path to an XML application manifest, relative to the project root. Embedded byte for byte. |
 
 - **`bin`** links an executable at the resolved `out` path.
 - **`static`** materialises a real `ar` archive at the resolved `out` path — the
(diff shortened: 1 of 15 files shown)


octalide:
Superseded by #2561, which rebuilds this work cleanly on current dev with checked resource layout/encoding and expanded integration coverage. This branch remains available as historical reference.

PR #2518, opened 2026-08-05
feat: #[embed] compile-time file embedding

### LINKED ISSUE #2507: sema: #[embed] — compile-time file embedding

octalide:
Adds `#[embed("path")]`, a declaration attribute that sources a `val`'s bytes from a file at compile time.

```mach
#[embed("assets/logo.qoi")]
val LOGO: [_]u8;          # length taken from the file's byte count

#[embed("boot/sector.bin")]
val SECTOR: [512]u8;      # length pinned; a size change fails the build
```

## What it does

- Resolves the path against the **declaring source file's** directory.
- Places bytes in read-only data as an ordinary constant byte array: no runtime I/O or copy. Empty files remain present read-only initializers rather than falling into `.bss`.
- Publishes each file as a content-digested `Q_EMBED_FILE` build input. Editing the asset invalidates the embedding module; an unchanged warm rebuild reuses its exact `Q_LOWER` result and borrowed buffer.
- Shares one placement between byte-identical embeds within a module, matching bytes, length, section, kind, and alignment. Ordinary globals are never merged because their distinct addresses are observable.
- Diagnoses missing files, directories, permission-denied files, files beyond the compiler's u32 array/section limit, malformed or duplicate decorators, invalid targets, and size mismatches without inferred-length cascades.
- Works through public cross-module references and every artifact format, including freestanding flat images.

## Language design

The issue's original spelling used `const LOGO: []u8`, but Mach has neither `const` nor a slice type. This uses `val` and adds `[_]T`: an inferred array length legal only on an `#[embed]` declaration. That leaves `[]T` available for a future real slice type.

An explicit `[N]u8` is also accepted and checked against the file. This deliberately supports fixed-size contracts such as boot sectors and ROM images.

## Follow-ups

- Mach cannot currently ask a value for its type's length. The `$size_of` versus `$len_of` language decision is tracked in #2536 and documented as a limitation.
- Deduplication is currently per module. The object boundary does not retain the embed identity, so identical content embedded by two modules occupies two final placements. The format-neutral linker carrier and cross-module coalescing work is specified in #2541. The integration fixture now exercises the public cross-module path so this boundary stays visible.

## Verification

All local verification ran on **x86_64 Linux** in `/tmp/codex-embed-2518.V1a3mh`, isolated from the shared Claude scratch directory that previously suffered build-output collisions.

**Executed natively:**

- Debug and release self-host fixpoints converge (`b == c`).
- Both fixpoint compilers pass **1095/1095** tests.
- Full `int/run.sh --target linux`: **129** profile/case runs pass.
- `int/lib/check-determinism.sh`: warm no-op and post-edit incremental builds both match clean.
- The embed surface case executes in debug and release, including inferred and explicit lengths, same-module dedup, empty embeds, and a public embed imported from another module.
- The freestanding raw image exits with a byte read from an embed.
- Final CI passed debug/release builds, units, incrementals, and Linux/arm64/riscv64/Windows integrations in [run 31040506909](https://github.com/briar-systems/mach/actions/runs/31040506909); the native Darwin release gate is skipped by PR policy.

**Cross-built and inspected, not executed:**

- The compiler builds valid images for `linux-arm64`, `linux-riscv64`, `windows-x86_64`, `darwin-x86_64`, and `darwin-aarch64`.
- The expanded embed fixture builds valid ELF, PE, and Mach-O images for all six configured targets, including empty and cross-module embeds.
- Its 17-byte sequence appears exactly twice in every final image: one placement shared by the three declarations in `main`, plus the second-module placement tracked by #2541.

No emulator or compatibility layer was used. Windows, aarch64, riscv64, and Darwin execution remain CI/real-runner validation; the PR matrix runs the first three at PR cadence, while Darwin runs at the dev-to-main release gate.

## Tests added

- `int/surface/embed`: byte round-trip, explicit-length agreement, same-module dedup, empty-file emission, and a public cross-module embed.
- `int/surface/embed-freestanding`: raw flat-image execution with a byte read from the embedded file.
- A format-neutral `ObjectImage` dedup test, including content, section, kind, alignment, ordinary-global, and zero-length cases.
- Driver tests for annotation rules, nested rejection behavior, unreadable paths, asset invalidation, exact warm `Q_LOWER`/buffer reuse, and same- versus different-length edits.
- Parser coverage for `[_]T`, `[N]T`, and `_` as an ordinary expression.
- Sema length-state coverage pins `u32::MAX` and `u32::MAX - 1` so legal file sizes cannot collide with compiler state sentinels.

Closes #2507


octalide:
Independent final review at `4f90350d`: approve-quality; no remaining blocker.

The review found and fixed two substantive issues before approval: full legal `u32` embed lengths collided with sema state sentinels (`83d9702b`), and lowering could silently substitute zero-filled storage after cache/path/allocation incoherence (`3681ada7`). A second independent pass verified cache ownership and teardown order, incremental invalidation, public cross-module use, empty embeds, exact per-module placement compatibility, malformed/error cases, and the explicit #2541 cross-module dedup boundary.

Repository parity was also completed in `4f90350d`: `.github/skills/mach/SKILL.md` now documents `#[embed]` and the inferred `[_]u8` form alongside the authoritative language reference.

Final evidence on current `dev`:

- debug and release self-host fixpoints converge byte-identically
- 1,095/1,095 tests pass with each reviewed fixpoint compiler
- embed and embed-freestanding integrations pass in debug and release
- all 16 substantive CI checks pass at the final head, including Linux, arm64, riscv64, and Windows integrations
- the expected Darwin release gate is skipped by policy for a PR targeting `dev`
- worktree and diff check are clean

No unresolved PR conversation, CI annotation, or review finding remains.

Issue #2520, opened 2026-08-05
link: mis-relocates foreign C objects — links cleanly, crashes at runtime


octalide:
mach mis-relocates vendored C objects on linux: an object that runs correctly when linked by `cc` crashes when linked by mach.

## Repro

In briar-systems/mach-audio (`vendor/mad.c`, miniaudio v0.11.25), build and run the `play` example against any WAV. Decode and mix succeed; the process segfaults in the device-open path.

The decisive comparison, run by the reporter: the **identical** `miniaudio.o` linked by `cc` into a small pure-C harness opens the device, reports 44100 Hz / 2 ch and starts cleanly. Same object, same machine, same environment — works under `cc`, crashes under mach. The difference is the linker, not the C.

## Expected

A C object linked by mach behaves as it does when linked by `cc`.

## Actual

Segfault in glibc `__strlen_avx2`, called from `_IO_vfprintf`, reached from miniaudio's `ma_log_postf("Loading library: %s\n", ...)` inside `ma_dlopen` — before any backend `.so` is opened. `%rdi` holds x86 instruction bytes rather than a pointer, i.e. a relocated reference resolves into code instead of to the intended string datum.

## Environment

linux x86_64, host glibc. Reproduces under mach 4.3.5, 4.4.0 and 4.5.0 — not a recent regression. `mach test .` in mach-audio passes 36/36 because no test exercises the shim.

## Why this is critical

Static linking of vendored C archives is the foundation of the ecosystem's cross-platform distribution plan (mach-glfw#31, mach-audio#11, later MoltenVK). Any consumer linking a real C library gets a binary that links cleanly and then crashes at runtime, which reads as the library's bug rather than the toolchain's. It also means mach-audio's shipped README claim of end-to-end linux playback does not currently hold.

## Investigation notes

Start by diffing what mach and `cc` produce from the same object: which relocation types appear in `miniaudio.o`, which of them mach applies, and where the resolved value diverges. Suspect the relocation kinds a large real-world C object uses that mach's own emitted code never exercises — string/rodata references through GOT-relative and PC-relative forms, section-relative addends, and `.rodata.str*` mergeable-section handling are the places to look first. A wrong addend or a section-index mismatch on a mergeable string section fits the observed symptom exactly.

Fix the relocation handling properly; do not special-case miniaudio.

## Acceptance

- The `play` example in mach-audio runs and produces audio when linked by mach on linux.
[...]

PR #2531, opened 2026-08-05
fix(elf): admit x86_64's SHT_X86_64_UNWIND .eh_frame in the object reader

### LINKED ISSUE #2516: link: the ELF reader drops clang's SHT_X86_64_UNWIND .eh_frame, so every clang-emitted object fails to link
The ELF object reader drops any section whose `sh_type` is not `SHT_PROGBITS` or `SHT_NOBITS`, so a clang-emitted `.eh_frame` — which carries the architecture-specific type `SHT_X86_64_UNWIND` — is skipped, and the `.rela.eh_frame` entry that targets it resolves to no section. Linking any clang-compiled x86-64 object fails; the gcc-compiled equivalent of the identical source links fine.

Found while authoring `doc/distribution.md` (#2511), whose cross-building section needs a vendored-C archive built per target. `clang --target=` is the natural one-compiler answer for emitting ELF, COFF, and Mach-O from one host, and it is exactly the case that fails.

## Repro

```sh
$ printf 'int qz_answer(void) { return 42; }\n' > qz.c
$ gcc   -c -O2 -o gcc.o   qz.c
$ clang -c -O2 -o clang.o qz.c
```

A project whose entry declares `ext fun qz_answer() i32;` and returns it:

```
$ mach build . --profile release gcc.o
0 errors / 0 warnings

$ mach build . --profile release clang.o
error: linker: relocation names an out-of-range section
```

The same holds when the object is wrapped in an `ar` archive, so it reaches a real build through an ordinary `[link.X]` `local` entry and a `[step.X]` that runs clang.

## Expected

Both objects link. `SHT_X86_64_UNWIND` is valid ELF for `.eh_frame` on x86-64, and clang has emitted it for years; nothing about the object is malformed.

## Actual

`error: linker: relocation names an out-of-range section`, from `src/lang/be/linker.mach:2944`.

The two objects differ in exactly one relevant way — the `sh_type` of `.eh_frame`:

```
gcc.o:   .eh_frame sh_type = 0x00000001   (SHT_PROGBITS)
clang.o: .eh_frame sh_type = 0x70000001   (SHT_X86_64_UNWIND)
```

Both objects carry a `.rela.eh_frame` whose `sh_info` names that section (`sh_info=7` for gcc, `sh_info=5` for clang).

The chain, in `src/lang/target/of/elf.mach`:

- the counting pass (`:3965`) and the populate pass (`:4021`) both admit a section only when `ty == SHT_PROGBITS || ty == SHT_NOBITS`;
- `sec_map[sh]` is therefore left at its `-1` initialiser (`:4018`) for clang's `.eh_frame`;
- the relocation pairing (`:4149`) reads `target_sec = sec_map[sh_info]` and takes that `-1`;
- the linker's bounds check (`src/lang/be/linker.mach:2944`) sees the resulting out-of-range index and rejects the object.

The file declares no constant for `0x70000001`; `SHT_PROGBITS` and `SHT_NOBITS` are at `:61` and `:67`.

Confirmed by elimination: `clang -fno-asynchronous-unwind-tables` (no `.eh_frame` at all) links; stripping `.llvm_addrsig` or passing `-fno-addrsig` does not help, ruling out the other section clang emits that gcc does not.

## Fix

Admit `SHT_X86_64_UNWIND` (`0x70000001`) alongside `SHT_PROGBITS` in the reader's section predicate, in both the counting and populate passes so the allocation and the fill stay in agreement. It carries bytes exactly like `SHT_PROGBITS` and its `SHF_ALLOC`/`SHF_EXECINSTR`/`SHF_WRITE` flags classify through the existing `kind` ladder unchanged.

Worth deciding at the same time whether an *unrecognised* section type that a relocation targets should stay a silent skip that surfaces later as a bounds failure, or be rejected where the mapping is lost, naming the section and its type. The current diagnostic describes a corrupt object, which this object is not, and it names neither the section nor the input file — a reader hitting it has no path to the cause.

## Acceptance

- A clang-compiled x86-64 object and archive link, and the program observes the C function's return value.
- The gcc path is unchanged.
- A surface case covering a clang-emitted object, so the reader's section-type set is pinned rather than tracked by whichever compiler CI happens to have.

## Environment

- triple: `x86_64-linux`, `sysv64`
- toolchain: `mach 4.5.0` built from `dev` at `08cc7178`
- OS: Linux 7.1.5 (CachyOS); gcc 16.1.1, clang 22.1.8

octalide:
## Summary

clang/lld type `.eh_frame` `SHT_X86_64_UNWIND` (`SHT_LOPROC + 1` = `0x70000001`) rather than gcc/gas's `SHT_PROGBITS`. The ELF reader's section-type admission predicate (`parse_object`'s counting and populate passes) admitted only `SHT_PROGBITS`/`SHT_NOBITS`, so a clang-emitted `.eh_frame` was silently dropped, leaving its `.rela.eh_frame` relocation pointing at no section — the linker then rejected the object with `relocation names an out-of-range section`.

- Root cause confirmed independently (not just from the filed diagnosis): compiled the repro's `qz.c` with both gcc and clang and inspected section headers with `readelf` — `.eh_frame` is `PROGBITS` for gcc, `X86_64_UNWIND` for clang.
- Fix: a new `section_has_bytes(ty, machine)` predicate replaces the two `ty == SHT_PROGBITS || ty == SHT_NOBITS` admission sites. It admits `SHT_PROGBITS`/`SHT_NOBITS` unconditionally, and admits a processor-specific type (`SHT_LOPROC..SHT_HIPROC`) only through an explicit per-`e_machine` allowlist — currently just `SHT_X86_64_UNWIND` on `x86_64` — never the whole range.
- Checked the other two ISAs mach targets for an analogous rejection: cross-compiled the repro with `clang --target=aarch64-linux-gnu` and `--target=riscv64-linux-gnu`. Both type `.eh_frame` as plain `SHT_PROGBITS` — no fix needed there. riscv64's one real processor-specific section, `SHT_RISCV_ATTRIBUTES` (`.riscv.attributes`), is non-`SHF_ALLOC` structured attribute data (not PROGBITS-equivalent program bytes) and nothing relocates into it, so it correctly stays unadmitted — it's unrelated to this bug and out of scope.
- Regression test in the existing `parse_object` unit-test style: builds a synthetic object with an `.eh_frame`-named section, retypes it to `SHT_X86_64_UNWIND` post-emit (mirroring what clang actually produces), and asserts the section survives and its relocation resolves to the real section rather than the `SECTION_EXTERNAL` sentinel a lost `sec_map` entry produced before the fix.

## Verification

- `mach build . -o a && ./a build . -o b && ./b build . -o c && cmp b c` — fixpoint converges (matches CI's `build-linux` job).
- `./c test .` — 1081 passed, 0 failed (including the new test; `mach.lang.target.of.elf` went from 27 to 28).
- End-to-end, using the fixed compiler against a minimal project (`ext fun qz_answer() i32;`):
  - unfixed installed `mach` (4.4.0): `clang.o` reproduces `error: linker: relocation names an out-of-range section`; `clang.a` (archive-wrapped) reproduces the same.
  - fixed compiler: `clang.o`, `clang.a`, and `gcc.o` all link and the program exits with `42` (the C function's return value) in every case.

Closes #2516

### DIFF
--- a/src/lang/target/of/elf.mach
+++ b/src/lang/target/of/elf.mach
@@ -4603,6 +4603,110 @@ test "mach.lang.target.of.elf.parse_object:unknown_reloc_type" {
     ret 0;
 }
 
+# clang/lld type `.eh_frame` SHT_X86_64_UNWIND (0x70000001) rather than gcc/gas's
+# SHT_PROGBITS. before the fix the reader admitted only SHT_PROGBITS/SHT_NOBITS,
+# so the section was dropped and `.rela.eh_frame`'s relocation resolved to
+# SECTION_EXTERNAL instead of the real section - which the linker then rejected
+# as an out-of-range section (#2516)
+test "mach.lang.target.of.elf.parse_object:x86_64_unwind_section_admitted" {
+    var alloc: A.Allocator;
+    page.make(?alloc);
+    var i: intern.Interner = intern.init(?alloc);
+
+    var of_reg: of.OfRegistry = of.registry_init();
+    val rr: R.Result[bool, str] = register(?of_reg);
+    if (R.is_err[bool, str](rr)) { ret 1; }
+
+    var seam: isa.RelocSeam = t_seam(t_x64_elf_reloc_type::*u8);
+    var isa_vt: isa.IsaVTable = t_isa(isa.ARCH_X86_64, EM_X86_64::u32, ?seam);
+
+    var text_bytes: [16]u8;
+    var k: usize = 0;
+    for (k < 16) { text_bytes[k] = 0x90; k = k + 1; }
+
+    var eh_bytes: [16]u8;
+    k = 0;
+    for (k < 16) { eh_bytes[k] = 0; k = k + 1; }
+
+    var secs: [2]of.Section;
+    secs[0].name = t_intern(?i, ".text"); secs[0].kind = of.SK_TEXT;
+    secs[0].bytes = ?text_bytes[0]; secs[0].len = 16; secs[0].align = 16;
+    secs[1].name = t_intern(?i, ".eh_frame"); secs[1].kind = of.SK_RODATA;
+    secs[1].bytes = ?eh_bytes[0]; secs[1].len = 16; secs[1].align = 8;
+
+    var syms: [1]of.Symbol;
+    syms[0].name = t_intern(?i, "qz_answer"); syms[0].section = 0; syms[0].offset = 0;
+    syms[0].size = 16; syms[0].flags = of.SYM_OBJ_FLAG_GLOBAL | of.SYM_OBJ_FLAG_FUNCTION;
+
+    # the FDE's initial-location field: PC32 against the function, the shape
+    # clang/gcc both emit into `.rela.eh_frame`
+    var relocs: [1]of.Relocation;
+    relocs[0].section = 1; relocs[0].offset = 8; relocs[0].kind = of.RK_PC32;
+    relocs[0].symbol = syms[0].name; relocs[0].addend = 0;
+
+    var img: of.ObjectImage;
+    img.alloc = ?alloc;
+    img.interner = ?i;
+    img.name = t_intern(?i, "test");
+    img.sections = ?secs[0]; img.section_count = 2;
+    img.symbols = ?syms[0]; img.symbol_count = 1;
+    img.relocations = ?relocs[0]; img.reloc_count = 1;
+
+    val tf_r: R.Result[fs.TempFile, str] = fs.temp_create(?alloc, "elf_unwind");
+    if (R.is_err[fs.TempFile, str](tf_r)) { ret 1; }
+    var tf: fs.TempFile = R.unwrap_ok[fs.TempFile, str](tf_r);
+    val tpath: str = fs.temp_path(?tf);
+
+    val em: R.Result[bool, str] = emit_object(?isa_vt, ?img, tpath::*u8);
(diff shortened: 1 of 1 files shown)


octalide:
Scope note, verified with a proper matrix rather than assumption:

- This fix resolves the **link-time** rejection reported in #2516 (`relocation names an out-of-range section` on a clang-typed `.eh_frame`). Confirmed across all four combinations of {bare CLI input, `[link.X]` local + `[step.X]` manifest path} x {loose `.o`, `ar` archive}: mine directly (CLI, both loose object and archive), @audio-backends against a real 767KB vendored object (mach-audio's miniaudio shim) through the manifest path with a loose object, and I closed the last untested cell myself afterward (manifest path + archive, via a `[step.X]` that runs `clang -c` then `ar rcs`) - baseline reproduces #2516's exact error, this branch builds and the program runs correctly.
- It does **not** address runtime correctness of the resulting binary. #2520 (relocations against section symbols resolving against the wrong output section base — see [the corrected root cause](https://github.com/briar-systems/mach/issues/2520#issuecomment-5188359743)) is a separate, pre-existing bug. Before this fix, a clang object hit #2516's link error and #2520 was never reached; after this fix, clang objects link successfully but can still surface #2520 as a runtime crash on affected relocations — GCC objects already exhibited this independently of #2516.

So: a runtime crash (e.g. exit 139) on a clang-linked binary built from this branch is expected until #2520 lands - it is not a regression introduced here. Leaving this comment so it doesn't get misread against this PR's diff.

PR #2533, opened 2026-08-05
fix(link): link foreign clang objects on Mach-O

### LINKED ISSUE #2521: link: foreign clang objects fail on Mach-O (section-relative relocs, missing _ prefix on ext refs)
Linking a foreign (clang-produced) object into a Mach-O target fails two independent ways, both blocking vendored C on darwin.

## What

Reported by an agent cross-linking a stub C object for darwin from linux.

**1. Section-relative relocations are unsupported.** Any normal clang object — one carrying debug info or `__compact_unwind`, i.e. essentially every object produced without `-g0 -fno-unwind-tables` — fails with `macho: section-relative relocation unsupported`. Mach-O `X86_64_RELOC_*`/`ARM64_RELOC_*` entries can reference a section rather than a symbol (`r_extern = 0`), and mach's reader/applier does not handle that form.

**2. `ext` references are emitted without the Darwin underscore prefix.** mach underscores its own Mach-O symbols (e.g. `_M5audio6deviceN4open`) but emits `ext fun` references unprefixed, so a reference to a C function never matches the `_`-prefixed name the C object actually defines: `undefined symbol: mad_device_open` while the object defines `_mad_device_open`.

Item 2 is a plain defect in the platform symbol-naming convention. Item 1 is missing capability in the Mach-O relocation path. Both must be fixed for any vendored C to link on darwin; fix them properly rather than requiring consumers to compile with `-g0 -fno-unwind-tables`.

## Scope note

This issue covers linking a foreign object into a Mach-O target. It does NOT cover *producing* that object: cross-compiling C against Apple frameworks from linux is impossible (zig bundles libc headers, not frameworks; the Apple SDK is not redistributable), so framework-dependent objects must be built on macOS regardless. Non-framework C is cross-compilable from linux and is the near-term test path here.

## Acceptance

- A clang object compiled normally (with debug info and unwind tables) links into a Mach-O executable.
- An `ext fun` reference resolves against the `_`-prefixed symbol a C object defines, and mach's own symbol naming is unchanged.
- Regression tests in the object-format suite cover both, cross-linked from linux so CI exercises them without a mac runner.
- `doc/` updated if the underscore convention is documented anywhere that is now wrong.

octalide:
Closes #2521

Two independent defects blocked linking a normally-compiled clang object into a Mach-O target. Both reproduced first with a stub C object cross-compiled from linux (`clang -target x86_64-apple-macos11 -c shim.c -o shim.o -g -O1`), then fixed.

## 1. Section-based relocations (`macho: section-relative relocation unsupported`)

A Mach-O relocation entry with `r_extern = 0` names a **section** rather than a symbol, and its in-place field carries the target's address in the object's own address space instead of a plain addend. Every normally-compiled clang object uses that form — `__compact_unwind` and every `__debug_*` reference into `__text`, `__const` and `__bss` — so parsing one failed outright and only `-g0 -fno-unwind-tables` objects got through.

The parser now gives each named section one synthetic local anchor symbol at its offset zero, named by the section's unique `<segname>,<sectname>` identity, and resolves the entry through that anchor with the target's in-section offset as the addend. It is the same shape COFF's full-section weak-carrier anchor already takes, so nothing outside `macho.mach` learns a second target form. Only sections a section-based entry actually names get an anchor, so an object without one grows no symbols.

Addend recovery is exact per form: an absolute field holds the whole target address, an x86_64 pc-relative field holds a next-instruction-relative displacement (the patch site's own object address is added back and the abstract `- 4` convention reapplied). An arm64 page / page-offset / branch field carries only a fragment of an address packed into an instruction immediate — the symbol-based form takes its addend from a paired `ARM64_RELOC_ADDEND`, not the field — so a section-based entry of those kinds is **rejected by name** rather than guessed at. No compiler emits one.

## 2. Darwin C symbol prefix (`undefined symbol: mad_device_open`)

An `ext fun` / `ext val` names a C declaration, so its link name is whatever the platform's C toolchain emits for that identifier. Darwin's C ABI prefixes an underscore, so `ext fun mad_device_open` has to bind the `_mad_device_open` a C object defines; mach emitted the bare name and the reference never resolved.

The prefix is a new `OsVTable` datum (`c_symbol_prefix`: `"_"` on darwin, `""` elsewhere), applied where the `ext` default link name is recorded, so a definition and every reference read the one recorded name and no source has to spell a platform's convention by hand.

`#[symbol("...")]` deliberately does **not** route through it. That decorator names the object symbol *literally* — which is what lets the runtime's inline asm reach `_rt_argc` and the OS entry point — and mach-std already spells those per platform by hand (`_start` on linux, `start` on darwin). Confirmed: a darwin binary with no `ext` declaration is **byte-for-byte identical** before and after this change.

## Verification

- Both defects reproduced against `dev` before the fix; the two new int cases fail on the unfixed compiler with `error: macho: section-relative relocation unsupported`.
- The resolved values are checked, not just the absence of an error: in a linked darwin binary, the section-based reference into `__const` resolves to the exact final VA of the `"hello"` string (located independently by searching for the string bytes), and both `__compact_unwind` entries land exactly on the `push rbp` opening each C function with their spacing preserved.
- `int/surface/macho-foreign-{x86_64,aarch64}` cross-link a checked-in clang object (debug info and unwind tables intact) into a darwin binary **on the linux leg**, so CI covers this without a mac runner. `shim.c` sits beside each `shim.o` with the exact regeneration command.
- Three new object-format unit tests cover the translation precisely (anchor identity/shape, absolute and pc-relative addend recovery on x86_64, the arm64 anchor path, the arm64 page-form rejection, and an out-of-range section index), plus one pinning each OS's declared `c_symbol_prefix`.
- `1090 passed, 0 failed`; full int suite green on linux; self-host fixpoint converges (`b == c`). All re-run after merging current `dev`.

## Out of scope, filed separately

#2540 — linking a `-g` clang object into a darwin binary built with `debug = true` produces DWARF `llvm-dwarfdump --verify` rejects. Mach-O DWARF cross-section references (`DW_FORM_strp`, `DW_AT_stmt_list`, abbrev offsets) are unrelocated absolute offsets, so the name-keyed debug merge shifts them and nothing remaps them. Distinct from either defect here: the link succeeds and the code and data relocations resolve correctly. The same binary without the foreign object verifies `No errors.`


octalide:
#2537 has merged, and this branch is now reconciled with its symbol-index relocation contract in merge commit `b15379c4`. The sole Mach-O conflict now maps external entries to their native symbol index and section-relative entries to the appended synthetic-anchor index; the section-relative addend recovery remains unchanged. I also migrated #2533's focused object tests to assert through `Relocation.sym` rather than the removed name field.

Fresh verification after the merge:

- debug self-host fixpoint byte-identical
- release self-host fixpoint byte-identical
- 1,092/1,092 in-source tests in both debug and release
- both x86_64 and AArch64 foreign Mach-O integration cases pass in debug and release with each fixpoint compiler
- target matrix: 270 case × leg pairs valid
- worktree clean

Fresh GitHub CI is now running. The PR remains unmerged pending an independent post-reconciliation review.

octalide:
Independent review at `b15379c471e79f0cc0ee3c6fc3179034f5ac983c`: approve-quality; I found no blocking issue and made no code changes.

I specifically re-audited the post-#2537 relocation contract. External Mach-O relocations retain the native `symnum` as the owning-image symbol index; section-relative relocations target `nsyms + sec_anchor[target_section]`; executable linking resolves those local indices directly; and relocatable carry remaps them through `sym_out`. The reconciliation preserves the intended semantics.

I also independently verified the emitted artifacts rather than relying only on successful builds:

- Mach-produced caller objects reference `_mad_device_open` on both x86_64 and AArch64 Darwin targets.
- The x86_64 linked `__debug_info` relocation resolves exactly to the mapped `hello` address (`0x1000099b6`), and AArch64 resolves exactly to `0x1000109b6`.
- The first x86_64 compact-unwind function pointers resolve to `0x1000072d0` / `0x1000072f0`, and the AArch64 pointers to `0x10000a750` / `0x10000a770`; each maps to the expected function instructions.

Fresh verification with a compiler built from this head:

- `m test . --jobs 8 -O0`: 1092 passed, 0 failed
- `m test . --jobs 8 -O2`: 1092 passed, 0 failed
- full Linux integration suite: passed, including both foreign Mach-O fixtures in debug and release
- target matrix: all 270 case/leg pairs passed
- all 16 substantive GitHub checks are green (the Darwin release gate is skipped as configured)

One adjacent, pre-existing capability gap remains: clang can emit x86_64 `SIGNED_1`, `SIGNED_2`, and `SIGNED_4` relocation forms, which are still rejected. That is not one of #2521's two defects and does not invalidate this PR's acceptance coverage, so I filed #2546 with a minimal real-world reproducer and dispositioned it as follow-up work after #2533 rather than expanding this PR.

Issue #2535, opened 2026-08-05
link: x64 has no addr32nb applier, so no COFF object with .pdata links


octalide:
The x86-64 relocation applier has no arm for `RK_ADDR32NB`, so linking a COFF object that carries `.pdata` unwind data fails with `error: relocation 'addr32nb' not supported`. Every x64 Windows compiler emits `.pdata` for any function that needs unwind info, so in practice **no vendored C compiled for windows links** unless its unwind tables are suppressed.

Found while authoring `doc/distribution.md` (#2511). The page's windows story is "vendor your C, ship DLLs beside the exe", and it only holds today because the sample C is a leaf function returning a constant — which emits no `.pdata` and so never exercises the missing arm.

## Repro

```c
/* declared locally so the case needs no Windows SDK */
void __stdcall Sleep(unsigned long ms);

int qz_answer(void) {
    Sleep(0);          /* a call is enough: the frame now needs unwind info */
    return 42;
}
```

```sh
$ clang --target=x86_64-pc-windows-msvc -c -O2 -o qz.o qz.c
$ ar rcs libqz.a qz.o
```

Link it into a windows artifact through an ordinary `[link.X]` `local` entry:

```
$ mach build . --profile release --target windows-x86_64
error: relocation 'addr32nb' not supported
```

The object is unremarkable — this is the standard x64 unwind function table:

```
$ llvm-objdump -r qz.o
RELOCATION RECORDS FOR [.text]:
0000000000000007 IMAGE_REL_AMD64_REL32    Sleep

RELOCATION RECORDS FOR [.pdata]:
0000000000000000 IMAGE_REL_AMD64_ADDR32NB .text
0000000000000004 IMAGE_REL_AMD64_ADDR32NB .text
0000000000000008 IMAGE_REL_AMD64_ADDR32NB .xdata
```

Three image-relative fields: the function's begin RVA, its end RVA, and the RVA of its `.xdata` unwind info.

Passing `-fno-unwind-tables -fno-asynchronous-unwind-tables` drops `.pdata` and the same project links and produces a correct import directory, which confirms `.pdata` is the whole trigger. That is not a fix — it discards the unwind data the Windows ABI expects, breaking SEH and debugger backtraces through the vendored frames.

## Expected

A COFF object with `.pdata` links, and its unwind entries are relocated into the image's exception directory.

## Actual

`error: relocation 'addr32nb' not supported`, from `src/lang/be/linker.mach:4043`.

[...]

PR #2537, opened 2026-08-05
fix(link): key relocations by symbol index, not by symbol name

### LINKED ISSUE #2520: link: mis-relocates foreign C objects — links cleanly, crashes at runtime

octalide:
Closes #2520

## Root cause

`of.Relocation` identified its target symbol by interned name. ELF gives every `STT_SECTION` symbol `st_name = 0`, so all section symbols in an object intern to the same empty string. The linker's local-symbol index kept the first definition of a name, causing every section-relative relocation to resolve against whichever anonymous section symbol appeared first, normally `.text`.

The real miniaudio object contains 934 such relocations and eight anonymous section symbols. Its backend-name table therefore resolved into executable bytes; the resulting invalid `const char *` caused the observed crash in `__strlen_avx2`.

The reduced cases confirmed the same failure across:

- `R_X86_64_32S` against `.rodata`
- `R_X86_64_PC32` against `.rodata` / `.data.rel.ro.local`
- `R_X86_64_64` against `.rodata.str1.1`

Named-symbol controls using the same relocation forms remained correct.

## Fix

Relocations now bind their target by index into their owning image's symbol array, matching the native representation in ELF `r_info`, COFF `SymbolTableIndex`, and Mach-O `r_symbolnum`:

```text
symbol: intern.StrId  ->  sym: u32
```

Local targets resolve directly from their indexed definition. Global and undefined targets continue to resolve by name through the cross-object symbol table.

The representation change is format-generic and propagated through:

- ELF, COFF, and Mach-O readers and writers
- codegen and DWARF relocation producers
- the generic linker, liveness analysis, and relocatable-output carry path
- COFF call-target inference and weak-section splitting
- parallel-codegen image rehoming

Codegen records relocations by name only while its symbol table is incomplete, then binds every record to a symbol index in one pass. This also declares genuinely undefined targets and replaces the two older extern-backfill sweeps.

## Regression coverage

- A flat-link test carries two anonymous ELF-style section symbols with the same empty name and compares the section-relative result against a named anchor at the exact same byte.
- A PE/COFF-path test carries two distinct `.rdata` section symbols with the same name. Four ABS64 sites prove each section-symbol target equals its own named anchor while the two target sections remain distinct. Reintroducing the old name-keyed lookup makes this test fail.
- Existing ELF, COFF, and Mach-O emit/parse tests now exercise indexed relocation targets.

`IMAGE_REL_AMD64_ADDR32NB` application and exact RVA assertions remain scoped to #2535. COFF already parses and emits that kind, but the x86-64 applier does not yet implement `S + A - ImageBase`; this PR deliberately does not fold #2535 into #2520.

## Verification

Local verification after merging current `dev`:

- Original isolated #2520 acceptance gate: both PIC and non-PIC reduced probes pass exact-value assertions across 32S, PC32, and ABS64; the named-symbol regression controls also pass.
- mach-audio `play` builds and runs to completion against the real vendored miniaudio object.
- mach-audio: 36 passed, 0 failed.
- Debug self-host fixpoint: byte-identical `b == c`.
- In-source tests: 1088 passed, 0 failed.
- Release self-host fixpoint: byte-identical `b == c`; focused release-built regression passes.
- Incremental determinism: clean, warm no-op, and post-edit builds match.
- Integration target matrix: 268 case/leg pairs validated.
- Linux integration lane passes.
- RISC-V integration lane passes; two local dependency-clone races were rerun individually and passed in both debug and release.

### DIFF
--- a/src/lang/be/codegen.mach
+++ b/src/lang/be/codegen.mach
@@ -329,37 +329,36 @@ fun pack_image(s: *session.Session, tgt: *target.Target, irmod: *ir.Module,
         ret R.err[of.ObjectImage, str](R.unwrap_err[bool, str](res_symbols));
     }
 
-    val res_relocs: R.Result[bool, str] = pack_relocs(?image, enc, text_sec, placements, placement_count);
+    # relocations are recorded by target name and bound to symbol indices only
+    # after pack_globals, because a global's aggregate initializer may point at a
+    # global that pass has not packed yet. binding also declares any target no
+    # pass defined, which is what makes it genuinely external.
+    var deferred: obj.DeferredRelocs = obj.deferred_init(s.alloc);
+
+    val res_relocs: R.Result[bool, str] = pack_relocs(?deferred, enc, text_sec, placements, placement_count);
     # re-home the retained PC<->source rows the same way symbols are, while the
     # placements are still live, so the #1699 line producer reads final section
     # offsets. infallible (in-place offset math), so no error to thread.
     pack_rows(enc, placements, placement_count, text_sec);
     if (placements != nil) { A.deallocate[SectionPlacement](s.alloc, placements, enc.symbol_count::usize); }
     if (R.is_err[bool, str](res_relocs)) {
+        obj.deferred_dnit(?deferred);
         obj.dnit(?image);
         ret R.err[of.ObjectImage, str](R.unwrap_err[bool, str](res_relocs));
     }
 
-    val res_globals: R.Result[bool, str] = pack_globals(s, tgt, ?image, irmod);
+    val res_globals: R.Result[bool, str] = pack_globals(s, tgt, ?image, ?deferred, irmod);
     if (R.is_err[bool, str](res_globals)) {
+        obj.deferred_dnit(?deferred);
         obj.dnit(?image);
         ret R.err[of.ObjectImage, str](R.unwrap_err[bool, str](res_globals));
     }
 
-    val res_reloc_externs: R.Result[bool, str] = pack_reloc_externs(?image, enc);
-    if (R.is_err[bool, str](res_reloc_externs)) {
+    val res_bind: R.Result[bool, str] = obj.flush_deferred(?image, ?deferred);
+    obj.deferred_dnit(?deferred);
+    if (R.is_err[bool, str](res_bind)) {
         obj.dnit(?image);
-        ret R.err[of.ObjectImage, str](R.unwrap_err[bool, str](res_reloc_externs));
-    }
-
-    # an aggregate constant's `&G` / function-pointer leaf may target a symbol
-    # defined in another module (cross-object reference). every such data-section
-    # relocation target that this object has not defined is declared undefined so
-    # the linker binds it - the data analog of pack_reloc_externs.
-    val res_data_externs: R.Result[bool, str] = pack_data_reloc_externs(?image);
-    if (R.is_err[bool, str](res_data_externs)) {
-        obj.dnit(?image);
-        ret R.err[of.ObjectImage, str](R.unwrap_err[bool, str](res_data_externs));
+        ret R.err[of.ObjectImage, str](R.unwrap_err[bool, str](res_bind));
     }
 
     # when `-g` is on and the target resolved a debug-info producer (elf / mach-o ->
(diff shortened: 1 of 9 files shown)


octalide:
Independent review verdict: approve-quality; no blocking correctness findings.

Evidence reviewed and reproduced:

- Audited the full representation migration from ObjectImage producers through parallel rehoming, ELF/COFF/Mach-O parsing and writing, atom liveness, final patching, COFF weak-COMDAT splitting, merged-image remapping, and relocatable carry. Each relocation index remains scoped to its owning symbol array and is explicitly remapped whenever that array changes. Invalid input indexes are rejected before dereference.
- Debug and release self-host builds both converged byte-for-byte. Full in-source tests passed in each mode: 1088 passed, 0 failed.
- Focused anonymous ELF section-symbol and duplicate COFF section-symbol exact-value tests passed in debug and release. ELF and COFF object emit/parse and malformed-target paths passed as part of the full suite.
- The preserved #2520 gate passed exact values for non-PIC R_X86_64_32S + R_X86_64_64 and PIC R_X86_64_PC32 + R_X86_64_64, including the named-symbol controls: 18 / 8 / 115 / 100 in both modes.
- The real vendored miniaudio object built and the play example ran to completion; mach-audio passed 36/36.
- Incremental determinism passed. Native integration cases affected by one accidental overlapping-run dependency-clone race all passed when rerun serially. The authoritative GitHub Linux, ARM64, RISC-V, and Windows integration lanes are green.
- GitHub had no reviews, issue comments, or inline comments at the reviewed d747fe00 head; every substantive CI job was green. The current 71df9c69 head adds only corrected comments describing indexed relocation binding and passed a byte-identical compiler build, both focused regressions, and the full 1088-test suite locally.

Merge-order note for #2533: merge #2537 first. Both branches touch Mach-O populate_relocs and merge-tree reports one explicit content conflict. #2533 must preserve the new index contract by setting the relocation target to symnum for r_extern entries and nsyms + sec_anchor[target_section] for synthetic section anchors. The conflict is loud rather than silently miscompiling, but #2533 must rerun its x86_64/aarch64 foreign-object exact-address tests after reconciliation.

Residual scope: #2535 still owns IMAGE_REL_AMD64_ADDR32NB application and exception-directory integration; this PR correctly preserves and distinguishes COFF section-symbol identity without expanding into that relocation-kind work. Darwin section-relative relocations remain #2533.

PR #2542, opened 2026-08-05
feat(link): read COFF short import records from import libraries

### LINKED ISSUE #2525: link: read COFF short import records so mingw/LLVM import libraries can be consumed
## What

An `llvm-dlltool`-generated import library fails with `error: coff: bigobj object too small`. Root cause is misclassification, not a bigobj defect: a COFF short import record and `ANON_OBJECT_HEADER_BIGOBJ` share the same 4-byte magic (Sig1=0x0000, Sig2=0xFFFF). The member's first 8 bytes are `0000 ffff 0000 8664` and it is 45 bytes, so mach classifies it as bigobj then rejects it as too small. mach cannot read the short import record at all.

`coff.mach` already has `bigobj_class_id_matches()` to discriminate — bigobj carries a ClassID GUID, a short import record does not. A short import record carries (symbol name, DLL name, ordinal/hint, import type), so supporting it lets mach consume standard mingw/LLVM import libraries.

This is the scalable successor to hand-listing symbols in the `symbols = [...]` manifest key added by #2510, and the recommended starting point of the PE foreign-object work.

## Acceptance

- An llvm-dlltool import library is consumed and its symbols bind to the correct DLL
- Classification distinguishes short import records from bigobj via the ClassID discriminator
- Regression test cross-linked from linux

octalide:
## What

Mach misclassified COFF short import records as bigobj files and could not consume standard LLVM or MinGW import libraries. This branch parses those records as authoritative symbol-to-DLL declarations and feeds their dependencies and attribution into the existing dynamic-link plan. Consumers need no source decorator or manifest symbol claim.

Bigobj detection now uses its complete version-plus-ClassID discriminator. Short import hint and ordinal metadata is preserved through DynamicInfo, including PE import-by-ordinal thunks.

## Design notes

- Import-library members can introduce DLL dependencies even though those DLLs are not explicit link inputs.
- Traditional all-idata bookkeeping members are ignored because Mach synthesizes the PE import directory from the dynamic plan.
- A short import declaration that contradicts source or manifest attribution is a hard error.
- ADDR32NB support for ordinary object relocations remains separately tracked by #2535.

## Verification

- Real llvm-dlltool archive linked from Linux; llvm-readobj reports Sleep and ExitProcess bound to kernel32.dll.
- Hermetic integration case covers two DLLs, by-name imports, and an import by ordinal.
- 1088 unit tests pass.
- Full Linux integration lane passes all 127 cases.
- Target matrix validates 269 case/leg pairs.
- Debug self-host fixpoint is byte-identical.

Closes #2525

### DIFF
--- a/int/lib/produce.sh
+++ b/int/lib/produce.sh
@@ -170,7 +170,7 @@ produce_panic_exit() {
 }
 
 # read_le_uint <file> <offset> <size>
-# print the unsigned little-endian integer of <size> bytes (2 or 4) at <offset>.
+# print the unsigned little-endian integer of <size> bytes (1, 2, or 4) at <offset>.
 # od reads in host byte order; every CI runner is little-endian.
 read_le_uint() {
     dd if="$1" bs=1 skip="$2" count="$3" 2>/dev/null | od -An -tu"$3" | tr -d ' \n'
@@ -290,13 +290,19 @@ produce_pe_imports() {
 
         t=0
         while :; do
-            entry=$(read_le_uint "$bin" $((thunk_off + t * 8)) 8)
-            [ "$entry" -eq 0 ] && break
-            # bit 63 set marks an ordinal import, which names no symbol
-            if [ $((entry >> 63)) -eq 1 ]; then
-                echo "$dll:#$((entry & 0xFFFF))" >>"$out"
+            thunk=$((thunk_off + t * 8))
+            entry_lo=$(read_le_uint "$bin" "$thunk" 4)
+            entry_hi=$(read_le_uint "$bin" $((thunk + 4)) 4)
+            [ "$entry_lo" -eq 0 ] && [ "$entry_hi" -eq 0 ] && break
+            # Inspect bit 63 as a byte. A by-ordinal thunk is an unsigned u64
+            # above the shell's signed arithmetic range, so reading the whole
+            # value into `$((...))` would overflow before the test.
+            top=$(read_le_uint "$bin" $((thunk + 7)) 1)
+            if [ $((top & 128)) -ne 0 ]; then
+                ordinal=$(read_le_uint "$bin" "$thunk" 2)
+                echo "$dll:#$ordinal" >>"$out"
             else
-                hn_off=$(pe_rva_to_off "$bin" "$sec" "$nsec" $((entry & 0x7FFFFFFF))) || return 2
+                hn_off=$(pe_rva_to_off "$bin" "$sec" "$nsec" "$entry_lo") || return 2
                 echo "$dll:$(read_cstr "$bin" $((hn_off + 2)))" >>"$out"
             fi
             t=$((t + 1))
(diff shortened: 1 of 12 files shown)


octalide:
Independent review found and fixed three merge blockers in `9790c241`:

1. The short-import `NameType` was reduced to a by-name boolean, so the PE writer emitted the public/link symbol even when the loader export name differed. A real x86-64 `llvm-dlltool` input containing `PublicName == ExportName` reports `Name type: export as`; the pre-fix compiler emitted `demo.dll:PublicName`, while the import record requires `demo.dll:ExportName`. The parser now implements NAME, NAME_NOPREFIX, NAME_UNDECORATE, and NAME_EXPORTAS (including its third payload string), keeps public and loader names distinct, and validates import/name/reserved bits. This matches the [PE/COFF import name types](https://learn.microsoft.com/en-us/windows/win32/debug/pe-format#import-name-type) and [LLVM's `applyNameType` / `getExportName` implementation](https://github.com/llvm/llvm-project/blob/main/llvm/lib/Object/COFFImportFile.cpp).
2. Import records were copied into the session-global attribution map from inside the linker. That made one artifact's import library capable of contaminating a later artifact link in the same build. Declarations now remain authoritative but link-local.
3. Conflict validation compared raw interned spellings, so a source/manifest attribution using logical alias `kernel32` contradicted a record naming its equivalent loader `kernel32.dll`. It now compares canonical dependency identity and retains hard errors only for different loaders.

The hermetic integration fixture now includes the exact EXPORTAS shape. It is a proven guard: the prior PR compiler fails both profiles with this diff:

```diff
-demo.dll:ExportName
+demo.dll:PublicName
```

Verification at `9790c241`:

- `mach test .`: 1,090 passed, 0 failed
- complete Linux integration leg: all cases passed, including `pe-import-library` in debug and release
- release self-host fixpoint: stage B == stage C, SHA-256 `0dd4a5eee15b5403ac251c48649f757119cbd5145afdb248180c7f7589bc0eaa`
- `git diff --check`: clean

I did not merge. New CI is pending on the pushed review fix.

octalide:
Follow-up: merged current dev (including #2537) into the branch at 3f7444cb and re-ran verification on the combined linker. Source suite: 1,092 passed / 0 failed. Complete Linux integration leg: all cases passed. Release self-host fixpoint: stage B == stage C, SHA-256 49d0627b879e795bb44bcf5d390db958315d48a542eb8598c546b25740e15d50. The focused export-as fixture still passes in both profiles. CI has restarted on the merged head.

octalide:
Final review state at `3f7444cb`: all 16 substantive CI checks pass across Linux, Windows, aarch64, x86_64, and riscv64; the Darwin release gate is intentionally skipped. The worktree is clean, GitHub reports the branch mergeable, and no code or test blocker remains from this review. The PR remains draft and review-required; I have not merged it.

Issue #1374, opened 2026-06-12
coff: native Windows loader rejects the emitted PE (ERROR_BAD_EXE_FORMAT); wine masks it


octalide:
Found by the #1351 windows-lane work (PR #1367, run 27426094065): the released `mach.exe` — and any exe built by the current PE emitter — cannot be launched by the native Windows loader. `CreateProcess` fails with `ERROR_BAD_EXE_FORMAT` ("%1 is not a valid Win32 application") on `windows-latest`, while the same binary runs under wine and `dumpbin` parses its headers cleanly. **Impact: every published `mach-<version>-x86_64-windows.zip` artifact is broken on real Windows; the wine-based release smoke test cannot catch this class.**

Diagnostics already gathered on the runner (throwaway CI step, since reverted):
- A known-good native exe runs in the same job (`cmd /c ver` → 0) — not an environment problem.
- Forced/mandatory ASLR is NOT the cause (ForceRelocateImages / RequireInfo / BottomUp / HighEntropy all `NOTSET`).
- `dumpbin`: x64 PE32+, **relocations stripped** (`IMAGE_FILE_RELOCS_STRIPPED`, no `.reloc` section), **no `DYNAMIC_BASE`** (DllCharacteristics `0x100`), fixed image base `0x13FFF0000`, first section RVA `0x10000`, `SizeOfHeaders` `0x400`; sections contiguous, `SizeOfImage` exact.

So the rejection is a loader header-invariant that wine doesn't enforce. Leading suspects in `src/lang/target/of/coff.mach`:
- the non-relocatable image (relocs stripped + no `.reloc` + fixed base) — likely fix: emit base relocations and set `DYNAMIC_BASE`;
- the `PE_IMAGE_BASE_RESERVE = 0x10000` design (`coff.mach:143`, `:1946`) placing the first section at RVA `0x10000` — check the SectionAlignment/FileAlignment invariants the real loader enforces (e.g. alignment pairing rules, header/section RVA consistency).

Pinpointing requires bisecting header fields against a live Windows loader; the `windows-latest` runner works for that loop (the #1351 worker iterated diagnostics via a temporary CI step on the PR branch).

Blocks #1351 (PR #1367 is held as draft until this lands — the lane itself is finished and merges as-is once the compiler emits a loadable PE).

PR #2545, opened 2026-08-05
fix(link): publish foreign x64 unwind tables

### LINKED ISSUE #2535: link: x64 has no addr32nb applier, so no COFF object with .pdata links

octalide:
Closes #2535

## Summary

- applies x64 `IMAGE_REL_AMD64_ADDR32NB` fields as `S + A - ImageBase`, using the same explicit image base for relocation and PE emission
- preserves foreign COFF `.pdata` as format-neutral exception metadata, then validates, combines, de-duplicates, sorts, and publishes it through the PE exception directory
- covers the path end to end with a real clang-produced object carrying three `ADDR32NB` relocations and 8 KiB-aligned text
- keeps the first PE section adjacent to the headers while prefix-padding over-aligned linker content to its exact RVA, preserving the native-loader invariant from #1374/#1376

## Verification

- source tests: `1112 passed, 0 failed`
- release self-host fixpoint: stage 2 and stage 3 byte-identical
- full Linux integration suite: debug and release green
- target-matrix validator: `280 case x leg pair(s) OK`
- ordinary PE output stability: `surface/pe-aslr` debug and release byte-identical to the pre-padding writer
- high-alignment regression: first section RVA `0x1000`, entry/content RVA `0x2003`, foreign runtime row `0x4000..0x4016`, exact ImageBase `0x140000000`
- all 16 required CI checks green at final head `426fd20e`

## Review status

Implementation is complete and independently reviewed. No remaining blockers were found.


octalide:
Independent review complete at `426fd20e56cd1030562f8dce3b0a139c6bf0ea63`: approved, with no remaining blockers.

I audited the full `origin/dev...HEAD` diff, public writer API/docs, generic-linker exception-metadata channel, COFF parser/relocation behavior, PE layout, runtime-function validation/de-duplication/sorting, the clang fixture producer, PR/issue history, and final CI.

One release blocker was found during review and fixed before approval. The 8 KiB text alignment originally placed the first PE section at RVA `0x2000` with 4 KiB SectionAlignment, recreating the header-to-first-section gap that #1374/#1376 proved native Windows rejects. Commit `426fd20e` now starts the section at RVA `0x1000` and prefix-pads its content to the exact linker RVA. This keeps the Windows loader adjacency invariant while preserving the explicit ImageBase used by `ADDR32NB`, generated unwind rows, base relocations, and import fixups. The focused unit asserts section RVA `0x1000`, virtual size `0x100e`, and text bytes at the padded `+0x1000` content offset.

Evidence:

- source tests: `1112 passed, 0 failed`
- release self-host fixpoint: stage 2 == stage 3, SHA-256 `8c30085a666182afa88ae28bd519d9cd6120a774536850a09eb6e7605c74355e`
- full Linux integration suite: all debug/release cases passed
- target matrix: `280 case x leg pair(s) OK`
- real clang fixture: three `IMAGE_REL_AMD64_ADDR32NB` relocations, 8 KiB text alignment, exact foreign `.xdata`
- emitted high-alignment PE: ImageBase `0x140000000`, first section/BaseOfCode `0x1000`, entry/content `0x2003`, foreign row `0x4000..0x4016 -> 0x5000`
- pre-existing `surface/pe-aslr` outputs remain byte-identical in debug and release
- `git diff --check origin/dev...HEAD`: clean
- all 16 required final-head CI checks passed, including Windows and every integration lane

Verdict: ready to merge into `dev` with a merge commit.

PR #2549, opened 2026-08-05
fix(link): support PE dllimport indirection

### LINKED ISSUE #2526: link: support __imp_* dllimport indirection
## What

`__imp` appears zero times in all of `src/`. Every Win32 function is `__declspec(dllimport)` in the mingw headers, so a `zig cc -target x86_64-windows-gnu` object references `__imp_Sleep`, not `Sleep`. Attribution is keyed by link symbol name, so `#[library("kernel32")]` on `Sleep` registers `Sleep` and never matches the object's `__imp_Sleep` reference; the result is `error: dynamic import '__imp_Sleep' has no #[library] attribution`.

Correct behavior: recognize the `__imp_` prefix, emit the import under the undecorated name, and resolve `__imp_X` to the address of the IAT slot rather than to the function.

This blocks vendored C on windows (briar-systems/mach-audio#11) and `symbols = [...]` from #2510 gives zero coverage for `__imp_`-prefixed references. It pairs naturally with issue #<1> but depends on nothing in it.

## Acceptance

- A zig cc / mingw windows object referencing `__imp_*` symbols links and the emitted import table names the undecorated symbols
- Regression test cross-linked from linux

octalide:
Closes #2526

## Summary

- canonicalize COFF `__imp_X` undefined symbols to loader export `X` while preserving every source spelling in the import map
- generalize deferred GOT work into format-neutral import-address fixups with preserved addends
- emit PE IAT cells for both function and data/indirect imports, while retaining call stubs only for function imports
- patch COFF import-address relocations to the canonical IAT cell with checked bounds and signed `rel32` arithmetic
- keep partial dynamic-plan cleanup capacity-correct on attribution errors
- document the COFF dllimport spelling and add a real clang-produced COFF integration fixture

## Regression evidence

- Current `dev` rejects the fixture at `dynamic import '__imp_Sleep' has no #[library] attribution`.
- This branch deduplicates direct `Sleep` and indirect `__imp_Sleep` into one loader import/IAT cell.
- The same fixture carries indirect-only `__imp_GetTickCount`, proving a data-style IAT entry is emitted without a call stub.
- Structural PE inspection recovers both foreign signed displacements and proves they land on the two expected IAT cells.

## Verification

- [x] debug unit suite: 1,117 passed
- [x] release unit suite: 1,117 passed
- [x] debug and release byte-identical self-host fixpoints
- [x] complete Linux integration suite
- [x] fresh CI green on Linux, aarch64, Windows, incremental, release, and integration lanes
- [x] independent review; import-plan capacity cleanup and indirect-only IAT coverage addressed
- [x] disposable combined validation with #2551 advances the real mach-glfw consumer beyond dllimport/CRT handling to the separate foreign-CodeView relocation gap tracked in #2553

Follow-up #2552 tracks COFF `DefinedLocalImport` semantics when `X` is defined inside the same link graph; that is distinct from this issue's loader-owned IAT path.


octalide:
Final independent audit before merge: branch was clean and contained current `dev`; the PR diff, linked #2526 acceptance, and all issue/PR conversation were reviewed; focused canonicalization, partial-plan cleanup, mixed-IAT-layout, and checked-rel32 unit tests were rerun; and the real clang-produced PE fixture passed under both debug and release profiles. All 16 required hosted checks passed, the complete local suite/fixpoint/integration evidence is recorded in the PR body, and no actionable finding remained. The repository-owner ruleset bypass was used only because the sole author cannot satisfy its second-identity approval requirement.

PR #2550, opened 2026-08-05
fix(macho): parse x86_64 SIGNED_1/2/4 relocations

### LINKED ISSUE #2546: macho: parse x86_64 SIGNED_1/2/4 relocations from clang objects
Follow-up to #2521 / #2533. Do this after #2533 merges because both changes touch the Mach-O relocation parser.

A normal x86_64 clang object using an immediate store to a global emits `X86_64_RELOC_SIGNED_4`, which mach currently rejects even after #2533:

```sh
clang -target x86_64-apple-macos11 -x c -c -o signed4.o -g -O1 - \n  <<< "int g; int mad_device_open(int a) { g = 42; return a; }"
llvm-readobj --relocations signed4.o
# __text + 0x8: X86_64_RELOC_SIGNED_4 against _g

mach build <darwin-x86_64 project linking signed4.o>
# error: macho: unsupported relocation type 8
```

`SIGNED_1`, `SIGNED_2`, and `SIGNED_4` are the ordinary PC-relative variants used when one, two, or four instruction bytes follow the displacement. They should parse into the existing indexed `RK_PC32` contract while preserving their effective field bias; plain `SIGNED` behavior and #2533 section-anchor handling must remain unchanged.

Acceptance:

- Parse all three x86_64 signed variants from real or hermetic Mach-O records.
- Link a clang-produced `SIGNED_4` object and assert the global store resolves to the exact target.
- Cover symbol-based and, where valid, section-based addend recovery.
- Run after #2533 so the implementation builds on its reconciled section-anchor/index contract.

octalide:
Closes #2546

## What

- parse `X86_64_RELOC_SIGNED_1`, `_2`, and `_4` into the indexed `RK_PC32` contract
- preserve LLVM/LLD embedded-addend and relocation-PC bias semantics for symbol- and section-based records
- normalize recovered in-place fields so parse → emit → parse folds the addend exactly once
- reject malformed width/PC-relative shapes, out-of-bounds fields, and section-target address overflow
- retain the section-anchor behavior introduced by #2533

## Real-object proof

- checks in ordinary C and its byte-reproducible `clang -target x86_64-apple-macos11 -c shim.c -g -O1 -fdebug-compilation-dir=.` object containing all three relocation variants
- extends the existing Darwin x86_64 foreign-object case with a structural producer that parses the linked Mach-O image, decodes each immediate-store displacement, and compares its exact linked target VA with an independently located initialized-data marker
- the pre-fix compiler fails this fixture with `macho: unsupported relocation type 8`; this branch passes it in debug and release profiles

## Verification

- debug self-host fixpoint (`b == c`) and 1,112/1,112 tests
- release self-host fixpoint (`b == c`) and 1,112/1,112 tests
- full Linux integration suite, including the exact SIGNED_1/2/4 Darwin cross-link
- target matrix: 280 case × leg pairs
- incremental determinism: warm no-op and post-edit rebuilds match clean builds
- clang fixture regeneration is byte-identical
- independent review found no production correctness blocker; its defensive test-hygiene suggestion is incorporated

Native macOS execution is not available on the local Linux runner. The checked-in real clang object and exact-target structural cross-link exercise the relevant Darwin linker behavior without executing the resulting binary.


octalide:
Independent review complete on merged head `96dbd038`.

I checked the production parser/addend algebra against LLVM/LLD's x86_64 Mach-O behavior (`SIGNED_N` contributes `N` to both the embedded addend and relocation PC), reviewed symbol- and section-based recovery plus bounds/overflow handling, and verified neutralizing the recovered in-place field preserves the indexed relocation contract across parse → link and parse → emit. No production correctness blocker found.

Independent reproduction:
- checked-in clang object contains `SIGNED_1`, `_2`, and `_4` and regenerates byte-identically from the documented command
- released Mach 4.6.0 fails both profiles with `macho: unsupported relocation type 8`
- reviewed head passes 1,112/1,112 tests and the exact-target foreign-object case in debug/release
- the defensive round-trip test ordering suggestion was incorporated in `96dbd038`

Merge is approved once the remaining post-merge hosted checks are green.

PR #2560, opened 2026-08-05
fix(driver): root dependency step outputs

### LINKED ISSUE #2558: driver: path-dependency steps resolve project.out through the symlink target
## Summary

A build step exported by a local path dependency can write `{project.out}` outside the consuming project's output tree. The driver expands it to a relative climb such as `../../out/linux-x86_64/debug`, then runs the command after `cd` into `dep/<alias>`. Path dependencies are materialized as symlinks, so external tools resolve that relative path from the symlink target's physical directory rather than the consumer's logical `dep/<alias>` path.

The driver pre-creates and later links the expected consumer output, while the step silently writes somewhere else. The build then fails with a missing link input; it can also leave untracked/generated files outside the consumer.

## Reproduction

1. Create a project with `[dep.lib] path = "../.."`.
2. Have the dependency export a local link input produced by a demanded `[step]` under `{project.out}`.
3. Run `mach dep pull` (creating `dep/lib` as a symlink) and build the consumer.

Observed in mach-glfw's consuming-project test:

```
step build-glfw: tools/build-glfw.sh ../../out/linux-x86_64/debug/vendor/glfw
error: cannot find link input 'out/linux-x86_64/debug/vendor/glfw/libglfw.a'
```

The archive was actually written two directories above the dependency checkout's physical target.

## Expected

`{project.out}` in every dependency step resolves to the consuming project's concrete output directory independent of whether the dependency is git-backed or a local path symlink, and independent of the invoker's working directory.

The driver should provide an absolute rooted output path (or an equivalent physical-path-safe representation) rather than requiring every step script to implement symlink-aware logical path normalization.

## Acceptance

- an integration fixture uses a local path dependency whose exported local archive/object is generated by a demanded step
- the step writes into the consumer's expected `{project.out}` and the final artifact links
- no generated output appears beside or above the dependency's physical source checkout
- relative and absolute project invocation paths behave identically
- git dependency steps, fingerprints/stamps, output existence checks, and root-project steps remain green
- the manifest/build-step documentation states that `{project.out}` is rooted to the consuming output for exported dependency steps

## Disposition

Assigned immediately in the current game-readiness recovery batch. Blocks the external-consumer acceptance for briar-systems/mach-glfw#32.

octalide:
## Summary

- expand dependency-step `{project.out}` to an absolute consumer-rooted output path
- normalize ordinary relative consumer roots before command expansion so `./subdir` and its normalized absolute spelling share one fingerprint/stamp
- resolve absolute step inputs, outputs, and glob roots without prefixing the dependency checkout
- add an end-to-end local path-dependency fixture that generates and links a foreign object in debug and release
- document the physical output-home and cache-key contract

## Verification

- 1,139/1,139 tests, x86_64 Linux debug
- 1,139/1,139 tests, x86_64 Linux release
- byte-identical debug fixpoint (`8987bc671d9641b4f4c941eed6713a387a4765b03483f50f7f084d02402054b2`)
- byte-identical release fixpoint (`01dbb9a54499388de65776a13de738d7aadbc56227ae5f56695ab762d1c35fd5`)
- `int/run.sh --target linux --filter 2558-dep-step-home` (debug and release)
- nontrivial relative invocation (`./int/regression/2558-dep-step-home`) followed by its normalized absolute invocation reuses the same output/stamp without rerunning the step
- existing root-step controls: `2543-bsd-ar-extended-name` and `pe-import-library` (debug and release)
- `int/lib/check-target-matrix.sh` (285 case/leg pairs)
- `sh -n int/regression/2558-dep-step-home/provider/build.sh`

Closes #2558


octalide:
Independent review complete at exact head 0ceff695e6c2678a9c31b87be780c76fabc36d7b. I reviewed the full diff, linked issue, PR conversation, and CI. Review found that ./project and its normalized absolute spelling initially produced different dependency-step command fingerprints; commit 0ceff695 resolves that gap and adds focused coverage. I independently rebuilt this head, passed the dep_out_home test in debug and release, and passed regression/2558-dep-step-home in both profiles. All 16 hosted checks pass, with only the expected Darwin release gate skipped. No actionable findings remain.

PR #2581, opened 2026-08-07
test(int): assert the darwin x86_64 entry argument block on both load commands

### LINKED ISSUE #2576: runtime/int(darwin): x86_64 entry reads argc off the stack, but --pie emits an LC_MAIN image that passes it in registers
Blocks briar-systems/mach-std#415.

A `--pie` darwin-x86_64 executable is emitted as an MH_PIE / LC_MAIN image, but mach-std's
x86_64 darwin entrypoint still reads `argc`/`argv` off the stack using the LC_UNIXTHREAD
convention, so it captures the return address as `argc`. Latent today because the only
fixture exercising the path never touches `argv`; it goes live the moment darwin-x86_64
links libSystem.

The code fix lands in `briar-systems/mach-std/src/runtime/darwin/x86_64.mach`; the fixture
gap is in this repo, which is why it is tracked here.

## Repro

`int/surface/macho-abs-bind-x86_64` already builds this exact configuration
(`case.conf`: `build-flags: --pie`, `build-target: darwin-x86_64`) and runs it natively on
the `macos-15-intel` leg. Add an `argv[0]` read to its `main` and the case fails:

```mach
#[symbol("main")]
fun main(argc: usize, argv: **u8) i64 {
    p.printlnf("argc={}", argc::i64);   # garbage: the return address
    ret 0;
}
```

## Expected

`argc` is the real argument count and `argv` the real argument vector, matching what dyld
passes an LC_MAIN entry.

## Actual

`argc` is the return address dyld pushed, `argv` points one slot past it, and `envp` is
derived from both — so `_rt_envp` is garbage too, which corrupts every
environment-variable read in `std.process.env`.

**Cause.** `mach/src/lang/target/os/darwin.mach:35` — `darwin_pie_exec` is aarch64-only, so
darwin-x86_64 defaults to the static writer. But `mach/src/lang/be/linker.mach:838` makes
the link dynamic on `dynlib_count > 0 || pie || …`, and
`mach/src/lang/target/of/macho.mach:3640-3656` emits LC_MAIN whenever `pie` is set. So `--pie`
on darwin-x86_64 already produces an LC_MAIN image today, while
`mach-std/src/runtime/darwin/x86_64.mach` still does:

```
mov rax, rsp      # rax = initial stack pointer
mov rdi, [rax]    # argc -> rdi     <-- under LC_MAIN this is the return address
lea rsi, [rax+8]  # argv -> rsi
```

Under LC_MAIN, dyld calls the entry like `main(argc, argv, envp, apple)` with the arguments
in `rdi`/`rsi`/`rdx`/`rcx`. The aarch64 entry was already converted to the register
convention in briar-systems/mach-std#324; x86_64 was explicitly left unchanged in that
commit because it was still on the static LC_UNIXTHREAD path at the time.

**Why it hides.** `int/surface/macho-abs-bind-x86_64/src/main.mach` declares
`main(argc, argv)` but reads neither — it only calls `mach_abs_bind_check()` and prints its
result. The fixture proves dyld binding works and says nothing about the entry contract.

**Why it stops being latent.** briar-systems/mach-std#415 adds a libSystem dependency, which
sets `dynlib_count > 0` and makes darwin-x86_64 take the dynamic writer unconditionally —
no `--pie` needed. Every darwin-x86_64 binary then has a broken `argc`/`argv`/`envp`.

## Environment

- triple: `darwin-x86_64`, `--pie`
- toolchain: mach 4.7.0 / mach-std 0.20.2, both `dev`
- OS: macOS 15 (Intel), the `macos-15-intel` runner in `.github/workflows/darwin-lane.yml`
- unaffected: `darwin-aarch64` (already on the register convention), all linux and windows
  targets

## Fix

Convert `mach-std/src/runtime/darwin/x86_64.mach` to capture `argc`/`argv`/`envp` from
`rdi`/`rsi`/`rdx`, mirroring the aarch64 entry. The stack dyld hands an LC_MAIN entry is
already 16-byte aligned at the call boundary, so the existing `and rsp, -16` should be
re-checked rather than carried over blindly.

In this repo: extend `int/surface/macho-abs-bind-x86_64` (or add a sibling case) so its
`main` reads `argc` and `argv[0]` and asserts them, closing the coverage hole that let this
survive. Whichever case lands must run on the native `macos-15-intel` leg, not build-only.

Note that darwin-x86_64 will need to reach the LC_MAIN path one way or another once #415
lands; whether `darwin_pie_exec` should also start returning true for x86_64 (making PIE
the default there rather than a `--pie` opt-in) is a separate call that can be made with
#415, not here.

octalide:
Closes #2576

Runtime fix: briar-systems/mach-std#421

## The real failure in #2576

`int/surface/macho-abs-bind-x86_64` already builds `darwin-x86_64 --pie` and executes it on the native Intel leg — the exact configuration that produces an MH_PIE / LC_MAIN image. It passed anyway, because its `main` declares `(argc, argv)` and reads neither. The suite therefore had a green case sitting directly on top of a broken entry contract.

## What this adds

Two `run: exec` cases on `targets: darwin-x86_64`, differing only in `build-flags`:

| case | build-flags | image | entry contract |
|---|---|---|---|
| `macho-entry-pie-x86_64` | `--pie` | MH_PIE, LC_MAIN | argc/argv/envp in rdi/rsi/rdx |
| `macho-entry-static-x86_64` | — | LC_UNIXTHREAD | argument block on the stack |

Each asserts three things the bug corrupts:

```
argc=1
argv0_is_prog=1
env_path=1
```

- `argc` — under the bug this is dyld's return address.
- `argv[0]` — the harness runs the artifact as `out/int/prog`, so the basename is fixed while the directory is not.
- `PATH` via `std.process.env` — `_rt_envp` is derived from argc/argv, so a misread block poisons the environment too. This is the half of #2576 that a pure argc check would still miss.

Two directories rather than one: `build-flags` is per-case, and the point of the pair is that the entrypoint has to satisfy **both** load commands. A single case would let the next change fix one contract by breaking the other — which is how the original regression happened (aarch64 moved to LC_MAIN in mach-std#324, x86_64 was left behind).

## That the pair actually catches it

Both cases resolve `[dep.mach-std] ref = "branch/main"`, i.e. the latest mach-std *release*, which still carries the broken entry. So on this branch, before mach-std#421 is released:

- `macho-entry-pie-x86_64` **must fail** on the `darwin-x86_64` leg
- `macho-entry-static-x86_64` **must pass** (the non-PIE path is unchanged)

That split is the demonstration — a red PIE case next to a green static case is the bug, isolated to the contract it belongs to. Dispatch is `int-main.yml` on this branch (see below).

## Where it runs

`int/targets.conf` maps the `darwin-x86_64` leg to `macos-15-intel`, and `int-main.yml` generates its matrix from that file, so both cases land on native Intel silicon with no workflow edit. `check-target-matrix.sh` passes (295 case x leg pairs).

Note: the int suite for darwin runs at **main cadence** (`int-main.yml`), not per-PR. `darwin-lane.yml`'s `macos-15-intel` job runs `./c test .` — the in-source unit corpus — and does not invoke `int/run.sh`, so it is not where an int case can be added.


octalide:
## The fixture is not blind — native `macos-15-intel` evidence

Two `int-main.yml` dispatches on the `darwin-x86_64` leg, identical except for which mach-std commit the cases resolve.

**1. Against released mach-std (`c500b97`, broken entry)** — [run 31155287617](https://github.com/briar-systems/mach/actions/runs/31155287617), this branch as-is:

```
PASS surface/macho-abs-bind-x86_64 [darwin-x86_64/debug]   (mach-std@c500b97)
PASS surface/macho-abs-bind-x86_64 [darwin-x86_64/release] (mach-std@c500b97)
    produce.sh: line 121: 7903 Segmentation fault: 11  "$bin"
FAIL surface/macho-entry-pie-x86_64 [darwin-x86_64/debug]   (mach-std@c500b97) (producer exit 139)
    produce.sh: line 121: 7936 Segmentation fault: 11  "$bin"
FAIL surface/macho-entry-pie-x86_64 [darwin-x86_64/release] (mach-std@c500b97) (producer exit 139)
PASS surface/macho-entry-static-x86_64 [darwin-x86_64/debug]   (mach-std@c500b97)
PASS surface/macho-entry-static-x86_64 [darwin-x86_64/release] (mach-std@c500b97)
```

The PIE case dies with **SIGSEGV**, both profiles — the argument block is garbage, so walking `envp` for `PATH` runs off into unmapped memory. The static case is green in the same run. And `macho-abs-bind-x86_64` — same target, same `--pie`, same native execution — is green next to a segfaulting binary, which is precisely the blindness #2576 describes.

**2. Against this fix (mach-std`@aa74840`)** — [run 31155794652](https://github.com/briar-systems/mach/actions/runs/31155794652), throwaway branch `tmp/2576-verify2` = this branch with every case's `[dep.mach-std] ref` repointed at the fix commit (deleted after; nothing to review here):

```
PASS surface/macho-abs-bind-x86_64 [darwin-x86_64/debug]   (mach-std@aa74840)
PASS surface/macho-abs-bind-x86_64 [darwin-x86_64/release] (mach-std@aa74840)
PASS surface/macho-entry-pie-x86_64 [darwin-x86_64/debug]   (mach-std@aa74840)
PASS surface/macho-entry-pie-x86_64 [darwin-x86_64/release] (mach-std@aa74840)
PASS surface/macho-entry-static-x86_64 [darwin-x86_64/debug]   (mach-std@aa74840)
PASS surface/macho-entry-static-x86_64 [darwin-x86_64/release] (mach-std@aa74840)
```

| | old runtime (`c500b97`) | fixed runtime (`aa74840`) |
|---|---|---|
| `macho-entry-pie-x86_64` (LC_MAIN) | **FAIL** — SIGSEGV | PASS |
| `macho-entry-static-x86_64` (LC_UNIXTHREAD) | PASS | PASS |

The static case being green in *both* columns is the other half of the point: the fix adds the LC_MAIN contract without disturbing the LC_UNIXTHREAD one.

## Merge order

These cases resolve `[dep.mach-std] ref = "branch/main"`, i.e. mach-std's latest **release**. So `macho-entry-pie-x86_64` will stay red on the darwin-x86_64 leg until briar-systems/mach-std#421 is merged **and released to mach-std `main`**. That is inherent to how every int case pins mach-std, not something to work around here — but it means this PR should land after that release, or land knowing the leg is red until it.

Note the leg is main-cadence, so this does not gate PR CI.

## Unrelated pre-existing failure

`surface/macho-got-x86_64` fails on the darwin-x86_64 leg in **both** runs above (`int: macho-got: native PIE execution failed`, both profiles) — so it is not caused by this PR and is **not** fixed by mach-std#421 either. It passed on the 2026-08-05 int-main run on `main`. Reported separately; untouched here.

octalide:
Correction to my evidence comment above: the line "It passed on the 2026-08-05 int-main run on `main`" about `macho-got-x86_64` is wrong — that case was created at `141e0dd3` (2026-08-05T23:35:34Z), ~18h *after* the last green run, so it never passed. It is not a regression. See briar-systems/mach#2586 for the correction; nothing about this PR's own evidence changes.

Issue #2578, opened 2026-08-07
BUG: [step] build steps never execute on a native Windows host (cmd.exe spawned without PATH resolution)


octalide:
`[step.X]` build steps cannot execute on a native Windows host: the step shell is spawned by bare name and `exec.run` does no PATH resolution, so `CreateProcessA` never finds `cmd.exe`.

## Repro

Any project with a build step, built by a native Windows `mach`:

```toml
[artifact.app]
kind = "bin"
entry = "main.mach"
out = "bin/app.exe"
targets = ["*"]
link = []
need = ["gen"]

[step.gen]
cmd = "echo hello"
in = []
out = ["out/gen.txt"]
need = []
```

```
mach build . --target windows
```

This is exactly what happened in CI: `int/surface/pe-resources-native` declared a
`[step.icon]` and failed on the `int windows` leg of run 31058386257
(commit e32ac920) with

```
FAIL surface/pe-resources-native [windows/debug] (build)
    error: failed to execute build step
```

## Expected

The step's `cmd` runs under `cmd.exe /c` and its exit status is honored, exactly
as `/bin/sh -c` works on POSIX hosts.

## Actual

The step never launches. `src/lang/driver/config.mach:955-959` selects the shell
by bare name on Windows:

```mach
var shell_name: *u8 = "/bin/sh";
var shell_flag: *u8 = "-c";
$if ($mach.build.os == $mach.os.windows) {
    shell_name = "cmd.exe";
    shell_flag = "/c";
}
```

and hands it straight to `exec.run` (`src/lang/driver/config.mach:997`).
`exec.run` -> `os.spawn` -> `spawn_redirected` passes `pathname` as
`CreateProcessA`'s `lpApplicationName`
(`mach-std src/system/os/windows/shared.mach:1642` / `:1676`). With a non-NULL
`lpApplicationName`, `CreateProcess` does not search `PATH` — a partial name is
completed against the current drive and directory only. `cmd.exe` is not in the
project's working directory, so the create fails with `ERROR_FILE_NOT_FOUND`,
`spawn` returns a negative errno, and `exec.run` reports `spawn failed`, which
`config.mach:1009` surfaces as `failed to execute build step`.

The absolute POSIX `/bin/sh` needs no lookup, which is why this is Windows-only.

The repo already knows about this constraint and solves it elsewhere:
`src/cli/cmd/dep.mach:1775-1785` resolves `git` through `util.resolve_cmd`
[...]

Issue #2587, opened 2026-08-07
BUG: Windows build steps are handed a cmd.exe command line quoted with the CRT argv convention


octalide:
Depends on #2578.

Even with the step shell resolved to a concrete path (#2578), every `[step]` still fails on a native Windows host: the `/c` argument is quoted with the CRT argv convention, which `cmd.exe` does not parse.

## Repro

Any project with a build step, built by a native Windows `mach`, once #2578's
PATH resolution is in place:

```
step marker: copy assets\seed.txt out\gen\marker.txt
Path not found.
error: build step exited non-zero
```

## Expected

The step's `cmd` runs with the project root as its working directory.

## Actual

`cmd.exe` receives a mangled command string and fails to `cd`, so the step body
runs in the wrong directory (or not at all).

`src/lang/driver/config.mach` builds the step command as

```mach
full_cmd = sfmt(a, "cd /d \"{}\" && {}", project_root, expanded_cmd);
```

and passes `argv = [<cmd.exe path>, "/c", full_cmd]`. mach-std's
`build_cmdline` (`src/system/os/windows/shared.mach:783-843`) then joins argv
using the **CRT/MSVCRT** convention: an argument containing spaces or quotes is
wrapped in `"`, embedded `"` become `\"`, and backslashes before a quote are
doubled. `full_cmd` always contains both spaces and quotes, so `cmd.exe`
receives

```
cmd.exe /c "cd /d \"D:\a\mach\mach\int\...\" && copy ..."
```

`cmd.exe` does not implement backslash escaping. It strips one outer quote pair
and treats the rest literally, so `cd` is handed the path `\"D:\a\mach\mach\...\"`
— quotes and backslashes included — and reports `Path not found.`. Because the
`cd /d "<root>" && ` prefix is unconditional, this affects **every** step on
Windows, not just ones with unusual paths.

The CRT convention is correct for ordinary programs; it is only wrong for
`cmd.exe /c`, which has its own parsing rules.

## Confirmation

Isolated with a space-free command, so no quoting is applied anywhere:

```
A: cmd "/ccd /d Z:\tmp\claude-1000\wshell&&cd&&copy seed.txt sub\out.txt"
   -> Z:\tmp\claude-1000\wshell
      1 file(s) copied          # works

B: cmd /c "cd /d \"Z:\tmp\claude-1000\wshell\" && cd && copy seed.txt sub\out.txt"
   -> Path not found.           # fails
```

[...]

PR #2589, opened 2026-08-07
[on hold] fix(driver): resolve the build-step shell to a concrete path

octalide:
**On hold — do not merge. Superseded in shape; the branch is kept for its history.**

This was the first half of #2578: a `mach.lang.driver.shell` that resolved the step shell by walking PATH. Two things changed after it was opened.

1. **PATH is the wrong mechanism for the shell.** mach-std's own exec tests resolve the interpreter from `%ComSpec%` (`src/process/exec.mach`, `shell_path()`), falling back to the canonical path. The shell is not a user-named program — it has a canonical platform location. The three PATH lookups elsewhere in mach (`dep.mach`, `run.mach`, `testing.mach`) are locating user-named programs, which is a genuine caller policy decision; this one never should have been a PATH lookup at all. Tracked separately as mach-std#425.
2. **It does not fix anything on its own.** Once the shell spawns, #2587 takes over: the `/c` argument is quoted with the CRT convention, which `cmd.exe` does not parse, so every step still fails.

## What replaces it

A single mach PR after the mach-std chain lands:

- mach-std#423 — `exec.run_in`, a working directory on the spawn contract (ready for review)
- mach-std#424 — `exec.run_shell`, owning the per-platform command quoting

With both, `config.mach` becomes one `exec.run_shell(cmd, project_root, env)` call: no ComSpec lookup, no `/bin/sh` constant, no `cd` prefix on either platform, and `mach.lang.driver.shell` dissolves entirely rather than shrinking.

The Windows-leg `[step]` fixture ships with that PR, along with the space-in-root and quote-in-command cases, each demonstrated failing against unfixed code.

## What still stands from this branch

The diagnosis and the layering finding. `lang` cannot import `cli` (`src/cli/util.mach` already imports `mach.lang.manifest`), so any shared helper has to live on the `lang` side — that constraint carries over.

Refs #2578, #2587


octalide:
Superseded by #2602, merged as `38298fc`.

The replacement landed in the shape this PR's own description predicted: with mach-std 0.23.0's `exec.run_shell` carrying the ComSpec lookup, the child working directory, and the per-platform quoting, `config.mach` became a single call and `mach.lang.driver.shell` was never needed at all rather than shrinking.

Closing unmerged. The diagnosis and the `lang` cannot import `cli` layering finding both carried over.

Closes #2578 and #2587 via #2602.

Issue #2510, opened 2026-08-05
link: pin symbol→DLL import mapping for static archives on the PE target


octalide:
## What

The vendored-static route for C dependencies (first consumer: mach-glfw's planned static GLFW archive) merges `.a` members into the executable; the archive's undefined symbols — Win32 API calls — must then resolve to PE imports. PE imports are per-DLL: the import table requires a symbol→DLL mapping, unlike ELF's flat `DT_NEEDED` list or Mach-O's recorded dylibs. It is not established how resolution learns that mapping when cross-linking from a linux host, where the target system DLLs are not on disk to consult.

Investigate and pin the behavior:

1. Trace `-l`/`[link]` PE resolution for undefined ext symbols that originate in static archive members (`doc/cli.md:159-184` documents the intent; verify the implementation).
2. If the mapping requires consulting real DLL export tables at link time, cross-linking this case from linux is broken. Design the proper fix rather than a workaround: `[link.X]` entries already carry a per-OS `name` (e.g. `glfw3.dll`) — extend resolution so a `[link]` entry claims its symbols, either explicitly or via a declared export list. Vendored import stubs per consumer repo are the fallback design; a manifest-declared mapping is preferred (no external stub files to version).
3. Implement the small change (if any) the investigation demands under this issue; escalate to a dedicated issue if it proves structural.

## Acceptance

- A test that cross-links, from a linux host, a static archive containing undefined Win32 symbol references produces a PE whose import table binds those symbols to the declared DLLs, asserted via the PE test harness.
- The resolved behavior (existing or newly implemented) is documented in `doc/cli.md`'s static-vs-dynamic resolution section.

octalide:
Concrete data from the first consumer (mach-glfw#31 / mach-glfw#32), in case it is useful for pinning the design.

GLFW 3.4's win32 backend, compiled with `zig cc -target x86_64-windows-gnu` and archived with `zig ar`, leaves **148** undefined non-GLFW symbols. They fall into two classes, and the second one is not covered by this issue's framing:

[...]

PR #2614, opened 2026-08-07
fix(link): name both attribution routes in the unattributed-import error

### LINKED ISSUE #2528: diag: unattributed-import error recommends a remedy that cannot work for archive-origin symbols
## What

`unattributed_import_message` at `src/lang/be/linker.mach:3962-3974` says to annotate the declaration with `#[library("...")]`. For a symbol arriving from a static archive member there is no declaration to annotate — that is the premise of #2510, which added the `symbols = [...]` manifest key precisely for this case. A user hitting this is pointed at the one remedy that cannot work, with no mention of the new key.

## Fix

Name both routes in the diagnostic — the decorator for symbols with a Mach declaration, the manifest key for archive-origin symbols.

## Acceptance

- The diagnostic names both
- A test asserts the archive-origin wording

octalide:
Closes #2528

## What

`unattributed_import_message` told the user to annotate the declaration with `#[library("...")]`. A symbol arriving from a static archive member has no declaration to annotate, which is the entire premise of #2510 and why the manifest `symbols = [...]` key exists. The one remedy offered was the one that cannot work.

The message now names both routes, and spells the manifest claim with the symbol's own name so it is directly actionable:

```
error: dynamic import 'some_unattributed_import' has no #[library] attribution; a two-level-namespace target requires every import to name its providing library - annotate the declaration with #[library("...")], or, when the symbol arrives from a static archive member and has no declaration to annotate, claim it with symbols = ["some_unattributed_import"] on the [link.<entry>] naming its library; candidate libraries: kernel32 (kernel32.dll)
```

The degraded form taken when the name cannot be resolved names both routes too, with the bare `symbols = [...]` spelling since it has no symbol to insert.

Docs already describe both routes correctly (`doc/cli.md`, `doc/language/ext-fun.md`), so no prose change was needed.

## Tests

New unit test `mach.lang.be.linker.unattributed_import_message:names_both_routes` asserts the archive-origin wording (`static archive member`, `no declaration to annotate`, `symbols = ["Sleep"]`, `[link.<entry>]`) on both the named and the degraded form.

With the test in place and `unattributed_import_message` reverted to its pre-fix body:

```
mach.lang.be.linker                          36 ok     1 FAIL     94ms
  FAIL  mach.lang.be.linker.unattributed_import_message:names_both_routes  ./src/lang/be/linker.mach:7223  (exit 1)

1167 passed, 1 failed, 1168 total
```

With the fix: `1168 passed, 0 failed, 1168 total`.

The `surface/pe-import-attribution` int golden was re-blessed; its diff is the before/after of the user-visible diagnostic.

## Verification

- `mach test .` green (1168/1168)
- `int/run.sh --target linux` green


octalide:
Self-host fixpoint verified on this branch: stage2 `cmp` stage3 identical. Full `int/run.sh --target linux` green.

PR #2618, opened 2026-08-07
fix(link): refuse conflicting #[library] attributions instead of last-writer-wins

### LINKED ISSUE #2529: link: duplicate #[library] attributions silently resolve last-writer-wins
## What

`bind_import_libraries` at `src/lang/driver/union.mach:433-437` inserts attributions with no conflict check, and the map's own docstring states "Registering the same key overwrites." So two modules each declaring `ext fun Sleep` under different `#[library]` values resolve by module order, silently.

#2510 added hard errors for claim-vs-claim and claim-vs-decorator conflicts, and its docs (`doc/language/ext-fun.md`, `doc/cli.md`) now present "claimed only once" as a global invariant — the prose asserts a symmetry the code does not have. Pre-existing, not introduced by #2510.

Additionally, the adjacent silent-drop shape in the same function: `if (is_some(lib)) { if (is_some(name)) { register } }` has no else, so a `#[library]` on a decl with no `export_name` would be silently dropped. It is currently unreachable — `register_ext_fun`/`register_ext_global` unconditionally default every ext decl's name and `scan_export_directives` recurses into comptime-if — but it is a silent-drop shape on an attribution path, which the project forbids.

## Fix

- Make decorator-vs-decorator conflict a hard error naming both declarations, matching the claim paths, then the docs become true as written
- Make the unreachable silent-drop branch an internal error

## Acceptance

- Conflicting decorators error naming both
- The unreachable branch is an internal error
- Docs accurate

octalide:
Closes #2529

## What

`bind_import_libraries` (`src/lang/driver/union.mach`) inserted every `#[library]` attribution into the session's link-name-keyed `import_libraries` map with no conflict check. That map overwrites on a repeated key, so two modules each declaring `ext fun Sleep` under different `#[library]` values resolved by module discovery order, silently. #2510 made claim-vs-claim and claim-vs-decorator hard errors and its docs present "claimed only once" as a global invariant, so the prose asserted a symmetry the code did not have.

Each link name's first attribution is now carried across the whole registration walk (a local `ImportAttribution` map: module FQN, source identifier, library), and a second declaration naming a different library is a hard error identifying both:

```
error: import 'Sleep' is attributed to library 'kernel32' by declaration 'Sleep' in module 'dup.a' and to library 'user32' by declaration 'Sleep' in module 'dup.b'; a symbol may name only one providing library
```

An identical repeat is accepted, matching `register_link_claims`: two declarations naming the same library name one provider.

### Silent-drop branch

The adjacent `if (is_some(lib)) { if (is_some(name)) { register } }` had no else, so a `#[library]` on a decl with no `export_name` was dropped without a word. It is unreachable today (`register_ext_fun` / `register_ext_global` default every `ext` decl's link name, and `scan_export_directives` recurses into comptime-if), but a silent drop on an attribution path is not an acceptable shape here. It is now `internal: ext decl carries a #[library] attribution but no link name`.

### Docs

`doc/language/ext-fun.md` and `doc/cli.md` now state the decorator-vs-decorator case explicitly, so "claimed only once" is true as written rather than describing only the claim paths.

## Tests

Two new tests in `src/lang/driver/tests.mach`:

- `mach.lang.driver:duplicate_library_attribution_names_both` — a three-module scaffold where `uniontest.a` and `uniontest.b` both declare `ext fun Sleep` under different libraries. Asserts the build fails and that the message names the symbol, both declarations, both modules, and both libraries.
- `mach.lang.driver:identical_library_attribution_accepted` — the same scaffold with both modules naming `kernel32`. Asserts the build succeeds **and** that `session.import_library("Sleep")` is `kernel32`, so the test verifies the attribution actually landed rather than only that nothing errored.

With both tests in place and `src/lang/driver/union.mach` reverted to `origin/dev`:

```
mach.lang.driver.tests                      178 ok     1 FAIL     3.2s
  FAIL  mach.lang.driver:duplicate_library_attribution_names_both  ./src/lang/driver/tests.mach:11341  (exit 1)

1168 passed, 1 failed, 1169 total
```

`identical_library_attribution_accepted` passes before the fix, as it must: it guards against the check over-firing, not the bug.

With the fix: `1169 passed, 0 failed, 1169 total`.

## Verification

- `mach test .` green (1169/1169)

### DIFF
--- a/doc/cli.md
+++ b/doc/cli.md
@@ -258,9 +258,13 @@ Two declarations supply it:
 Both write the same attribution, so a symbol may be claimed only once: a
 `symbols` entry that contradicts another entry's claim, or a `#[library]`
 decorator's, is an error naming both claimants rather than a silent
-order-dependent win. Repeating an identical claim is accepted, which is what an
+order-dependent win. The same holds between two `#[library]` decorators: two
+`ext` declarations resolving to one link name under different libraries are an
+error naming both declarations, so no attribution is decided by module load
+order. Repeating an identical claim is accepted, which is what an
 `export = true` entry reaching a consumer through both the cascade and its own
-manifest does.
+manifest does, and what two modules declaring the same binding under the same
+library do.
 
 On ELF the loader resolves imports by global search, so neither declaration
 changes the emitted binary; both are still validated against the link's
--- a/doc/language/ext-fun.md
+++ b/doc/language/ext-fun.md
@@ -147,6 +147,14 @@ ext fun WSAStartup(ver: u16, data: *u8) i32;
   manifest instead, by the `symbols` key on the `[link.X]` entry that supplies
   it; see [manifest.md](../manifest.md#linkname--link-requirements). The two
   declarations write the same attribution, so a symbol may be claimed only once.
+- "Only once" is checked across every declaration, and two `#[library]`
+  decorators are no exception. Two `ext` declarations that resolve to the same
+  link name but name different libraries are a hard error identifying both
+  (`import '<sym>' is attributed to library '<a>' by declaration '<d>' in module
+  '<m>' and to library '<b>' by declaration ...`), so which library the import
+  table names never depends on module load order. Two declarations naming the
+  *same* library are accepted: they name one provider, which is what a shared
+  binding declared in two modules does.
 
 `library` composes with `symbol`: the rename sets the imported symbol's name,
 `library` sets the dependency it is imported from. On one decl the import is emitted
--- a/src/lang/driver/tests.mach
+++ b/src/lang/driver/tests.mach
@@ -11333,3 +11333,85 @@ test "mach.lang.driver:embed_buffer_is_stable_across_warm_rebuilds" {
     session.dnit(?s);
     ret bad;
 }
+
+# two `ext` declarations pinning one link name to different libraries must be a
+# hard error naming both, not a silent load-order win (#2529). the session's
+# attribution map overwrites on a repeated key, so before the check which library
+# the import table named was decided by module discovery order
+test "mach.lang.driver:duplicate_library_attribution_names_both" {
+    var alloc: A.Allocator;
+    if (O.is_some[str](page.make(?alloc))) { ret 1; }
+    val sr: R.Result[session.Session, str] = session.init(?alloc);
+    if (R.is_err[session.Session, str](sr)) { ret 1; }
+    var s: session.Session = R.unwrap_ok[session.Session, str](sr);
+
+    var names: [3]str;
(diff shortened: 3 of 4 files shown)


octalide:
Follow-up commit: the missing-source-text branch now degrades the diagnostic instead of erroring. It is read only to name a declaration in the message, and every sibling reader in `union.mach` already degrades on an absent `SourceFile`. The conflict check is keyed on the link name, so detection and refusal are unaffected either way. Re-verified: `mach test .` 1169/1169, full linux int suite green, fixpoint holds.

octalide:
## Correction pushed: mutually exclusive comptime arms

While re-verifying I found that the conflict check as originally written rejected a legitimate and expected pattern — a target-conditional attribution:

```mach
    #[library("kernel32")]
    ext fun Sleep(ms: u32);
}
$or {
    #[library("c")]
    ext fun Sleep(ms: u32);
}
```

The arms are mutually exclusive, so calling them a contradiction asserts something untrue. Each declaration now carries the arm path it sits under, and a conflict is reported only between declarations that no `$if` chain separates.

The rule is decided from the source alone, never by evaluating a condition. I tried evaluating them (descending only into the selected arm, as lowering does) and it is the wrong place: this pass runs before sema, so `mach.lang.driver:layout_intrinsic_remaining_positions` started failing — a `$size_of` in a declaration-scope `$if` condition is sema's diagnostic to report, and evaluating here front-ran it.

### A separate pre-existing defect, not fixed here

On `dev`, independent of this PR, a target-conditional attribution binds to the **wrong arm**. The projection runs before a target selects an arm, so the last arm walked wins. Building the fixture above for windows on `dev`:

```
```

`c` is the `$or` arm — the one windows does not take. This PR leaves that exactly as it was on purpose; repairing it means projecting attributions after arm selection, which is a pass-ordering change well beyond #2529. Worth its own issue.

### Tests

- `mach.lang.driver:exclusive_comptime_arms_are_not_a_conflict` — the chain above builds clean and still registers an attribution.
- `mach.lang.driver:comptime_arm_does_not_mask_cross_module_conflict` — a second module contradicting the chain shares no arm with it, so that is still a conflict naming both. This is what keeps the exemption from becoming a blanket amnesty for anything near a `$if`.

Demonstrated failing before the exemption (with `arms_exclusive` removed so every meeting reports):

```
  FAIL  mach.lang.driver:exclusive_comptime_arms_are_not_a_conflict  ./src/lang/driver/tests.mach:11574  (exit 1)
1189 passed, 1 failed, 1190 total
```

After: `1190 passed, 0 failed, 1190 total`. Full linux int suite green, fixpoint holds.

PR #2623, opened 2026-08-07
fix(link): refuse an import whose loader name is an import-address alias

### LINKED ISSUE #2527: link: refuse to emit an import whose name begins with __imp_
Silent miscompile: using `#[symbol("__imp_Sleep")]` to force attribution produces a PE whose import table contains `kernel32.dll:__imp_Sleep`, which does not exist.

## Repro

Use `#[symbol("__imp_Sleep")]` in a Windows target to force attribution, or use the `symbols = [...]` key from #2510 with an `__imp_*`-prefixed name.

## Expected

An import named `__imp_*` is rejected with a diagnostic pointing at the real fix.

## Actual

Emitted into the import table, producing an image that cannot load on Windows with a missing-entry-point error. The image links clean and passes CI, failing only at load time.

## Environment

linux x86_64 host, cross-linking for windows-x86_64, mach 4.5.0. Established statically from emitted bytes plus the PE import-by-name contract; nothing was executed.

octalide:
Closes #2527

## What

`__imp_X` is PE's spelling for the address cell the import table *itself* synthesizes for export X. It is never an export, so no library provides a symbol under that name. An image importing `kernel32.dll:__imp_Sleep` links clean, passes CI, and fails only when Windows loads it, with a missing-entry-point error. That is the silent miscompile.

`canonical_import_name` strips exactly one `__imp_` from an *object symbol*, which is why `#[symbol("__imp_Sleep")]` on its own no longer reproduces on `dev`. Two routes still reach the loader plan with the prefix intact:

1. **The manifest claim.** A doubled source spelling canonicalizes down to the alias, which a `symbols = ["__imp_Sleep"]` claim then attributes. (A `#[library]` decorator cannot do this: it keys on the source spelling the canonical name is derived from, so it always misses.)
2. **An import-library short record.** `ImportDecl.import_name` overrides the canonical name after canonicalization has run, so a `.lib` record naming `__imp_X` goes straight into the table unchecked.

The guard sits where the row's loader-facing name is settled (after the `decl != nil` override in `build_dynamic_info`), so it covers both. It is written against the format's own `import_address_prefix` rather than a literal `__imp_`, so it is inert on ELF and Mach-O, where a `__imp_`-named symbol carries no such meaning.

```
error: dynamic import '__imp_Sleep' names the import-address alias of 'Sleep' rather than an export: '__imp_' is the convention for the address cell the import table itself synthesizes, so no library exports a symbol under that name and the image would fail to load. import 'Sleep' instead - the linker emits that import and resolves every alias reference to its address cell
```

A bare prefix with no suffix names no export and is not an alias, so it is left alone.

## Tests

Two, one per route:

- `int/surface/pe-imp-alias-refused` — `run: build-fails`, cross-built win64 on the linux leg. Covers the manifest-claim route end to end.
- `mach.lang.be.linker.build_dynamic_info:rejects_import_address_alias` — an `ImportDecl` whose `import_name` is `__imp_Sleep` over an undecorated object symbol, so canonicalization is not involved. Asserts the link errors, that **the plan carries no import row at all**, and that the message names both the alias and the export.

The issue notes a test that only checks the link succeeds is exactly what let this through, so here is the miscompiled output. With the guard removed, the same fixture links clean and emits:

```
$ python3 peimp.py out/windows/debug/bin/case
import dir rva=0x3000 size=60
kernel32.dll:__imp_Sleep
kernel32.dll:ExitProcess
```

`kernel32.dll:__imp_Sleep` is the bad row: kernel32 exports `Sleep`, never `__imp_Sleep`.

The int case reports that directly:

```
FAIL surface/pe-imp-alias-refused [linux/debug] (build succeeded; expected a link error)
FAIL surface/pe-imp-alias-refused [linux/release] (build succeeded; expected a link error)
int: 2 failure(s) on linux
```

And the unit test, with only the guard block deleted:

```
  FAIL  mach.lang.be.linker.build_dynamic_info:rejects_import_address_alias  ./src/lang/be/linker.mach:7266  (exit 1)

1167 passed, 1 failed, 1168 total
```

With the fix, both pass and the emitted table is refused rather than written.

The pre-existing `pe-dllimport-indirection` and `pe-local-import-pointer` cases stay green, so the legitimate Clang `__imp_X` object reference still resolves to X's IAT cell and still emits the undecorated import.

## Verification

- `mach test .` green (1168/1168)


octalide:
Merged current `dev` (which now carries #2614, #2618, #2625). The conflict was in `src/lang/be/linker.mach`: `dev`'s #2528 test and this branch's #2527 test were added at the same anchor, so both are kept.

Re-verified on the merged tip, including re-confirming the guard is load-bearing after the reconstruction. With only the guard block removed:

```
mach.lang.be.linker                          37 ok     1 FAIL     86ms
  FAIL  mach.lang.be.linker.build_dynamic_info:rejects_import_address_alias

FAIL surface/pe-imp-alias-refused [linux/debug] (build succeeded; expected a link error)
FAIL surface/pe-imp-alias-refused [linux/release] (build succeeded; expected a link error)
```

Restored: `mach test .` 1197/1197, full linux int suite green, self-host fixpoint holds.

PR #2625, opened 2026-08-07
fix(link): read the relaxable GOTPCREL spellings

### LINKED ISSUE #2534: link: R_X86_64_REX_GOTPCRELX (type 42) rejected as unsupported
## Summary

With `-fPIC`, a global referenced through the GOT emits `R_X86_64_REX_GOTPCRELX` (relocation type 42) and mach rejects it outright.

## Repro

Compile a C object with `-fPIC` that references a global through the GOT, and link it with mach.

## Expected

The relocation is applied (or correctly relaxed — GOTPCRELX is the relaxable form of GOTPCREL, and a linker may either honor the indirection or relax it to a direct PC-relative reference when the symbol is local).

## Actual

`error: elf: unsupported relocation type 42`

## Environment

linux x86_64, cross-linking not required — reproduces natively. mach 4.5.0.

## Notes

Found while isolating #2520. **Distinct from #2520**: this is a hard error, not silent corruption, so it fails loudly rather than producing wrong data. Same ELF reader though, and any consumer vendoring C built with `-fPIC` will hit it — which is the default on most distributions.

## Disposition

Queued behind #2520 — same file, same subsystem, and #2520 is the critical one. Do not start until it lands, to avoid two agents in the same relocation code.

octalide:
Closes #2534

## What

`-fPIC` reaches a global through the GOT and emits `R_X86_64_REX_GOTPCRELX` (42); under `-fno-plt` an indirect call emits `R_X86_64_GOTPCRELX` (41). The ELF reader mapped neither, so both were rejected outright: `error: elf: unsupported relocation type 42`. `-fPIC` is the distribution default, so every project vendoring C reached this.

## Honoured, not relaxed

Both types carry GOTPCREL's own arithmetic (`G + GOT + A - P`). They differ from `R_X86_64_GOTPCREL` only by *permitting* a linker to rewrite the instruction at the patch site into a direct PC-relative reference when the target is not preemptible. The psABI makes relaxation an option, never an obligation, so both map onto the existing `RK_GOTPCREL` kind and the indirection is honoured as written.

That was a deliberate choice over relaxing:

- The patch is byte-identical to plain GOTPCREL's, and mach already synthesizes and fills the GOT slot the reference names, so the slow path is complete and provably correct today.
- Relaxing means decoding the bytes *preceding* the patch site to identify the REX/opcode/ModRM form and rewriting `mov sym@GOTPCREL(%rip), %reg` into `lea`, or `call *sym@GOTPCREL(%rip)` into `call rel32` plus padding. A wrong rewrite is silent and executes. The payoff is one elided load per site.

A correct slow path beats an unsound fast one, so I did not implement relaxation.

The map is now many-to-one, which makes it lossy in one direction: a relaxable form carried into a merged relocatable object (`--emit obj`) re-emits the plain `R_X86_64_GOTPCREL`. That drops only the downstream linker's permission to relax, never the arithmetic, and is asserted in the test.

The numbers are read against `e_machine`, so riscv, which gives 41 and 42 entirely different meanings, is untouched.

## Tests

- `mach.lang.target.of.elf.reloc_kind_for:gotpcrelx_reads_as_gotpcrel` — both types read as `RK_GOTPCREL`, the forward map still emits the unrelaxable type, and 41/42 are still rejected for `EM_RISCV`.
- `int/surface/elf-gotpcrelx` — `run: exec`, linux only. A real `cc -fPIC -fno-plt` object reads and writes a global defined in a separate TU, plus an indirect call, all through the GOT.

The int case's observable is the **values**, not the link status:

```
read=7
add=12
bump=13
final=13
```

The sequence is order-dependent and the final read must observe the earlier writes through the same slot, so a slot left zero or pointed at the wrong symbol faults or prints a wrong number. A link-succeeded check would have passed with a garbage slot, which is the failure mode this has to exclude.

The build step re-reads its own object and fails if either relocation type is absent, so a toolchain that stopped emitting them breaks the case loudly instead of leaving it silently covering nothing.

With only the two-type mapping reverted:

```
=== unit before ===
failures:
  mach.lang.target.of.elf.reloc_kind_for:gotpcrelx_reads_as_gotpcrel  ./src/lang/target/of/elf.mach:4575  (exit 1)
1172 passed, 1 failed, 1173 total

=== int before ===
    error: elf: unsupported relocation type 42
FAIL surface/elf-gotpcrelx [linux/release] (build)
    error: elf: unsupported relocation type 42
int: 2 failure(s) on linux
```

That `unsupported relocation type 42` is verbatim the issue's reported symptom. With the fix both pass.

## Note

`R_X86_64_CODE_4_GOTPCRELX` (43), the EVEX/APX-encoded form added by newer binutils, has the same resolution semantics and would map the same way, but I could not produce one with this host's toolchain, so it is deliberately left unmapped rather than added untested. Worth a follow-up issue.

## Verification

- `mach test .` green (1173/1173)


octalide:
Note: this branch had a `Merge branch 'dev'` pushed to it from outside this session while I was re-verifying. I reconciled rather than force-pushed, so that merge is preserved. Re-verified on the reconciled tip: `mach test .` 1194/1194, full linux int suite green, self-host fixpoint holds.

PR #2630, opened 2026-08-07
fix(link): exempt mutually exclusive comptime arms from the attribution conflict

octalide:
Fixes a regression **currently live on `dev`**, introduced by #2618 (my own PR).

## What happened

#2618 was merged at `f72fdc2f`. I pushed three correction commits after that point, so they were never picked up. `dev` therefore carries the version of the `#[library]` conflict check that rejects a legitimate and expected pattern: a target-conditional attribution.

```mach
    #[library("kernel32")]
    ext fun Sleep(ms: u32);
}
$or {
    #[library("c")]
    ext fun Sleep(ms: u32);
}
```

Against `dev` today:

```
error: import 'Sleep' is attributed to library 'kernel32' by declaration 'Sleep' in module 'ctif.a' and to library 'c' by declaration 'Sleep' in module 'ctif.a'; a symbol may name only one providing library
```

The arms are mutually exclusive, so calling them a contradiction asserts something untrue. This PR is those three commits, cherry-picked onto current `dev`.

## The fix

Each declaration carries the comptime-arm path it sits under, and a conflict is reported only between declarations that no `$if` chain separates.

The rule is decided from the source alone, never by evaluating a condition. I did try evaluating them — descending only into the selected arm, as lowering does — and it is the wrong place: this pass runs before sema, so `mach.lang.driver:layout_intrinsic_remaining_positions` began failing, because a `$size_of` in a declaration-scope `$if` condition is sema's diagnostic to report and evaluating here front-ran it.

## A separate pre-existing defect, deliberately not fixed here

Independently of #2529 and of this PR, a target-conditional attribution binds to the **wrong arm**. The projection runs before a target selects an arm, so the last arm walked wins. With this PR applied, building the fixture above for windows gives:

```
```

`c` is the `$or` arm, the one windows does not take. That is `dev`'s behaviour from before #2529 and is left exactly as it was; repairing it means projecting attributions after arm selection, which is a pass-ordering change. It deserves its own issue.

## Tests

- `mach.lang.driver:exclusive_comptime_arms_are_not_a_conflict` — the chain above builds clean and still registers an attribution.
- `mach.lang.driver:comptime_arm_does_not_mask_cross_module_conflict` — a second module contradicting the chain shares no arm with it, so that is still a conflict naming both. This is what stops the exemption becoming a blanket amnesty for anything near a `$if`.

With `arms_exclusive` removed so every meeting reports:

```
  FAIL  mach.lang.driver:exclusive_comptime_arms_are_not_a_conflict  ./src/lang/driver/tests.mach:11574  (exit 1)
1189 passed, 1 failed, 1190 total
```

And the end-to-end shape, against `dev` as it stands, is the diagnostic quoted at the top of this PR.

## Verification

- `mach test .` green (1199/1199)
- The real cross-module conflict still errors naming both declarations

### DIFF
--- a/doc/cli.md
+++ b/doc/cli.md
@@ -261,7 +261,9 @@ decorator's, is an error naming both claimants rather than a silent
 order-dependent win. The same holds between two `#[library]` decorators: two
 `ext` declarations resolving to one link name under different libraries are an
 error naming both declarations, so no attribution is decided by module load
-order. Repeating an identical claim is accepted, which is what an
+order. Declarations under different arms of one `$if` chain are exempt: no target
+selects both arms, so they never both apply. Repeating an identical claim is
+accepted, which is what an
 `export = true` entry reaching a consumer through both the cascade and its own
 manifest does, and what two modules declaring the same binding under the same
 library do.
--- a/doc/language/ext-fun.md
+++ b/doc/language/ext-fun.md
@@ -165,6 +165,9 @@ ext fun WSAStartup(ver: u16, data: *u8) i32;
   table names never depends on module load order. Two declarations naming the
   *same* library are accepted: they name one provider, which is what a shared
   binding declared in two modules does.
+- Declarations under **different arms of one `$if` chain** are exempt, because no
+  target selects both arms, so they never both apply. A target-conditional
+  attribution is written the obvious way and is not a conflict.
 
 `library` composes with `symbol`: the rename sets the imported symbol's name,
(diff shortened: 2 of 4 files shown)


octalide:
Merged current `dev` (which now also carries #2623). Re-verified on the merged tip: `mach test .` 1200/1200, full linux int suite green, self-host fixpoint holds. Target-conditional `#[library]` no longer errors as a conflict; the cross-module conflict still errors naming both declarations.

Issue #2636, opened 2026-08-07
link: a manifest symbols = [...] claim keys on the source spelling while #[library] keys on the link name, so claims do not attribute on Mach-O


octalide:
Found by the darwin release gate on the 4.9.0 cut, where two `#[library]` attribution tests failed on the macOS legs and passed everywhere else. The tests were host-blind and are fixed separately; the divergence they exposed is this, and it is real.

## What

The two routes that attribute an import to a library key the session map on **different spellings**.

| route | key | where |
|---|---|---|
| `#[library("...")]` decorator | the declaration's **link name**, carrying the target's symbol prefix | `union.bind_import_libraries` via `session.export_name` |
| `symbols = [...]` manifest claim | the string **as written in `mach.toml`** | `emit.register_link_claims`, `req.symbols[i]` verbatim |

On ELF and PE these coincide, because the object symbol for `ext fun Sleep` is `Sleep`. On Mach-O the object symbol is `_Sleep`, so they do not.

Demonstrated on `dev` by cross-building the same source three ways. The decorator route's own diagnostic shows the key it used:

```
$ mach build . --target linux
error: import 'Sleep' is attributed to library 'kernel32' ... and to library 'user32' ...
$ mach build . --target windows
error: import 'Sleep' is attributed to library 'kernel32' ... and to library 'user32' ...
$ mach build . --target darwin-x86_64
error: import '_Sleep' is attributed to library 'kernel32' ... and to library 'user32' ...
```

## Consequences on Mach-O

The linker looks the attribution up under the **object symbol** (`linker.mach`, `session.import_library(s, cn.canonical)`, where `cn.canonical` derives from `sym.name`). So on Mach-O:

1. **A `symbols = ["Sleep"]` claim does not attribute anything.** It registers `Sleep`; the linker asks for `_Sleep`; the import stays unattributed and, since Mach-O is a two-level namespace, the link fails with the very "no `#[library]` attribution" error the claim exists to prevent. #2510's feature does not work on darwin for a symbol spelled the way a user would naturally write it.
2. **The claim-vs-decorator conflict check silently cannot fire.** The two routes occupy disjoint key spaces, so a manifest claim contradicting a decorator is not detected on Mach-O at all. #2529 has just made the decorator-vs-decorator case a hard error and the docs describe "claimed only once" as a global invariant, which is again not true as written on this target.
[...]

PR #2644, opened 2026-08-07
fix(link): project #[library] attributions after the target selects its arm

### LINKED ISSUE #2639: link: a target-conditional #[library] attribution binds to the arm the target does not take
Found while fixing #2630 (the comptime-arm exemption to #2529's conflict check). Independent of both: it reproduces on `dev` from before #2529 landed, and #2630 deliberately left it exactly as it was.

## What

A target-conditional `#[library]` attribution binds to the **wrong arm**. `union.register_import_libraries` projects attributions onto link names before any arm is selected, walking every branch of a declaration-scope `$if` chain, so the **last arm walked wins** regardless of which arm the target actually takes.

## Repro

```mach
    #[library("kernel32")]
    ext fun Sleep(ms: u32);
}
$or {
    #[library("c")]
    ext fun Sleep(ms: u32);
}
pub fun call() { Sleep(0); }
```

Built for **windows**, with a `[link.kernel32]` entry declared and no `c`:

```
```

`c` is the `$or` arm, the one a windows build does not take. The attribution the windows arm declares is overwritten by the arm that was never selected.

## Why it matters

This is the idiomatic way to write a cross-platform `ext` binding, and it is exactly the shape #2630's fixture uses. The failure is loud when the losing arm's library is absent from the link, which is the common case and how this was found. It is **silent** when both arms' libraries happen to be present, where the import is simply attributed to the wrong one and the loader is asked for a symbol from a library that does not export it.

## Fix

Project attributions **after** arm selection rather than before, so only the selected arm's declarations register.

That is a pass-ordering change, which is why #2630 did not attempt it. Note the constraint #2630 documented while investigating: evaluating comptime conditions inside `register_import_libraries` is the wrong repair. That pass runs before sema, and descending only into the selected arm made `mach.lang.driver:layout_intrinsic_remaining_positions` fail, because a `$size_of` in a declaration-scope `$if` condition is sema's diagnostic to report and evaluating early front-ran it. So the fix is to move when the projection happens, not to teach the projection to evaluate.

## Acceptance

- A target-conditional attribution registers the library named by the arm the target selects, and no other.
- The repro above builds for windows against a `[link.kernel32]` entry with no `c` dependency declared, and for linux against `c` with no `kernel32`.
- A test asserts the attribution for **both** targets from one source, so a fix that merely reverses the winner is not mistaken for a fix.
- #2529's cross-module conflict detection and #2630's exclusive-arm exemption both still hold.

octalide:
Closes #2639

## What

`union.register_import_libraries` ran at project load and walked **every** branch of a declaration-scope `$if` chain, so the last arm walked won regardless of which arm the target takes.

Against `dev`, building the issue's repro for windows with `[link.kernel32]` declared and no `c`:

```
```

`c` is the `$or` arm. With this PR the same build succeeds and binds `kernel32`.

As the issue notes, the loud form is the lucky one. When both arms' libraries happen to be present in the link, the import is silently attributed to the wrong provider and the loader is asked for a symbol that library does not export.

## The fix is *when*, not *what*

The projection moves from `build_project_impl` (project load) to the head of `passes.run_lower_pass`. Two preconditions have to hold there and only there:

- **After sema**, so the build target's comptime-arm selection has been settled and the walk can descend into the one live arm.
- **After PH_SEMA's diagnostic gate** (`run_phase` gates on `diagnostic.has_errors` between phases), so a condition sema rejected is reported by sema rather than re-evaluated here.

That second point is the constraint #2630 recorded: evaluating conditions *inside the old position* was the tempting repair and the wrong one, because it front-ran `mach.lang.driver:layout_intrinsic_remaining_positions` — a `$size_of` in a declaration-scope `$if` condition is sema's diagnostic to report. Moving the pass rather than teaching it to evaluate is what makes folding safe.

`run_lower_pass` is the seam every consumer already goes through: the engine's `lower_phase` and every test helper call it directly, so there is one place the projection happens rather than one per driver.

## #2630's machinery is deleted, not left behind

Selecting the arm subsumes the exclusive-arm exemption: two arms can never both register, so `ArmStep` / `OwnedPath` / `arms_exclusive` / `push_arm` and the per-attribution arm-path allocation are all gone. #2630 was explicitly the interim repair; leaving its machinery as unreachable surface would be worse than removing it. The behaviour it protected is asserted more strongly now — see the test below.

## Tests

`ut_attr_probe` is a new helper that drives a scaffolded fixture through load, sema **and lower** for a named target, then reports the library a link name ends up attributed to. The four existing attribution tests moved onto it, because a probe that stops at project load now sees an empty map and would pass vacuously — exactly the "reports more than it verifies" shape.

- **`mach.lang.driver:conditional_library_binds_the_selected_arm`** (new) — one source, asserted for **both** targets: `windows` must yield `kernel32` and `linux` must yield `c`. Asserting both from one source is what stops a fix that merely reverses the winner from passing.
- `comptime_arm_does_not_mask_cross_module_conflict` — a second module contradicting the chain coexists with it under every target, so it is still a conflict naming both.
- `duplicate_library_attribution_names_both` and `identical_library_attribution_accepted` — #2529's cross-module detection, unchanged in behaviour.

End-to-end, with a compiler built from `dev`:

```
$ mach-dev build . --target windows
```

With this PR, that build exits 0.

At the unit level, with only the comptime-if descent reverted to walk every arm:

```
failures:
  mach.lang.driver:conditional_library_binds_the_selected_arm  ./src/lang/driver/tests.mach:11873  (exit 1)

1219 passed, 1 failed, 1220 total
```

With the fix: `1220 passed, 0 failed, 1220 total`.

## Note on scope

`register_import_libraries` still iterates every loaded module rather than `p.topo`. Iterating the compiled set would arguably be more correct (a module no artifact compiles should contribute no attribution), but that is a separate change in blast radius and is not what #2639 describes, so I left it.

## Verification

- `mach test .` green (1220/1220)


octalide:
Merged current `dev`. One conflict in `src/lang/driver/tests.mach`, worth flagging because it touches the same concern as #2636.

`dev` had separately fixed the host-blindness in `duplicate_library_attribution_names_both` and `identical_library_attribution_accepted` by *loosening* the assertions — matching `Sleep' is attributed to...` with the prefix left out, and trying both `Sleep` and `_Sleep` as the lookup key — because the link name carries the host format's symbol prefix.

This PR's `ut_attr_probe` builds for a **named** target rather than the host's, so the link spelling is pinned and the assertions can stay exact instead of accommodating whichever host ran them. I kept that form and carried `dev`'s reasoning into the comments, since the underlying observation is right and worth keeping recorded.

Re-verified on the merged tip: `mach test .` 1221/1221, full linux int suite green, self-host fixpoint holds.

octalide:
Merged current `dev` again. The conflict is worth describing because it is a design overlap, not a textual one, and I removed two functions that had just landed.

`dev` has since added `ut_names_import` and `ut_import_attribution`, which accommodate the host's link spelling by trying both `Sleep` and `_Sleep`. They exist because a link name carries the target's symbol prefix and these tests were running against whatever target the host resolved to.

`ut_attr_probe` in this PR removes the cause instead: it builds for a **named** target, so the spelling is a property of the target under test rather than of the machine running the suite, and the assertions can be exact. That left both helpers with no callers, so I deleted them — the same call I made for #2630's `arms_exclusive`, which the real fix also made unnecessary. Flagging it explicitly since the code is a day old and I would rather be told if you want it kept.

To make sure this does not *lose* coverage — the helpers were reached on the darwin legs, and pinning targets could have meant darwin's prefix path stopped being exercised here — `conditional_library_binds_the_selected_arm` now asserts three targets from one source:

```
windows       -> Sleep  attributed to kernel32   (the $if arm)
linux         -> Sleep  attributed to c          (the $or arm)
darwin-x86_64 -> _Sleep attributed to c          (the $or arm, prefixed key)
```

So the prefixed spelling is asserted directly and on every leg, rather than only where the host happens to be a Mac. Removing the two helpers also reunited `ut_str_contains` with its doc comment, which they had been inserted between.

Re-verified: `mach test .` 1221/1221, full linux int suite green, self-host fixpoint holds.

PR #2648, opened 2026-08-07
fix(link): key a manifest symbols claim on the target's link name

### LINKED ISSUE #2636: link: a manifest symbols = [...] claim keys on the source spelling while #[library] keys on the link name, so claims do not attribute on Mach-O

octalide:
Closes #2636

## What

The two routes that attribute an import to a library keyed the session map on different spellings: a `#[library]` decorator on the declaration's **link name**, which carries the target's C symbol prefix, and a `symbols = [...]` claim on the string **as written in `mach.toml`**. On ELF and PE those coincide. On Mach-O the object symbol is `_Sleep`, so they do not.

Reproduced against `dev` with a cross-built darwin image whose only unattributed import is claimed as `symbols = ["Sleep"]`:

```
error: dynamic import '_Sleep' has no #[library] attribution; ... claim it with symbols = ["_Sleep"] on the [link.<entry>] naming its library; candidate libraries: System (/usr/lib/libSystem.B.dylib), curses (/usr/lib/libcurses.dylib)
```

The claim registered `Sleep`, the linker asked for `_Sleep`, so #2510's feature simply did not work on darwin — and the conflict check could not fire either, the two routes occupying disjoint key spaces.

## The manifest is the side that moved

A claim is a source-level name, so it now goes through the same prefix rule a declaration's link name goes through. That rule is a single `project.link_name_for`, which both routes call; `union.ext_link_name` is now `intern_span` plus that call. Two spellings of one rule is what produced the defect, so there is now one.

### Already-prefixed claims: normalized, not accepted or rejected

The prefix is only ever **added**, and applied **unconditionally**. I did not special-case a claim that already looks prefixed, because "already prefixed" is not decidable from a spelling: `_exit` is a real C symbol whose Mach-O object spelling is `__exit`, so rejecting or stripping a leading `_` would break the source spelling of every symbol whose C name starts with one.

Unconditional application is the only rule that keeps one manifest spelling correct on every target. Writing the mangled form yourself therefore does not work — on darwin `symbols = ["_Sleep"]` claims `__Sleep`, which nothing imports, and fails loudly as an unattributed import. Accepting both spellings, the alternative, is what would make manifests silently non-portable again.

### The #2528 diagnostic no longer sends users into that trap

It named the **object** symbol in its `symbols = [...]` suggestion, so on darwin it said `symbols = ["_Sleep"]` — following the error produced a manifest correct on darwin and wrong everywhere else, and after this change correct nowhere. It now suggests the source spelling via `source_spelling`, the exact inverse of the add-one-prefix rule. The message still *names* the object symbol, since that is what the link asked for.

### Two in-tree manifests were carrying the mangled spelling

`macho-got-x86_64` and `macho-abs-bind-x86_64` claimed `___stderrp` / `___stdoutp` — the darwin-mangled form, which is what the verbatim registration required. Their C fixture declares `__stderrp`. Both are now written portably as `__stderrp` / `__stdoutp`. That those two existed is the defect in miniature.

## Tests

**`int/surface/macho-import-claim`** (new) — the acceptance's Mach-O case. Cross-builds darwin-x86_64 on the **linux** leg and reads the bind tables host-side, so it runs per-PR rather than only where someone happens to author on a Mac. That was the point the issue makes about why this survived.

It needed a producer, so `macho-imports` is the Mach-O analogue of `pe-imports`: every `<dylib>:<symbol>` row from the bind and lazy-bind tables. Golden:

```
libcurses:_Beep
libSystem:_ExitProcess
libSystem:_Sleep
```

Two dylibs are deliberate, for the reason `pe-import-claim` gives: with one dependency, a linker that bound every import to the first would still produce the expected table, so the case would pass while routing was broken. `Sleep` is claimed by `[link.System]`, `Beep` by `[link.curses]`, and the decorator path (`ExitProcess`) must land in the same dylib as the manifest-claimed one. Neither name is spelled with a `_` anywhere in the fixture.

**`mach.lang.build.emit.register_link_claims:pins_and_rejects_conflicts`** — extended to run the claim route against a real darwin target vtable as well as linux. On darwin it asserts the same manifest string registers under `_Sleep`, does **not** register under `Sleep`, and that the conflict check fires — which it could not while the key spaces were disjoint.

**`mach.lang.be.linker.unattributed_import_message:names_both_routes`** — asserts the suggestion is `symbols = ["Sleep"]` and specifically *not* `symbols = ["_Sleep"]` when the object symbol is `_Sleep`, plus that a symbol not carrying the prefix is suggested unchanged.

The new int case, run against a compiler built from `dev`:

```
FAIL surface/macho-import-claim [linux/debug] (build)
    error: dynamic import '_Sleep' has no #[library] attribution; ... claim it with symbols = ["_Sleep"] ...
FAIL surface/macho-import-claim [linux/release] (build)
int: 2 failure(s) on linux
```

That is both halves of the issue in one line: the claim did not attribute, and the diagnostic pointed at the non-portable spelling.

## Verification

- `mach test .` green (1221/1221)


octalide:
CI caught a defect in my own test, which is worth recording because it is the same class as the issue.

`int linux` failed on the golden while the case passed locally. The content was identical; only the order differed:

```
@@ -1,3 +1,3 @@
-libcurses:_Beep
 libSystem:_ExitProcess
 libSystem:_Sleep
+libcurses:_Beep
```

That is locale collation. Under C, `libSystem` sorts before `libcurses` (`S` is 0x53, `c` is 0x63); under a UTF-8 locale it is the other way. My producer used a bare `sort -u`, so the golden depended on the runner's environment rather than on the emitted image — a test that passes where it was authored and fails elsewhere, which is exactly what #2636 is about.

`produce_pe_imports` already pins `LC_ALL=C sort`; mine now does too, and the golden is re-blessed to the C order. Verified under both collations:

```
$ LC_ALL=en_US.UTF-8 int/run.sh --target linux --filter macho-import-claim ...
PASS surface/macho-import-claim [linux/debug]
PASS surface/macho-import-claim [linux/release]
```

Re-verified: `mach test .` 1221/1221, full linux int suite green, self-host fixpoint holds.

Issue #2538, opened 2026-08-05
diag: IR verifier failures are emitted with no span, making middle-end errors nearly undiagnosable


octalide:
Split out of #2523, where this was a secondary note. The typing bug there is one instance; the diagnostic behavior is the general problem and outlives any single defect.

## What

IR verifier failures are reported with no span. The message names the check that fired and nothing else:

```
error: constant typing: constant carries an incompatible type
```

No file, no line, no function. Every other diagnostic class in the compiler carries a location — during the same session, resolver and sema errors all pointed at source:

```
error: no symbol `allocator` exported by `std.allocator.page`
  --> ./src/compress/probe.mach:99:35
error: unresolved identifier `true`
  --> ./src/compress/repro.mach:5:19
```

so the verifier is the outlier, not the norm.

Why it is wrong: a diagnostic exists to tell a developer where to look. This one tells them a property was violated somewhere in the program. In a single-function test case that is a minor annoyance; in a real module it is close to undiagnosable. Localizing #2523 in a 900-line module took a scripted bisection of roughly forty compiles — truncating the module function by function to find which function carried it, then cutting that function branch by branch. The message is identical no matter which function is at fault, so there is nothing to grep for and nothing to narrow by inspection. A developer without a scripting harness would reasonably conclude the compiler was broken rather than their code.

This is not the verifier author's oversight. The root cause is structural and already documented in the debug info milestone: the source-location chain is severed at the AST to IR seam, "IR Instruction/Value carry no location and the builder emit_* API takes no span". The verifier cannot report a span because by the time it runs, there is no span to report. Diagnostics are therefore a second consumer of the same location plumbing that debug info needs, and worth naming as one — the foundation work described in that milestone would fix this as a side effect, and until then every middle-end check that fires will have the same problem.

## Fix

Candidate designs:

1. **Carry source locations on IR instructions** and have the verifier report the location of the offending instruction. This is the correct fix and is the same plumbing the debug info foundation requires; doing it once serves both. Cost: touching the builder `emit_*` API and every construction site.
2. **Carry a location on the enclosing function only**, and have the verifier report "in function `f` (file:line)" without pinpointing the instruction. Much cheaper than (1), and closes most of the gap — naming the function turns a forty-compile bisection into a five-second read. Could ship ahead of (1) and be subsumed by it later.
[...]

PR #2663, opened 2026-08-07
fix(diag): name the opcode, function and position in a backend refusal

### LINKED ISSUE #2656: diag: the backend's unsupported-instruction refusal names neither the opcode nor the function
Reported twice from the SPIR-V track, where it made the first hour of each change mostly bisection.

## What

Every unsupported shape in the SPIR-V backend produces one line:

```
error: instruction is not yet supported by the SPIR-V target
```

It names neither the opcode nor the function. In a module of any size there is nothing to grep for and nothing to narrow by inspection, so localizing means truncating the source until it stops.

This is the same complaint #2538 made about the IR verifier, which has since been fixed and now names the module, function and position. The backend refusal is the remaining instance.

## Why it is not just the SPIR-V backend's problem

The fix wants an opcode-name function on `mir.mach`, which does not have one. That is a shared seam, which is why the SPIR-V track reported it rather than adding it. Any other backend refusal wanting to name an opcode has the same gap.

## Fix

Add an opcode-name function to `mir.mach` and have the backend refusal name the opcode and the enclosing function. #2538 established that the location plumbing already exists on IR instructions, so a position may be available too.

## Acceptance

- An unsupported instruction is refused with the opcode name and the function it appears in.
- `mir.mach` exposes an opcode-name function other backends can use for the same purpose.
- A test asserts the opcode name appears, so it cannot regress to the bare message.

octalide:
Closes #2656

## What

Every unsupported shape in the SPIR-V backend produced one sentence with no opcode, no function and no position. Localizing one meant truncating the source until it stopped failing.

Before:
```
error: addressed memory is not yet supported by the SPIR-V target: ...
```

After:
```
error: addressed memory is not yet supported by the SPIR-V target: ... (MIR_ALLOCA in sum_pair, at ./src/main.mach:15:5)
```

## The seam

`mir.opcode_name(op) str` — the source spelling of a MIR opcode. Total over the space with a fallback, so no caller guards, and a new opcode that forgets its arm degrades to `<unknown MIR opcode>` rather than reading as some other instruction.

It returns the constant's own name (`MIR_DIV_S`) rather than a mnemonic, deliberately: the reader of a backend refusal is usually the person about to implement the missing arm, and that is the token they grep the compiler for. There was no existing MIR opcode-name table to reuse — the IR printer's is for IR opcodes, the ISA printers' are for machine opcodes — so this is new rather than duplicated.

`mir.instr_refusal(alloc, srcmap, fn_name, mi, what) str` composes `<what> (<opcode> in <function>, at <path>:<line>:<col>)`, the shape #2538 settled on for the IR verifier.

## On your two checks

**1. The location does reach here.** `MirInstr.loc` survives lowering to the backend, and `MirModule` already carries the `srcmap`, so the refusal names an exact position and not merely a function. Where a location is genuinely absent (synthesized MIR, or no source map) the message says `no source location on this instruction` rather than dropping the clause — the degradation #2538 ruled out, because a message that quietly omits the position reads as though it was never plumbed.

One layering choice worth flagging: `instr_refusal` takes the function name **already resolved** rather than a `MirFunction`. A MIR function carries only its mangled link name (`_M4case4mainN8sum_pair`), and the SPIR-V backend holding the IR beside it can resolve the readable `sum_pair`. MIR owns the opcode and the location; it does not own the vocabulary. The backend falls back mangled-name → stand-in, since this is only ever read by a diagnostic.

**2. The second caller is the mos6502 legalize pass.** Its `unsupported_message` claimed *in its own doc comment* to name the opcode and width, and named neither — `width` was entirely unused. It now says both through `mir.opcode_name`, so the two passes name an instruction identically and a reader who greps one finds the other. Its family prose is kept, because that says what is missing in a way an opcode alone does not: `MIR_SHL` is unsupported there only through certain operand shapes.

That needed an allocator threaded through five counting functions (`expansion_count`, `*_piece_count`), all of whose callers already had one, so it is mechanical. `legalize.run` also holds the `MirModule`, so it *could* carry a location too — I stopped short of that, since it is more plumbing than #2656 asks for.

## File ownership

Everything is in `mir.mach` (additive), `spirv/emit.mach`, and `legalize.mach`. **No edits in `mir/lower.mach`**, so no sequencing needed. I checked for open PRs touching any of the three before starting and found none.

The thirty SPIR-V refusal sites are unchanged: the emitter stamps the function and instruction it is working on onto `Emit` at its two chokepoints, the way MIR itself stamps `loc` at `push_instr`. The terminator path stamps too — without it a terminator refusal would name the last *body* instruction, which is a wrong attribution rather than a missing one, and worse than the bare message.

## Tests

- **`mir.opcode_name:total_and_distinct`** — sweeps all 83 declared opcodes and asserts none reaches the fallback and no two share a name. Naming the *wrong* instruction is worse than naming none, so totality-and-distinctness is the property, not "some names come back".
- **`mir.instr_refusal:names_opcode_and_function`** — the opcode and function reach the message, and a locationless instruction produces the stated degradation rather than a silently shortened message.
- **`int/surface/spirv-unsupported-names-opcode`** — the real diagnostic end to end.

The int fixture indexes its array with a **runtime** value on purpose. With a constant index the release pipeline promotes the array into registers and the refusal never fires, so the case asserted the message in `debug` and asserted nothing at all in `release`. I found that when the first blessing failed on the release leg.

The int case against a compiler built from `dev`:

```
@@ -1 +1 @@
-... (MIR_ALLOCA in sum_pair, at ./src/main.mach:15:5)
+... and only as a whole-variable read or write
int: 2 failure(s) on linux
```

The unit test, with one arm deleted from the table the way a forgotten opcode would leave it:

```
failures:
  mach.lang.be.codegen.mir.opcode_name:total_and_distinct  ./src/lang/be/codegen/mir.mach:1659  (exit 1)
1229 passed, 1 failed, 1230 total
```

(My first attempt at that deletion silently didn't match on whitespace and the suite stayed green — which is the reason for demonstrating failure rather than assuming it.)

## Verification

- `mach test .` green (1230/1230)

### DIFF
--- a/int/surface/spirv-unsupported-names-opcode/case.conf
+++ b/int/surface/spirv-unsupported-names-opcode/case.conf
@@ -1,0 +1,12 @@
+# a SPIR-V refusal must say WHICH instruction it refused (#2656). Every unsupported
+# shape used to produce one sentence with no opcode, no function and no position,
+# so localizing one in a module of any size meant truncating the source until it
+# stopped failing - which is how the SPIR-V track spent the first hour of each of
+# its four changes.
+#
+# The observable is the diagnostic itself: the opcode name, the enclosing function,
+# and the source position of the offending instruction. Asserting only that the
+# build fails would pass against the bare message this case exists to rule out.
+run: build-fails
+targets: linux
+build-target: spirv
--- a/int/surface/spirv-unsupported-names-opcode/expect.spirv.txt
+++ b/int/surface/spirv-unsupported-names-opcode/expect.spirv.txt
@@ -1,0 +1,1 @@
+error: addressed memory is not yet supported by the SPIR-V target: a module-scope variable is reachable only when it carries a shader interface role (`#[input]`, `#[output]`, `#[builtin]`, or `#[uniform]`), and only as a whole-variable read or write (MIR_ALLOCA in sum_pair, at ./src/main.mach:15:5)
--- a/int/surface/spirv-unsupported-names-opcode/expect.spirv.txt
+++ b/int/surface/spirv-unsupported-names-opcode/expect.spirv.txt
@@ -1,1 +1,1 @@
-error: addressed memory is not yet supported by the SPIR-V target: a module-scope variable is reachable only when it carries a shader interface role (`#[input]`, `#[output]`, `#[builtin]`, or `#[uniform]`), and only as a whole-variable read or write (MIR_ALLOCA in sum_pair, at ./src/main.mach:15:5)
+error: a local allocation is not supported by the SPIR-V target: a shader has no stack, so a value needing addressable storage has nowhere to live (MIR_ALLOCA in sum_pair, at ./src/main.mach:15:5)
--- a/int/surface/spirv-unsupported-names-opcode/mach.toml
+++ b/int/surface/spirv-unsupported-names-opcode/mach.toml
@@ -1,0 +1,34 @@
+[project]
+id = "case"
+version = "0.0.0"
+src = "src"
+# a fixed output root: the delivery is a module TREE, not a single binary, so the
+# producer finds the modules under the same directory `-o` names rather than
+# through a per-target/profile template it would have to reconstruct.
+out = "out/int"
+
+# no `of` key. that is the whole point of the case: the target spells only the
+# tuple, and the toolchain must resolve the object format that carries a finished
+# SPIR-V module on its own (#2245).
+[target.spirv]
+isa = "spirv"
+os  = "freestanding"
+abi = "spirv"
+
+[profile.debug]
+opt = 0
+debug = false
+simd = "scalarize"
+
+[profile.release]
+opt = 2
+debug = false
+simd = "scalarize"
+
+[artifact.case]
+kind = "bin"
+entry = "main.mach"
+out = "bin/case"
+targets = ["*"]
+link = []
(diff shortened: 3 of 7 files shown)


octalide:
CI caught a golden collision, and the shape of it is worth recording because it is the design working rather than failing.

The SPIR-V track has since rewritten the alloca refusal's prose on `dev`, from *addressed memory is not yet supported…* to *a local allocation is not supported by the SPIR-V target: a shader has no stack…*. My golden still carried the old head, so the case diffed. The actual output on CI was:

```
error: a local allocation is not supported by the SPIR-V target: a shader has no stack, so a value needing addressable storage has nowhere to live (MIR_ALLOCA in sum_pair, at ./src/main.mach:15:5)
```

The located suffix — the thing this PR adds and the case exists to pin — was intact and correct throughout. Only the sentence the backend supplies changed, which is exactly the seam: `instr_refusal` composes with whatever the backend says, so their prose improvement and this located form are independent.

Merged `dev` and re-blessed onto the new head. Their new wording is better than the one it replaced, so the golden is stronger for it.

Re-verified on the merged tip: `mach test .` 1235/1235, full linux int suite green, self-host fixpoint holds. My stamps all survived the merge into `spirv/emit.mach`, which took 299 lines of their changes.

Issue #2662, opened 2026-08-07
isel(x86_64): float immediates are still rebuilt per use, though #2248 is closed


octalide:
#2248 is closed, and the symptom it describes is still live on x86-64.

## Evidence

`src/lang/target/isa/x64/encode.mach:3212` still says so in the code:

```
# x86-64 has no SSE immediate form, so this pair is unavoidable until
# the constants live in a pool (#2248).
```

Every float immediate is still materialized as a `MOVABS` into the R11 GP scratch followed by a `MOVQ` into the xmm register, per use. Measured independently while working #2315: the same `movabs`/`movq` pair appears three times per iteration in `rec_particle`.

It also blocks work downstream. #2640's first stage (PR #2661) removed the vector-literal reinterpret but could not remove the stack round trip, and one of the three reasons is this: the constant case wants a rodata load and there is no constant pool to load from.

## What I am not claiming

Not that #2248 was closed wrongly. It may have been closed for scope, superseded, or resolved for a narrower case than its title covers. I have not read its history closely enough to say, and whoever closed it will know. This issue exists so the live symptom is tracked somewhere open, rather than resting on a closed issue that a reader will reasonably assume is done.

If #2248 was closed deliberately with the symptom accepted, close this too and say so, and the code comment should stop pointing at a closed issue as though it were pending work.

## Why it is worth doing

Three independent consumers want it:

- **Float immediates** on x86-64, the original case: two instructions and a GP scratch per use, in loops.
- **Vector literals** (#2640): the constant case wants a rodata load, and without a pool the fallback is a stack round trip.
- **SIMD constants** generally, for the same reason.

## Acceptance

- A float immediate on x86-64 loads from a constant pool rather than being rebuilt per use, or the decision not to is recorded and the code comment stops citing a closed issue as pending.
- Emitted instruction counts before and after for a loop carrying a float constant, measured rather than asserted.
- If a pool lands, #2640's constant case is re-checked against it, since that is the other consumer already waiting.

octalide:
**Answered, and the answer is that #2248's closure was deliberate — so I am leaving the decision to you rather than acting on it.**

I opened this saying I had not read #2248's history closely enough to know whether it was closed wrongly. It was not. You closed it yourself, `NOT_PLANNED`, on 2026-07-25:

[...]

Issue #2655, opened 2026-08-07
lower: float loads and stores unconditionally materialize the address, destroying the symbol reference a logical-addressing target needs


octalide:
Third instance of the same family as #2640 and #2649, found while assessing what separates the SPIR-V backend from a genuine shader path. New, and the cheapest of the three.

## What

A scalar `f32` or `f64` interface variable does not work on SPIR-V, in either direction. A `f32x4` does, and an `i32` does:

```
REFUSED | scalar f32 output: STORE only
REFUSED | scalar f32 input:  LOAD only
OK      | scalar i32 load+store (control)
```

At IR level all three are identical: `load f32 @sym`.

## Cause

`lower_load` / `lower_store` materialize the address into a vreg **when the value is float**, because x86 cannot take a float operand off a symbol, and rewrite the operand to `op_mem_value(base, 0)`. That destroys the direct symbol reference SPIR-V needs.

It is unconditional. The materialization is a machine constraint of one target family applied to every target, exactly as #2649's displacement folding and #2640's memory round-trip are.

## Why it is worth doing first of the three

An `f32` varying is completely ordinary in a shader: alpha, a fog factor, a depth value, any scalar interpolated between stages. So half the scalar interface surface is dead, not an exotic corner. And the fix is the same conditional the other two need, on a smaller site.

## Fix

Make the float address materialization conditional on the target requiring it, rather than unconditional. A target that can name a symbol directly should keep the direct reference.

Worth doing as one piece of work with #2640 and #2649 rather than three, since all three are the same decision (MIR discarding structure for a machine reason that a logical-addressing target then cannot reconstruct) at three sites, and each also removes wasted work on native targets.

## Acceptance

- A vertex stage writing a scalar `f32` output and a fragment stage reading a scalar `f32` input compile and pass `spirv-val --target-env vulkan1.3`.
- A native target's emitted code is unchanged where it genuinely needs the materialization, proven rather than asserted.

Issue #2649, opened 2026-08-07
lower: MIR folds a GEP's index list into a byte displacement, which a logical-addressing target cannot invert


octalide:
Found while implementing #2571 (SPIR-V interface variables and descriptors). Like #2640 it is not a SPIR-V defect: it blocks that target, but the fix is in the middle end.

## What

MIR folds an aggregate member walk into a **byte displacement** before the back end sees it:

```
%0 = gep ptr @camera, 0: i64, 1: i64     ->   MIR_GEP with disp = 16
%1 = load f32x4 %0
```

The index list `0, 1` is gone by then; only `16` survives.

SPIR-V reaches a member through `OpAccessChain`, which names the member's **ordinal**, not its offset. So a target under the logical addressing model needs the indices the lowering discarded.

## Why the displacement cannot simply be inverted

An offset does not determine an ordinal. Unions overlap by construction, so several members share offset 0. Nested aggregates share offsets with their first members. Padding belongs to no member at all. Recovering an index list from a displacement would be guessing dressed up as arithmetic, and the guess would be silently wrong exactly where it matters, which is why PR #2645 refuses by name instead of inverting.

## Consequence today

A uniform block is **declarable and bindable but not readable member-wise**. PR #2645 landed the declaration side and it is validator-clean: `Block`, per-member `Offset`, `DescriptorSet` and `Binding` all verified with `spirv-val`. A vertex stage reading a `Location` input and writing `BuiltIn Position` works end to end. What is refused is reading a **member** of a uniform block, which is most of what a uniform block is for.

So #2571 is half-met and stays open on this.

## Fix

Keep the GEP's index list through MIR lowering for a target that needs it, rather than collapsing to a displacement unconditionally. The displacement is the right representation for a flat-addressing target and should stay the default there; a logical-addressing target needs the structural form preserved.

Worth deciding deliberately, since it is a representation change in the middle end: whether the index list rides alongside the displacement on every target, or whether the collapse is gated on a machine-model fact the way `has_mul` and `has_var_shift` already are. The second matches the existing precedent for "this target cannot do the usual thing".

## Related

- #2640, the same shape one layer down: a vector literal lowers through an array `alloca` read back as a vector, which logical addressing also cannot express. Both are cases of MIR discarding structure that a logical-addressing target requires, and whoever takes one should read the other.

## Acceptance

- A fragment stage reading a member of a uniform block emits an `OpAccessChain` with the member's ordinal and passes `spirv-val`.
[...]

Issue #2640, opened 2026-08-07
lower: a vector literal round-trips through an array alloca, which SPIR-V's logical addressing cannot express


octalide:
Found while implementing #2569 (SPIR-V vector types). It is not a SPIR-V defect: it blocks that target, but the fix belongs in the middle end and the benefit is not target-specific.

## What

A vector literal and an uninitialized vector local both lower **through memory**:

```mach
val v: f32x4 = f32x4{1.0, 2.0, 3.0, 4.0};
var z: f32x4;
```

```
%0 = alloca [4]f32 1: i64
     store f1: f32, ...        (once per lane)
%10 = load f32x4 %0: ptr
```

That allocates an **array** and then reads it back as a **vector**, reinterpreting the allocation through its pointer.

## Why it is a problem beyond an inefficiency

On a native target this is a stack round-trip for what is usually a compile-time constant: four stores and a load to materialize a value the backend could have emitted as a constant. That alone is worth removing, and it is on the same axis as #2315, where a `Result` reached memory for no reason.

On **SPIR-V it is not expressible at all.** The logical addressing model gives a pointer exactly one pointee type and provides no pointer bitcast, so there is no legal encoding of "allocate as `[4]f32`, load as `f32x4`". This is not an emitter gap that a more complete emitter would close: no SPIR-V back end can accept this shape. The backend currently refuses it up front with a diagnostic naming the construct, rather than reporting the `alloca` that happens to begin the sequence.

Consequently #2569's acceptance item "an `int/surface/` case covering f32x4 **construction**, lane arithmetic, and extract" cannot be met. Lane arithmetic and extract landed in PR #2638; construction is blocked here.

## Fix

Lower a vector literal to a value rather than to storage: a constant when every lane is constant, otherwise a chain of `OP_VEC_INSERT` starting from a zero or undef vector. An uninitialized vector local wants the same treatment as any other uninitialized scalar rather than an array allocation reinterpreted.

The constant case is the common one and is also the one that removes the stack traffic on every native target, so it is worth doing even independently of SPIR-V.

## Acceptance

- `f32x4{...}` with constant lanes lowers to a vector constant, with no `alloca` and no store/load round trip.
- `f32x4{...}` with runtime lanes lowers to an insert chain, still with no memory.
- `var z: f32x4;` does not allocate an array.
- The `int/surface/` SPIR-V case in #2569 covering construction becomes writable, and #2569's remaining acceptance item is met.
[...]

PR #2671, opened 2026-08-07
feat(mir): keep a GEP's index list on a logical-addressing target

octalide:
Refs #2649 — deliberately **not** `Closes`. This is the enabling half; building the access chain is backend work, detailed below.

## Cause

MIR folded an aggregate member walk into a byte displacement before the back end saw it: `gep ptr @camera, 0, 1` arrived as `disp=16` and nothing else. On a byte-addressed machine that *is* the address. SPIR-V reaches a member through `OpAccessChain` naming its **ordinal**, and an offset does not determine one — unions overlap, a nested aggregate shares an offset with its first member, and padding belongs to no member. Inverting it would be guessing dressed as arithmetic, wrong exactly where it matters.

## Fix

`lower_gep` branches on `flat_addressing`. Flat targets take the identical path. A logical target takes `lower_gep_structural`, which emits the walk unchanged.

**The operand layout is the contract**, and it is what the emitter reads to tell the two forms apart:

| form | operands | count |
|---|---|---|
| flat | `[dst, mem]` | exactly 2 |
| structural | `[dst, base, idx0, idx1, ...]` | more than 2 |

The source pointee type is deliberately **not** carried. It is recoverable at the consumer from what the base names — a global's declared type, or the allocated type of the `MIR_ALLOCA` a local base defines — and duplicating it would be a second source of truth for a fact the base already fixes.

This is the second consumer of `flat_addressing` (#2655) and it needed no reshaping, which was the point of landing that as an axis rather than as a conditional.

## Flat-target output unchanged, proven not asserted

The strongest form available: **the entire 8.2 MB compiler binary, built from identical source by each compiler, is byte-identical.** That is every gep in the compiler's own body.

Plus gep-heavy source — nested records, arrays with runtime indices, a union, pointer-parameter walks, reads and writes:

| target | debug | release |
|---|---|---|
| x86_64 | identical (72 instrs) | identical (74) |
| aarch64 | identical (82) | identical (95) |
| riscv64 | identical (101) | identical (100) |

## No silent miscompile in the interim

Worth stating because it was the real risk of landing the enabling half alone. A member read addresses through the GEP's **vreg**, not a symbol, so `interface_target` never resolved it as a whole-variable access and cannot start to now. Verified directly: `camera.view` still refuses by name after this change rather than emitting a whole-block load.

I corrected the two refusal messages that named the fold as the cause, since that is no longer true on that target and they would now send a reader to the wrong problem — the same failure mode their own comment warns about.

## Tests

`mach.lang.driver:gep_index_list_survives_on_logical_addressing` lowers a member walk **all the way to MIR** and asserts the operand count on both sides: more than 2 on spirv, exactly 2 on x86_64, aarch64 and riscv64. Asserting both sides is what makes it a regression test rather than a feature test.

**Mutation evidence — both directions of the gate are load-bearing:**

| mutation | result |
|---|---|
| drop the gate (never keep the list) | 1233 passed, **1 failed** |
| invert the gate (keep it on flat targets too) | 1233 passed, **1 failed** |

I added `ut_manifest_spirv` rather than adding a spirv target to `ut_manifest`, so a new target here cannot perturb the ~180 tests sharing that fixture.

## What the SPIR-V backend needs (not done here)

`addressed_memory_refusal` now names it precisely. Concretely: on `MIR_GEP` with `operand_count > 2`, map the base symbol to its variable id, map operands 2 onward to constant ids, walk the base's IR type with those ordinals to get the member type, intern a pointer to it in the right storage class, and emit `OpAccessChain`. The subsequent load or store then addresses the chain result.

Per #2665's finding this unblocks three consumers: uniform block member reads, storage buffer member access, and — after #2661 removed the type-punned allocation — the per-lane geps that are all that remains of #2640's vector literal blocker.

**On the seed question:** this does **not** make a vector nameable as an SSA value. It is about addressing, not constants. `VAL_CONST_AGG` is still global-initializer-only and `const_type_ok` still rejects `VAL_CONST_INT` with a vector type, so the insert-chain seed remains blocked independently, and #2662 still gates the native constant case.

## Status

- `mach test .` — 1239 passed, 0 failed
- `int/run.sh --target linux` — all cases passed, all five spirv cases included
- release self-host fixpoint, three generations — R2 and R3 byte-identical

### DIFF
--- a/src/lang/be/codegen/mir/lower.mach
+++ b/src/lang/be/codegen/mir/lower.mach
@@ -2537,6 +2537,15 @@ fun lower_gep(ctx: *lctx.LowerCtx, mb: *mir.MirBlock, inst: *instr.Instruction, 
         ret R.err[bool, str]("mir.lower_ir: gep with indices is missing its source pointee type");
     }
 
+    # a logical-addressing target reaches a member by ORDINAL, not by byte, so the
+    # walk is kept rather than folded (#2649). the fold below is a byte-space
+    # transformation and is not merely slower there, it is unrecoverable: an offset
+    # does not determine an ordinal, because unions overlap, a nested aggregate
+    # shares an offset with its first member, and padding belongs to no member.
+    if (!ctx.tgt.model.flat_addressing) {
+        ret lower_gep_structural(ctx, mb, inst, dst, base);
+    }
+
     # walk the indices, accumulating a constant displacement and at most one
     # scaled runtime index. a struct level needs a constant field ordinal.
     var disp:      i64 = 0;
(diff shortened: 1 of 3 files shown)


octalide:
Reviewed from the consumer side — I own `spirv/emit.mach` and will be the one building the access chains from these operands.

**The shape is right and I verified the central claim rather than trusting it.** Reusing the existing `flat_addressing` fact from #2655 instead of adding a parallel flag is the right call, and distinguishing structural from flat by `operand_count` alone means a consumer needs no new discriminator.

Machine-target output is genuinely untouched. I built **identical source** (dev @ `46477b2e`) with a dev-built compiler and with this branch's compiler: byte-identical, 8249344 bytes both. Plus 1239 tests, full linux int suite, and fixpoint `b == c`.

I also confirmed all four SPIR-V refusal shapes still behave correctly under the new operand layout — uniform member read, aggregate index, vector literal, vector zero-init. `address_origin` reads the base at operand 1, which the structural form preserves, so nothing silently changed meaning.

Thank you for updating the two refusal messages in my file rather than leaving them asserting the ordinals are lost. That is exactly the staleness this track keeps getting bitten by.

---

## One correction: the pointee-type claim is half wrong

The doc comment says:

> The source pointee type is NOT carried here. It is recoverable at the consumer from what the base names — a global's declared type, **or the allocated type of the `MIR_ALLOCA` a local base defines**

The global half is true. **The alloca half is not.** `lower_alloca` uses `aux_ty` only to compute size and alignment and then discards it:

```
val elem_size:  u32 = ir_type.byte_size(ctx.types, inst.aux_ty, ctx.tgt);
val elem_align: u32 = ir_type.byte_align(ctx.types, inst.aux_ty, ctx.tgt);
...
add_alloca_slot(ctx, sk, total, elem_align);      # size + align only
```

`add_alloca_slot` records `(size, align, SLOT_ALLOCA)`. The type is gone, and `MIR_ALLOCA` itself is `[dst, MEM(slot, 0)]`.

**Why it matters rather than being pedantic.** A local base is exactly the vector-literal case — `alloca f32x4` plus per-lane geps — which is #2640's literal half, one of the things #2649 is meant to unblock. With the ordinals but no pointee type I would know "index 2" and not "of what", and I cannot type the `OpVariable` for the allocation either. So the follow-up hits a wall precisely where this PR is aimed.

Nothing is broken today, since the back end refuses all of these anyway. But the claim will get copied forward, so it is worth fixing now, one of two ways:

1. **Have `MIR_ALLOCA` retain its allocated type.** Probably better — it is one fact recorded where the allocation is, it keeps the "recoverable from the base" contract true for both kinds of base, and #2669 wants it independently (a Function-storage `OpVariable` with a null initializer needs the type too).
2. Carry the pointee type on the structural GEP after all, at the cost of a second source of truth for globals.

Either is fine by me; option 1 keeps the contract you documented actually true.

PR #2678, opened 2026-08-07
wip(inline): carry an asm payload across a clone (do not merge)

octalide:
**Draft, do not merge.** Opened so the mechanism and its measurements are
reviewable rather than sitting on a branch. #2231 is not closed by it.

Based on `1a9a1881` and now 28 commits behind `dev`; it needs a rebase
before it means anything, and one of its stated blockers has since been
removed. Recording accurate state rather than the commit message's.

## What it does

`ir.clone_asm_payload` copies an OP_ASM's `Function.asm_blocks` entry
onto the clone, remapping the map key **and** each `{name}` binding's
alloca through the inliner's instruction map, and deep-copying the
bindings array since `module_dnit` frees it per OP_ASM. That removes the
only reason `should_inline` refused an asm-bearing callee — the
restriction was mechanical, not semantic.

It also drops the same gate from the self-recursion peel, whose snapshot
preserves instruction ids so the payload rides across with an identity
remap.

`instruction:asm_is_at_least_as_conservative_as_call` pins the property
that makes this safe at all: inlining removes the CALL and leaves the
OP_ASM, so ordering survives only if the asm statement is as strong a
barrier as the call it replaced. It is — `is_pure` excludes both, which
is what CSE, DCE and LICM gate on, and `licm` and `dependence` name both
directly. Nothing else in the tree fails if someone makes OP_ASM pure,
so that test does.

## It works, and it is measured

A same-module callee holding `lock xadd` inlines into its caller with no
call left. An int case round-trips values through inlined load / store /
fetch-add helpers, identically at debug and release, on all three ISAs.

x86-64, interleaved runs, min of twelve: atomic increment 48 → 45 ms
(−6.3%), syscall-plus-atomic mix 46 → 41 ms (−10.9%). Compiler binary
+4 KB.

## Why it is not merged

**One blocker is gone.** riscv64 used to fail every release build of
anything depending on mach-std, because inlined `{name}` slots pushed a
frame past its ±2048 reach. #2674 fixed that at the cause by laying
asm-bound slots nearest the frame pointer, and riscv64 now passes its
full int suite with this branch applied. The commit message predates
that and is stale.

**One blocker remains.** aarch64's reach for a `{name}` is the unscaled
imm9's **−256** — thirty-two eight-byte slots — because slots sit below
the frame pointer and the scaled imm12 form is unsigned-only. A caller
inlining several `{name}`-heavy syscall wrappers exceeds that;
`darwin/shared.mach`'s `syscall2` has five binds and lands at −288, so
`macho-foreign-aarch64` and `macho-pie-aarch64` fail to build. #2674
raised the ceiling substantially but not past this.

The path is addressing a `{name}` from **SP rather than FP**: offsets
become positive, the scaled imm12 applies, and the reach goes from 32
slots to 4096. That is a change to `SlotFn`'s contract — its doc says
"frame-pointer-relative", and the per-target divergence it abstracts is
exactly this — plus an argument that SP is stable across an asm block.
It wants its own issue.

**And it still does not reach the atomics.** `std.sync.atomic` is
cross-module from every caller, and `should_inline` excludes extern
callees by design; lifting that needs the `Q_LOWERED_SURFACE` fingerprint
extended or importers silently reuse stale IR (#1859). The reach this
branch actually buys today is same-module callers — `shared.mach`'s
syscall wrappers have 134 such call sites — not `std.sync.atomic`.

## No orderings

Untouched. The issue's litmus-test bar (message-passing, store-buffering,
IRIW per ISA) is unmet and nothing here claims anything about the memory
model beyond the barrier-equivalence test above.

### DIFF
--- a/src/lang/me/ir.mach
+++ b/src/lang/me/ir.mach
@@ -1252,6 +1252,86 @@ pub fun clone_dbg_metadata(m: *Module, src_fn: *Function, dst_fn: *Function, src
     ret R.ok[bool, str](true);
 }
 
+# copy the inline-asm payload the OP_ASM `src_iid` carries in `src_fn` onto the cloned
+# OP_ASM `dst_iid` in `dst_fn`, remapping each `{name}` binding through `instr_map`
+#
+# The side table that made an asm-bearing callee un-inlinable (#2231): the payload lives
+# in `Function.asm_blocks` keyed by the CALLEE-side instruction id, and the instruction
+# clone carries only the pool entry, so an inlined body would have arrived with its asm
+# statement present and its body text gone. This is the `clone_dbg_metadata` move for
+# the other side table.
+#
+# TWO THINGS MUST BE REMAPPED, not one, and the second is the one worth naming. The map
+# KEY moves from the callee's instruction id to the clone's. So does every binding's
+# `instr`, which names the OP_ALLOCA owning a `{name}` local's storage: after cloning
+# that alloca is a different instruction in a different function, and a binding still
+# pointing at the callee-side id would resolve `{name}` against a slot in the wrong
+# frame - an address nobody wrote, which is exactly the silent-miscompile shape the
+# whole asm effect model is built to avoid.
+#
+# The bindings array is DEEP-COPIED because it is owned: `module_dnit` walks each
+# function's pool and frees the binds of every OP_ASM it finds, so two entries sharing
+# one array would be freed twice.
+# ---
+# m:         module owning the allocator
+# src_fn:    the callee the payload is read from
+# dst_fn:    the caller the payload is copied into
+# src_iid:   the callee OP_ASM id
+# dst_iid:   the cloned OP_ASM id in the caller
+# instr_map: callee instruction index -> cloned caller instruction id, or nil when the
+#            clone preserves ids (the peel's snapshot) and no remap is needed
+# map_len:   number of entries in `instr_map`, ignored when it is nil
+# ret:       ok(true), or an allocation error
+pub fun clone_asm_payload(m: *Module, src_fn: *Function, dst_fn: *Function,
+                          src_iid: id.InstructionId, dst_iid: id.InstructionId,
+                          instr_map: *u32, map_len: u32) R.Result[bool, str] {
+    var key: id.InstructionId = src_iid;
+    val res: R.Result[*IrAsm, str] = map.get[id.InstructionId, IrAsm](?src_fn.asm_blocks, ?key);
+    if (R.is_err[*IrAsm, str](res)) { ret R.ok[bool, str](true); }
+    val src: *IrAsm = R.unwrap_ok[*IrAsm, str](res);
+
+    var dst: IrAsm;
+    dst.isa        = src.isa;
+    dst.body       = src.body;
+    dst.binds      = nil;
+    dst.bind_count = 0;
+    dst.bind_cap   = 0;
+
+    if (src.bind_count > 0) {
+        val rb: R.Result[*IrAsmBind, str] = A.allocate[IrAsmBind](m.alloc, src.bind_count::usize);
+        if (R.is_err[*IrAsmBind, str](rb)) { ret R.err[bool, str](R.unwrap_err[*IrAsmBind, str](rb)); }
+        dst.binds      = R.unwrap_ok[*IrAsmBind, str](rb);
+        dst.bind_cap   = src.bind_count;
+        dst.bind_count = src.bind_count;
+
+        var i: u32 = 0;
+        for (i < src.bind_count) {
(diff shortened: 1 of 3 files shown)


octalide:
Rebased onto current `dev` (`51005032`) and re-measured everything against
a `dev` baseline built from the same commit, so the numbers below are not
the stale ones from the original branch point.

Builds clean, `mach test .` 1246 green.

## The two checks the issue asks for

**Three-generation fixpoint, with inlining live:** `B == C`.

**Byte-identical output at every `--jobs` count, with inlining live:**
1 / 2 / 4 / 8 / 16 all identical.

This is the version of that check that actually means something, and I
should have said last time that I had not run it. The one I reported
earlier was on the frame-layout branch that became #2674, where inlining
was not enabled — so it was measuring the wrong thing. Inlining is what
could plausibly perturb a memory model, so it is the branch the check
belonged on.

## Instruction count and runtime

Atomic increment, `bump` calling a same-module `lock xadd` helper 20M
times, release:

| | calls in `bump` | min | mean |
|---|---|---|---|
| `dev` | 1 | 48 ms | 48.0 ms |
| this branch | **0** | **45 ms** | **45.7 ms** |

Syscall-plus-atomic mix (300k iterations of `getpid` plus a `lock xadd`):

| | min | mean |
|---|---|---|
| `dev` | 46 ms | 47.9 ms |
| this branch | **42 ms** | **43.5 ms** |

15 interleaved runs each, min and mean. Both binaries produce identical
output. So the call really is gone (1 → 0 in the loop body) and it is
worth about 6% on the isolated atomic and 9% on the mixed workload.

The modest size of that is itself informative: an uncontended `lock xadd`
is ~20 cycles and dominates the call it replaces, so removing the call
cannot buy more than this. A contended counter would show a smaller
relative win still, since the cache-line transfer would dominate both.

## The blocker is unchanged on current dev

`surface/macho-pie-aarch64` still fails to build at release:

```
darwin/shared.mach:315:9: encode: inline-asm ldr/str offset 288 is out of
range for both immediate forms
```

`syscall2` has five `{name}` binds, and aarch64's reach for a `{name}` is
the unscaled imm9's −256 because slots sit below the frame pointer and the
scaled imm12 form is unsigned-only. #2674 raised the ceiling a long way
and unblocked riscv64 entirely, but a caller inlining several
`{name}`-heavy wrappers still exceeds thirty-two bound slots.

Still not mergeable. The path remains SP-relative `{name}` addressing.

octalide:
Superseded by #2782, which closes #2231.

**The design here is adopted, not reworked.** Commit `ffa4bccc` cherry-picked cleanly onto current `dev` @ `2b1a0464` - `ir.clone_asm_payload`, the `should_inline` and peel gate removals, and `instruction:asm_is_at_least_as_conservative_as_call` are all carried across unchanged and credited in #2782's body.

**Re-verified rather than assumed**, since the caution about 138 stale commits was fair:

- **#2689 does not break the remap.** A binding names an alloca, the remap follows that alloca into the caller, and the SP-relative addressing is computed from the caller's own frame layout afterwards, so the change is downstream of everything this commit touches. #2735's frame-layout change is likewise downstream.
- **riscv64, which this commit said it broke, is fine.** #2674 fixed the `{name}` reach at the cause. All six targets cross-build clean and the riscv64 int leg is green.
- The measurements did need redoing, and the current numbers are on #2782.

**What this branch correctly identified as the missing half is what #2782 adds**: `std.sync.atomic` is cross-module from every caller, so the mechanism here never reached the issue's headline case. #2782 makes `#[inline]` a cross-module contract - a dependency's `#[inline]` body is monomorphized into its importer as a `PUB | WEAK` definition, the same machinery a generic instance uses with zero type arguments - and extends `Q_LOWERED_SURFACE` to include such bodies so importers cannot reuse stale IR.

Measured end to end: **25.0 ms -> 21.1 ms (-15.5%)** on an atomics-heavy loop, identical output, self-host fixpoint byte-identical.

Recording the conclusion plainly since it was asked for: **this branch was worth having.** Rebasing one commit and re-measuring cost an hour; rediscovering the payload-clone mechanism, and in particular the two-things-must-be-remapped point about the binding allocas, would have cost far more. This PR can close.

octalide:
Closing as superseded. #2231 landed via #2782, which took this branch's design — `ir.clone_asm_payload` copying an `OP_ASM`'s `Function.asm_blocks` entry onto the clone, remapping the map key and each `{name}` binding's alloca — rebased onto current `dev` and re-measured (25.0ms to 21.1ms, -15.5%, identical output).

This branch was explicitly draft/do-not-merge, opened so the mechanism and its measurements were reviewable rather than sitting unseen on a branch. It served that purpose. It is now 60+ commits behind and its content is on `dev`, so there is nothing left to merge.

Issue #2687, opened 2026-08-07
types: resolve a vector as the form TxN bounded by the target's vector width, not a table of ten names


octalide:
Owner decision (2026-08-07): take the `TxN`-as-a-form approach, **provided there is no long-term problem**. There is one, and it has a clean boundary, so this issue is the bounded version.

## Why a form rather than more seeded names

The type representation **already supports it**. `TYPE_VECTOR` carries `(element, lanes)` as data, and `typename.mach` already renders a vector generically as `<element>x<lanes>`. Nothing about the type system hardcodes ten vectors.

What hardcodes them is `resolve.mach:842`, which seeds exactly `VEC_COUNT = 10` spellings into a registry as `SYM_VECTOR` symbols. It is a **name lookup, not a parse**. Resolving `<prim>x<lanes>` by parsing is the natural generalization and matches what the rest of the compiler already believes.

## The boundary, which is the point of this issue

Generalizing `N` splits into two cases with very different costs.

**Sub-128-bit (`f32x3`, `f32x2`, `u32x3`) — no long-term problem.** The value is narrower than the register that holds it. `has_v128` stays meaningful and true, the register classes are unchanged, and `me.pass.scalarize`'s "a vector is `[lanes]elem`" representation generalizes without change. The real work is layout and ABI: `$size_of` becomes lane-dependent rather than always 16, aggregate layout and alignment follow, and `abi.mach`'s `is_vector` — documented today as *"true if the argument is a 128-bit SIMD vector"* — needs to classify a 12-byte or 8-byte vector.

**Super-128-bit (`f32x8`, `f64x4`) — a genuine long-term problem, and out of scope here.** It needs register classes that do not exist (ymm/zmm, SVE), and `has_v128` is a **boolean capability** that says nothing about 256-bit units, so it needs a per-width capability fact rather than a wider one. Every target lacking the unit must then scalarize or split. That is a project, not an extension.

**So: `TxN` as a form, with `N` bounded by the target's vector width.** The form is general; the bound is a capability check rather than a hardcoded list, which is what distinguishes this from simply seeding three more names.

## What it buys

Both live needs are entirely in the sub-128 case:

- **`vec3` vertex attributes.** Real vertex buffer layouts are full of them, and a mach shader currently cannot describe one. `[3]f32` is not a substitute: it is memory rather than a vector register, gets no lane arithmetic, and on SPIR-V emits `OpTypeArray` rather than `OpTypeVector`, so it does not interop either.
[...]

Issue #2658, opened 2026-08-07
spirv: a compute stage has no data path (no storage buffers) and its uvec3 built-ins cannot be typed


octalide:
Two gaps that together mean no real compute shader can be written, reported from the SPIR-V track after #2570 and #2571 landed.

## 1. Compute stages have no interface at all

Vulkan forbids `Output` storage in `GLCompute`. Compute data goes through **storage buffers**, which nothing in the SPIR-V work so far covers: #2571 added `Input`/`Output`/`Uniform` with `Location`, `BuiltIn`, `DescriptorSet` and `Binding`, and a uniform block is read-only by definition.

So `#[stage("compute")]` declares an entry point that can be dispatched and cannot read or write anything. The stage surface exists; the data path does not.

Needs a `Buffer`-decorated `OpTypeStruct` in `StorageBuffer` storage, plus whatever mach-side declaration surface names it. That surface is a design decision in the same family as the `#[uniform(set, binding)]` choice, and should probably reuse its shape.

## 2. Three compute built-ins cannot be typed

`GlobalInvocationId`, `LocalInvocationId` and `WorkgroupId` are `uvec3` in SPIR-V. Every mach vector is exactly 128 bits, so there is `u32x4` and no `u32x3`. The `#[builtin(...)]` decorator accepts these three names and the type cannot be spelled correctly.

Since a compute shader's first act is almost always reading its invocation id, this is not an edge case.

Related to the general absence of `f32x3`, which is interop friction for vertex attributes but not a blocker there, since vec4 is legal in every interface position. For these three built-ins the type is fixed by the specification, so widening is not available the way it is for a varying.

## Acceptance

- A compute shader can declare a storage buffer, read from it and write to it, and passes `spirv-val --target-env vulkan1.3`.
- The three `uvec3` built-ins are either spellable or refused with a diagnostic saying why, rather than accepted by the decorator and then mistyped.

octalide:
Split, with both halves accounted for.

**Done here.** The `uvec3` built-ins (`GlobalInvocationId`, `LocalInvocationId`, `WorkgroupId`) are refused by name with the reason, in PR #2665. They were being emitted mistyped and the validator was rejecting them, so this is now a clear refusal rather than a silent mistype. Making them spellable needs a `u32x3` that mach's fixed 128-bit vectors cannot express, which is a type-system question rather than backend work.

[...]

PR #2695, opened 2026-08-07
types: resolve a vector as the form TxN bounded by the target's vector width

octalide:
Refs #2687. **This is the type-and-layout half of #2687, not the whole of it.** ABI classification and the SPIR-V leg are outstanding, and `f32x3` is explicitly **not yet proven across a call boundary**. Deliberately not `Closes`.

## What this does

A vector type is now resolved by READING the form `<element>x<lanes>`, bounded by the target's vector width, and its layout is lane-derived rather than a flat 16 bytes.

The type representation never needed to change. `TYPE_VECTOR` already carried `(element, lanes)` and `typename.mach` already rendered a vector generically. Two things blocked `f32x3`:

1. `resolve.seed_vectors` put exactly ten names into a registry, so `f32x3` failed because nothing had seeded that name.
2. `layout.mach` folded a payload-free `NODE_VECTOR` as `extent(16, 16)`, so even once resolved it would have laid out as `f32x4`.

Both are gone.

## Where the bound lives, and why it is not `has_v128`

The bound is a new `MachineModel.vector_bits`, the widest vector a target admits by name. Deliberately **not** the existing `has_v128` capability:

- **They are independent in the direction that matters.** rv64gc and mos6502 both declare `has_v128` false and still admit every 128-bit shape today, which `me.pass.scalarize` turns into per-lane scalar code. A target with no vector unit still has a vector width, so gating the SPELLING on the capability would refuse names that compile correctly right now.
- **A boolean cannot express the growth axis.** A target that gains ymm / zmm / SVE raises `vector_bits` and the frontend admits the wider spellings with no change.

It reaches the frontend through `ComptimeCtx`, which already carries the per-build target facts the frontend needs, alongside `pointer_width`.

## Size and alignment

```
size  = element_size * lanes
align = size            when the vector fills the vector register
align = element_align   when it is narrower
```

| type | size | align |
|---|---|---|
| `f32x4` | 16 | 16 |
| `f32x3` | 12 | 4 |
| `f32x2` | 8 | 4 |
| `i16x4` | 8 | 2 |

Verified by compiling and running real programs, not by unit test alone. A `rec Vertex { pos: f32x3; uv: f32x2; }` measures **20 bytes**, packed, which is the motivating vertex-buffer case.

**The alignment is discontinuous at the register width, deliberately**, and the fold says so at the fold. A vector that fills the register aligns to its whole size because the machine's vector load requires it; a narrower one is a packed aggregate that scalarizes or loads piecewise. It reads as an inconsistency cold and would otherwise be "fixed" into uniformity, which would either over-align every vertex attribute or under-align every register-resident vector. Packing is the whole point: padding `f32x3` to 16 would make it indistinguishable from `f32x4` in memory and withhold the reason anyone wanted it.

`layout`'s dependence on the target is now a `layout.Machine` record rather than a loose `ptr_width`, since that dependence is an axis that grows and had just grown.

## Diagnostics

The refusal path matters as much as the accepted one, and the three outcomes stay distinct so a width problem is never described as a typo:

```
error: vector `f32x7` is 224 bits wide, more than this target's 128-bit vector width
error: vector `f64x3` is 192 bits wide, more than this target's 128-bit vector width
error: vector `f32x1` needs at least 2 lanes
error: unresolved type name `matrix4`     # not a vector form, falls through unchanged
error: unresolved type name `f32x04`      # leading zero is not a lane count
```

## The regression bar

**All ten pre-existing shapes are exactly 128 bits**, so they still measure 16 and 16 and this is a pure extension rather than an ABI change to shipped code. Pinned as a test at the fold and again at the IR layer, and asserted by walking `vec_shape` rather than by restating the spellings.

Tests assert what a form resolved TO, not merely that it resolved: element kind, lane count, and bit width. Refusals cover the exact boundary cases (`f32x5`, `i8x17`, `f64x3`), the degenerate `f32x1`, and `u64x4294967295` proving an absurd lane count reports too wide rather than wrapping u32 into a shape that appears to fit.

Every new assertion was checked by mutation rather than trusted:

- reverting the fold to `extent(16, 16)` — **fails**
- collapsing the alignment discontinuity to uniform — **fails**
- disabling the width bound with `if (false)` — **fails**

While doing that I found the existing `intern_vector:dedup_size_align` test declared **no machine model at all**, so its vector width was 0 and it could not have caught either layout mutation. Fixed, and extended with the sub-128 and 256-bit-target cases.

Full suite green (1249). Self-host fixpoint verified byte-identical (stage2 == stage3).

## What is outstanding

- **ABI classification** of a 12-byte or 8-byte vector. `f32x3` is not yet proven across a call boundary.
- **The SPIR-V leg**: `f32x3` reaching `OpTypeVector %float 3`, and the `uvec3` compute built-ins from #2658.

`doc/language/types.md` is updated: the "every vector is exactly 128 bits" guarantee is replaced by the real rule and the discontinuity is documented, rather than merely relaxed in prose.

## One behaviour change worth calling out

A vector spelling takes priority over scope in type position, which is how the ten already worked, so a user record named `u8x16` was already shadowed. Reading the form widens that set: a user type named `f32x3` is now shadowed too. Consistent with the existing rule, but it is a source-compatibility surface the issue does not mention.

### DIFF
--- a/doc/language/types.md
+++ b/doc/language/types.md
@@ -21,25 +21,52 @@ There is no compiler `bool`. `bool` is a stdlib `def bool: u8;` with `true` /
 
 ## SIMD vectors
 
-Ten 128-bit vector types are compiler-seeded, each a fixed number of lanes of a
-primitive numeric element:
+A vector type is a **form**, not a fixed list: any primitive numeric element
+followed by `x` and a lane count.
 
 ```mach
-f32x4  f64x2                    # float lanes
+f32x4  f32x3  f32x2  f64x2      # float lanes
 i8x16  i16x8  i32x4  i64x2      # signed integer lanes
 u8x16  u16x8  u32x4  u64x2      # unsigned integer lanes
 ```
 
-These ten are the complete set. The spelling is `<u|i|f><width>x<count>` with a
-**single** `x` — there are no other vector types and no multi-`x` shapes. A name
-like `f32x4x4` is not a vector type (it resolves as an ordinary identifier); a
-matrix is an algorithm over vectors and belongs in a library over vector
-elements, not the language. A vector spelling is recognized only in type
-position, so a value may still be named `f32x4` without colliding with the type.
+The spelling is `<u|i|f><width>x<count>` with a **single** `x`. A name like
+`f32x4x4` is not a vector type (it resolves as an ordinary identifier); a matrix
+is an algorithm over vectors and belongs in a library over vector elements, not
+the language. A vector spelling is recognized only in type position, so a value
+may still be named `f32x4` without colliding with the type.
 
-Every seeded vector is exactly 128 bits: `$size_of` and `$align_of` are both
-`16`. Arrays of vectors (`[4]f32x4`) and pointers to vectors (`*f32x4`) are
-ordinary composite types over a vector element.
+Two rules bound the form:
+
+- **At least 2 lanes.** `f32x1` is refused; a one-lane vector is just its scalar.
+- **No wider than the target's vector register.** Every current target is 128
+  bits, so `f32x7` (224 bits) is refused by name. The error names the width you
+  asked for and the width the target has, rather than reporting an unknown type.
+
+`ptr` is not a lane element: its width is target-defined rather than a scalar bit
+count, so `ptrx2` is not a vector spelling.
+
+### Size and alignment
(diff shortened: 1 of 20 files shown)


octalide:
-

Issue #2697, opened 2026-08-07
codegen: a value has one width, but a sub-register-width vector has two (register 16, memory 12)


octalide:
Split out of #2687, which exposed this and is holding a frontend refusal until it is fixed.

## What

`op_width_of` answers **one** number for what are really two different quantities: the width a value occupies in a register, and the footprint it occupies in memory. That conflation was invisible while every vector was exactly 16 bytes, because the two numbers were the same. Lane-derived vector layout (#2695) separates them, and the conflation becomes a silent miscompile.

```
i32x3{10,20,30} + i32x3{1,2,3}   ->  lanes 11, 22, 0
```

**The affected set is wider than 12 bytes, and the rule is not "non-power-of-two".** Measured by running a program on x86-64 (`gpr_width` 8), reporting each shape's computed value:

| shape | size | got | want | |
|---|---:|---:|---:|---|
| `i8x3` | 3 | 6 | 6 | ok |
| `i16x3` | 6 | 6 | 6 | ok |
| `i32x2` | 8 | 3 | 3 | ok |
| `i16x4` | 8 | 10 | 10 | ok |
| `i8x9` | 9 | 1 | 10 | **wrong** |
| `i16x5` | 10 | 4 | 6 | **wrong** |
| `i32x3` | 12 | 30 | 60 | **wrong** |
| `i16x7` | 14 | 1 | 8 | **wrong** |
| `i32x4` | 16 | 10 | 10 | ok |

So: 3 and 6 are non-power-of-two and **pass**, while 9, 10 and 14 also **fail**. Falling through to `gpr_width` only loses data when the vector exceeds the GP word, so the hole is **any size strictly greater than `gpr_width` and less than the vector register width**.

That is **target-dependent**, which matters for the fix: written against the constant 12 it looks complete on x86-64 and leaves the hole open on a target with a narrower word. On mos6502 (`gpr_width` 1) the hole would be sizes 2 through 15.

Note also that the damage is not a uniform "last lane dropped" — `i8x9` returns 1 of an expected 10 and `i16x7` returns 1 of 8, so how much survives varies by shape. The headline is the size range, not a consistent truncation.

*(A note on measuring this: the first version of the probe packed one bit per shape into the process exit status, which is 8 bits, so bits 8 and up were silently truncated and `i32x4` was never actually measured while the table printed "ok" for it. The numbers above come from a per-shape stdout report with a must-pass and a must-fail control, so "everything failed" cannot read the same as "the program never ran".)*

[...]

Issue #2688, opened 2026-08-07
spirv: no GLSL.std.450 extended instruction set, so a shader has no sqrt, normalize, dot or mix


octalide:
The SPIR-V backend emits no `OpExtInstImport`, so a shader has no `sqrt`, `normalize`, `dot`, `min` / `max` / `clamp`, `pow` or `mix`. A trivial vertex plus fragment pair does not need them; anything past hello-triangle does.

Owner decision (2026-08-07): **compile specific math functions made for shader work, living in their own project rather than the stdlib.** `briar-systems/mach-shader` now exists for it.

## Why not the alternatives

**Not intrinsics.** A `$normalize` is new language surface per function, and a `$`-name meaningful on exactly one backend is the wart the decorator channel was chosen to avoid.

**Not compiling `std.math` through.** `std.math.sqrt` promises IEEE semantics; `GLSL.std.450 Sqrt` promises what the driver does. Silently substituting one for the other across a target boundary is the failure class this repo keeps finding — a value that is *nearly* right, where nothing errors.

Keeping the shader set in its own module makes the semantic difference visible at the call site:

```mach
use sh: shader.math;

#[stage("fragment")]
fun frag_main() {
    val n: f32x4 = sh.normalize(in_normal);
    val d: f32   = sh.max(sh.dot(n, light_dir), 0.0);
    out_colour   = sh.mix(ambient, albedo, f32x4{d, d, d, d});
}
```

## Scope

- `OpExtInstImport "GLSL.std.450"` emitted once per module, on demand rather than unconditionally.
- An `OpExtInst` lowering path keyed on the function being called.
- The `mach-shader` module declaring the set. Whether each function also carries a native body, so the same source runs on a CPU target, is a design call worth making deliberately rather than by omission.

## Open questions worth settling in the design, not during implementation

- **Which functions.** The GLSL.std.450 set is large. Start from what a real shader needs rather than transcribing the spec.
- **Native fallback.** If `shader.math` has native bodies it is a normal library with a fast path; if it does not, it is a shader-only module that fails to link elsewhere. Both are defensible; silently being the second while looking like the first is not.
- **Precision.** Several GLSL.std.450 entries are explicitly implementation-defined in precision. Whichever way that is documented, it should be documented, since it is the whole reason this is not `std.math`.

## Acceptance

- A fragment shader calling at least `normalize`, `dot`, `max` and `mix` compiles and passes `spirv-val --target-env vulkan1.3`.
[...]

Issue #2721, opened 2026-08-07
isa(arm64): the NEON integer-multiply comment states the wrong legality rule, and only one arrangement is tested


octalide:
Found while gathering evidence for #2488.

## What

`arm64/encode.mach` declares the NEON integer-multiply base word with a comment stating a legality rule that is **wrong in both directions**:

```mach
val VMUL: u32 = 0x4E209C00;  # integer multiply, legal only on .8h (size 1)
```

NEON `MUL (vector)` is legal on `8B`, `16B`, `4H`, `8H`, `2S` and `4S`. It is illegal on `1D` and `2D`. So the comment:

- **forbids arrangements the hardware allows** (`.2s` / `.4s`, and the byte arrangements), and
- **does not state the constraint that is real** (no 64-bit integer lanes), which is the one a future widening would need to respect

The encoder itself is not wrong today, because nothing selects `VMUL` at a size other than 1. The comment is what a reader consults before widening the selector, and it currently points away from the true boundary. #2488 explicitly carries "64-bit integer lanes must stay refused on arm64 regardless of direction; a widened rule that forgets this trades a conservative refusal for an invalid encoding" as a note to preserve, and the in-source comment contradicts the half of that which matters.

Confirmed by reading the selector at `arm64/encode.mach:2599`, which routes `VEC_MUL` to `VMUL` for the integer case with no arrangement guard of its own, so the size ultimately encoded is whatever the caller passes.

## Coverage

One encode test exercises this word, at size 1 only:

```mach
if (pack_neon_3same(VMUL, 1, 0, 0, 1) != 0x4E619C00) { bad = 1; }
```

So `.4s` is not merely unverified, it is unrepresented. That is the exact shape #2037's arrangement collapse lived in: a base word plus a size field that *looks* general, with nothing pinning what each arrangement actually encodes. A widening that trusts the word's apparent generality would produce a silent wrong-arrangement miscompile rather than a refusal.

## Fix

Correct the comment to state the real rule (legal on the 8/16/32-bit lane arrangements, illegal on 64-bit integer lanes), and add byte-exact `pack_neon_3same` assertions for the arrangements the hardware permits, checked against the ARM ARM encoding rather than against what the encoder currently emits. A test that asserts the encoder agrees with itself would restate the defect instead of catching it.

[...]

Issue #2488, opened 2026-08-04
rfc(simd): should the legality of a vector operation be target-dependent? (integer lane multiply)


octalide:
Spun out of #2349's design phase. The narrow question there ("move the vectorizer's integer-multiply exclusion behind a capability") turns out to be undecidable without answering a language-surface question first, which is this RFC.

## Problem

Today `i32x4 * i32x4` is refused on **every** target, uniformly, at the language level (`fe/sema/infer.mach:1443`: "`*` is float lanes, or 16-bit integer lanes only (pmulld is out of baseline)"). Verified: compiling `fun mul32(a: i32x4, b: i32x4) i32x4 { ret a * b; }` for aarch64 today errors, even though AArch64 hardware supports `MUL` on `.4s`.

The rule's stated reason is x86-64's SSE2 baseline (`pmulld` is SSE4.1, and mach has no feature-level mechanism to gate on). So a real hardware capability on one target is refused because a *different* target lacks it.

Making it per-target is mechanically easy — a declared `MachineModel` bool, the `has_lane_ops`/`has_mul` precedent, and the vectorizer already receives the target. The question is whether it *should* be:

**Should the legality of a vector operation be target-dependent?** Today mach's vector surface is uniform: source that compiles for one target compiles for all. Per-target capability introduces a portability cliff — `i32x4 * i32x4` would compile on aarch64 and fail on x86-64, discovered only when someone cross-compiles.

## Candidate directions

1. **Keep it uniform (status quo, reason corrected).** The rule stays a floor: the vector surface is what every supported target can do. Portable by construction; leaves aarch64 performance on the table. #2349's measured value is low (`i32_madd` at 1.11x), which weakens the case against.
2. **Per-target capability, refused where absent.** Honest about hardware, and the code that wants it gets it. Cost: the portability cliff above, plus the surface becomes something a user must consult per target.
3. **Uniform surface with defined scalarization.** Accept `i32x4 * i32x4` everywhere; on targets lacking a lane multiply, lower it to scalar lanes. Preserves portability AND unlocks the capable target — the same model the SIMD epic (#1965) already chose for operations not present on every ISA ("defined scalarization elsewhere"). Cost: a performance cliff instead of a portability one, silent unless surfaced.
4. **Feature levels.** Gate on an x86-64 feature level (SSE4.1+), making the capability a *build* choice rather than a target fact. Largest change; the only one that lets one target have two answers.

[...]

PR #2722, opened 2026-08-07
fix(abi): pass and return a vector narrower than the vector register

octalide:
Closes the ABI half of #2687. `f32x3` now crosses a function boundary.

```mach
fun addv(a: f32x3, b: f32x3) f32x3 { ret a + b; }
```

did not compile before this. The FP move-width classifier accepted the fixed set
{4, 8, 16} and a twelve-byte vector matched no member of it, so the whole call
boundary refused. That made vec3 usable inside one function and nowhere else.

## The defect

The same conflation #2697 fixed one layer down, repeated one layer up and again
below it. A value's **register width** and its **memory footprint** are two numbers,
and they agreed for every vector until lane-derived layout (#2695) made them differ.
Three layers each answered with one fixed-size ladder, and each produced a different
symptom:

| layer | with a 12-byte vector |
|---|---|
| `op_width_of` (#2697, fixed) | fell through to the GP word, lanes lost |
| `fp_move_width` | refused outright, honest and total |
| `encode_fp_mov` (x64) | silently defaults any non-{4,8,16} size to 8 |

The third is why widening the accepted set is not the fix and was not enough. Once
the ABI stopped refusing, a stack-passed `i32x3` carried a width-12 move into an
encoder that quietly narrowed it to 8, and the top lane vanished with no diagnostic
anywhere - the exact class this repo keeps hitting, arrived at from a different
direction.

## The fix

The ABI asks the two questions separately, exactly as the IR lowering already does:

- a **register** move carrying a vector is the vector register's width, derived from
  `MachineModel.vector_bits` and not the constant 16
- every access **between that register and memory** - an outgoing stack argument, an
  incoming stack parameter, the win64 by-reference marshal temp - is the value's own
  footprint and goes through the scratch round trip #2697 already built, because no
  ISA here can issue a twelve-byte access

Those two helpers moved out of the store/load lowering into the shared context so
the ABI reuses them rather than growing a second answer. That is the first commit,
inert on its own.

## Per target, from MachineModel facts

Not assumed to follow x86-64.

| target | placement |
|---|---|
| x86-64 SysV / win64 | one XMM argument register, XMM0 for the return. The psABI classifies both eightbytes of a 12-byte vector SSE, which is one register; the top four bytes are undefined |
| AAPCS64 | one V register. AAPCS64's short-vector types are 8 and 16 bytes, so a 12-byte vector is not one and no conformance rule applies. It fits the register, so it rides it |
| lp64 (rv64gc) | unchanged. No vector register file, so by hidden reference like any memory-class value, at whatever size it is |
| mos6502 | unchanged. The stack, and an sret return |

Nothing is refused. `x86-64 linux` and `aarch64 linux` both work.

**darwin-aarch64 is where the obvious fix would have been wrong.** Apple's arm64 ABI
gives a fixed stack argument its natural size, so an `i32x3` slot there is exactly
twelve bytes with the next argument packed against it, and the sixteen-byte write
that works everywhere else destroys the neighbour. The footprint-exact write is the
reason that leg is in the case's target list.

## Proof, by running

`int/surface/vector-abi` covers pass by value, return by value, a returned vector fed
straight into another call, register exhaustion forcing two vectors onto the stack, a
vector inside a 20-byte record passed and returned by value, and both neighbour
orders. **Every case carries a guard on the adjacent argument or field**, since the
whole risk of a wrong slot width is that it lands next door.

Every callee is `#[noinline]`. At opt 2 the inliner otherwise deletes the call the
case exists to exercise, which is measured rather than assumed: before that, the
release profile stayed green with the stack half reverted.

**Measured per leg by reverting each half and re-running**, not reasoned from the
target models:

| leg | without the width split | without the scratch round trip |
|---|---|---|
| linux | does not compile | `many` = 42477, want 54477 |
| linux-arm64 | does not compile | `many` = 32477, want 54477 |
| linux-riscv64 | passes | passes |

How much survives differs by ISA, which is why the observable is a value and not a
"did it run". **linux-riscv64 passes unpatched on both profiles**: `has_v128` is
false there, so these shapes scalarize and pass by hidden reference and none of this
path runs. The case still belongs on that leg and says exactly this in its comment,
but it is not evidence about this fix there.

## Gates

- `mach test .` 1267 passed, 0 failed
- full int suite green on linux, linux-arm64, linux-riscv64 (the two local arm64
  failures are `c-variadic` under forced qemu and `narrow-stack-args` linking a
  host-`cc` x86-64 object, both failing identically on unpatched `dev`)
- `check-target-matrix.sh` 372 pairs OK
- self-host fixpoint holds, debug and release, stage2 byte-identical to stage3
- `check-determinism.sh` clean
- cross-builds clean for windows and darwin-aarch64

## Still open

`Part of #2687`, not `Closes`. The SPIR-V leg remains: `OpTypeVector %float 3`, the
`uvec3` built-ins from #2658, and `Cross`, which #2688 cut pending exactly this ABI
work. That work is now unblocked.


octalide:
Third commit, after the vector surface contract in #2488 landed as a decision: the legality of a vector type is target-independent and only its realization is not, so `vector_bits` is a capability input to lowering and never a legality gate. #2727 lifts the frontend bound and puts a 256-bit vector on a 128-bit register, which is this same question asked in the other direction.

**Nothing in this PR was shaped around 12 bytes, around 16, or around "smaller than the register."** The quantities are the value's own extent and the target's own vector register width throughout, so a shape that is neither a multiple nor a divisor of the register (`f32x5`, whose split has a sub-width tail) needs no different question asked of it.

**Where the seam is, stated exactly.** `fp_move_width` answers for ONE register's worth. An over-width vector realizes as a split, and a split is already a multi-piece `abi.ParamSlot` that `materialize_pieces` walks today for an HFA run and for a register-plus-stack aggregate. So the extension point for #2727 is each convention's `classify_arg` / `classify_return` deciding a piece count, and the move-width question is asked once per piece and does not change. The refusal `fp_move_width` carries for an over-width vector is a "this question is malformed for that value", not "that shape is illegal", and it says so.

**One real gap closed rather than documented.** `subwidth_vector_memory` is an inequality, so it is true in BOTH directions. An over-width vector reaching the scratch round trip would have been spilled at the register's width into a scratch it outgrows and then had more bytes copied out than were ever written - a silent wrong value, landing the day #2727 lands and nowhere near the change that caused it. Both helpers now refuse it by name. That was a latent miscompile, not a doc issue, and it is the one thing in this commit that changes behaviour.

**What remains genuinely narrow, said plainly rather than retrofitted.** `op_width_of` (from #2697, upstream of this PR) returns `vector_bits / 8` for every vector regardless of extent, so it caps at one register. That is the single place #2727 has to change shape rather than extend, and it is not this PR's to change: widening it without the multi-piece placement underneath would turn today's honest refusals into truncated moves. The refusals added here are what make that ordering safe.

Re-verified after the change: `mach test .` 1267/0, vector int cases green on all three PR legs, release self-host fixpoint holds (stage2 == stage3).

Not touched and not trusted: the aarch64 `VMUL` arrangement comment in #2721. Nothing here goes near integer lane multiply selection.

Issue #2727, opened 2026-08-07
simd: a vector wider than the target's register is refused by name, so arbitrary lane counts do not work


octalide:
Decided in #2488. Depends on #2687.

## What

`vector_bits` is used as a **legality gate on the surface** when it is a **capability fact about realization**. Measured on `dev`, every target refuses any vector wider than its vector register, by name, at the frontend:

```
error: vector `f32x8` is 256 bits wide, more than this target's 128-bit vector width
```

| shape | arm64 | riscv64 (no SIMD unit) |
|---|---|---|
| `f32x5` | refused, 160 bits | **refused, 160 bits** |
| `f32x8` | refused, 256 bits | **refused, 256 bits** |
| `f32x16` | refused, 512 bits | **refused, 512 bits** |

The riscv64 column is the incoherence. That target has no vector unit, scalarizes every vector operation it is given, and still refuses `f32x8` on the ground that 256 bits exceeds a 128-bit SIMD width **it does not have**. The shape would lower to eight ordinary scalar float adds.

Per @octalide in #2488: vector types should work regardless of backend and whether or not they are scalarized, and arbitrarily sized vector types should just work even when not always backed by SIMD.

This is the same defect as #2726 on a different axis. That one gates an *operation* on a realization fact; this one gates a *type*. Both make the surface target-dependent for reasons that belong to lowering.

## Fix

`vector_bits` stops deciding legality and becomes an input to realization, keeping the growth-axis role it was introduced for in #2687: a target gaining ymm/zmm/SVE raises it and gets **better codegen**, rather than admitting new spellings.

The frontend admits `TxN` for any element type and any `N >= 2`. The backend realizes it in preference order:

1. one vector register, when the shape fits `vector_bits` and the op has a packed form
2. a **split** across several registers or chunks, when it does not fit
3. per-lane scalar code, via the existing `me.pass.scalarize`

Tier 3 already exists and carries whole targets. Tier 2 is the new machinery, and it is the same split-and-recombine that #2687 is currently building for the sub-width case one axis over - a 96-bit vector on a 128-bit register and a 256-bit vector on the same register are the same question asked in two directions. **This must sequence behind #2687 rather than race it**, or the two land conflicting answers in `layout.mach`, the ABI classifier and `mir/context.mach`.

[...]

PR #2736, opened 2026-08-07
feat(simd): vector op legality is target-independent, so an unpacked lane op scalarizes

### LINKED ISSUE #2726: simd: vector operation legality is target-independent, so an unsupported lane op scalarizes instead of being refused
Decided in #2488. Depends on #2721.

## What

The legality of a vector operation is target-independent; only its realization is target-dependent. Today `*` on integer lanes is refused uniformly at the language level for a reason that is a *realization* fact about one target:

```
error: vector `*` on this lane width is not supported (integer multiply is 16-bit lanes only)
```

Measured on `dev`, this refuses `i32x4 * i32x4` on **riscv64**, a target with no vector unit, which scalarizes `i32x4 + i32x4` one line up and would lower the multiply to four ordinary 32-bit scalar multiplies. It also refuses it on **aarch64**, whose NEON `MUL` supports `.4s` in hardware. The cited constraint is x86-64's SSE2 baseline (`pmulld` is SSE4.1).

The project already has the mechanism for exactly this situation. `me.pass.scalarize` turns vector ops into per-lane scalar code for targets with `has_v128 = false`, and `isa.mach`'s own contract comment describes that as the model. Per-operation uniform refusal is the one place it is not applied.

The cost usually charged against this - a silent performance cliff - is already paid. `simd` is a **required** per-profile manifest key, and `simd = "require"` already produces a precise diagnostic:

```
error: simd = "require": vector add in '_M4case4mainN5add32' would scalarize on riscv64
       (target has no 128-bit SIMD); set simd = "scalarize" to build with the scalar expansion
```

## Fix

The rule has **three homes** today - `vectorize.supported_vector_op`, `fe/sema/infer.mach` (the message above), and `be/codegen/vector_runtime.mach`. Fixing one alone is incoherent, which is why #2349 was closed against a comment-truth fix rather than a behaviour change.

They collapse onto **one authority** answering a realization question ("does this target have a packed form for this op at this lane width"), consulted by the backend to choose between the packed form and the scalar expansion. Sema stops asking it entirely, because it is not a legality question.

The scalar shortfall must route through the same path `simd = "require"` already observes, so a project that cannot afford it is told, and its message should name the operation and lane width as the existing one names the operation and target.

**64-bit integer lanes stay refused on aarch64** regardless: NEON `MUL` has no `.2d` form. That is a genuine encoding constraint rather than a baseline preference, and #2721 covers pinning the permitted arrangements byte-exact before anything reaches `.4s`. A widened selector that forgets this trades a conservative refusal for an invalid encoding, which is #2037's failure.

## Acceptance

- `i32x4 * i32x4` compiles on x86-64, aarch64 and riscv64, and **runs**, with lane values checked by hand rather than read off the lowering.
- On aarch64 it selects the packed `.4s` form; on riscv64 it scalarizes. Verified by what executes, not by what is emitted.
- A 64-bit integer lane multiply is still a defined refusal on aarch64, with a test.
- `simd = "require"` reports the scalarized multiply, naming the operation and lane width.
- The three sites consult one authority, and a test asserts they cannot disagree. Asserting the table against itself does not count - the agreement must be observable.
- `mach test .` green, `int` green including the native aarch64 leg, and the self-host fixpoint holds.

octalide:
Closes #2726. Stacked on #2731 (#2721), which must land first — retarget to `dev` once it does.

## The incoherence

`i32x4 * i32x4` was refused at the language level citing x86-64's SSE2 baseline (`pmulld` is SSE4.1). Measured on `dev`, that refused it on **riscv64**, a target with **no vector unit at all**, which scalarizes `i32x4 + i32x4` one line up and would lower the multiply to four ordinary scalar multiplies. It also refused it on **aarch64**, whose NEON `MUL` supports `.4s` in hardware. The rule refused work on the targets least able to benefit from the refusal.

Per #2488's decision: the legality of a vector type and of a vector operation are **target-independent**; only the **realization** is target-dependent.

## The authority

One question, `isa.packed_width(model, op, is_float, lane_bits) -> u32`: how wide is this target's packed form for this operation at this lane width, or **0** for none. Gaps are declared per ISA beside the other capability facts; the mid-end reaches it through the single door `me.vecform`, which owns the one `InstrKind -> VecOp` translation. **Sema does not call it**, because it is not a legality question.

Three design choices and what each buys:

- **It returns a width, not a bool.** #2727 needs a number to split *by*. A bool would answer "is there a packed form" and leave "how wide" to be re-derived somewhere else, which is how two facts drift apart. It also lets an ISA whose packed float ops are wider than its packed integer ops (AVX1 over x86-64) be *described* rather than rebuilt around.
- **It is keyed on the lane width, never the lane count or total width.** So splitting a vector wider than the register into register-sized chunks does not change the answer — each chunk asks this same question and gets this same result. **Nothing here changes when #2727 removes the width gate.**
- **Gaps, not presence.** Presence would be ~80 rows per target and the row that mattered would be the one nobody noticed was missing. Only rows something *consults* are declared: integer divide has no packed form on either capable target, but no lowering asks, and a row nothing reads is exactly how `fp_scratch_reg` came to declare one register while the encoder used another.

Declared gaps, with the evidence:

| target | gap | why |
|---|---|---|
| x86-64 | int `*` at 8, 32, 64 | SSE2 packs only `pmullw` (`66 0F D5`). `pmulld` is `66 0F 38 40`, the SSE4.1 opcode map. `pmullq` has no legacy encoding (AVX512DQ). No SSE generation has a byte multiply. |
| aarch64 | int `*` at 64 | `MUL (vector)` leaves size 3 unallocated (#2731). |
| SPIR-V | none | `OpIMul` is defined over vectors of any component type and count. The empty list is the member proving the authority does not assume a gap must exist. |
| riscv64, mos6502 | n/a | `has_v128 == false` answers every query with 0. |

## Realization

Two halves, both reading that one authority:

- **`scalarize.expand_gap_ops`** (new) runs on a target that *has* a vector unit and rewrites only the gap operators. **Vector values survive** — `i32x4 + i32x4` still selects `paddd` next to an `i32x4 * i32x4` that does not. Built from `OP_VEC_EXTRACT` / `OP_VEC_INSERT`, which already have a realization on every target (native `UMOV`/`INS` on NEON, memory-imaged by `expand_lane_ops` on x86-64), so it needs no memory machinery of its own.
- **`scalarize.run`** is unchanged and still owns the whole module where there is no vector unit.

`simd = "require"` moves ahead of both and applies on **every** target.

## Verification — run, not read

**Lane values, computed by hand and checked against three targets.** `int/surface/vector-mul-widths`, one shared golden:

```
i32.mul 2100000000 210003 -2100000 -1
i32.add 121000 70004 699997 131072
u32.mul 2100000000 210003 4294967293 4294967295
i8.mul 44 106 -116 0 5 16
i16.mul 512 9 -20 -32536
i64.mul 4294967327064771079 -3703703670369
```

Every product is designed to be unreachable by a wrong lowering: `i32` lane 0 needs the full 32-bit product; `i32` lane 3 is `65537 * 65535 = 2^32 - 1`, wrapping to `-1`, which a 16-bit lane multiply or a dropped high half cannot land on; every `i8` product wraps inside its byte so a 16-bit arrangement would carry into its neighbour; `i64` lane 0 exceeds `2^32`.

**Which strategy was taken**, since a value check cannot see correct-but-scalarized. `simd = "require"` is the observable, and the full matrix was measured:

| shape | x86-64 | aarch64 | riscv64 |
|---|---|---|---|
| `i32x4 *` | fires | **does NOT fire** | fires |
| `i8x16 *` | fires | **does NOT fire** | fires |
| `i16x8 *` | does not fire | does not fire | fires |
| `i64x2 *` | fires | **fires** | fires |
| `i32x4 +` | does not fire | does not fire | fires |

Pinned as int cases: `vector-mul-scalar-shortfall` (build-fails under `require`, per-target goldens, all three legs) and `vector-mul-packed-4s` (aarch64 legs, builds green under `require` — which *is* the assertion that the packed form was chosen, plus lane values so a green build cannot hide a wrong word).

**At the machine level**, disassembled from the aarch64 binary:

```
414810: 4ea19c02  mul v2.4s, v0.4s, v1.4s
41481c: 4e219c02  mul v2.16b, v0.16b, v1.16b
414854: 4e619c02  mul v2.8h, v0.8h, v1.8h
```

and `i64x2` takes the scalar path (`umov` / `mul x` / `ins`). Zero invalid encodings in the binary. On x86-64 only `pmullw` appears — no `pmulld` — while `paddd` survives, which is the "vector values survive, only the gap operator expands" claim in emitted code.

**The three sites cannot disagree**, and it is observable rather than asserted against a table. `scalarize:require_and_expansion_read_one_authority` relates two *independent* observations of the same module: how many operators `detect` reports (the number the `require` error is raised from) and how many survive `expand_gap_ops` (what the compiler actually did). They must be N and 0. The same `i32x4 * i32x4` goes to all three targets, so the only variable is the target.

**Proof the tests fail without the patch.** A compiler built from `origin/dev` (`c8dba96`) was used to run every new case:

| case | verdict on `origin/dev` |
|---|---|
| `vector-mul-widths` | FAIL (build) — `error: vector \`*\` on this lane width is not supported` |
| `vector-mul-scalar-shortfall` | FAIL (diff) — emits the old sema refusal instead of the shortfall |
| `vector-mul-packed-4s` | FAIL (build) — same refusal, on the target whose hardware has the instruction |

**Mutation testing.** Deleting aarch64's `.2d` gap row makes `require_and_expansion_read_one_authority` fail at the aarch64 `i64x2` assertion, and end-to-end the mutated compiler produces:

```
error: encode: NEON integer multiply has no 64-bit lane arrangement (MUL (vector) leaves size 3 unallocated)
```

which is #2731's backstop firing exactly as intended — a defined error rather than the invalid word `0x4EE29C20`. That is the two layers working as designed: the authority routes it away, and the encoder catches a wrong authority.

## Note on scope, and one thing I changed deliberately

**`i64x2 * i64x2` is no longer refused on aarch64.** Both issue bodies said "64-bit integer lanes stay refused on aarch64", but under the governing contract that is self-contradictory: a missing NEON `.2d` MUL is a realization fact of exactly the same category as SSE2 lacking `pmulld`, and refusing it on aarch64 *only* reintroduces the per-target language surface the decision deletes — pointing the wrong way, since the target *with* a vector unit would be the one that refuses. No target has a packed 64-bit lane integer multiply at all. Confirmed with @octalide, who has corrected both issue bodies. The load-bearing half — never emit an invalid encoding — is #2731's encoder backstop, proved above.

**`vectorize-emit`'s `pred_guard_calc` was fixed rather than re-blessed.** Its guard used `x*3`, chosen because integer multiply was not vectorizable; on aarch64 it now is, so the case would have kept passing while silently no longer testing "an unsupported guard tree must decline". Its guard is now a shift, which is out of the increment-1 table on every target, and a new `pred_guard_mul` covers the target-dependent behaviour on purpose — its goldens *disagree* across legs (`simd` on aarch64, `scalar` on x86-64 and riscv64), which is the observable that the vectorizer only forms what the target can pack.

**The new multiply coverage went to `int/` rather than `vector_runtime.mach`**, for bootstrap: that file is compiled by the *seed*, so it can only spell what 4.14.0 already accepts. Once a release carrying this is the seed it belongs there beside the 16-bit case, and the comment says so.

## Gates

- `mach test .` — **1279 passed, 0 failed**
- `int` green on `linux`, `linux-riscv64`, and `linux-arm64`. Two arm64 failures (`c-variadic`, `narrow-stack-args`) are **pre-existing** — verified failing identically with the `origin/dev` compiler — and are local-environment artifacts of running that leg under qemu on an x86-64 host. **CI's native `ubuntu-24.04-arm` leg is the authority here.**
- Self-host fixpoint holds, stage2 byte-identical to stage3: `f2ab3fb46c30264a021190b8656d275e5ab8596cb2ae9e894360d3cc175f909a`
- `int/lib/check-target-matrix.sh` — 396 case x leg pairs OK
- `doc/language/operators.md` and `doc/manifest.md` updated; CHANGELOG entry added.


octalide:
Retargeted to `dev` so CI runs (it only triggers on PRs into `dev`/`main`). This branch carries #2731's commit as well until that lands — merge #2731 first, and this reduces to its own commit.

Issue #2745, opened 2026-08-07
spirv: the uvec3 compute built-ins and GLSL.std.450 Cross are refused on a rule #2687 deleted, and no built-in's type is checked


octalide:
Part of #2742. Depends on #2687.

Two places in the SPIR-V back end still refuse a shape on the ground that "every mach vector is exactly 128 bits", which stopped being true when #2687 landed. Both refusals are now false statements about the language, and both block something #2687 and #2658 named as the reason to do the work.

## Repro

**The three compute built-ins.**

```mach
#[builtin("global_invocation")] var gid: u32x3;
rec B { v: f32x4; }
#[storage(0, 0)] var b: B;

#[stage("compute")] #[workgroup(64, 1, 1)]
fun c() { b.v = f32x4{1.0, 1.0, 1.0, 1.0}; }

fun main() i32 { ret 0; }
```

```
error: this compute built-in cannot be typed in mach yet: SPIR-V specifies
`global_invocation`, `local_invocation` and `workgroup_id` as three-component
unsigned vectors, and every mach vector is exactly 128 bits, so `u32x4` exists and
`u32x3` does not. ...
```

`u32x3` does exist. The refusal is `unspellable_builtin` in `spirv/emit.mach`, reached from `declare_interface` before the variable's type is even looked at.

**`Cross`.**

```mach
#[spirv_op("GLSL.std.450", "Cross")]
fun cross(a: f32x3, b: f32x3) f32x3;
```

```
error: that instruction set defines no instruction of this name, or the SPIR-V
target does not carry it; see `mach.lang.spirvop` for the accepted names and why
each omission is one
```

`mach.lang.spirvop` records why:

> - `Cross` is defined only for 3-component float vectors. mach has no `f32x3`
>   (briar-systems/mach#2687), so there is no type to declare it at. It becomes a
>   row the day that lands.

That day has come. The mechanism is already proven for three-component operands — a locally declared `#[spirv_op]` over `f32x3` emits correctly and validates today:

```
%16 = OpExtInst %v3float %17 Normalize %15
%20 = OpDot %float %18 %19
```

## Expected

- A `#[builtin("global_invocation")]` variable declared `u32x3` declares an `OpVariable` in `Input` storage decorated `BuiltIn GlobalInvocationId`, and the module passes `spirv-val --target-env vulkan1.3`. Same for `local_invocation` and `workgroup_id`.
[...]

Issue #2744, opened 2026-08-07
spirv: a sub-register-width vector interface variable is unreachable, and the refusal names the wrong cause


octalide:
Part of #2742. Depends on #2687.

A `#[input]`, `#[output]` or `#[builtin]` variable whose type is a vector narrower than the vector register cannot be read or written at all, and the refusal blames addressed memory rather than the lowering that produced it.

## Repro

```mach
#[input(0)]  var a: f32x3;
#[output(0)] var b: f32x3;
#[builtin("position")] var position: f32x4;

#[stage("vertex")]
fun v() { b = a; position = f32x4{0.0, 0.0, 0.0, 1.0}; }

fun main() i32 { ret 0; }
```

```
$ mach build . --target spirv
error: addressed memory is not yet supported by the SPIR-V target: a module-scope
variable is reachable only when it carries a shader interface role (`#[input]`,
`#[output]`, `#[builtin]`, or `#[uniform]`), and only as a whole-variable read or
write (MIR_STORE in v, at ./src/main.mach:6:15)
```

The variable *does* carry an interface role, and the access *is* a whole-variable write, so the sentence contradicts the program it is reported against. Substituting `f32x4` for `f32x3` in both declarations builds and validates.

The same variable is fine everywhere the value does not pass through a stage's interface:

| shape | result |
|---|---|
| `f32x3` as a function parameter and return | builds, valid |
| `f32x3` as a `#[uniform]` / `#[storage]` block member | builds, valid, `OpTypeVector %float 3` with correct offsets |
| `f32x3` as an `#[input]` / `#[output]` / `#[builtin]` variable | **refused** |

So this is not a typing gap — the interface declaration path already builds the right `OpTypeVector` for these shapes. It is a lowering gap in front of it.

## Expected

`b = a;` on a pair of `f32x3` interface variables emits `OpLoad %v3float %a` / `OpStore %b %…`, exactly as the `f32x4` pair does. `#[input(0)] var a: f32x3;` is the whole reason #2687 lists `vec3` vertex attributes as one of its two live needs, and a vertex attribute that cannot be read is not one.

## Actual

[...]

Issue #2742, opened 2026-08-07
epic(spirv): the sub-register-width vector leg of #2687


octalide:
Depends on #2687.

> ## Status (2026-08-07)
>
> Not started. The type half already works: `f32x3` reaches the emitter as `OpTypeVector %float 3`, it is laid out correctly inside a `#[uniform]` and a `#[storage]` block, and it passes and returns across a function boundary as a value. The **operation** half does not. Every arithmetic, comparison and lane-write on a sub-register-width vector emits a module whose result type is four lanes wide whatever the vector's real lane count is, with no diagnostic; construction fails with an unlocated internal error; and an `#[input]` / `#[output]` variable of one of these types cannot be read or written at all. None of it is covered by a test, so the eleven `int/surface/spirv-*` cases are green over all of it.

## What

#2687's acceptance ends with a SPIR-V bullet:

> `f32x3` reaches SPIR-V as `OpTypeVector %float 3`, and the `uvec3` built-ins in #2658 become spellable.

The ABI half of #2687 has landed and the type half of that bullet is satisfied. This epic is the rest of it, and it is larger than the bullet reads, because the SPIR-V back half has a second place where a vector's lane count is decided and that place still believes every vector is 128 bits.

The model that ties the children together is **which fact the lane count is read off**. There are two:

- the **IR type**, `IRT_VECTOR(element, lanes)`, which is correct for every shape. Everything routed through `ir_value_spv_type` — signatures, interface variables, uniform block members, `#[spirv_op]` operands — is right today, which is exactly why the type half already works.
- the **MIR lane descriptor**, `MirInstr.vec_lane`, which packs an element kind and an element byte width into one `u8` and *derives* the count as `16 / elem_bytes`. That derivation is the 128-bit assumption #2687 deleted from the language, still standing in MIR. Everything routed through it — the type of a computed vector result, a comparison's mask type, a build's expected operand count — is wrong for a sub-width shape.

SPIR-V is the only back end that reads a lane count off the descriptor at all, because it is the only one that has to *name* a vector type rather than pick an instruction for it. So the defect is invisible on every machine target and total on this one.

The third child is separate from that model: it is the two places in the back end that still refuse a shape by citing the rule #2687 removed.

## Children

- [x] #2743 the MIR lane descriptor's 128-bit lane-count derivation, which mistypes every sub-width vector operation
- [x] #2744 a sub-width vector interface variable is unreachable, and the refusal names the wrong cause
[...]

PR #2757, opened 2026-08-07
fix(codegen): carry the lane count in the MIR vector descriptor

### LINKED ISSUE #2743: spirv: a MIR lane descriptor derives its lane count from a 128-bit assumption, so every operation on a sub-width vector emits a mistyped module
Part of #2742. Part of #2742. The vector work in #2687 is otherwise complete: the type form, the layout rule, the codegen width split and the ABI classification all landed and are verified running on x86-64, aarch64 and riscv64. This is the remaining leg.

Every arithmetic, comparison, lane-write and construction on a vector narrower than 128 bits emits a SPIR-V module whose result type has four lanes rather than the vector's own, with no diagnostic. The Khronos validator rejects each one.

## Repro

```mach
fun add3(a: f32x3, b: f32x3) f32x3 { ret a + b; }
fun main() i32 { ret 0; }
```

```
$ mach build . --target spirv          # succeeds, no diagnostic
$ spirv-val out/int/bin/probe/main.spv
error: line 26: Expected arithmetic operands to be of Result Type: FAdd operand index 2
  %17 = OpFAdd %v4float %15 %16
```

The disassembly shows the whole shape of it:

```
    %v3float = OpTypeVector %float 3
          %5 = OpTypeFunction %v3float %v3float %v3float
    %v4float = OpTypeVector %float 4
          %1 = OpFunction %v3float None %5
          %6 = OpFunctionParameter %v3float
          %7 = OpFunctionParameter %v3float
         %14 = OpVariable %_ptr_Function_v4float Function   <- backing store for an f32x3
         %15 = OpLoad %v3float %11
         %16 = OpLoad %v3float %12
         %17 = OpFAdd %v4float %15 %16                      <- operands 3 lanes, result 4
         %18 = OpLoad %v4float %14
               OpReturnValue %18                            <- v4float out of a v3float function
```

The signature is right and the computation is wrong, in the same function, which is the signature of the root cause below.

The same failure, measured on `dev` at `0dc93031`:

| source | build | `spirv-val` |
|---|---|---|
| `fun f(a: f32x3, b: f32x3) f32x3 { ret a + b; }` | ok | `OpFAdd %v4float` — invalid |
| `fun f(a: f32x2, b: f32x2) f32x2 { ret a * b; }` | ok | `OpFMul %v4float` — invalid |
| `fun f(a: i32x2, b: i32x2) i32x2 { ret a & b; }` | ok | `OpBitwiseAnd %v4uint` — invalid |
| `fun f(a: f32x3, b: f32x3) u32x3 { ret a < b; }` | ok | `OpFOrdLessThan %v4bool` — invalid |
| `fun f(v: f32x3, x: f32) f32x3 { var r: f32x3 = v; r[1] = x; ret r; }` | ok | `OpCompositeInsert %v4float` — invalid |
| `fun f(k: f32) f32x3 { ret f32x3{k, k, k}; }` | **fails** | — |
| `fun f(v: f32x3) f32 { ret v[2]; }` | ok | valid |
| `fun f(a: f32x3) f32x3 { ret a; }` (call, no arithmetic) | ok | valid |
| `fun f(a: f32x4, b: f32x4) f32x4 { ret a + b; }` (control) | ok | valid |

The construction row fails with

```
error: spirv.emit: vector build lane count does not match its vector type
```

which carries no source location, no function name and no position, because it is an internal `R.err` rather than a located refusal — the one place the mismatch is checked is the one place it is reported as a compiler bug rather than as a program one.

## Expected

A vector operation on any lane count the target admits emits a SPIR-V module whose result type is that vector's own type, and `spirv-val` accepts it. Where the count is one a Shader module cannot declare, the existing named refusal fires (`i8x16` / `i16x8` already do this correctly).

## Actual

The result type is `OpTypeVector <elem> (16 / elem_bytes)` regardless of the vector's declared lane count, so it is right only for the ten seeded 128-bit shapes. A build is the sole case that notices, and only because it counts operands.

## Root cause

`mir.vec_lane` is a `u8` packing an element kind in the high nibble and an element byte width in the low nibble (`mir.vec_lane_make`). It does not carry a lane count. The count is derived:

```mach
# the lane count of a vector descriptor (16 / element-bytes), or 0 for a non-vector
pub fun vec_lane_count(v: u8) u32 {
    val eb: u8 = vec_lane_elem_bytes(v);
    if (eb == 0) { ret 0; }
    ret 16 / eb::u32;
}
```

`16 /` is the 128-bit assumption, in the one representation #2687 did not reach. It was sound while every vector was exactly 128 bits, and it stayed sound on every machine target after #2687 because none of them reads a count off it — a machine back end selects an instruction from the descriptor's *class and lane width*, which are both still correct, and the register it writes is the same register whatever the count.

SPIR-V is the exception because it must name the result type. `spirv/emit.mach:value_type_for` routes a vector-tagged instruction to `vector_type_of(e, mi.vec_lane, ...)`, which ends in `types.type_vector(e.tt, lane_ty, mir.vec_lane_count(lane))`. That is where the four appears. It then propagates: the vreg's backing `OpVariable` is declared at the wrong type, every load and store of it is wrong, and `emit_vec_build` compares `mi.operand_count - 1` against `vector_shape` of the wrong type and is the only site that raises anything.

Everything on this target that is *right* bypasses the descriptor entirely and reads `ir_value_spv_type` off the IR type: signatures, `OpFunctionParameter`, interface variable declarations, uniform block members, and `#[spirv_op]` operands. That last one is worth stating, because it proves the type machinery is already correct for these shapes:

```mach
#[spirv_op("GLSL.std.450", "Normalize")] fun nrm3(v: f32x3) f32x3;
#[spirv_op("core", "OpDot")]             fun dot3(a: f32x3, b: f32x3) f32;
fun use(a: f32x3, b: f32x3) f32 { ret dot3(nrm3(a), b); }
```

emits `%16 = OpExtInst %v3float %17 Normalize %15` and validates.

## Fix

The lane count has to become a fact the descriptor carries rather than one it computes. Candidate designs:

1. **Widen the descriptor.** `vec_lane` becomes a `u16` (or a small record) carrying kind, element bytes and lane count, `vec_lane_count` returns the stored field, and `vec_lane_make` gains a count argument. Every construction site in `mir/lower.mach` already holds the IR vector type when it builds the descriptor, so the count is available at all of them.
2. **Derive the count from the instruction's width instead of a constant.** `MirInstr.width` is the vector's byte width, so `width / elem_bytes` is the count. Smaller diff, no representation change, but it makes the descriptor's meaning depend on a second field that other opcodes use differently, and it silently produces a wrong count for any instruction whose `width` is the register's rather than the value's — which is exactly the class of bug #2687's `refactor(codegen): split a value's register width from its memory footprint` was written to separate.
3. **Have the SPIR-V emitter read the IR type rather than the descriptor.** It cannot in general: a MIR instruction's result vreg has no IR type once lowering has run, which is why `value_type_for` exists.

**Preferred: (1).** It removes the assumption rather than relocating it, it is the representation the reader expects to find, and it is what #2727 will need anyway — an over-width vector split into register-sized chunks has a per-chunk lane count that no single global constant can express.

The `emit_vec_build` mismatch check should stay, but become a located refusal through `unsupported(e, ...)` rather than a bare `R.err`, so that if the two facts ever disagree again the message names the function and the source position instead of reading as an internal error.

## Acceptance

- Every row of the table above builds and passes `spirv-val`, including the construction row.
- A new `int/surface/spirv-vector-narrow` case, `build-target: spirv`, `run: spirv-val`, covering `f32x2`, `f32x3`, `i32x3` and `u32x3` across add, multiply, compare, lane read, lane write and construction, plus a call passing and returning each. The existing `spirv-vector` case stays 128-bit only and its comment stating so stays true.
- `spirv-val` is the observable, not a golden alone: the failure mode here is a module mach's own reader accepts and a consumer rejects, which is the exact thing a golden cannot see.
- The eleven existing `spirv-*` cases stay green on `linux/debug` and `linux/release`, byte-identical goldens.
- Self-host fixpoint holds, since the descriptor's representation changes under every target.

octalide:
Closes #2743. Part of #2742.

## What was wrong

`mir.MirInstr.vec_lane` was a `u8` packing a lane class and an element byte-width, with the lane count **derived**:

```mach
pub fun vec_lane_count(v: u8) u32 {
    val eb: u8 = vec_lane_elem_bytes(v);
    if (eb == 0) { ret 0; }
    ret 16 / eb::u32;
}
```

`16 /` is the 128-bit assumption #2687 removed from the language, still standing in the one representation it did not reach. It is right for the ten shapes that are exactly one 128-bit register and wrong for every other one.

SPIR-V is the only target that must **name** a vector type, so it is the only place a lane count is read off the descriptor at all, and the mistyping was total there and invisible everywhere else.

## The new shape, and why it survives #2727

`vec_lane` is a `u32`: class in bits 24-27, lane byte-width in bits 16-23, **lane count in bits 0-15**.

Widened rather than restructured into a record, deliberately. The descriptor is compared with `!= 0`, assigned from `VEC_LANE_NONE`, and equality-compared against a `vec_lane_make(...)` in the isel tests, in around sixty places across five backends. Keeping it one scalar with one identity leaves every one of those correct by construction, and there is no arrangement of a record that also makes "a descriptor with a class but no count" unrepresentable.

Sixteen bits of count, not the four a 128-bit register would need. The hardest plausible member is the **over**-width vector of #2727, not the sub-width one in front of me: `f32x8` is 8 lanes, `f32x1024` is 1024, and #2727 explicitly leaves a surface cap open. A count spanning 1..16 would have had to be rebuilt the moment that lands. Class and width keep whole fields of their own so a lane width above 8 does not interact with the packing either.

`vec_lane_make(kind, elem_bytes, lanes)` takes the count **positionally**, so a descriptor cannot be built without one - an omission is an arity error at build time rather than a plausible default reaching the emitter. `ir_type.vector_lane` gained an `out_lanes` for the same reason: a caller holding the class and width but not the count is precisely the shape that invented one, and two callers were already digging `data.vector_.lanes` out by hand beside it.

## Producers audited

One real producer: `mir/context.vec_lane_for_type`, which reads the IR type. Everything else is either a clear to `VEC_LANE_NONE` (`legalize`, `regalloc`, `frame`, `looprotate`, `mos6502/encode`), a whole-instruction copy (`rules.emit_piece`, `mir.instr_like`), one full-register `.16b` NEON `VORR` in `arm64/encode`, or an isel/printer/encoder unit test. Every call site was visited by the arity change; the test sites all construct 128-bit shapes and now say so.

## Consumers audited

`vec_lane_count` has exactly three non-test consumers and all three are in `spirv/emit.mach`: `vector_type_of` (the result type of every computed vector), `emit_vector_compare` (the mask type and the intermediate bool vector), and `unsupported_value_type` (the refusal). x86-64, aarch64, riscv64 and mos6502 read `vec_lane_kind` and `vec_lane_elem_bytes` and never the count - they select an instruction from the class and width and write a whole register - which is exactly why they were correct on the bug and why a green machine suite proved nothing.

`wide_vector_refusal` now keys on the count, which is the fact SPIR-V actually refuses, and pins each named shape to its lane width so an over-width `f32x8` from #2727 cannot be reported as a `u16x8`.

## Evidence

`spirv-val` is the observable throughout, not our own expectations.

**Before**, compiler built from `origin/dev` at `914e5dfb`:

```
$ spirv-val out/int/obj/case/main.spv
error: line 45: Expected arithmetic operands to be of Result Type: FAdd operand index 2
  %21 = OpFAdd %v4float %19 %20
```

The new int case run against that same compiler does not even build:

```
FAIL surface/spirv-vector-narrow [linux/debug] (build)
    error: spirv.emit: vector build lane count does not match its vector type
```

and with the construction rows stripped so the rest gets to the validator, the module contains **71** `%v4`-typed lines for a source with no four-lane vector in it.

**After**: `spirv-val` clean, and **0** `%v4` lines in the same module. The `f32x3` repro now reads

```
%19 = OpFAdd %v3float %17 %18
%48 = OpFOrdLessThan %v3bool %46 %47
%53 = OpSelect %v3uint %48 %50 %52
%65 = OpCompositeInsert %v3float %64 %63 2
```

`int/surface/spirv-vector-narrow` covers `f32x2`, `f32x3`, `i32x2` and `u32x3` through arithmetic, bitwise, compare-to-mask, lane read, lane write, construction, control flow and calls.

**CPU targets, lane values checked by hand** (both profiles on x86-64, aarch64 and riscv64, sub-width vectors through arithmetic and across function boundaries):

```
f32x3 add 11 22 33          (1,2,3) + (10,20,30)
f32x3 chain 121 484 1089    (a+b)*(a+b) across two calls
f32x2 mul 21 55             (3,5) * (7,11)
i32x2 add 2 10              (-4,9) + (6,1)
u32x3 and 240 3840 240      0xF0F0,0xFF00,0x00FF & 0x0FF0
u32x3 mask 4294967295 4294967295 0
f32x3 setlane 11 99 33
```

Identical on all three, both profiles. The machine targets were green on the bug, so this is values rather than a suite verdict.

## Suites

- `mach test .` - 1286 passed, 0 failed
- `int` linux - all cases passed
- `int` linux-riscv64 - all cases passed
- `int` linux-arm64 - `c-variadic` and `narrow-stack-args` fail on this developer host, **identically on `origin/dev`** (a host `cc` cross-compiling for aarch64, and qemu). The native `linux-arm64` leg is `pr` cadence, so CI settles it.
- self-host fixpoint: stage2 == stage3, `sha256 818b907de6527273735f1ea71a64e55f80988a0e2b6e86e607026854d78d974f`

## Siblings

Neither #2744 nor #2745 falls out of this. What does fall out for free is the third row of the #2743 table: `f32x3{k, k, k}` was the only spelling that failed loudly, as an unlocated internal error, because `emit_vec_build` is the one site that compares operand count against the type - it builds and validates now.

One **new** defect surfaced while writing the coverage, and it is not any of the three children: an **uninitialized** sub-width vector local is refused on this target as addressed memory.

```
fun z() f32x3 { var v: f32x3; ret v; }
error: addressed memory is not yet supported by the SPIR-V target ... (MIR_STORE in z)
```

It fails for `f32x2`, `f32x3`, `i32x2` and `u32x3`, passes for `f64x2` and `f32x4`, and behaves identically before and after this change - a lowering question, not a typing one, in the same family as #2744. Filed separately; the case says why the shape is absent from it.


octalide:
Filed #2758 for the sub-width zero-init defect this coverage surfaced, and linked it under #2742. It is not part of this PR: it reproduces identically on `dev` before and after this change, and it is a lowering question in the same family as #2744.

Issue #2288, opened 2026-07-25
epic: diagnostic surfaces must be able to represent the defects they are used to find


octalide:
> ## Closed 2026-08-08. Converted: the defects are trackable, the audit is not.
>
> This epic asked for an audit of every observability surface. That framing cannot
> terminate, and the reason is a measurement rather than a feeling: **every surface
> that was actually examined produced at least one defect**, and three surfaces are
> still unexamined. At roughly one new gap per surface, "audit everything" is a
> standing concern about how this codebase builds diagnostic surfaces, not a unit of
> work that closes.
>
> The individual defects do close. Five have, and the concrete remainder is filed as
> three ordinary issues below.
>
> **What was found, and how.** Every entry below was established by producing output
> and diffing it, never by reading the printer. That distinction is not procedural
> nicety — every defect here survived being read.
>
> | surface | outcome |
> |---|---|
> | `--emit-asm` (instance 1) | fixed, #2192 / #2197 |
> | the IR float printer's signed zero (instance 3) | fixed, #2274 |
> | goldens over emitted code (instance 2) | a statement, not a patch; it stands, and is now #2822's first entry |
> | `--emit-ir` float constants | **#2753** — `1.5` and `1.75` both `f1`, `+inf` as a finite-looking number, the smallest denormal as `f0`. Three collapses in seven values |
> | `--emit-asm` inline-asm symbol operands | **#2788** — named a *different real symbol* from the same module, and dropped a direct `call`'s operand. x86-64 only |
> | `--verify-ir` conversion widths | **#2808** — accepted a `BITCAST` changing width by a factor of four |
> | diagnostics that print a type | **#2819** — two same-named records from different modules rendered `expected P, found P` |
> | `--emit-mir` per ISA | **not examined** → #2821 |
> | the golden / `expect.txt` harness | enumeration **not written** → #2822 |
> | `mach build -v` phase readout | **measured and unfixed** → #2816 |
>
> **Three findings worth carrying forward, because they are about the class and not
> about any one bug.**
>
> *1. A collapse is usually pinned as correct.* #2753's truncation was deliberate,
> commented, and held in place by a passing test asserting `-1.5` renders `"f-1"`.
> #2788's dropped operand was held in place by two unit tests asserting
> `dst.kind == OPK_NONE`. When auditing a surface, assume its tests encode the
> collapse; do not read them as evidence.
>
> *2. Per-ISA means per-ISA.* #2788 was present on exactly one of three register
[...]

PR #2761, opened 2026-08-07
fix(ir): render float constants in shortest round-trippable decimal

### LINKED ISSUE #2753: ir: --emit-ir renders a float constant as its truncated integer magnitude, so 1.5 and 1.75 print identically and an infinity prints as a finite number
Part of #2288.

## What

`--emit-ir` renders a float constant as its **truncated integer magnitude**, so the surface cannot show any distinction that lives below the decimal point, at the top of the range, or at the bottom of it. Found while triaging #2288, which names `--emit-ir` and asks exactly this question of it.

Measured on `dev` @ `7ca88f1e`, one function per constant, `--emit-ir` on `linux-x86_64`:

| source | rendered |
|---|---|
| `ret 1.5;` | `ret f1: f64` |
| `ret 1.75;` | `ret f1: f64` |
| `ret 1.0e400;` (`+inf`) | `ret f9223372036854775808: f64` |
| `ret -1.0e400;` (`-inf`) | `ret f-9223372036854775808: f64` |
| `ret 9223372036854775808.0;` (finite) | `ret f9223372036854775808: f64` |
| `ret 5.0e-324;` (smallest denormal) | `ret f0: f64` |
| `ret 0.0;` | `ret f0: f64` |

Three collapses in seven values. `1.5` and `1.75` are indistinguishable. `+inf` is indistinguishable from a large finite value **and** renders as a plausible-looking finite number rather than announcing itself. The smallest denormal is indistinguishable from zero.

**The compiler is right about every one of them.** Bit patterns read back at run time through `:~u64`, all three targets:

```
1.0e400  -> 9218868437227405312    (0x7FF0000000000000, +inf)
5.0e-324 -> 1                      (the smallest denormal)
9223372036854775808.0 -> 4890909195324358656
```

So this is purely a defect of the view, which is what makes it #2288's subject rather than a codegen bug.

## Why it is wrong by design rather than an oversight

`src/lang/me/ir/printer.mach:330` states the decision and its reason:

```mach
# std has no float formatter; print the truncated integer magnitude
# tagged `f` so the dump stays readable (it is not reloadable anyway).
```

and `mach.lang.me.ir.printer.print_value:signed_zero` **pins the truncation as correct**:

```mach
bad = bad | float_renders_as(?m, -1.5, ty, "f-1");
bad = bad | float_renders_as(?m, 2.5, ty, "f2");
```

That test is the shape #2288 warns about in its own bar. It is not wrong about anything it asserts, and it makes the missing distinction look like an intended one, so a printer fix has to change the test rather than merely pass it.

**The stated reason is stale.** `std.format.write_f64` writes "shortest round-trippable decimal" and works today (`printlnf("{} {} {}", 1.5, 1.75, 0.1)` prints `1.5 1.75 0.1`). The compiler already imports `std.format` in eight modules, including `be/codegen/encode.mach` and `me/lower/mangle.mach`, so `printer.mach` reaching it is a `use` line rather than a dependency decision.

The secondary claim is also weak now. "It is not reloadable anyway" argues that fidelity does not matter because the dump is not an input. #2288's whole thesis is that the dump is what someone reads **when they already suspect a bug**, and a float bug is precisely where the fraction is the evidence. #2274 was a float constant defect debugged through this printer, and this is the same printer withholding four more digits than the one it withheld then.

## Fix

Render the constant through `std.format.write_f64`, in shortest round-trippable decimal, and give infinities and NaN their own spellings rather than letting `f::u64` produce a number for them.

Candidates:

1. **`write_f64` plus explicit `inf` / `-inf` / `nan` spellings.** Preferred. One `use`, one call, and the three non-finite classes stop rendering as integers. Signed zero keeps rendering `f-0` / `f0`, which #2274 established and which shortest-round-trip decimal preserves anyway.
2. Render the raw IEEE bit pattern in hex (`f0x7FF0000000000000`). Total by construction and unreadable for the common case, which is the property the current code was trying to protect.
3. Both: decimal with the bit pattern in a trailing comment. More output than the dump wants.

**Preferred: (1).** It is the smallest change that makes every distinction the middle end keeps apart visible, and (2)'s totality is not worth losing readability the printer explicitly optimises for.

A NaN payload is worth deciding rather than inheriting: `nan` alone collapses distinct payloads, and the middle end does keep float constants apart by bit pattern (`cse.mach` value-numbers them that way). Spelling a non-default payload explicitly is the coherent answer.

## Acceptance

- `1.5` and `1.75` render distinguishably, proved by producing both and diffing the output rather than by reading the printer.
- `+inf`, `-inf` and NaN render as themselves and never as an integer magnitude.
- The smallest denormal renders distinguishably from `0.0`.
- `-0.0` still renders `f-0` and `0.0` still renders `f0`, so #2274's fix is not regressed.
- `mach.lang.me.ir.printer.print_value:signed_zero` is updated rather than worked around, and its `-1.5 -> "f-1"` / `2.5 -> "f2"` rows become the real renderings. A rename is appropriate since the test now pins float rendering generally.
- Every emitted-IR golden that moves is explained as a previously-invisible distinction becoming visible, per #2288's bar. Objects stay byte-identical: this must not move codegen.
- `mach test .` green, `int` green, self-host fixpoint holds.

## Environment

- toolchain: `mach 4.15.0`, built from `origin/dev` @ `7ca88f1e` (ahead of the v4.15.0 tag)
- triples: measured on `linux-x86_64`, bit patterns confirmed identical on `linux-arm64` and `linux-riscv64`
- OS: Linux x86-64

octalide:
Closes #2753. Part of #2288.

## Before, on the unpatched compiler

Built `origin/dev` @ `6991413e`, one function per constant, `--emit-ir` on `linux-x86_64`:

```
ret f1: f64                       # 1.5
ret f1: f64                       # 1.75
ret f9223372036854775808: f64     # +inf
ret f-9223372036854775808: f64    # -inf
ret f9223372036854775808: f64     # 9223372036854775808.0, a finite value
ret f0: f64                       # 5.0e-324, the smallest denormal
ret f0: f64                       # 0.0
ret f-0: f64                      # -0.0
ret f0: f64                       # 0.1
ret f1: f32                       # 1.25
```

Three collapses, exactly as the issue reports: `1.5` = `1.75`, `+inf` = the finite 2^63, denormal = zero.

## After

```
ret f1.5: f64
ret f1.75: f64
ret finf: f64
ret f-inf: f64
ret f9.223372036854776e18: f64
ret f5e-324: f64
ret f0: f64
ret f-0: f64
ret f0.1: f64
ret f1.25: f32
```

Every distinction the middle end keeps is now visible, and `f-0` / `f0` still differ, so #2274 is not regressed.

## What changed

`print_value` now renders a float constant through `std.format.write_f64` in shortest round-trippable decimal. That function already spells both signed zeros and both infinities, so the only class it collapses is NaN, whose payloads `cse.mach` value-numbers apart. A NaN whose bits are not the canonical quiet `0x7FF8000000000000` therefore renders `fnan:0x<bits>` rather than a bare `fnan`.

The stated reason for the old rendering, "std has no float formatter", was stale: `printer.mach` already imported `std.format`.

## The test that pinned the defect

`print_value:signed_zero` asserted `-1.5 -> "f-1"` and `2.5 -> "f2"`, which made a missing distinction look intended. It is replaced by `print_value:float_rendering`, which pins the real rendering across the IEEE classes. Operands are built from bit patterns (`float.bits_f64(0x7FF0000000000000)`, `float.bits_f64(1)`, `float.bits_f64(0x43E0000000000000)`) rather than decimal literals, and every expectation was derived from IEEE-754 before the code was run, not captured from output.

Mutation-checked: grafting the new test onto the pristine `origin/dev` printer fails it.

```
FAIL  mach.lang.me.ir.printer.print_value:float_rendering  ./src/lang/me/ir/printer.mach:820  (exit 1)
```

## Verification

- No golden moves. There are no `.ir` goldens in `int/`, and nothing outside `printer.mach` referenced the old test.
- Objects byte-identical. Built the repro project with a compiler from pristine `origin/dev` and with this branch's compiler; `diff -r` over the whole output tree reports no difference. This moves nothing but the dump.
- `mach test .`: 1288 passed, 0 failed.
- `int/run.sh --target linux --deps pin`: all cases passed.
- Self-host fixpoint holds, stage2 == stage3 == `6c3f4461064c517e4b20e33a16fce310d7db7307b6661d3cf22120d69db43ebd`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,15 @@ The refusal itself is unchanged. Its diagnostic and `doc/language/decorators.md`
 
 ### Fixed
 
+#### `--emit-ir` shows a float constant, not its truncated integer magnitude (#2753)
+Every float constant in an IR dump rendered as the integer part of its magnitude, so the surface could not show any distinction living below the decimal point, at the top of the range, or at the bottom of it. `1.5` and `1.75` both printed `f1`. A positive infinity printed `f9223372036854775808`, which is also what the finite value 2^63 printed, so an infinity read as a plausible finite number and the two were indistinguishable. The smallest denormal printed `f0`.
+
+The middle end never lost any of this: `me/pass/cse.mach` value-numbers float constants by bit pattern and `me/pass/algebraic.mach` declines to fold `x + 0.0`, precisely so the IR keeps IEEE distinctions apart. Only the view collapsed them, which matters because a dump is what someone reads when they already suspect a float bug and the fraction is the evidence.
+
+Constants now render through `std.format.write_f64` in shortest round-trippable decimal, so `f1.5`, `f1.75`, `f5e-324` and `f9.223372036854776e18` are each themselves, and the non-finite classes spell themselves as `finf`, `f-inf` and `fnan` rather than borrowing an integer. A NaN whose payload is not the canonical quiet one renders its full bit pattern as `fnan:0x…`, since a payload is a distinction the middle end keeps too. Signed zero still renders `f-0` and `f0`, which #2274 established.
+
+The reason recorded in the code for the old rendering - that std had no float formatter - had gone stale; `printer.mach` already imported `std.format`. The unit test that pinned `-1.5` to `f-1` and `2.5` to `f2` was enforcing the defect rather than any intended behaviour, and now pins the real rendering across the IEEE classes, built from bit patterns rather than decimal literals. Objects are byte-identical: this moves nothing but the dump.
+
 #### An over-aligned stack local is over-aligned at run time (#2735)
 `#[align(32)]` and above worked on a global and did nothing on a local. The slot's offset was a correct multiple of the requested alignment, but it was measured from the frame pointer, and the only alignment that reaches the frame pointer is the ABI's 16-byte call boundary. So the address was a multiple of 32 exactly when the process stack happened to start on one, which depends on the environment block and therefore changes between runs of the same binary. `$size_of` already reported the padded size, so the storage was sized for a promise the placement did not keep.
 
--- a/src/lang/me/ir/printer.mach
+++ b/src/lang/me/ir/printer.mach
@@ -23,6 +23,11 @@
 # as `phi <ty> [bb0 %0: i32, bb1 %3: i32, ...]`. Per-instruction modifier flags
 # render as opcode suffixes (`add.nsw`, `load.volatile`).
 #
+# Float constants render as `f` followed by the shortest round-trippable decimal
+# (`f1.5`, `f1.75`, `f5e-324`), with `finf` / `f-inf` / `f-0` for the classes an
+# integer magnitude cannot spell and `fnan:0x<bits>` for a NaN whose payload is
+# not the canonical quiet one.
+#
 # Every value operand at a use site carries its own IR type as a `<value>: <ty>`
 # suffix, so a mis-typed operand (an aggregate passed as a bare pointer) is
 # visible without chasing def-use chains. An extern declaration renders its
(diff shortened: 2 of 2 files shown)


octalide:
Integration legs, run locally with `--deps pin` against this branch's compiler:

| leg | result |
|---|---|
| `linux` (native) | all cases passed |
| `linux-riscv64` (qemu, as `targets.conf` specifies) | all cases passed |
| `linux-arm64` | 4 failures, all pre-existing and host-artifacts |

The `linux-arm64` leg is `native` in `targets.conf` and I have no aarch64 host, so I ran it under `--runmode qemu`. `surface/c-variadic` dies on SIGILL under qemu-user and `surface/narrow-stack-args` fails its build because its `cc` probe step compiles for the x86-64 host. **Both fail identically when the same filtered run is done with a compiler built from pristine `origin/dev` @ `6991413e`**, so neither is this branch. The native `ubuntu-24.04-arm` leg in CI is the authority here, per `targets.conf`'s own note that a qemu leg is not equivalent.

Issue #2755, opened 2026-08-07
debuginfo: every -g image emits ~200 duplicate file_names warnings from dwarfdump --verify


octalide:
Found during the #1594 triage.

## What

Every `-g` image, ELF and Mach-O alike, produces about **200 warnings** from `llvm-dwarfdump --verify`:

```
.debug_line[...].prologue.file_names[1] is a duplicate of file_names[0]
```

Errors are clean. The warnings are not, and they are systematic rather than incidental.

## Why it is worth fixing rather than tolerating

#1594 was closed on the strength of `--verify` being clean, and "clean" meant **no errors**. A standing wall of warnings degrades that signal in exactly the way this repo keeps getting caught by: the next real warning arrives into a stream nobody reads, and a verification tool whose output is routinely noisy is one that has stopped verifying. The value of running a validator at all is that its output is expected to be empty.

The duplicate itself is also a real if minor waste. DWARF 5 changed the file-name table so that entry 0 is the primary source file rather than a reserved slot, and emitting the same file again at index 1 is the shape of a producer that was written against the DWARF 4 convention and not revisited. Worth confirming that is the cause rather than assuming it.

## Fix

Emit the file-name table with entry 0 carrying the primary source file and no duplicate at index 1, per the DWARF 5 rule. Check whether any consumer in the tree indexes file entries assuming the DWARF 4 layout, since changing the table changes those indices.

## Acceptance

- `llvm-dwarfdump --verify` is clean of both errors **and** warnings on a `-g` build, for ELF and Mach-O.
- A `--verify` run is exercised in CI so the stream stays empty rather than drifting back.
- Line-table lookups still resolve: an `addr2line` round trip matches the DIE it should, checked against a known source position.

octalide:
## The DWARF 4 theory does not hold, and the fix as stated would regress #2582

The issue asks to confirm the cause rather than assume it, so I implemented the proposed change and measured it: `FileTable` 0-based, entry 0 the primary source, no repeat at index 1, `ft.count` entries emitted.

`llvm-dwarfdump --verify` went completely clean, errors and warnings both. And binutils stopped being able to read the section at all:

```
$ addr2line -f -e ./dbg2 $(readelf -hW ./dbg2 | awk '/Entry point address:/{print $NF}')
addr2line: DWARF error: mangled line number section (bad file number)
...(one per CU, 118 of them)...
```

That is #2582 verbatim. #2582 is closed, and its fix is what introduced this duplicate, on purpose.

[...]

Issue #2582, opened 2026-08-07
debuginfo: line table rejected by addr2line (bad file number) while llvm-dwarfdump --verify passes


octalide:
## What

`addr2line` cannot read the line-number program in a `-g` build. On a release `-g`
self-build of the compiler:

```
$ addr2line -f -e mach.g 0x400de9
addr2line: DWARF error: mangled line number section (bad file number)
...(repeated per CU)...
unwrap
./././dep/mach-std/src/types/option.mach:54
```

Function/file/line resolution does eventually come back, but every CU emits the
error first, so at minimum the file-index encoding in `.debug_line` is wrong (a
file number outside the declared file table, or a v4/v5 file-index base-off-by-one).

Found incidentally while localizing #2572; unrelated to that bug's cause or fix.

## Why this matters beyond the bug

**`llvm-dwarfdump --verify` passes on the same binary.** Our only automated debug-info
check is `int/surface/debuginfo`, which runs exactly that verifier — so the case is
green while a standard, extremely common consumer (`addr2line`, and by extension
`perf`, `gdb`'s addr2line paths, and most crash symbolizers) rejects the section.

The verification gap is at least as valuable as the defect: `--verify` checks
structural/reference integrity, not that the line program decodes against the file
table the way a consumer will read it.

## Acceptance

- `addr2line -f -e <mach -g build> <addr>` resolves with no DWARF error on every CU.
- `int/surface/debuginfo` gains a consumer-side assertion so this class cannot pass
  unnoticed again — running `addr2line` over a known address and requiring clean
  stderr, rather than relying on `--verify` alone.
- Root cause names the encoder site in `src/lang/be/codegen/dwarf.mach`.

octalide:
Related: #2583 (also debug-info fidelity, also found while fixing #2572). Both share the same verification gap — `llvm-dwarfdump --verify` accepts output that real consumers misread, and `int/surface/debuginfo` runs only that verifier.

Issue #1594, opened 2026-06-24
debuginfo: full DWARF 5 debug info that survives optimization (ELF + Mach-O)


octalide:
> ## Status (2026-08-07): all ten children are closed and the acceptance is measured green, with one open nit
>
> This epic had no status block and no `## Children` checklist. Both added. Every one
> of the ten native sub-issues is closed.
>
> Verified today by building and running against `dev` @ `7ca88f1e`, not by reading:
>
> | check | result |
> |---|---|
> | ELF linked exe, `-g`, x86-64 release | `llvm-dwarfdump --verify` **No errors** |
> | Mach-O linked exe, `-g`, `darwin-aarch64` | `llvm-dwarfdump --verify` **No errors** |
> | DWARF version | `version = 0x0005`, `DW_UT_compile`, `DW_AT_producer ("mach 4.15.0")` |
> | in-binary `__DWARF` on Mach-O (#1702) | present in the linked executable, verify clean |
> | `.debug_loclists` / `.debug_rnglists` (#1705, #1707) | both sections present in the linked ELF |
> | inlining transparent on optimized output (#1707) | 48,373 `DW_TAG_inlined_subroutine` DIEs in the release `-g` compiler |
> | `addr2line` round-trip | `0x41ad20` -> `negz32` at `./src/main.mach:4`, matching the DIE's `DW_AT_low_pc` and `DW_AT_decl_line` |
>
> **The one thing that is not clean.** `--verify` reports 200 warnings on both the ELF
> and the Mach-O image, all the same shape:
>
> ```
> warning: .debug_line[0x00000aa3].prologue.file_names[1] is a duplicate of file_names[0]
> ```
>
> The line-table prologue emits the primary source file twice, once as entry 0 and once
> as entry 1. It is a warning rather than an error and consumers tolerate it, so the
> acceptance line "llvm-dwarfdump --verify clean" is met on errors and not on warnings.
> Worth its own small issue against the `.debug_line` producer rather than holding this
> epic open.
>
> A second gap in the acceptance that I could not test here: `break file:line`,
> single-step and `frame variable` need a real debugger session, and nothing in this
> repo's harness drives one. The DWARF is structurally valid and self-consistent, which
> is not the same claim.
>
> **Not closing this.** Every child is done and the byte-level acceptance measures
> green, but the debugger-behaviour half of the acceptance is unverified and the
> `file_names` duplication is unresolved. Both are owner calls.
>
> ---

## What
[...]

PR #2763, opened 2026-08-08
fix(int): read the dwarfdump --verify stream, not its exit status

octalide:
Refs #2755. Part of #2288. **Deliberately not `Closes`** - the producer change #2755 asks for is wrong, and the measurement is below.

## The DWARF 4 theory does not hold

#2755 proposes the duplicate is "the shape of a producer written against the DWARF 4 convention and not revisited", and asks to confirm rather than assume. I implemented the change - 0-based `FileTable`, entry 0 the primary source, no repeat at index 1 - and measured it.

`llvm-dwarfdump --verify` went completely clean, errors and warnings both. And every libbfd consumer stopped being able to read the section:

```
$ addr2line -f -e ./dbg2 $(readelf -hW ./dbg2 | awk '/Entry point address:/{print $NF}')
addr2line: DWARF error: mangled line number section (bad file number)
...(x118, one per CU)...
```

That is #2582 verbatim, which is closed and whose fix introduced this duplicate on purpose.

The cause is that DWARF 5 contradicts itself. §6.2.4 numbers file entries from 0 and makes entry 0 the primary source. §6.2.2 nonetheless still gives the line state machine's `file` register an initial value of **1**. binutils resolves that in favour of §6.2.2 and validates the initial register against the table before any `DW_LNS_set_file`, so a table whose only entry is slot 0 is rejected outright.

Confirmed against both reference producers on this host, so it is not a mach-specific reading:

| producer | file table | `llvm-dwarfdump --verify` | `addr2line` |
|---|---|---|---|
| clang 21 `-gdwarf-5` | one entry, slot 0 only | clean | **`mangled line number section (bad file number)`** |
| gcc 16.1.1 `-gdwarf-5` | slot 0 and slot 1, both the primary | **`file_names[1] is a duplicate of file_names[0]`** | clean |
| mach `dev` | slot 0 and slot 1, both the primary | same warning | clean |

So mach emits gcc's shape, llvm-dwarfdump warns on gcc's output identically, and the warning is about the spec, not about us. Dropping the duplicate trades a cosmetic warning from one validator for a real decode failure in addr2line, perf, and gdb's bfd path.

Also worth recording: #2582's commit message says the reserved-slot shape is "the shape gcc and clang both emit". The gcc half is right and the clang half is wrong - clang emits one entry, and binutils cannot read it. `dwarf.mach`'s comment now says so, so the next person does not repeat the experiment I just ran.

## What this PR does deliver

The half of #2755 that is a real defect and is squarely #2288's subject: **the observable could not represent a warning.**

`produce_debuginfo` read `--verify`'s exit status, which is zero on warnings. "No diagnostics at all" and "no errors, and ~200 warnings on every `-g` image" both rendered `dwarfdump_verify=clean`. The golden asserted the first while the tree was in the second - the epic's question ("can two states the compiler distinguishes render identically?") answered yes, on the surface #1594 was closed on.

It now reads the stream: `errors` / `warnings` / `clean`, plus the distinct warning texts, sorted and deduplicated with only the per-CU `[0x...]` offset elided. So a class that is present and explained pins itself in the golden, and any new class is a diff.

## The golden move, derived not blessed

```
-dwarfdump_verify=clean
+dwarfdump_verify=warnings
+dwarfdump_warning=.debug_line[].prologue.file_names[1] is a duplicate of file_names[0]
```

Both lines are the truth about today's producer, and `clean` was the collapse. There is exactly **one** warning class and no other, which I checked independently of the harness on `-g` images for `linux-x86_64`, `linux-arm64`, `linux-riscv64` and `darwin-aarch64`. The Mach-O count in my fixture is 12 warnings, one per CU, same class - so the issue's report that ELF and Mach-O both warn is one cause, not two.

## Verification

- `surface/debuginfo` passes on all three ELF legs, `linux`, `linux-arm64`, `linux-riscv64`, in both profiles.
- Against the unpatched harness the case reports `dwarfdump_verify=clean` while the stream carries the warning wall; against this one it names it.
- `mach test .`: 1288 passed, 0 failed.
- `int/run.sh --target linux --deps pin`: all cases passed.
- The `dwarf.mach` change is comment-only, and the compiler this branch builds is **byte-identical** to the one `origin/dev` @ `6991413e` builds.

## What is left on #2755

Its acceptance asks for a stream clean of warnings as well as errors. That is not reachable from the producer without breaking binutils, so it needs a decision rather than a patch. Written up on the issue.


octalide:
## Correction: the warning is validator-version dependent, and the golden had to change

CI's int legs failed on the first push, and the reason is worth recording because it changes what this PR can honestly assert.

The runner installs Ubuntu's `llvm` package, **llvm-dwarfdump 18**, which does not report the duplicate-`file_names` warning at all. My host has 22, which does. So the stream I measured and pinned in the golden was a statement about my llvm package, not about the compiler:

```
-dwarfdump_verify=warnings
-dwarfdump_warning=.debug_line[].prologue.file_names[1] is a duplicate of file_names[0]
+dwarfdump_verify=clean
```

`bc4ca28` fixes it the right way round. The producer still reads the stream rather than the exit status, but it filters exactly one class - the documented DWARF 5 §6.2.2 / §6.2.4 interop duplicate - with the binutils measurement written beside it, and reports everything else. The golden returns to `dwarfdump_verify=clean`.

That looks like the line it started on, and it is not the same claim. Before this PR, `clean` meant `--verify` exited zero, which it does with any number of warnings. After it, `clean` means the stream carries no error and no warning outside the one explained class, and a warning of any other kind fails the case and prints its own text into the diff. Verified by mutation: with the allowlist pattern changed so it matches nothing, the producer reports `warnings` and names the class.

The version dependence is also a third argument against changing the producer, on top of the two in the PR body: the wall of warnings #2755 opens with is not visible to the validator CI actually runs.

octalide:
Also made the warning-filter pipeline explicit about an empty result being normal (`|| true`), since `run.sh` runs under `set -e` and the filter legitimately removes every line. It only did not bite because `produce` is invoked as an `if` condition, which suspends errexit inside it, and depending on that is not something the next producer should have to know.

octalide:
## CI has not run against the fix, and it is not this branch

GitHub Actions stopped creating workflow runs repo-wide at `2026-08-08T00:16:46Z`. Neither the `bc4ca28` push, the close/reopen, the force-push of `88d0f19`, nor an empty retrigger commit produced a run, and no run has been created for **any** branch in the repo since that timestamp - `status=queued` and `status=in_progress` are both empty and the repo is not disabled. That looks like an org-level Actions quota rather than anything about this PR. Flagging it because several agents were pushing concurrently and it will be blocking their PRs too.

So the only red CI on this PR is run `31228888789`, which is against `d78e1fe2` - the superseded commit whose golden recorded the llvm-22 warning text. `88d0f19` is exactly the fix for that failure.

Verified locally against the final head instead:

| check | result |
|---|---|
| `mach test .` | 1288 passed, 0 failed |
| `int --target linux --deps pin` | all cases passed |
| `int --target linux-riscv64 --deps pin` | all cases passed |
| `int --target linux-arm64 --deps pin` | the 4 pre-existing qemu/host failures only, identical on pristine `origin/dev` |
| self-host fixpoint | stage2 == stage3 == `74575252924008db12bf950669bd4212151d0b8375ffcaf543366eefe61928f7` |
| `int/lib/check-deps.sh` | 98 declarations OK |
| `int/lib/check-target-matrix.sh` | 445 case x leg pairs OK |
| `int/lib/check-determinism.sh` | warm and post-edit incremental both match clean |

`surface/debuginfo` passes on all three ELF legs in both profiles. Please re-run CI when Actions is available before merging.

PR #2765, opened 2026-08-08
fix(target): refuse an unencodable float width instead of substituting one

### LINKED ISSUE #2733: isa(x86_64): an unsupported FP move width is silently narrowed instead of refused
Found while implementing the ABI half of #2687 (PR #2722), reported rather than fixed there because it is a separate defect with a wider blast radius than that PR's scope.

## What

`x64/encode.mach` handles an unexpected floating-point move width by **silently substituting one**:

```mach
if (w != 4 && w != 8) { w = 8; }
```

A caller asking for a 12-byte move gets an 8-byte move and no diagnostic. The bytes that were not moved are simply not moved, and the value is silently truncated.

This is what made #2687 hard to diagnose, and it is worth recording how. There were **three** layers each answering one fixed-size ladder for two different questions (register width and memory footprint):

| layer | behaviour with a 12-byte vector |
|---|---|
| `op_width_of` | fell through to the general-purpose word, losing lanes (fixed, #2697) |
| `fp_move_width` | refused outright, honest and total |
| `encode_fp_mov` | **silently narrows to 8** |

The middle layer's refusal was **masking** the bottom one. Widening `fp_move_width` alone compiled cleanly and then dropped the top lane of every stack-passed vector with no error at all. The refusal was the only thing standing between this default and a miscompile.

PR #2722 makes it unreachable for vectors by asking the two questions separately. **The default itself still stands**, at `x64/encode.mach:546`, with the same shape at two further sites (3371, 3432, which do refuse rather than substitute - worth confirming which behaviour each site actually has).

## Why it is wrong independent of #2687

A defensive fallback that produces *wrong output* is worse than no fallback. This one converts "the compiler was asked something it cannot do" into "the program silently computes the wrong answer", which is the single worst failure mode a backend has. The repo's standing rule names it directly: a defensive fallback is a signal that something upstream is shaped wrong, and this one was concealing exactly that for as long as it existed.

It is also a **trap for the next widening**. #2727 will make vectors wider than the register legal, and any path that reaches this site with a size it does not expect gets truncation rather than a compile error. The refusals added in #2722 are what make that ordering safe today, and they are the only thing doing so.

## Fix

Refuse rather than substitute. An unsupported move width is a compiler bug, and it should say so with the width it was given and the site that asked, rather than emitting a narrower move.

Audit the other two sites and any sibling in the aarch64 and riscv64 encoders for the same shape. The question to ask at each is whether an unexpected value produces a diagnostic or a quietly different program.

Relatedly, `aapcs64` hardcodes an element width of 16 in `make_slot_hfa(reg, 1, 16, size)` on the vector path. That is inert today because the pieces are never materialized for a non-aggregate, but it is a literal sitting on a path #2727 will read.

## Acceptance

- An unsupported floating-point move width is a defined compiler error naming the width, not a substituted narrower move.
- The refusal is proved to fire by driving the encoder directly, not by arguing no caller reaches it. Arguing unreachability is what let this sit.
- The other two sites in `x64/encode.mach` and the aarch64 and riscv64 equivalents are audited, and the finding is recorded either way.
- `mach test .` green, `int` green, self-host fixpoint holds.

octalide:
Closes #2733

## The defect

All three register machines answered a float-move width they do not implement by **substituting 8** and encoding anyway:

```
if (w != 4 && w != 8) { w = 8; }
```

the same line in the same role in x86-64's `encode_fp_mov`, aarch64's `encode_fmov` and riscv64's `encode_fmov`. A move the compiler could not encode became an 8-byte move that dropped everything above the low lane, with no error anywhere. A defensive fallback that produces **wrong output** is worse than no fallback, and this one had already cost real debugging: during #2687 the middle layer's honest refusal was the only thing masking it, and widening that refusal alone compiled cleanly and then silently dropped the top lane of every stack-passed vector.

## Census across all three encoders

| site | before | after |
|---|---|---|
| `x64 encode_fp_mov` | **silent narrow** | refuses |
| `x64 encode_float_const_mov` | **silent narrow**, not in the issue's census | refuses |
| `x64 encode_float_conv` | **silent**, no ladder: `dst_w == 4` selects, SD otherwise | refuses per arm |
| `x64 encode_float_arith` / `_neg` / `_cmp` | refused | refused, now driven by a test |
| `arm64 encode_fmov` | **silent narrow** | refuses |
| `arm64 encode_fp_conv` | **silent**, `dw == 8` / `sw == 8` selects, S otherwise | refuses per arm |
| `arm64 encode_fp_arith` / `_neg` / `_cmp` | refused | refused, now driven by a test |
| `riscv64 encode_fmov` | **silent narrow** | refuses |
| `riscv64 encode_fp_conv` | **silent**, same shape | refuses per arm |
| `riscv64 encode_fp_arith` / `_neg` / `_cmp` | refused | refused, now driven by a test |

The conversions are the finding triage could not have made by grepping for the ladder: they carry no ladder at all, just a boolean selector off an unvalidated float width, and which side is the float side depends on the opcode - so the check sits per arm, and the tests pin that the integer side's legitimate 1 and 2 still pass.

## The refusals are driven, not argued

Each refusal is called directly from a unit test with a width it must refuse, and **the assertion is that the byte sink stayed empty**. "Returned an error" and "emitted nothing" are different properties, and only the second separates a refusal from an emit-then-error. Every legitimate width is checked to still encode on every target, including aarch64's and x86-64's 16-byte vector arms, which return above the scalar check and must not be caught by it.

x86-64's `encode_fp_mov` now reports its backend bugs rather than ending the process itself, which is what makes it checkable at all. Behaviour is unchanged: the process-terminating `panic` moved to its single caller in `encode_mov`.

All four fixes were mutation-verified - each was reverted individually and the corresponding test observed to fail.

## aapcs64

Both `make_slot_hfa(reg, 1, 16, size)` literals are now written as the **V register** width they are, named `VEC_REG_BYTES`. The obvious "fix" of passing `size` is wrong and the PR says so in the comment: `f32x3` is twelve bytes and correctly rides a whole register. That was measured, not supposed - the first attempt failed `int/surface/vector-abi` on the twelve-byte `i32x3` argument.

What the literal did hide is the case one register cannot serve, which #2727 makes legal: a 32-byte vector would have been handed V0 and a piece describing half of it. An argument now spills by value and a return takes the indirect result, both memory-correct since the slot carries the true size.

## Verification

- `mach test .`: 1292 passed, 0 failed
- `int` linux: all cases passed
- `int` linux-riscv64: all cases passed
- `int` linux-arm64 (qemu locally): identical to `origin/dev` baseline - `c-variadic` and `narrow-stack-args` fail on both, and both are host-environment artifacts of running an aarch64 leg on an x86-64 box. **The native aarch64 leg is CI's to settle**, per targets.conf and the qemu-fidelity note in `int/run.sh`.
- Self-host fixpoint converged, stage2 byte-identical to stage3: `25f7f92cd9b1de43c02d7fdeeb45f66cccbdbd7e69bea907103f2371b2fc49fe`
- The corpus re-run under the fixpoint compiler: 1292 passed, 0 failed

**The compiler's own source trips none of the new refusals** - it built itself three times to a fixpoint.


octalide:
CI green on every leg, including the **native aarch64 leg** on `ubuntu-24.04-arm` (17 pass, darwin skipped as the release gate). That settles the two failures I saw locally under qemu on an x86-64 host: `c-variadic` and `narrow-stack-args` fail identically on `origin/dev` there and are host-environment artifacts, not regressions.

The audit turned up one finding outside this PR's boundary, filed as #2766: the float **memory access** width is substituted rather than refused on aarch64 and riscv64, the same shape one family over. `riscv64 fp_mem_funct3` answers 4 with `flw`/`fsw` and everything else with `fld`/`fsd`, and `arm64 emit_fp_mem` falls through its Q arm to `use_d = width == 8`, so a 32-byte access takes the four-byte S form. Both reached with an unvalidated `mi.src_width`. It is not folded in here because the defaulting helpers (`ldst_lo12_kind`, `width_shift`) are shared with the integer access path, so the correct fix is a contract change on that family rather than a refusal at the float callers - and putting the check only at today's callers would repeat the boundary mistake this PR is about.

Issue #1944, opened 2026-07-04
codegen: release-mode -g changes machine code (additivity break)


octalide:
### Invariant

`-g` must add **only** debug sections. The machine code — every PT_LOAD segment — must be byte-identical with and without `-g`, at **every** optimization profile. This is the additive-only guarantee the debug-info program is built on (see `dbg/verify.sh` step 5, `elf_seg_identical`).

### Bug

At `--profile release`, `-g` changes the emitted machine code. Building any nontrivial fixture (or `mach-std`) `--profile release -g` vs `--profile release` yields differing `.text` / LOAD-segment bytes.

### Repro (clean dev)

```
mach build <fixture> --profile release -g   -o /tmp/g
mach build <fixture> --profile release      -o /tmp/nog
# LOAD segments differ:
cmp <(objdump -d /tmp/g) <(objdump -d /tmp/nog)   # diverges
```

### Signature

A single large disassembly shift (~12k lines) beginning at one early flipped function — consistent with **one** `-g`-sensitive decision cascading through instruction/layout selection, not a diffuse perturbation.

### Why this was invisible

`dbg/verify.sh` step 5 (`elf_seg_identical`) builds both images at `--profile debug`, where the release-only optimization passes (inline + the late algebraic/CSE/constfold/dce sweep) never fire. The break only manifests once those passes run.

### Known cause #1 (already being fixed — NOT this issue)

The inliner's `callee_instr_count` (`src/lang/me/pass/inline.mach:223`) sums `blk.instr_count`, which includes `OP_DBG_VALUE` instructions. Those exist only under `-g`, so the instruction total — and therefore `should_inline`'s `< INLINE_INSTR_THRESHOLD` test — flips between the two builds. Fixed in **PR #1932**.

### This issue: the residual

After excluding `OP_DBG_VALUE` from the inliner's count, release `-g` vs non-`-g` **still** differs. That residual `-g`-sensitive decision in the release pipeline is the target of this issue. Root cause and exact decision site to follow.

octalide:
[...]

PR #2767, opened 2026-08-08
fix(codegen): stamp the frame-pointer-side reservation once and read it in DWARF

### LINKED ISSUE #2759: codegen: a riscv64 -g variable location is off by the frame's reserved top, because DWARF reads the raw slot offset and the encoder biases it
Every `-g` variable that lives in a frame slot is described at the wrong address on riscv64, by exactly the number of bytes the prologue reserves at the top of the frame.

Found while fixing #2735, which touches the same two files. It is independent of that fix and predates it.

## Repro

`src/lang/be/codegen/encode.mach` records a `VARLOC_FRAME` binding as the RAW slot offset:

```
if (fn.frame.slots[si].value == vreg) { kind = VARLOC_FRAME; off = varloc_slot_off(fn, si); }
```

which for a non-realigned frame is `fn.frame.slots[si].offset`. The producer emits that as `DW_OP_fbreg <off>` against a `DW_AT_frame_base` naming the frame-pointer register (`dwarf.frame_base_reg`).

But riscv64 does NOT address a slot at `s0 + slot.offset`. It pins `s0` at the frame TOP, with the ra / s0 record and the callee-saved GP and FP areas occupying the bytes immediately below it, and biases every slot access down past that region (`riscv64.encode.frame_slot_offset`):

```
fun frame_slot_offset(f: *mir.MirFunction, v: u32) O.Option[i64] {
    ...
    ret codegen_frame.slot_offset(?f.frame, v, frame_reserved_top(f));
}
```

`frame_reserved_top` is `16 + gp_bytes + fp_bytes`, measured over an identical corpus in #2291 as nonzero on all 3 865 riscv64 lookups and ranging 16 to 200 bytes.

Build any riscv64 binary with `-g`, break in a function with a callee-saved register live, and print a local: the debugger reads `s0 + slot.offset`, the machine wrote `s0 - reserved_top + slot.offset`, and the two differ by `reserved_top`.

x86-64 and aarch64 pass bias 0, so both are correct and the defect is riscv64's alone.

## Expected

A `DW_OP_fbreg` offset resolves to the same address the emitted load and store use, on every target. The debugger prints the variable.

## Actual

It prints whatever sits `reserved_top` bytes above the variable, which in a function with callee-saved registers is the saved `s0`, the saved `ra`, or another callee-save slot. Silent: the value has the variable's type and a plausible magnitude, and in a leaf with no callee-saves the bias is 16 and the neighbour is a real local.

This is the same defect class as #2715, #2725 and #2735 - one layout fact derived in two places, agreeing in the two targets where the correction happens to be zero and disagreeing in the one where it is not. `frame.slot_offset`'s own comment names the bias as "the ONE thing that differs between targets", and the DWARF path is the reader that does not apply it.

## Fix

The offset the producer emits and the base register it names are two halves of one answer, exactly as `isa.asm.SlotFn` says for an inline-asm `{name}`. Two candidate designs.

**A. Ask the target.** `emit_var_bindings` is per-target code reached through the encoder, so it can route the slot through the same `frame_slot_offset` the target's own accesses use rather than reading `fn.frame.slots[si].offset` directly. That deletes the second derivation instead of correcting it, which is the reason to prefer it. It needs the per-target hook to be reachable from the shared `encode.mach` binding recorder, which today it is not - the recorder is shared and the resolver is not.

**B. Carry the bias as a frame fact.** #2735 added `MirFrame.base_dist`, the distance from the stack pointer to the slot region's base, stamped once per function by the target's own encoder. `reserved_top` is the frame-pointer-side twin of that number and could be stamped beside it, so the shared recorder adds it without knowing which target it is on.

**A is preferred** where it is reachable, because it leaves one derivation rather than two that agree. B is the fallback if threading the resolver into the shared recorder turns out to require a wider contract change than the defect is worth, and it is strictly better than the status quo either way.

Whichever is chosen, `dwarf.frame_base_reg`'s comment should say what the base register actually guarantees about the offsets measured from it, since the current text describes the base without naming the bias.

## Acceptance

- A riscv64 binary built with `-g` resolves a frame-slot variable to the address the emitted code uses, asserted by reading the variable's value out of a debugger or by decoding the `DW_OP_fbreg` offset and comparing it against the load the encoder emitted for the same slot - not by inspecting the offset alone, which is self-consistent under the bug.
- The function under test has at least one callee-saved GP register and one callee-saved FP register live, so `reserved_top` is not the 16-byte leaf value. A leaf-only case passes under a fix that corrects the record but not the save areas.
- x86-64 and aarch64 `.debug_info` is byte-identical before and after, since their bias is 0 and nothing about them should move.
- `.text` byte-identical with and without `-g` on every target - the #1944 additivity invariant, which this must not spend.
- Three-generation self-host fixpoint, self-built against self-built.

## Environment

- triple: `linux-riscv64`, cross-built from `linux-x86_64`; x86-64 and aarch64 are unaffected
- toolchain: `dev` at 96c251f3, and on the #2735 branch
- OS: Linux 7.1.5

octalide:
Closes #2759.

## Before, on the unpatched compiler

Fixture: `probe(p: Pair, n: i64)` takes a register-passed aggregate, so `p` is reconstructed into a frame slot and gets a `DW_OP_fbreg` location, and it keeps enough values live across calls that six callee-saved GP and three callee-saved FP registers are saved. `frame_reserved_top` is therefore `16 + 48 + 24 = 88`, not the 16-byte leaf value.

Built from `origin/dev` @ `6991413e` for `linux-riscv64`, `-g`:

```
DW_AT_frame_base  (DW_OP_reg8 X8)
DW_AT_name        ("p")
DW_AT_location    (DW_OP_fbreg -16)
```

And the code the same compiler emitted for the same slot:

```
4076a0: addi  sp, sp, -0x80
4076a4: sd    ra, 0x78(sp)
4076a8: sd    s0, 0x70(sp)
4076ac: addi  s0, sp, 0x80
4076b0: sd    s1, 0x68(sp)        # ... s2..s6, fs0..fs2 follow
4076d4: sd    a0, -0x68(s0)       # p, at s0 - 104
4076d8: sd    a1, -0x60(s0)
```

`p` lives at `s0 - 104`. DWARF said `s0 - 16`, which is `sp + 0x70` - **the saved `s0`**. Off by exactly 88, the reservation.

## After

```
DW_AT_location    (DW_OP_fbreg -104)
```

Which is `sd a0, -0x68(s0)` exactly.

## The fix

Design **B** from the issue, taken further than the issue frames it, because done this way it beats **A** on the issue's own criterion.

`MirFrame.fp_dist` - the bytes between the frame pointer and the slot region's base - is stamped once by the target's own encoder, right beside `base_dist`, its stack-pointer-side twin, before any operand is resolved. x86-64 and aarch64 stamp 0; riscv64 stamps `frame_reserved_top(f)`. Then:

- `frame.slot_offset` **loses its `bias` parameter** and reads `frame.fp_dist`. The comment calling the bias "the ONE thing that differs between targets" was right, and a caller-supplied argument is exactly what let one caller not supply it.
- `encode.varloc_slot_off` subtracts `fp_dist` for the frame-pointer case, matching what it already did with `base_dist` for the realigned case.
- riscv64's own `frame_slot_offset` now goes through the same shared lookup, so it no longer calls `frame_reserved_top` a second time either.

That leaves **one** derivation with two readers, rather than A's "route the DWARF path through the target's resolver", which would still have left the resolver deriving and the recorder calling. `dwarf.frame_base_reg`'s comment now says what the base register does and does not guarantee about the offsets measured from it, as the issue asks.

## The regression case

`int/regression/2759-riscv64-fbreg-bias`, run kind `varloc-fbreg`. Per the acceptance, it does **not** inspect the offset alone - a producer with its own derivation is self-consistent under the bug. It crosses the two halves: for each subprogram, every `DW_OP_fbreg` offset must appear as a displacement in some instruction of that same function naming the register `DW_AT_frame_base` names.

Two ISA-independent facts, so the golden is shared:

```
varloc_fbreg_checked=nonzero
varloc_fbreg_unbacked=0
```

`checked` is stated because `unbacked=0` would otherwise be vacuous if extraction broke.

**Against the unpatched compiler**, which is the point:

| leg | unpatched | patched |
|---|---|---|
| `linux` | pass | pass |
| `linux-arm64` | pass | pass |
| `linux-riscv64` | **`unbacked=13` (debug), `unbacked=6` (release)** | pass |

Green on the two targets whose reservation is zero and red on riscv64 alone - the harness now states the thing that made this invisible.

## Invariants

- `.text` byte-identical **with and without `-g`** on all three targets, the #1944 additivity invariant, unspent.
- `.text` byte-identical **before and after** on all three targets. This moves no codegen.
- `.debug_info` and `.debug_loclists` byte-identical before and after on `linux-x86_64` and `linux-arm64`.
- On `linux-riscv64` they move, and the textual `--debug-info` diff is **only** `DW_OP_fbreg` offsets plus the DIE-offset shift their ULEB widths cause. Nothing else changed.

## Verification

- `mach test .`: 1288 passed, 0 failed. `frame.slot_offset:alloca_and_spill_lookup` gains a nonzero-`fp_dist` arm, since the zero case is the one that could not see this.
- `int/run.sh --deps pin`, full suite: `linux` all passed, `linux-riscv64` all passed. `linux-arm64` has the 4 pre-existing failures noted below.
- Self-host fixpoint holds, stage2 == stage3 == `d78409d7803fe3042f44b27d567fb49f0d30c9e890183ae69c093559fd0b7a67`.

### On the aarch64 leg

`targets.conf` marks `linux-arm64` native and I have no aarch64 host, so I ran it under `--runmode qemu`. `surface/c-variadic` dies on SIGILL under qemu-user and `surface/narrow-stack-args` fails its build because its `cc` probe compiles for the x86-64 host. Both fail identically with a compiler built from pristine `origin/dev` @ `6991413e`, so neither is this branch. The new case passes on that leg in both profiles. The native CI leg is the authority.


octalide:
**Rebased onto current `dev` and re-verified locally, because GitHub Actions is not creating workflow runs.**

No run has been created for any branch in this repo since 2026-08-08T00:16:46Z. Pushes register (the PR head matches the pushed commit) and workflows all report `active`, but no run appears. GitHub's own status page reports Actions operational, so this is specific to this org rather than a platform incident, and the likely cause is exhausted Actions minutes: today ran 30+ CI runs at 17 jobs each, including macOS runners that bill at a 10x multiplier.

**This PR's earlier green run does not cover its current head.** The `dev` merge brought in the other work that landed today, so the CI-verified commit predates a materially different tree. Reporting the old run as coverage would be the exact failure this repo keeps finding.

Verified locally on the merged head instead:

- `mach test .` **1290 passed, 0 failed**
- `int` suite **green on linux**
- `int` suite **green on linux-riscv64** (qemu), which is the target this PR actually fixes
- **self-host fixpoint holds**, stage2 byte-identical to stage3, self-built against self-built

**Not covered locally and still needed before merge:** the native aarch64 leg, Windows, and the darwin gate. The aarch64 one matters most, since it is the only leg that settles machine-level facts in this repo.

Holding the merge until Actions resumes.

Issue #2741, opened 2026-08-07
int: a case with a C [step] assumes the host cc targets the leg, and fails as a relocation overflow


octalide:
Found while running the `int` suite for #2728 and #2729.

## What

`int/surface/narrow-stack-args` cannot be cross-built to `linux-arm64` from an x86-64 host. Its `[step]` invokes the bare host `cc` on `probe.c`, so an **x86-64 object is produced and then linked into an aarch64 image**. It surfaces as:

```
error: relocation 'plt32' to 'float_tail' overflows
```

CI is unaffected, because the arm64 leg builds natively on `ubuntu-24.04-arm` where the host `cc` happens to target the leg. So this only bites a developer cross-building locally, which is the ordinary way to reproduce a leg failure without waiting for CI.

## Why it is worth fixing rather than documenting

The failure mode is the problem, not the inconvenience. **The diagnostic names nothing about the cause.** A relocation overflow reads as a codegen or linker defect in mach, and someone chasing it would go looking in the encoder. The actual cause is that a case with a C step silently assumes `cc` targets the leg being built, and nothing checks that assumption or reports it when it does not hold.

That is the same shape as the defects this repo keeps finding: a step that is silently wrong rather than absent, producing a plausible failure that points away from itself. A developer who hits this learns the wrong thing about the compiler.

It also means a case with a C step is **not actually cross-target coverage** even where the harness presents it as such. It happens to work on the native leg and cannot work anywhere else, and nothing in the case says so.

## Fix

The harness knows the target triple for the leg it is building. A `[step]` that shells out to a C compiler should either be given a cross-capable compiler for that leg, or the case should declare that it requires a host-matching toolchain and be **skipped with a stated reason** on legs where that does not hold.

A skip with a reason is the honest minimum: it keeps the run accounting truthful, which a silent link failure does not. Prefer the cross-capable route if the toolchain is reasonably available, since a skipped case is still absent coverage.

Whatever is chosen, the condition should be detected and named rather than discovered through a relocation error.

## Also, trivially

`m` is not in the root `.gitignore`, although `mach build . -o m` is the documented CI bootstrap command, so it shows as an untracked file in every working tree that has followed the instructions.

## Acceptance

- A case with a C `[step]` either cross-builds correctly for the leg, or is skipped with a reason naming the toolchain requirement.
- The skip or failure is visible in the run accounting rather than surfacing as a link error.
- No case silently presents host-only coverage as cross-target coverage.
- `m` is gitignored.

Issue #2735, opened 2026-08-07
codegen: an over-aligned stack local is not over-aligned at run time; the frame pointer only guarantees 16


octalide:
An `#[align(N)]` record with `N > 16` gets its declared alignment as a type, but a **local** of that type does not get it at run time. The offset is a multiple of `N` from the frame pointer, and the frame pointer is only guaranteed 16-byte aligned, so the address is a multiple of `N` only when the two happen to agree.

Found while fixing #2725. It is independent of that fix and predates it.

## Repro

```mach
use std.types.size.usize;
use p: std.print;

#[align(64)]
rec Big64 { a: i64; }

#[align(32)]
rec Big32 { a: i64; }

fun probe() i64 {
    var pad: i64 = 1;
    var x: Big32;
    var y: Big64;
    x.a = 1; y.a = 2;
    p.printlnf("x_mod32={} y_mod64={} sz32={} sz64={}",
        (?x)::usize & 31, (?y)::usize & 63,
        $size_of(Big32), $size_of(Big64));
    ret x.a + y.a;
}

#[symbol("main")]
fun main(argc: usize, argv: **u8) i64 { p.printlnf("r={}", probe()); ret 0; }
```

`mach build . --target linux --profile release`, then run it three times:

```
x_mod32=16 y_mod64=48 sz32=32 sz64=64
x_mod32=16 y_mod64=16 sz32=32 sz64=64
x_mod32=0  y_mod64=32 sz32=32 sz64=64
```

## Expected

`x_mod32=0` and `y_mod64=0`, every run, on every target. `#[align(N)]` on a type is a promise about every object of that type, and `$size_of` already reports 32 and 64, so the storage is sized for the promise even though the placement does not keep it.

## Actual

Both are usually nonzero, and the value is not even stable across runs of the same binary: it tracks where the process stack happened to start. `x_mod32=0` in the third run above is the coincidence, not the rule.

## Cause

[...]

Issue #2689, opened 2026-08-07
asm: a {name} slot is addressed FP-relative, so aarch64's reach is 32 slots and an inlined caller exceeds it


octalide:
Found while working #2231 (PR #2678), which is blocked on it. Sibling to #2667 and #2674, and the last of that family.

## What

An inline-asm `{name}` slot is addressed **frame-pointer-relative**. Slots sit below the frame pointer, so the offset is negative, and aarch64's scaled `imm12` form is **unsigned only**. The whole reach for a `{name}` is therefore the unscaled `imm9`'s **−256** — thirty-two eight-byte slots.

That is not a theoretical limit. `darwin/shared.mach`'s `syscall2` has five `{name}` binds; a caller inlining a few such wrappers lands at **−288**, and `int/surface/macho-foreign-aarch64` and `macho-pie-aarch64` fail to build.

#2674 raised the ceiling substantially by laying asm-bound slots nearest the frame pointer, which is why this is now the *remaining* blocker rather than one of several. It did not raise it past this case.

## Why it cannot be legalized

This is the same constraint #2667 established. A compiler-emitted access whose offset is out of range is legalized by materializing the address through a register the backend reserves. A `{name}` is reached by an **author-written** access, and there is no register the compiler may borrow inside someone else's asm body — borrowing IP0 is exactly the silent clobber #2667 fixed.

So the slot has to be reachable by construction. #2674 did what ordering can do; this is what is left.

## Fix

Address a `{name}` from **SP rather than FP**. Offsets become positive, the scaled `imm12` form applies, and the reach goes from 32 slots to 4096.

Two things that make this more than a change of base register:

- **`SlotFn`'s contract says "frame-pointer-relative"**, and the per-target divergence it abstracts is precisely this. Changing it is a contract change, not a local edit.
- **It needs an argument that SP is stable across an asm block.** A body that pushes, or that calls, moves SP under the offset. Whether that is refused, detected, or already impossible in this grammar should be settled explicitly rather than assumed — a wrong answer here is a silent wrong address, which is the failure mode this family keeps producing.

## Acceptance

- A `{name}` is reachable at any frame depth the target's scaled form admits, on every ISA.
- `int/surface/macho-foreign-aarch64` and `macho-pie-aarch64` build with #2678's inlining applied.
- Whether SP can move within an asm body is settled and documented, and if it can, the case is refused rather than mis-addressed.
- `SlotFn`'s doc comment describes what it actually promises.

[...]

PR #2782, opened 2026-08-08
feat(inline): flow a cross-module `#[inline]` body into its importer

### LINKED ISSUE #2231: perf: inline atomics through parsed asm — remove the call and stack-slot staging

octalide:
Closes #2231

## What was actually left

The issue names two costs, the call and the stack-slot staging. #2258 removed the staging. Measured on `dev` @ `2b1a0464`, `atomic.fetch_add` was still a `call` and nothing else, because **every caller of `std.sync.atomic` is in another module** and a dependency's regular function body never reached an importer at all.

Both blockers are closed (#2689, #1859), so this is the one change left plus the orderings, which are untouched and stay sequential-consistency-everywhere.

## The axis, not a special case

A dependency's body flows into an importer's IR only when the function is a **template** - generic, comptime-param, or pack-tailed - because those have no single concrete body the defining module could emit. `#[inline]` is now the fourth member of that same axis, and the first one an author asks for rather than the type system forcing: it declares that the body is part of the function's public surface, so an importer may materialise it.

It reuses monomorphization's machinery with **zero type arguments**: the instance worklist, the drain that binds the context to the origin module, and `append_function`'s existing upgrade of the call site's signature-only extern into the definition, which is what makes the already-recorded `value.fn(index)` name a body the inliner can see.

## The linkage is what keeps the function one function

The copy is emitted `PUB | WEAK` under the **origin's** mangled name, not as a private clone. The defining module's strong definition wins at link time and every reference in every object - including inside the importer holding the weak copy - binds to that single address. A private clone would give `?atomic.load` a different answer per module, which is a silent wrong answer rather than a missed optimisation. Same linkage a generic instance uses, for the same reason.

## The incremental firewall gets a correctly shaped hole

`Q_LOWERED_SURFACE` (#1859) excludes regular bodies so an impl-only edit in a dependency does not re-lower importers. An `#[inline]` body is now **included**, because an importer holding IR built from the old body is holding stale code - a miscompile that appears only on an incremental build and never on a clean one.

`driver:incremental_lower_inline_body_pins_no_cross_module_inline` existed to pin the old behaviour and said in its own comment that it must flip when this landed. It now asserts the invalidation, as `incremental_lower_inline_body_invalidates_importers`. **Proved it has teeth**: reverting only the fingerprint arm while leaving the flow in place fails exactly that test and nothing else.

## On PR #2678

**Adopted, rebased, re-verified.** `ir.clone_asm_payload`, the `should_inline` / peel gate removal, and `instruction:asm_is_at_least_as_conservative_as_call` are that branch's single commit `ffa4bccc` cherry-picked onto current `dev`. It applied cleanly and, contrary to the caution that it might not survive #2689 and #2735, it is still correct: a binding names an alloca, the remap follows that alloca into the caller, and the SP-relative addressing #2689 introduced is computed from the caller's own frame layout afterwards. Verified by building all six targets, riscv64 included - the target that branch said it broke, which #2674 has since fixed at the cause.

Credit where it is due: that design is most of the mechanism and rediscovering it would have cost a day. #2678 should close with this.

What #2678 could not do, and this adds, is the cross-module reach it correctly identified as the missing half.

## Measurements

x86-64, release, a loop of 3M iterations over `store` / `load` / `fetch_add` / `fetch_sub` with six values live across each barrier. Min-of-9 in-process timings, 8 interleaved before/after runs:

| | min (ms) |
|---|---|
| before | 25.00 |
| after | **21.13** |

**-15.5%**, spread under 0.5% either side, and the checksum `4499998500000` is identical from both binaries and hand-derived (`5*N(N-1)/2` style totals, verified by hand). Compiler binary 8,531,968 -> 8,536,064 (+4 KB).

The staging that remains inside the inlined body is inherent to a `{name}` binding, which denotes a frame slot. Removing that is a register-constrained asm operand, a different feature.

## Coverage, and why it is two cases

`int/regression/2231-cross-module-inline-asm` runs the program and checks every value at both profiles on all three ISAs. **It passes identically against a compiler built from `origin/dev`**, which is the point rather than a flaw: inlining is value-preserving by definition, so a value observable cannot see this feature at all. That is this repo's recurring defect class, so it is stated in the case's own comments.

`int/regression/2231-cross-module-inline-shape` reads the **emitted call count per function** through a new `call-shape` producer, a new `calls` mode on the existing `emit_scan` shape-observable framework. Against `origin/dev` it fails on all three ISAs with exactly:

```
-via_inline_asm calls=0
-via_inline_plain calls=0
+via_inline_asm calls=1
+via_inline_plain calls=1
 via_noinline calls=1
 via_undecorated calls=1
```

The two unchanged lines are the contract's other direction: `#[noinline]` still opts out, and an undecorated body still does not cross the boundary. That is why the observable is a **count** and not a boolean - "inlined everything in reach" must fail here, not pass.

The producer filters to the case's own project prefix, so mach-std's template instances (which int resolves at its latest **release**) cannot move the golden.

## Verification

- `mach test .` **1297 passed, 0 failed**.
- `int` green on **linux** and **linux-riscv64**. On **linux-arm64** the only failures are `surface/c-variadic` and `surface/narrow-stack-args`, both **pre-existing** and reproduced on a compiler built from `origin/dev` @ `2b1a0464` - they are #2741, fixed by #2773. Verified my two new cases PASS on linux-arm64. The windows leg is not testable on this host (qemu-user cannot load PE).
- All six targets cross-build clean, including riscv64, darwin and PE, which is where the strong-plus-weak duplicate definition is a new combination.
- **Self-host fixpoint holds, stage2 byte-identical to stage3**: `2c9bf6b345275fa9c7716050ba576ada88dd2eee10f2ecf689d67336deef3302` against released mach-std (the CI configuration), and `fcd0778f0160ce4fe1a6bcd98a5bf0a1a5115cc38c3da9dea0b6a52f991a4232` against a mach-std carrying the `#[inline]` atomics.

## Not in this PR

The atomics themselves need `#[inline]` in mach-std, which is a separate repo and a separate PR. That change is purely additive and spellable by the 4.16.0 release mach-std builds against, so it is safe to land before this ships. **This compiler change is what makes it mean anything**; mach-std without it is inert, and mach-std with it and a released compiler is unchanged.

Memory orderings are untouched. The 16:26 disposition on #2231 stands: the litmus bar needs native aarch64 and riscv64 runners, not compiler work.


octalide:
Added `addr_ab=1 addr_am=1` to the exec case, verified on all three ELF ISAs at both profiles.

The PR body claims the `PUB | WEAK` linkage is what keeps an inlined cross-module function **one** function, and nothing pinned it. Three modules now take the address of the same `#[inline]` callee and all three must agree.

This is worth having because the alternative implementation is the tempting one: emitting a private clone per importer is simpler, avoids the strong-plus-weak duplicate definition entirely, and would pass **every other assertion in both cases** while making `?f == ?f` false across a module boundary. That is a silent wrong answer, not a missed optimisation.

octalide:
Rebased onto current `dev`. Two notes for the merge.

**The `call-shape` producer's ownership filter follows the dotted-mangling change (#2790).** A project now owns a symbol whose leading path component is its id - `case.main.mixed` - rather than a `_M4case` prefix. Both goldens survived unchanged, because the producer reports the demangled bare identifier and that did not move.

**One expected failure until #2796 lands: `surface/debugger-gdb` at release.** It is not this PR's defect. This PR inlines `panic_os`, which shifts segment sizes enough to tip gdb over the overlapping-`PT_LOAD` defect in mach's own ELF writer (#2795). Verified directly: merging #2796 into this branch makes the case pass at both profiles with nothing else changed.

So the merge order is **#2796, then this**. On the other two ELF legs and everything else, `int` is already green here.

Local verification on the rebased tree: `mach test .` 1312 passed / 0 failed; `int` green on linux-arm64 and linux-riscv64; self-host fixpoint byte-identical.

Issue #2777, opened 2026-08-08
abi(riscv): only lp64 exists, so mach-compiled riscv64 code is ABI-incompatible with the platform default lp64d


octalide:
## What

`abi.mach` declares exactly one RISC-V ABI, `ABI_LP64`, and `abi/lp64.mach` is its only implementation. The RISC-V psABI defines a family, and the member we implement is the one almost nobody runs.

| ABI | float args | notes |
|---|---|---|
| `lp64` | integer registers | soft float, what mach implements today |
| `lp64f` | `fa0`-`fa7`, single precision | requires F |
| **`lp64d`** | `fa0`-`fa7`, single and double | requires D, **the default on riscv64 Linux** |
| `lp64q` | as above plus quad | requires Q, rare |
| `lp64e` | integer registers, 16-register RV64E | embedded |

**The consequence is interop, not preference.** `riscv64-linux-gnu` distributions build on `lp64d`. A mach-compiled riscv64 object passes a `double` in `a0` while every system library expects it in `fa0`, so linking mach code against any C library on a real riscv64 Linux system produces silently wrong arguments rather than a link error. We have not noticed because the riscv64 leg runs under qemu-user against our own code and links nothing foreign.

It also explains why #2771 exists at all. Variadic float and double arguments reading as zero is an `lp64` float-in-integer-register question, and the whole class is under-exercised because we only implement the variant our own tests use.

Note `#2771` should be fixed on its own terms first: `lp64` is a legitimate ABI and must be correct even once its siblings exist.

## Fix

`AbiVTable` is already the registration seam - `abi.register(reg, ?lp64_vtable)` with an `ABI_*` id, a name and an `isa.ARCH_*`. The axis is real and already has three members across two other architectures, so adding RISC-V variants should be a drop-in rather than a layer change. Design it so that it is.

What each variant actually changes is narrow and worth stating rather than discovering: which register bank a float argument lands in, the variadic rule (**variadic floats pass as if integer even under `lp64d`**, which is a rule people get wrong and #2771 sits next to), the return-value bank, and for `lp64e` the register count.

**The ABI must be checkable against the ISA extensions.** `lp64d` on a target without D is not a configuration, it is an error, and it should be refused when the target is resolved rather than producing code that cannot execute. That implies extension selection is part of this, not separate: today `riscv64` is a single opaque arch with `has_v128 = false` and no way to say which of IMAFDCV is present.

[...]

Issue #2771, opened 2026-08-08
abi: riscv64 variadic float/double arguments read as 0


octalide:
Related to #2741.

## Repro

On `int/surface/c-variadic` (briar-systems/mach), the case is currently `exempt: linux-riscv64`, historically because its `[step.probe]` shelled out to the runner's bare `cc`, which produces an x86-64 object no riscv64 link can consume. Fixing that (see #2741 - route the step through `int/lib/cc.sh`, which cross-compiles with `clang -target riscv64-linux-gnu -mno-relax -fno-asynchronous-unwind-tables -fno-unwind-tables -fno-jump-tables -c`) lets the case build and link cleanly for `linux-riscv64` for the first time. It then runs under `qemu-riscv64` without crashing, but prints wrong values:

```
bash int/run.sh --target linux-riscv64 ./m
```
(`linux-riscv64`'s targets.conf run-mode is already `qemu`, no override needed.)

## Expected

`expect.txt`:
```
none=0
mixed n=3 7 8 2500
ptr n=1 4242
floats n=8 1000 4000 8000
pair n=1 309
wide n=1 1234
```

## Actual

```
none=0
mixed n=3 7 8 0
ptr n=1 4242
floats n=8 0 0 0
pair n=1 309
wide n=1 1234
```

Every fixed-register argument (the `n=` counts, `ptr`, `pair`, `wide`) is correct. Every VARIADIC float/double argument reads as exactly 0 - not garbage, not a reordered value, zero - while the surrounding fixed and pointer variadic arguments in the same calls are right. That pattern points at mach reading the variadic float from the wrong location rather than a general variadic-passing bug: RISC-V's lp64 calling convention passes a NAMED floating argument in FA0-FA7, but a VARIADIC one takes the integer/stack path per the ABI's variadic rule, the same shape of divergence this case already exists to catch on Apple arm64 (#2575) and narrow-stack-args catches on Apple arm64's stack tail (#2598). If mach's riscv64 lowering still expects a variadic float in FA0-FA7 regardless of the callee's own variadic-ness, it would read an uninitialized/zeroed FP register - consistent with what's observed.

linux, linux-arm64, darwin-x86_64 (SysV/AAPCS64 variadic rules) all pass this same case already, so this looks specific to the RISC-V lp64 variadic-float rule, not a general variadic-lowering defect.

## Environment

- affected triple: riscv64-linux (lp64 ABI)
[...]

PR #2784, opened 2026-08-08
fix(abi): place a riscv64 variadic float in the integer bank

### LINKED ISSUE #2818: codegen: read the machine word from MachineModel in the shared backend, and gate float ops out of width splitting
Part of #2778.

The shared backend carries widths that are hardcoded to 8 (or derived from a 64-bit assumption) where they should be read from `MachineModel`. Every one of them is invisible today because a hardcoded 8 that happens to be right cannot be caught, and every one becomes a wrong answer on a 32-bit register file.

**This is worth doing whether or not RV32 ever lands.** One defect of exactly this class has already shipped and been fixed (#2777: `capture_slot_reg` took a move width from the machine word rather than the value, emitting `fmv.d.x` where `fmv.w.x` was required, so RISC-V's NaN-boxing rule turned every soft-float `f32` into `NaN` — a silent miscompile, not a diagnostic). The list below was found by the same audit and is the rest of it.

Doing this first also shrinks #2778(b) to genuine target work rather than a target change tangled with a width audit.

### The width-from-the-word sites

Each of these needs its width from the model rather than a literal:

- `context.stack_slot_bytes` — outgoing stack argument granularity, hardwired 8. Should be the machine word.
- `abi.stack_arg_align` — passes a by-reference pointer as 8 bytes and clamps alignment up to 8.
- two `round_up_i64(stack_off, 8)` sites in the signature walk — the outgoing stack cursor rounds to 8.
- `context.chunk` — the aggregate-copy walk tests `remaining >= 8` and emits an 8-byte move, which on a 4-byte-word machine emits a move the target cannot make.
- the spill-slot alignment literal at `context.mach:1271`.
- riscv64's prologue/epilogue frame geometry — 9 literal-8 sites, none read from the model.

The right shape is the one #2777 established: read the width from `MachineModel` at the single place that owns the question, rather than passing a width down or adding a per-target branch. Where two halves of one policy exist (a placement side and a capture side), make them read the same authority — that asymmetry is what produced the `NaN` bug.

### The float-bank gate in `legalize` (the sharp one)

`legalize.instr_needs_legalize` excludes only vector ops from width splitting. It does not exclude **float** ops. So at `lane_width` 4, an `f64` register copy is a plain `MIR_MOV` that would be split into two 4-byte **integer** lane moves.

That is a **silent miscompile, not a diagnostic** — the same shape as the `NaN` bug, and it fires the moment any target sets a 4-byte word. It should be fixed on its own merits and pinned by a test, independently of whether RV32 exists.

Good news for sizing, confirmed while auditing: `legalize` is already width-generic rather than mos6502-hardcoded, and its own tests already exercise `lane_width` 4 against 8. So the splitting machinery is reusable and this is a gate, not a rewrite.

## Acceptance

- every site above reads its width from `MachineModel`, with no literal 8 standing in for the machine word in shared backend code
- the `legalize` float-bank gate lands with a test that fails if a float op is admitted to integer lane splitting at `lane_width` 4
- where a placement side and a capture side both exist, they read one authority, and a test pins that they agree
- x86-64, aarch64 and riscv64 output byte-identical (these are all no-op substitutions on a 64-bit word), or any change justified explicitly
- `mach test .` green, `int` green on linux, linux-arm64 and linux-riscv64, self-host fixpoint holds

## Notes for whoever takes this

- Not blocked by anything. It is a prerequisite for #2778(b) (`ARCH_RISCV32` + the `ilp32` family + encoder forks + frame geometry) but lands and ships on its own.
- #2778(c) — 64-bit div/rem on a 32-bit register file — is the genuinely new work: `legalize` refuses wide DIV/REM outright and there is no libcall path anywhere in `be/`. Keep it out of this issue and out of (b).
- The `ilp32` ABI family is already fully described by the `RiscvAbi` contract landed in #2777 and held by a test that builds an `ilp32d` descriptor and asserts the classifier reads XLEN rather than a constant. (b) does not need to design it, only to register it against a real `ARCH_RISCV32`.

octalide:
Closes #2818.

Twelve width decisions in shared backend code answered "how wide is the machine word" with a literal `8`, plus the `legalize` float-bank gate. Each site now reads `MachineModel`, and the judgement of **which quantity each literal was actually answering** is recorded at the site, because they are not all the same question.

## The motivation, in the codebase's own words

`src/lang/target/abi/riscv.mach:24` already documents the fact: *"slot granularity. 64 across the lp64 family, 32 across ilp32"*, and `:211` says *"XLEN in bytes (8 for lp64\*, 4 for ilp32\*)"*.

The ABI layer has written it down. The shared lowerer never asked it. That is the whole defect: a second copy of one decision, with the correct copy already present, agreeing by accident because every target mach ships today makes both eight. Credit to `rv32` for finding it while sweeping for RV32 blockers.

**To be plain about risk: none of this miscompiles a shipped target.** Every literal is correct today. What makes it the highest-value item in that sweep is the failure mode when it stops being correct: a program that compiles, links, runs, and passes wrong arguments across a foreign boundary, certified green by a test that shares the assumption.

## The 12 sites, and what each was really asking

Taxonomy: **(a)** the machine word / slot granularity, **(b)** the pointer width, **(c)** the value's own width.

### `be/codegen/mir/context.mach`

| # | Site | Was asking | Now reads |
|---|---|---|---|
| 1 | `stack_slot_bytes` size round-up and its `size == 0` default | (a) | `gpr_width` |
| 2 | `copy_chunk_width` ladder cap | (a) | `gpr_width` |
| 3 | `add_spill_slot` size padding | (a) | `gpr_width` |
| 4 | `add_spill_slot` slot alignment | (a) | `gpr_width` |
| 5 | `add_spill_slot_aligned` size padding | (a) | `gpr_width` |

Site 2's reason is sharper than "granularity" and worth stating: the chunk travels through a **scratch GP register**, load into it and store out of it, so the widest chunk the walk can move is the widest value that register holds. A 4-byte-word machine was being asked for a move it has no register for.

Site 5 is the clean illustration that these are separable: its **alignment** stays the caller's, question (c), while only its **padding** is the machine's. That asymmetry is correct and now documented.

### `be/codegen/mir/abi.mach`

| # | Site | Was asking | Now reads |
|---|---|---|---|
| 6 | `stack_arg_align`, `CLASS_STACK_BYREF` | **(b)** | `pointer_width` |
| 7 | `stack_arg_align`, `CLASS_VEC_STACK_BYREF` | **(b)** | `pointer_width` |
| 8 | `stack_arg_align`, `if (align < 8) ret 8` | (a) | `gpr_width` |
| 9 | fixed-parameter split-chunk cursor `round_up_i64(stack_off, 8)` | (a) | `gpr_width` |
| 10 | variadic-tail split-chunk cursor `round_up_i64(stack_off, 8)` | (a) | `gpr_width` |

**Sites 6-8 are the taxonomy in miniature: one function, one literal, two different questions.** The by-reference clauses align a *pointer* — the doc comment said so all along, "holds only an eight-byte pointer" — while the floor three lines below is the *outgoing-slot granularity*. Substituting one authority into both would have been the same defect wearing a fix's clothes.

Sites 9 and 10 now round to the same word `stack_slot_bytes` sizes each chunk to, so the cursor a caller aligns and the footprint it then advances by cannot disagree. That asymmetry is what produced #2777.

### `be/codegen/regalloc.mach`

| # | Site | Was asking | Now reads |
|---|---|---|---|
| 11 | `class_width`, GP arm | (a), as the spill-move width | `ctx.spill_width` |
| 12 | `spill_slot_align`, non-vector arm | (a) | `ctx.spill_width` |

These were **already inconsistent on a shipped target before this PR**. `regalloc.mach:537` sets `ctx.spill_width = tgt.model.spill_width` and the spill store/reload MOVs use it correctly, so on mos6502 (`spill_width` 1) every GP slot reserved eight bytes for a move that writes one. A parameterised fact and an unparameterised one, four lines apart, disagreeing.

### The `legalize` float gate

`instr_needs_legalize` excluded only **vector** ops from width splitting, never **float** ones. At `lane_width` 4 an `f64` register copy is a plain `MIR_MOV` of width 8, split into two 4-byte **integer** lane moves between GP-class lane vregs — so the value leaves the float bank entirely and the reassembled halves are not the float. Silent miscompile, and the exact mirror of #2777.

The gate is now the register **bank** (`instr_rides_fp_bank`), which strictly subsumes the vector flag, since a vector vreg is FP-class too. `mir.vreg_is_fp` is the same authority instruction selection already reads to identify a float register copy.

The **lane planner** now reads the same gate, so a plan cannot describe a split the rewriter never performs — the placement/capture symmetry the issue asked for.

A target with **no** float bank is unaffected: mos6502 declares one register class, a soft-float value rides GP vregs, and it splits exactly as before. That is why mos6502 can have floats at all, and it is asserted as a control.

## Sites that look like the machine word and are NOT

More valuable than the list above, because this is the part a future reader gets wrong. **None of these were changed.**

| Site | What the `8` (or `16`) actually is | Why substituting the word would be wrong |
|---|---|---|
| `riscv64/encode.mach` `save_callee_fp`, width 8 per callee-saved f register | **float register width** (FLEN/8) | Stays 8 under RV32D, becomes 4 under RV32F. Tracks FLEN, never XLEN. |
| `riscv64/encode.mach` `frame_total` `align16` | **`model.stack_align`** | 16 under both lp64 and ilp32d. Never the word. |
| `regalloc.mach` `class_width`, non-GP arm, 16 | **vector register width** (`vector_bits / 8`) | A third quantity again. Left literal, documented, and flagged as needing its own change with its own byte-movement check. |
| `abi/riscv.mach:550-560` `callee_saved_v` size 8 | **float register width** | rv32's finding, same class as `save_callee_fp`. Not touched — that file is owned by other agents. |

## What is deliberately NOT in this PR

**The riscv64 prologue/epilogue frame geometry (9 sites).** Two reasons, and the first is the better one:

1. Several of those `8`s are not the machine word at all (see the table above). Substituting `gpr_width` into `save_callee_fp` or `align16` would create a **new** instance of the defect this PR closes, while looking like the fix.
2. For the ones that genuinely are the word, a model read there does not produce a working RV32 encoder anyway: `emit_mem_access` at width 8 selects `sd`, which RV32 has no form for. The encoder fork has to land with it.

Handed to `rv32` on `feat/2778-riscv32` with the per-site judgement, which is worth more to it than the substitution.

**`target/abi.mach:713-716` and `:743`** (`piece_for_reg` / `make_slot` chunking a two-register aggregate at the word). Correctly identified as (a), and the prose at `:704-707` states the assumption out loud. Left out because `make_slot` is a pure constructor with no model in scope, so parameterising it means threading a word through **109 call sites across 9 files**, six of them convention classifiers including `abi/riscv.mach`, which two other agents are inside right now. That is a collision, not a hard problem. Filed separately.

### The tests fail against the unpatched compiler

Not "a test exists" — each one was run both ways. `int/` is frozen, so all coverage is in `mach test .`.

**Float gate.** Reverting only the `mir.vreg_is_fp` clause reproduces the pre-fix gate:

```
=== unpatched gate ===
  FAIL  mach.lang.be.codegen.legalize:a_float_is_not_split_into_integer_lanes  (exit 1)
0 passed, 1 failed, 1 total

=== patched gate ===
mach.lang.be.codegen.legalize     1 ok    334us
1 passed, 0 failed, 1 total
```

The test asserts **inertness** for an `f64` `MIR_MOV` at `lane_width` 4, with two controls that are what make it mean anything: the identical function over **GP-class** vregs at the identical machine must still split into two lane moves, and the same function on a machine with **no float bank** must still split into eight byte lanes. Without those, inertness would prove only that the fixture legalizes nothing.

**Machine word.** Pinning `gpr_width` and `pointer_width` back to the literal `8` and rebuilding:

```
=== literal-8 build ===
  mach.lang.be.codegen.mir.context.stack_slot_bytes                        (exit 1)
  mach.lang.be.codegen.mir.context.copy_chunk_width                        (exit 1)
  mach.lang.be.codegen.mir.abi:stack_slot_granularity_is_the_machine_word  (exit 1)
0 passed, 3 failed

=== model-read build ===
mach.lang.be.codegen.mir.context.stack_slot_bytes                          1 ok
mach.lang.be.codegen.mir.context.copy_chunk_width                          1 ok
mach.lang.be.codegen.mir.abi:stack_slot_granularity_is_the_machine_word    1 ok
3 passed, 0 failed
```

### The word and the value width genuinely differ

**mach ships no target where they differ, so this is constructed, and the PR says so rather than claiming coverage it does not have.** Each test asserts against synthetic `MachineModel`s:

- `stack_slot_bytes` and `copy_chunk_width` at words **8, 4 and 1**. A 4-byte-word machine must never be handed the 8-byte move it has no register for, which is the assertion a hardcoded ladder cannot pass.
- `stack_slot_granularity_is_the_machine_word` additionally builds a model where the **word is 4 and the pointer is 2**. That is the only thing that can separate sites 6-8: a site reading `gpr_width` for the hidden pointer's alignment answers 4 and must answer 2. Without a model where they disagree, both readings pass.
- The placement side and footprint side are walked together over three arguments on the 4-byte-word machine, pinning that they agree.

### Byte identity

Compiler built before and after, each compiling the full mach project across three targets at both profiles:

```
linux-x86_64/debug:    223 files (222 objects) compared, 0 differ
linux-x86_64/release:  223 files (222 objects) compared, 0 differ
linux-arm64/debug:     223 files (222 objects) compared, 0 differ
linux-arm64/release:   223 files (222 objects) compared, 0 differ
linux-riscv64/debug:   223 files (222 objects) compared, 0 differ
linux-riscv64/release: 223 files (222 objects) compared, 0 differ
TOTAL: 1338 files, 1332 objects, 0 differing
```

### mos6502 DOES change, and correctly

The one place where the literal `8` was **not** a no-op, and it took a full run to find.

mos6502 has `gpr_width = 1`, so its outgoing stack arguments now pack tightly instead of each consuming an 8-byte slot that machine never had, and its GP spill slots reserve the one byte the spill move writes. Both sides of the compiler agree on the new layout, and it is verified **by executing the emitted bytes on the reference 6502 core**, not by reading them.

The only thing that broke was `driver/tests.mach`, which held `UT_MOS_ARG1 = 8` for the second parameter's offset — **the same defect one level up**, a test harness carrying its own copy of the compiler's constant and therefore agreeing with it by construction rather than checking it. It now derives the offset.

### `mach test .`

```
1339 passed, 0 failed, 1339 total
```

### `int`

Filtered to the ABI, float, aggregate and stack-argument cases, all three legs, both profiles. `--runmode qemu` for the non-native legs.

- `linux`: all passed
- `linux-arm64`: all passed
- `linux-riscv64`: all passed, including `2777-riscv-abi-lp64`, `lp64d` and `lp64f`

### Self-host fixpoint

Each stage built into a **freshly removed `out/`**:

```
stage2: cdef8e4ae4949be9  5877760 bytes
stage3: 53df5f42b8f498c3  6465968 bytes
stage4: 53df5f42b8f498c3  6465968 bytes
FIXPOINT: stage3 == stage4
stage2 != stage3 (expected: stage1 is the released 4.16.0, not this branch)
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -318,6 +318,19 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y qemu-user
 
+      # linux-riscv64 is the only leg whose runner is not its own ISA, so it is the
+      # only one that reaches int/lib/cc.sh's CROSS path in CI - and a cross build
+      # needs the target's libc headers as soon as a case's C probe includes one.
+      # c-variadic's does (#2771), and without these clang resolved the runner's own
+      # x86-64 /usr/include instead. keyed on the target rather than the runmode:
+      # this package is riscv64's, and a future qemu leg of another ISA needs its
+      # own, not this one. cc.sh names the same path in its --sysroot.
+      - name: install riscv64 cross libc headers
+        if: matrix.target == 'linux-riscv64'
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y libc6-dev-riscv64-cross
+
       # the host-side validators the structural cases inspect artifacts with, all
       # reading the target artifact regardless of the runner's own ISA. debuginfo
       # (int/surface/debuginfo) verifies the emitted DWARF with llvm-dwarfdump and
--- a/int/lib/cc.sh
+++ b/int/lib/cc.sh
@@ -19,7 +19,9 @@
 # (native build, including every CI runner), and fall back to a cross-capable
 # `clang -target <triple>` only when it does not (a local cross-build). Either
 # path only ever needs to emit an object (`-c`) - mach does the link itself -
-# so `clang -target` needs no sysroot or cross-linker to be sufficient.
+# so the cross path needs no cross LINKER. It does need the target's libc
+# HEADERS the moment a probe includes one, which is what the riscv64 branch's
+# `--sysroot` supplies; see the triple table below.
 #
 # usage: cc.sh <compiler args...>  (same argv a bare `cc` would take)
 #
@@ -56,15 +58,27 @@ if [ "$(host_isa)" = "$MACH_TARGET_ISA" ] && [ "$(host_os)" = "$MACH_TARGET_OS" 
 fi
 
 # cross build: the host cannot produce this leg's ISA/OS natively, so route
-# through clang with an explicit target triple. compile-only (-c) needs no
-# sysroot or cross-linker for any of these, verified against every
-# targets.conf leg (mach#2741) - only a case that asked cc.sh to LINK a full
-# binary would need one, and none does today.
+# through clang with an explicit target triple.
+#
+# COMPILE-ONLY NEEDS NO CROSS LINKER. It DOES need the target's libc headers as
+# soon as a probe includes one, and this file used to claim otherwise - "needs no
+# sysroot or cross-linker for any of these, verified against every targets.conf
+# leg (mach#2741)". That claim was verified against a set that could not disprove
+# it: `c-variadic` was exempt on linux-riscv64, so no probe that cross-built here
+# had ever included a libc header. Un-exempting it (mach#2771) reached the case
+# that falsifies it, and clang resolved the HOST's /usr/include/stdint.h while
+# targeting riscv64 - failing inside it on a file no case mentions
+# ("bits/libc-header-start.h file not found", a multiarch host's glibc looking for
+# its own arch subdirectory). Silently succeeding would have been worse: an object
+# compiled against the wrong architecture's libc headers, which is the same class
+# of quiet divergence mach#2741 named. So the riscv64 branch names a sysroot, and
+# no branch pretends the headers are free.
 #
 # extend this table exactly when targets.conf gains a leg (same axis
 # check-target-matrix.sh already enforces one block per [target.*] on).
 triple=
 extra=
+sysroot_arg=
 case "$MACH_TARGET_OS-$MACH_TARGET_ISA" in
     linux-x86_64)    triple=x86_64-linux-gnu ;;
     linux-aarch64)   triple=aarch64-linux-gnu ;;
@@ -105,9 +119,27 @@ case "$MACH_TARGET_OS-$MACH_TARGET_ISA" in
     #                                  hits the exact same ADD32/SUB32 pairing.
     # suppressing emission at the compiler is the fix, not teaching mach's reader
     # relocation arithmetic it will never otherwise need (mach#2741).
+    #
+    # ...and a SYSROOT, because this is the one leg CI cross-builds: linux-riscv64
+    # is the only targets.conf row whose runner is not its own ISA, so it is the
+    # only leg whose probes need target libc headers in CI. `--sysroot` alone is
+    # sufficient and is what makes the resolution honest - clang finds
+    # <sysroot>/include and stops searching the host's /usr/include, so a probe
+    # either compiles against riscv64's own headers or fails naming the one it
+    # wanted. The path is where Ubuntu's `libc6-dev-riscv64-cross` installs, which
+    # ci.yml installs on this leg; MACH_RISCV64_SYSROOT overrides it for a host
+    # that keeps its cross headers elsewhere. Absent, the build stops here naming
+    # the package rather than falling back to the host's headers, which is exactly
+    # the silent wrong-headers outcome this exists to prevent.
     linux-riscv64)
         triple=riscv64-linux-gnu
+        sysroot=${MACH_RISCV64_SYSROOT:-/usr/riscv64-linux-gnu}
+        if [ ! -d "$sysroot/include" ]; then
+            echo "cc.sh: cross-building linux-riscv64 needs riscv64 libc headers, and none are at '$sysroot/include' - install Ubuntu's libc6-dev-riscv64-cross (or your distro's riscv64-linux-gnu glibc headers) and point MACH_RISCV64_SYSROOT at it if it lands elsewhere. compiling against the host's headers instead would emit an object built to the wrong architecture's libc (mach#2741, mach#2771)" >&2
+            exit 1
+        fi
         extra="-march=rv64gc -mabi=lp64d -mno-relax -fno-asynchronous-unwind-tables -fno-unwind-tables -fno-jump-tables"
+        sysroot_arg=$sysroot
         ;;
     windows-x86_64)  triple=x86_64-pc-windows-msvc ;;
     darwin-x86_64)   triple=x86_64-apple-darwin ;;
@@ -125,8 +157,10 @@ fi
 
 # $extra is deliberately unquoted: it is always either empty or a fixed set of
 # single-word flags built above, never user input, so word-splitting it here is
-# the intended expansion, not a quoting bug.
-clang -target "$triple" $extra "$@" || exit $?
+# the intended expansion, not a quoting bug. the sysroot is the opposite - it can
+# come from MACH_RISCV64_SYSROOT - so it rides its own quoted expansion, present
+# only for a branch that set one.
+clang -target "$triple" ${sysroot_arg:+--sysroot="$sysroot_arg"} $extra "$@" || exit $?
 
 # the -mabi/-march pin above is a REQUEST, and a toolchain that does not honour it
 # fails silently in the direction that looks like a mach bug: a soft-float probe
--- a/int/surface/c-variadic/case.conf
+++ b/int/surface/c-variadic/case.conf
@@ -10,12 +10,11 @@
 # compile the probe object through int/lib/cc.sh when the runner's own `cc` can't
 # target the leg (mach#2741), and both match the golden.
 #
-# exempt linux-riscv64: builds and runs cleanly through the same cc.sh path, but
-# every variadic float/double argument comes back as 0 - a real mach bug in
-# riscv64's lp64 variadic-float lowering, exactly the class of divergence this
-# case exists to catch, newly reachable now that cc.sh can cross-build the probe
-# for this leg at all. Tracked as #2771; re-enable once that lands rather than
-# reporting a wrong value as a pass.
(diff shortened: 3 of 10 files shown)


octalide:
Holding this merge, not because of anything wrong with the fix.

CI is green on this branch only because `d635afb7` re-exempted `surface/c-variadic` and `surface/narrow-stack-args` on `linux-riscv64` pending #2797. That was the right call to unblock CI, but it means the riscv64 legs are **skipped, not passing**.

#2771 is a riscv64-only defect. With those two cases exempt on that leg, nothing in the suite executes the variadic classifier fix on the only target it affects, so merging now would close the issue with zero executing coverage of its own fix. That is the failure shape this repo has been hit by repeatedly this week: green because the question was not asked.

Sequence instead:
1. #2797 lands (the linker collapsing nine pooled-constant `AUIPC`/`LO12` pairs onto one low-12 offset, which is what was actually failing these two cases).
2. The `exempt: linux-riscv64` lines come back off both cases.
3. This PR re-runs, and the riscv64 legs go green **by executing**.
4. Merge.

If step 3 shows the classifier fix is still needed, we will know it is real. If those cases pass on riscv64 without this PR once #2797 is fixed, then the variadic classifier change needs its own justification rather than inheriting one from a failure it did not cause.

octalide:
Rewritten and re-verified. The hold above is addressed on both counts.

- `surface/c-variadic` is **un-exempt on `linux-riscv64`** and the leg is green with it running, not skipped. `int linux-riscv64` passes in CI on this branch, and the full leg passes locally under qemu.
- `surface/narrow-stack-args` was already un-exempt on the leg (#2797 landed); it stays that way and passes.

The evidence the hold asked for is in the PR body: the case failing against a pre-fix compiler, the flat-zero symptom reproduced as a controlled experiment (lp64d 's `variadic_float_bits` set to its named `flen_bits` of 64), and the real values read back after.

Turning the case on also falsified a claim in `int/lib/cc.sh` about cross builds needing no sysroot; that is the second commit, and it is why `int linux-riscv64` needed a CI change rather than a re-exemption.

Issue #2788, opened 2026-08-08
asm: --emit-asm prints the wrong symbol for a memory operand and drops a call's operand entirely


octalide:
Found while investigating an unrelated inlining defect. Pre-existing on `dev`, not introduced by any current branch.

## What

`--emit-asm` renders symbol operands **inside an `asm` block** wrongly. Two distinct failures, both in the printed text only:

- a memory operand naming a symbol prints the **wrong symbol**: every `[rip + _rt_argc]` renders as `[rip + dbggdbcase]`, which is a different symbol entirely
- a call naming a symbol **loses its operand**: `call _rt_init` renders as a bare `call` with nothing after it

**The emitted object is correct.** Only the printed assembly is wrong, so this is a diagnostic-surface defect rather than a codegen one.

## Why it matters more than it looks

`--emit-asm` is what someone reaches for when debugging inline asm, which is exactly the code where they cannot fall back on reading the source and assuming the obvious. A reader chasing a symbol reference is shown a **different, real symbol name** rather than something obviously broken, so there is no signal that the output is untrustworthy. A bare `call` with no operand is at least visibly wrong; a plausible wrong name is not.

This cost real time during an investigation tonight: an agent chasing an inlining defect followed the printed operand and had to discover independently that the text disagreed with the object.

It belongs to the class epic #2288 exists for - a diagnostic surface that cannot represent what it is used to inspect. Here the surface actively misrepresents it.

## Likely shape of the cause

The operand carries a symbol reference that the printer resolves against the wrong table or index, since it produces a **valid symbol name from the same module** rather than garbage. The `call` case losing its operand entirely suggests the printer has no branch for that operand kind rather than a wrong lookup. Confirm both rather than assuming they share a cause - they may not.

## Acceptance

- A symbol operand inside an `asm` block prints the symbol actually referenced, checked against the emitted object's relocation for that site rather than against what the printer produces.
- A `call <symbol>` inside an `asm` block prints its operand.
- Covered for all three register targets, since the printers are separate per ISA and a fix in one says nothing about the others.
- A test that would fail if the printer regressed to naming a different real symbol, not merely to naming nothing.

PR #2802, opened 2026-08-08
fix(codegen): remove inline-asm operand scratch buffer

### LINKED ISSUE #2789: asm: an operand longer than a register-name scratch buffer is reported as malformed, on a limit nothing below it imposes
## What

Each ISA's inline-asm operand parser copies the operand token into a fixed stack buffer before deciding what kind of operand it is, and reports a **well-formed** operand as malformed when the copy does not fit:

```mach
var buf: [64]u8;
if (!iasm.copy_region(c.body, lo, hi, ?buf[0], 64)) {
    ret R.err[bool, str]("encode: malformed inline-asm operand"); }
val tok: str = (?buf[0])::str;
val ri: i32 = asm_reg_index(tok);
```

`tok` has **exactly one consumer**, the register-name lookup on the next line. Every branch after it - `%mod(sym)`, `off(base)`, immediate, bare symbol - works directly on `c.body, lo, hi` and never reads the copy. The symbol path in particular stores `name_off` / `name_len` as offsets into the body.

So the logic is: copy the operand into a scratch sized for **register names**, and if it does not fit, declare the operand malformed. Register names are three or four characters. A 108-byte symbol is perfectly well-formed and simply is not `a0`.

## The sizes are arbitrary, and nothing below them imposes a limit

- `riscv64/encode.mach` and `arm64/encode.mach` use `[64]u8`; `x64/encode.mach` uses `[128]u8`. Three copies of one scratch at two different sizes is itself the evidence that no one chose a value.
- **There is no ISA limit**: instructions encode addresses and offsets, never names. Symbol names are a linker and object-format concept.
- **There is no object-format limit**: ELF writes names verbatim into `.strtab` with no validation and no length cap, COFF routes anything over 8 bytes to its string table, and Mach-O neither inspects nor prefixes mach's own symbols.
- **There is no assembler**: mach emits objects directly.

## It is live today, not theoretical

Measured over the compiler's own objects: **1565 of 12477 symbols exceed 63 bytes**, longest **108**:

```
_M3std5types6resultN86unwrap_errIN4mach4lang5build4emit18LoadedImageOptionsEN4mach4lang5build7outcome4FailEE
```

Refusal points confirmed by bisection: arm64 and riscv64 at 64 bytes with 63 the last accepted, x64 at 128. So on two of three targets, roughly an eighth of the symbols in the compiler cannot be named from an inline-asm body, and the diagnostic blames the author's syntax.

## Fix

**Remove the buffer, do not enlarge it.** Give `asm_reg_index` the region (`body, lo, hi`) and let it compare in place. The pattern already exists in the same layer: `iasm.is_sym_region(s, lo, hi)` does exactly this kind of in-place check without copying.

Then there is no scratch, no size to pick, no overflow branch, and no misleading diagnostic - the class of defect disappears rather than moving to a larger number. A bigger buffer would leave the same wrong error waiting at a new threshold.

All three ISAs carry the pattern and all three should change together, since a fix in one says nothing about the others.

## Acceptance

- An inline-asm body can name a symbol of any length the object format accepts, on all three register targets, verified with a symbol longer than 108 bytes.
- No fixed-size operand scratch remains in any ISA's operand parser.
- A genuinely malformed operand still reports as malformed, with a test that would fail if the two cases were confused.
- `mach test .` green, `int` green including the native aarch64 leg, self-host fixpoint holds.

octalide:
## Summary
Each ISA's inline-asm operand parser copied the operand token into a fixed stack buffer before deciding what kind of operand it was, and reported a well-formed operand as malformed when the copy didn't fit (`[64]u8` on aarch64/riscv64, `[128]u8` on x86-64). The buffer's only consumer was the register-name lookup; every other branch already read the source region directly. Removed, not enlarged - `asm_reg_index`/`a64_reg`/`x64_reg` now compare the register table in place.

x86-64's `[symbol]` memory operand had a second, narrower instance of the same shape one layer down: `x64_mem`'s no-star bracket parser copied into a `[32]u8` meant for a base register, and a bare rip-relative symbol reference (`[rip + <symbol>]`) passed through the same copy, since x86-64's grammar makes a base register and a bare symbol genuinely ambiguous at that position until one is ruled out. Fixed the same way. Checked and confirmed arm64/riscv64 don't have this second problem: their memory syntax (`off(base)`, `[Xn, ...]`) requires a register in the base position syntactically, and a bare symbol is a structurally different form (`:mod:sym`) that never touches a register buffer.

## Note on scope
This branch originally also carried a fix for #2788, developed independently before I found that PR #2801 (merged) already fixed it with the same root-cause diagnosis and a stronger int-level test. Rebased onto `dev` after #2801 merged and dropped the #2788-specific changes - this PR is #2789 only now.

## Verification
- per-ISA unit test: same operand run twice varying only length (134 bytes, past every old limit on every ISA), the long one **succeeds** and encodes **byte-identically** to the short one (not merely "no error"), and the relocation carries the full untruncated name. A genuinely malformed operand still refuses, checked in the same test so the two cannot be confused.
- built a compiler from **unpatched** source with each new test appended and confirmed each fails there, on all three targets, before calling it done.
- `mach test .`: 1324 passed, 0 failed
- `int/lib/check-target-matrix.sh` / `check-deps.sh`: clean
- `int/run.sh --target linux ./m` and `--target linux-arm64 --runmode qemu ./m`: full suite green on both legs, rebased onto current `dev`
- self-host fixpoint holds: stage3 == stage4 byte-identical

Closes #2789


octalide:
Heads-up on overlap: the #2788 half of this landed already in #2801 (merged into `dev`), with the same root cause and the same fix shape — intern the symbol once in `x64_emit` before `x64_build`, thread the `StrId` onto the rip-relative operand and onto the direct transfer's `dst`, and reuse it in `x64_emit_sym_ref` instead of interning a second time. It also updated the same two unit tests that were pinning `dm.dst.kind == OPK_NONE`, and added `int/surface/asm-symbol-operands`, which pairs the names `--emit-asm` prints against the names the object relocates against across all three ELF legs.

So this branch will conflict in `x64_emit` / `x64_build` / `x64_to_isa` and in those two tests. The #2789 half (the fixed operand scratch buffers across all three ISAs) is untouched by #2801 and is the part still worth landing — rebasing onto `dev` and dropping the #2788 changes should leave it clean. Worth re-checking the `[32]u8` base-register buffer you found on the x86-64 `[symbol]` path though: #2801 did not touch that, so if it is still a live truncation it belongs in this branch.

Issue #2800, opened 2026-08-08
link: the pinned-import diagnostic reports a library that is in the link as absent


octalide:
> **Rewritten.** This issue was originally filed as *"a consumer's `#[library]` does not bind against a dependency's exported entry"*. That premise was tested against an unpatched `dev` compiler and is **false** — the cascade works. The real defect is the diagnostic, which states something untrue and sent the investigation a day in the wrong direction. The original report, its repro, and the findings that disproved it are all kept below.

The linker refuses a pinned import with:

```
error: import 'X' pinned to library 'glfw' not among the link's dependencies
```

when `glfw` **is** among the link's dependencies. It is a `source = "local"` entry — a static input — and a static entry's logical name never reaches the linker at all, so the check that produces this message searches only half of the link and reports the half it cannot see as absent.

The message is not merely imprecise. It asserts a fact about the manifest that is wrong, and it is the only signal the user gets.

## Where it comes from

`resolve_and_store_input` (`src/lang/build/emit.mach`) keeps a link input's logical `library` name only for `LINK_DYNAMIC` facts; a static input is reduced to a path. `dynlib_index_of` (`src/lang/be/linker.mach:5359`) and the `pinned_lib_message` call site (`src/lang/be/linker.mach:5278`) therefore search a set that structurally cannot contain a static entry's name.

Two notions of "the link" exist and attribution reads the narrower one: the **dynamic** dependency set, not the **effective** link set (own entries plus cascaded exports, static and dynamic alike).

## The wording is decided by accident

Which of two messages fires for the same broken program depends on what else is in the link:

- absent symbol pinned to a static entry's name, **and some other dynamic input present** → `import 'X' pinned to library 'L' not among the link's dependencies`
- the identical absent symbol pinned to the identical name, **no dynamic input in the link** → `undefined symbol: X`

The second is the truthful one. Which one a user sees is a consequence of unrelated link contents, which is how a message that states a falsehood survived.

## Original repro (kept)

`mach-glfw` exports its vendored archive and gives it the logical name:

```toml
[link.glfw-static]
source  = "local"
path    = "{project.out}/vendor/glfw/libglfw.a"
library = "glfw"
export  = true
```

`boom` depends on mach-glfw, has `link = []`, and declares one import of its own:

```mach
#[library("glfw")]
ext fun glfwGetInstanceProcAddr(instance: ptr, name: *u8) ptr;
```

```
[...]

PR #2804, opened 2026-08-08
test(int): pin that an exported link entry's logical name binds in a consumer

octalide:
Adds `int/regression/2800-cascade-library-attribution`.

## Why

`doc/manifest.md` states the rule twice — an entry with `export = true` also applies to any project that links this project's modules, and its `library` value is the stable logical name a consumer's `#[library("...")]` resolves against. Nothing in the suite asserted it.

It could not have been asserted anywhere cheaper. Every axis of the guarantee — the cascade, the logical name, the consumer-side decorator — needs two manifests to exist at once, so a unit test cannot reach it and neither can any single-project case. The consumer here declares `link = []` and nothing but a `dep` entry; its own `#[library("mathlib")]` binds against the dependency's `[link.m] library = "mathlib" export = true`, and the binary links and runs.

## This case passes today, and that is the point

It is not a regression guard for a fix in this PR — there is no fix in this PR. #2800 was filed believing this path was broken. Testing the premise against unpatched `dev` showed it works: the reporter's import named a symbol that does not exist (`glfwGetInstanceProcAddr`, for GLFW's `glfwGetInstanceProcAddress`), and a misleading linker diagnostic pointed the investigation at the cascade instead. Had this case existed, the day that took would have been a minute.

`case.conf` carries that account inline so a later reader does not read a passing case as dead weight and delete it.

## Verification

`int/run.sh --target linux --deps pin --filter '2800-*'` — PASS on debug and release, against a compiler built from unpatched `dev`.

Related to #2800, which is being rewritten as the diagnostics bug it actually is.


octalide:
Full `int` suite on linux, `--deps pin`, with this case included: 289 PASS, 0 FAIL, against a compiler built from this branch's base. The case itself passes on debug and release.

To be explicit about what marking this ready means here: there is no fix to run. The behaviour the case asserts already works, and the run above is the evidence that it does — which is the whole claim of the PR.

Issue #2749, opened 2026-08-07
simd: the realization authority answers a lane-width question, but a SPIR-V Shader module's ceiling is a lane count


octalide:
Blocks #2727.

`isa.packed_width` answers a question keyed on the lane **width** and returns a bit width the caller divides by that lane width to get a lane count. SPIR-V's constraint is not a bit width: a Shader module's `OpTypeVector` takes 2, 3 or 4 components and nothing else, at every element type. The authority cannot express that, so SPIR-V declares an empty gap list and the mismatch is currently masked by two other gates that #2727 removes.

## What

#2726 made vector operation legality target-independent and put the whole of the realization dependence in one function:

```mach
pub fun packed_width(m: *MachineModel, op: VecOp, is_float: bool, lane_bits: u32) u32
```

with `packed_gaps` empty for SPIR-V, documented as:

> empty on an ISA whose packed forms are total (SPIR-V)

That reading is right about *operations*: SPIR-V has a packed form for every `VecOp` at every lane width, and no `(op, lane_bits)` pair is missing. It is the wrong axis for the constraint SPIR-V actually has, which is on the lane **count**:

| shape | `packed_width` says | SPIR-V can type it |
|---|---|---|
| `f32x4` | 128 | yes |
| `f32x3` | 128 | yes |
| `i16x8` | 128 | **no** — 8 components need `Vector16`, a Kernel capability |
| `i8x16` | 128 | **no** — 16 components, same |

For 8-bit lanes the authority reports a 128-bit packed form, which is 16 lanes, which no Shader module may declare. The emitter refuses those two shapes by name:

```
error: the SPIR-V target has no vector form for `i8x16` / `u8x16`: a Shader
module's vector holds at most 4 components, and a 16-component vector needs the
Vector16 capability, which only a Kernel module may declare
```

So today the authority and the back end give different answers about the same shape, and the back end's is the one that lands. That is already a smaller version of the incoherence #2726 deleted: `i8x16 + i8x16` is refused on SPIR-V for a realization reason, when #2488's rule is that a vector type and a lane-wise operator are legal on every target and only the realization varies. The right answer for `i8x16` on SPIR-V is 16 scalar `OpIAdd`s, not a refusal.

## Why it is worth fixing before #2727 rather than after

[...]

Issue #2716, opened 2026-08-07
codegen: take the direct chunked access for sub-width vectors whose lanes align, instead of the scratch round trip


octalide:
Follow-on to #2697, which made sub-register-width vectors work by spilling and gathering through a scratch slot. This is the optimisation that removes the round trip, recorded because it is the obvious next step and because the obvious version of it is wrong.

## What

#2697 carries a sub-width vector between a register and memory through a register-width scratch slot, walking it in 8/4/2/1 chunks the way aggregate copies already do on every target. For a 12-byte `i32x3` that is a store to scratch, an 8-byte chunk, a 4-byte chunk, and a load — where the x64 encoder can legalise the same access directly as an 8-byte plus a 4-byte access against the real address, with no scratch and no round trip.

That is a real win on the shapes it applies to: `i32x3` is the shape `vec3` is made of, and it is the one most likely to appear in a hot loop.

## Why the obvious version is wrong

**It does not generalise, and the failure is silent.** The chunk decomposition has to respect lane boundaries, and for several shapes it does not:

- `i16x7` is 14 bytes. After an 8-byte chunk the tail is 6 bytes spanning lanes 4, 5 and 6 — a 4-byte chunk of that is the second half of one lane and the first half of another.
- `i8x9` is 9 bytes: 8 + 1, which is fine only because the lane width is 1.

So a decomposition written against `i32x3` and generalised by size arithmetic produces, for `i16x7`, an access that splits a lane across two chunks. Every byte still arrives, so a naive lane-sum test passes; what breaks is anything that observes a partial lane — which is why this is worth doing carefully rather than quickly.

The scratch round trip is uniform across every shape precisely because it never has to reason about lanes: it moves bytes into a slot the register width fits, and the lane structure is re-imposed by the vector load.

## Fix

Take the direct path **only where the chunk boundaries fall on lane boundaries**, and fall back to the scratch round trip otherwise. The predicate is a property of the shape, not of the size: a chunk decomposition is admissible when every chunk boundary is a multiple of the lane width.

For a 16-byte register and the 8/4/2/1 walk that means the direct path is available when the lane width divides 8, 4, 2 and 1 as the walk descends — so `i32x3` (4-byte lanes, 8+4) qualifies and `i16x7` (2-byte lanes, 8+4+2 with a 6-byte tail) does not at the 4-byte step.

Candidate shapes, least to most:

[...]

Issue #2794, opened 2026-08-08
spirv: no sampler or image bindings, so a textured shader cannot be written


octalide:
The GPU decorator set covers `#[input]`, `#[output]`, `#[builtin]`, `#[uniform]`, and `#[storage]`. There is no way to declare a sampled image, so a shader that reads a texture cannot be expressed.

## Impact

This is the remaining blocker for briar-systems/boom#27, which replaces boom's GLSL shaders with Mach ones. Every shader boom ships samples a texture:

- the sprite shader samples `u_texture0` and is what draws every 2D quad and all text
- the material shader samples a base-colour map
- only the plain lit fragment shader in `examples/animation` avoids it

So the shader migration can prove its build path today but cannot port the shaders that matter.

## What is needed

A binding that lowers to `OpTypeImage` plus `OpTypeSampledImage`, and a sampling operation lowering to `OpImageSampleImplicitLod`. Shape consistent with the existing decorators:

```mach
#[sampler(0, 1)] var albedo: sampler2d;

#[stage("fragment")]
fun frag_main() {
    out_colour = sample(albedo, in_uv);
}
```

Descriptor set and binding numbering matches `#[uniform]`, so nothing new is introduced at the descriptor level.

## Acceptance

- A fragment shader declaring a sampled image and sampling it compiles to a module `spirv-val` accepts.
- The emitted module declares `OpTypeImage` with the right dimensionality and a `Sampled` operand of 1.
- Depth-comparison sampling is out of scope for a first cut and should be named as such rather than half-supported.

octalide:
Implemented in #2803. The exact surface a consumer sees, so mach-vk / mach-shader / boom can bind against it.

## Source

```mach
#[spirv_op("core", "OpImageSampleImplicitLod")]
fun sample(s: sampler2d, uv: f32x2) f32x4;

#[sampler(0, 1)] var albedo: sampler2d;

#[input(0)]  var in_uv:      f32x2;
#[output(0)] var out_colour: f32x4;

#[stage("fragment")]
fun frag_main() {
    out_colour = sample(albedo, in_uv);
}
```

Sampling is a `#[spirv_op(...)]` declaration rather than a new language form — a sample IS one SPIR-V instruction the way `sqrt` and `dot` already are, so it rides the decorator the GPU set already had. Write the declaration once in a shader library and the call site reads exactly as this issue sketched it.

## Emitted module

`spirv-val --target-env vulkan1.3` accepts it.

```
               OpCapability Shader
[...]

PR #2806, opened 2026-08-08
lang: admit a vector of any lane count on every target

### LINKED ISSUE #2727: simd: a vector wider than the target's register is refused by name, so arbitrary lane counts do not work

octalide:
Closes #2727

**Stacked on #2805** (#2716). That branch is its base, so this diff shows both until #2805 merges; review it after.

## What

`vector_bits` was read as a legality gate on the surface when it is a capability fact about realization. Every target refused `f32x5` / `f32x8` / `f32x16` by name at the front end, and incoherently: rv64gc has no vector unit at all, scalarizes every vector it is handed, and still refused `f32x8` on the ground of a 128-bit SIMD width it does not have.

The form now admits `TxN` for any element and any `N` in [2, 65535], identically on every target - `type.vector_form` no longer takes a width at all. The only refusals left are the degenerate one-lane vector and a lane count past what MIR represents, and neither sentence mentions a target.

## The shape

**One authority for placement.** `isa.fits_vector_register` folds the register width and the lane-count ceiling exactly as `packed_width` already folds them for an *operation*, so no caller consults two facts. False has one meaning and it is not a refusal: the value is placed in memory and worked one lane at a time. That single answer covers a target with no vector unit, a shape wider than the register, and a shape past a lane-COUNT ceiling (SPIR-V's four components), which is what stops the over-width path and the no-SIMD path being two mechanisms that drift.

**`me.pass.scalarize` became per-value instead of per-target.** Its selection predicate was a bare `is_vector`, which was the same question only because the pass ran all-or-nothing. Now it claims a value the target cannot hold in a register: every vector on rv64gc (its original job, byte-identical) and only the over-width shapes on x86-64 / aarch64, so an `i32x4` beside an `f32x8` still selects one packed instruction.

**Placement runs ahead of the two operator stages.** Not a correctness gate - measured, both orders give the same lanes on all three legs - but the cost differs enormously, because reaching an over-width shape first the operator stages expand it against a value no register holds and placement then takes the result apart again. `@d = @a + @b` at `i32x8`, x86-64 release: **169 instructions / 128-byte frame** placement-first, **871 / 1072** placement-last. The comment says which of the two it is.

**Layout gains a third alignment rung.** An over-width vector aligns to the vector **register** width - `f32x8` is 32 bytes aligned to 16, not 32 - because it is placed as several register-width pieces and there is no 32-byte vector load on a 16-byte register for a larger alignment to serve. The reasoning is written into `layout.mach`.

**Every convention gives an over-width vector the memory class**: a hidden pointer in, the indirect-result pointer out. `aapcs64` had an unreachable by-value arm for this case; it would have handed over a slot describing a value the caller does not hold.

The two fixed 16-lane buffers the old bound had made safe (`MAX_GAP_LANES`, `MAX_VECTOR_LANES`) are allocations, and SPIR-V's `const_composite` / `emit_vec_build` lost their lane-count refusals with them.

## Testing

`int/surface/vector-wide` asserts **lane values after real arithmetic**, because #2687 shipped a silent miscompile behind a test that only checked `$size_of`. Per-lane (a sum cannot tell a rotation from a right answer), on all three ELF legs and both profiles:

- `i32x8` all eight lanes after `+`, plus `-` and `&`
- the non-power-of-two `i32x5` and `i16x7`, and `f64x4`
- a **loop-carried `i32x8` accumulator** through a phi, which is the shape #2749 found straight-line cases cannot see
- **guard fields on both sides** of a wide vector, and either side of an `i32x5`, so a footprint overrun is caught
- `$size_of` / `$align_of` / `$offset_of` agreeing, `[3]i32x8` strides, and a wide vector passed to and returned from a call

`f32x8` / `i32x8` / `f64x4` also compile and validate on SPIR-V, where the same authority routes them to the `OpTypeArray` realization an over-count shape already took.

**Failure demonstrated against unpatched source**, not just a green matrix:
- built from `dev`: every shape in the case is refused at the front end
- mutating the alignment rung back to `size`: 4 assertions flip to WRONG (`alignof-i32x8` 32, `offsetof-v-before` 32, `sizeof-WideAfter` 64, `offsetof-odd-guard1` 40)

## Verification

- `mach test .`: 1304 passed, 0 failed
- `int/run.sh --target linux-arm64 --runmode qemu`: all cases passed
- `int/run.sh --target linux-riscv64`: all cases passed
- self-host fixpoint: stage2 == stage3 byte-identically, each stage built into a freshly removed `out/`


octalide:
## A misattributed doc comment was corrected in this PR's rebase — recording it so the fix is not invisible

Noting after the merge, because it arrived as a conflict resolution rather than as intended work and would otherwise sit unannounced inside a rebase.

#2794 inserted its `image_form` test directly beneath `vector_form`'s doc comment in `src/lang/type.mach`, leaving that comment documenting a test it has nothing to do with:

```mach
# `vector_form` reads `<element>x<lanes>` and separates the three outcomes: a
# shape the target admits, a shape too wide for it, and a spelling that is not a
# vector at all. the ten formerly seeded spellings are exactly the 128-bit forms
test "mach.lang.type.image_form:reads_every_operand" {
```

It surfaced when this branch's own `vector_form` test collided with it. Rather than resolve around it, `image_form` now carries a header derived from its own body — what it asserts and why every operand is checked individually — and `vector_form`'s comment sits above `vector_form`'s test.

Worth naming rather than quietly fixing: a doc comment attached to the wrong test misleads the next reader **while looking authoritative**, which is the same failure #2288 is about one layer down. Nothing about it was load-bearing for this change; it is a documentation correction that happened to be adjacent.

Issue #2813, opened 2026-08-08
link: linked riscv64 ELF images carry e_flags = 0, claiming a soft-float non-compressed ABI they are not


octalide:
Found while writing the regression case for #2797.

## What

Every riscv64 executable mach links has `e_flags = 0x0` in its ELF header:

```
$ readelf -h out/int/build/bin/case | grep -i 'machine\|flags'
  Machine:                           RISC-V
  Flags:                             0x0
```

`0x0` on RISC-V means **soft-float ABI, no compressed extension**. Both halves of that are false for the image: mach's riscv64 back end passes floats in `fa` registers (lp64d), and the image contains RVC instructions - its own input objects say so, carrying `0x5` (`RVC, double-float ABI`) whether they come from mach's own writer or from clang.

## Why it matters

The immediate cost is that every RISC-V-aware reader believes the header over the bytes:

- `llvm-objdump -d` on a linked image decodes every compressed instruction as `<unknown>`, because the header told it the C extension is absent. Passing `--mattr=+c` works around it, which is what `int/lib/produce.sh`'s `riscv_pcrel_image_scan` now has to do - and a scan that silently measured nothing over `<unknown>` lines is exactly the failure shape that hides a defect.
- A tool that checks ABI compatibility before combining or loading an image reads a soft-float claim from a hard-float program. `#2797` was chased for a long time partly because a soft-float/hard-float mismatch produces precisely the zeros the real bug produced, so a header that lies about this is actively expensive to debug around.

Nothing at runtime consults it today (the kernel and qemu-user do not gate on `e_flags` for a static executable), so this is wrong-metadata rather than wrong-execution.

## Fix

The writer should carry the float-ABI and extension flags through from the merged inputs rather than leaving the field zero. The inputs already agree on a value - `readelf -h` on every contributing object shows `0x5` - so the merge rule is "all inputs must agree, and the agreed value is the image's", with a disagreement being a real error worth naming (a soft-float object linked into a hard-float image is not linkable).

Check the other ELF arches for the same gap while there: `e_flags` is arch-defined and unused on x86-64, but aarch64 has its own bits.

## Acceptance

- `readelf -h` on a linked riscv64 image reports `RVC, double-float ABI`, matching its objects.
- `llvm-objdump -d` decodes a linked riscv64 image's compressed instructions without `--mattr=+c`; `int/lib/produce.sh`'s workaround for it comes out.
- A case asserts the emitted header field, not just that the image runs.

PR #2815, opened 2026-08-08
fix(link): pair riscv64 pc-relative low-12 relocations with their own auipc

### LINKED ISSUE #2797: link: riscv64 collapses multiple AUIPC+fld/flw constant-pool pairs onto one low-12 offset
Found while investigating a CI-only failure on mach#2771/#2784 (int/surface/narrow-stack-args and int/surface/c-variadic failing on linux-riscv64 in CI only, passing on one developer's host). Ruled out cc.sh's cross-compiler flags as the cause first (see below) - this is a distinct, deeper, and more serious bug than the one #2771/#2784 were chasing.

## Repro

Build int/surface/narrow-stack-args for linux-riscv64 (any recent origin/dev): `mach build . --target linux-riscv64 -o /tmp/bin`. The probe's `float_tail` C function takes 8 named `f64` parameters (values 1.0 through 8.0, each from its own pooled float constant, loaded fa0 through fa7) plus a further `f32`/`f32`/`f64` tail that spills to integer registers once the fa bank is exhausted.

Disassemble the caller site in `main` (`llvm-objdump --triple=riscv64 -dr /tmp/bin`, find the `fld fa0..fa7` cluster immediately before the `jalr <float_tail>`):

```
419830: fld fa0, -0x2bc(t0)
419838: fld fa1, -0x2bc(t0)
419840: fld fa2, -0x2bc(t0)
419848: fld fa3, -0x2bc(t0)
419850: fld fa4, -0x2bc(t0)
419858: fld fa5, -0x2bc(t0)
419860: fld fa6, -0x2bc(t0)
419868: fld fa7, -0x2bc(t0)
419870: fld ft0, -0x2bc(t0)
```

Every one of these nine loads carries the IDENTICAL immediate (-0x2bc) off the SAME base register t0, confirmed at the raw instruction-word level (not a disassembler rounding artifact: the encoded imm[11:0] bit field, bits 31:20, is 0xd44 in all nine 32-bit words - only the destination-register field differs). Executing this loads whatever ONE address t0-0x2bc names into all nine registers - here, apparently zero-filled memory, since the observable is every float argument reading back as exactly 0 (`int/surface/narrow-stack-args`'s golden: `float n=11 regs=10 80` / `float tail=95 105 115`, actual: `regs=0 0` / `tail=0 0 0`).

The unlinked object (`out/linux-riscv64/debug/obj/case/main.o`, before linking) is correct: each of the nine loads has its OWN `auipc t0, 0x0` immediately before it, each auipc carrying its own `R_RISCV_PCREL_HI20` against a DISTINCT local symbol (`.Lconst.0` through `.Lconst.8`), each fld carrying the matching `R_RISCV_PCREL_LO12_I` against that same distinct symbol. The nine `.rodata.cst` sections holding the actual constant bytes are also individually correct (read directly with `readelf -x`: 0x3FF0000000000000 through 0x4020000000000000, i.e. 1.0 through 8.0, then 0x4027000000000000 = 11.5 for the tail's `f64`). So the defect is introduced during LINKING: nine separate, individually-correct AUIPC+LO12 pairs in the object become one shared `auipc` in the image (a legitimate relaxation - real distance between the nine constants is small enough for one page), but only the FIRST pair's resolved low-12 offset survives; the other eight instructions keep that same low-12 rather than each getting its own distance-from-the-shared-auipc recomputed.

`int/surface/const-pool` (mach#2700/#2248) already exercises pooled float constants on riscv64 and passes, but its heaviest case (`distinct3`) uses three constants, not nine back-to-back loads feeding one call's argument registers - the auipc-sharing path this hits may only engage past some density/count threshold, which would explain why the existing case never caught it.

## Not the cause (ruled out first)

Two real but insufficient fixes already landed on `int/lib/cc.sh` chasing this exact CI symptom before the true cause was found:
- Pinning `-mabi=lp64d` (mach#2771) - CI's toolchain defaulted to it differently than a developer's newer one.
- Pinning `-march=rv64gc` alongside it - CI's clang (18) needs the architecture profile stated explicitly for `-mabi=lp64d` to take effect at all.

Both were verified necessary and correct (confirmed via `readelf -h` on the produced object: `double-float ABI` flag set, matching what a native riscv64 toolchain produces) and are staying in `cc.sh`. But applying them did not fix CI, because the actual defect is downstream of the C object entirely - it is in how mach's OWN linker places mach's OWN pooled-constant loads, and reproduces with `cc.sh`'s cross-compiled object correctly built. Reproduced directly in a container matching the CI runner (Ubuntu 24.04, clang-18, qemu-user 8.2.2) to rule out a qemu-version or environment difference: same corrupted immediates, same wrong zeros.

## Fix

In `src/lang/be/linker.mach`'s riscv64 relocation resolution (or wherever the auipc-sharing/relaxation decision lives): when multiple `R_RISCV_PCREL_HI20`/`R_RISCV_PCREL_LO12_I` pairs against distinct symbols get merged onto one shared `auipc`, each `LO12_I` instruction must still get its OWN low-12 immediate recomputed against its OWN target symbol's actual placement relative to that shared `auipc` - not inherit the first pair's resolved value. Likely candidate designs: either don't share the `auipc` across pairs targeting different symbols at all (correctness over code size), or share it but track each `LO12_I`'s own symbol through to its own immediate computation.

## Acceptance

- A case with 4+ (or however many the real threshold is - find it) distinct pooled float constants loaded back-to-back on riscv64 reads each one back correctly, not the first one's neighbor's zero/wrong value.
- `int/surface/narrow-stack-args` and `int/surface/c-variadic` pass on `linux-riscv64` once this lands (currently re-exempted there with this issue as the stated reason).
- A regression case pins the exact shape: N >= 4 distinct float constants consumed in one function, each read back by value, not just "the build links and runs."

octalide:
Closes #2797

## Summary

A `R_RISCV_PCREL_LO12_I/_S` **does not name the address it loads**. Per the RISC-V psABI it names a *label at the paired `auipc`*, and the target is read off the `R_RISCV_PCREL_HI20` sitting at that label.

mach never followed that indirection. It took the low half's own symbol as the target — which for this spelling *is* the `auipc`'s address — and measured from `patch_va - 4`. The displacement collapsed to `auipc_va - (patch_va - 4)`: a number about where the two instructions sit, with the actual target nowhere in it.

**This broke every foreign object unconditionally, at every hi/lo distance.** Byte-adjacent halves — exactly the layout the `patch_va - 4` rule assumed — resolve to a flat `0`, which is not the assumption holding but the target having dropped out. Halves one instruction apart resolve to `-4`. Neither is the address the pair names. Verified at the byte level on a real clang 18.1.3 object with an adjacent pair: pre-fix `fld fa5, 0x0(a0)`, post-fix `fld fa5, -0x2dc(a0)` resolving to `0x41d070`, whose bytes are `00 00 00 00 00 00 29 40` = 12.5, the constant the paired HI20 names.

**Silent, and it read as zero.** The wrong address held a zero, so a pooled scale factor loaded as `0.0` and every product came out exactly `0` — a row of believable values, not a crash.

## Corrections to #2797's filing

The issue is carefully evidenced but its reading of that evidence is wrong in two ways that would mislead a future reader:

- **There is no `auipc` sharing and no merging.** Each of the nine `fld`s has its own `auipc`, still there in the linked image at an 8-byte stride. The immediates are identical because the constants are also 8 bytes apart, so every pair has the same displacement. That output is correct by construction.
- **There is no density or count threshold.** `int/surface/const-pool` passes because it links only mach's own objects, which never use the psABI spelling.

And a correction to my own earlier framing: this is **not** a story about compiler versions. `narrow-stack-args` passes under clang 22 only because that clang builds its `10.0` with `lui`+`slli` and emits **no relocation at all** — there is no pair to resolve wrongly. clang 18 pools the constant and gets one. Whether a probe trips this is a property of the shape its compiler emitted, never of the compiler being new enough.

## The fix

The pairing is now a **per-instruction fact carried on the relocation**, not a rule re-derived from adjacency. The producer folds this low half's own distance back to its `auipc` into the addend, so `S + A - P` names the `auipc`'s pc for a low half exactly as it names the patch site for a high half.

- **`isa/riscv64/reloc.mach`** — `LA_HI_LO_GAP` is gone; the applier measures from the plain patch site and assumes nothing.
- **`isa/riscv64/encode.mach`** — `push_pcrel_lo` states the fact at emission, carrying the paired high half's own addend forward so the two cannot drift apart. A low half with no `auipc` to pair with is refused. This also fixes mach's inline-asm `%pcrel_lo`, equally wrong for any pair with an instruction between the halves.
- **`of/elf.mach`** — `pair_riscv_pcrel_lo12` follows the psABI label at the boundary where that spelling is the only spelling, adopting the high half's **symbol index and addend** (so the symbol's own offset within its section resolves through the normal path like any other symbolic relocation).

## Testing

**New case `int/surface/riscv-pcrel-pair`**, riscv64 only. Reads four doubles and four longs through materialized bases, a pooled scale factor, and two values written back through the store spelling — **all by value**. Against unpatched `dev`:

```
-doubles n=4 125 250 375 500      +doubles n=4 0 0 0 0
-longs=47531                      +longs=0
-stored longs=121931 doubles=2500 +stored longs=0 doubles=0
-misaligned=0                     +misaligned=4
```

It is deliberately **non-trivial on every axis a pairing can get right by luck**, because `narrow-stack-args` — the case this was found through — is trivial on all of them. Under clang 18 that probe reaches exactly one pooled constant, `.LCPI1_0`, at section offset **0**, addend **0**, **4** bytes from its low half. Passing it proves the chase happens and nothing about what the chase carries. The new case covers:

| axis | coverage | why it matters |
|---|---|---|
| symbol offset | `scale`'s 12.5 is `.LCPI2_0` at **+8**, sharing a section whose **+0** holds `pcrel_doubles`' 1000.0 | a dropped offset resolves the +0 constant right by coincidence and only the +8 one wrong |
| high-half addend | the stores target `kl[2]`/`kd[3]`, which clang spells `kl + 10` and `kd + 18` on the HIGH half | a dropped addend is invisible whenever it is 0 |
| hi→lo gap | pairs at **4, 6 and 8** bytes | a fixed-gap rule is right for exactly one of these |
| both spellings | loads reach `_I`, stores reach `_S` | one resolution path, but a shared path is only covered where something exercises it |

Three emitted facts beside the values, each named for what it actually measures: `label-pairs` (the object still contains a psABI-spelled low half — clang folds a `const` array into immediates and the shape disappears); `misaligned` (no load in the whole image resolves off its own access width — **this is what catches the defect on the image alone**, reporting 4 and 8 pre-fix); and `text-relative` (none of those addresses is in .text — a true invariant that reads 0 both before and after this fix, kept because it fails closed for a pair resolved onto code, and reported as its own number rather than described as if it were doing the catching).

### On the `scale=0 0 0` that CI reported earlier

That was **not** an incomplete fix. It was this case's `mach.toml` still declaring `abi = "lp64"` while #2811 landed on `dev` and made `lp64` genuinely soft-float (`lp64d` is the hard-float one). CI builds the branch merged with `dev`, so it saw #2811 before my branch was rebased onto it. `scale` is the only observable here that passes a float **across** the mach→C boundary — the others pass a pointer and return integers — so it alone reports an ABI mismatch, in the same believable-zero shape. Demonstrated both ways: flipping only that one line reproduces `scale=0 0 0` with `doubles`/`longs` still correct, and flipping it back fixes it. The case now carries a comment saying exactly this, so the next person reads `abi =` before the linker. Independently ruled out against the incompleteness hypothesis by linking a real clang 18.1.3 object with the fixed compiler: `scale=25 100 1250`, resolving to the offset-**8** symbol and reading 12.5 from it.

### Verification

x86-64 Linux host, clang 22 and clang 18.1.3 (`ubuntu:24.04` container) both exercised, rebased onto `dev` at 76d671cb:

- `mach test .`: **1336 passed, 0 failed**
- `int --target linux-riscv64`: **all cases passed**, including `narrow-stack-args` (un-exempted, passing by executing) and the new case, both profiles
- `int --target linux-arm64 --runmode qemu`: all passed
- `int --target linux`: all passed **except** `surface/spirv-ext-inst` at both profiles, which fails identically on an unmodified checkout of this branch's base — a floating `mach-shader@0e538ca` no longer exporting `normalize`/`dot`/`mix4`. Unrelated to this PR, which touches no shader or SPIR-V code; flagged rather than folded in.
- `check-target-matrix.sh`: 522 pairs OK; `check-deps.sh`: OK
- **Self-host fixpoint**: stage2 == stage3, byte-identical, each into a freshly removed `out/`

## Landing notes

`int/surface/narrow-stack-args` is un-exempted here and needs `cc.sh`'s `-march=rv64gc -mabi=lp64d` pin, which is #2784's work **duplicated into this branch** so the un-exemption travels with what makes it green. Identical text; should resolve cleanly when #2784 rebases onto this.

`int/surface/c-variadic` stays exempt here deliberately. With this fix alone it still fails (`floats n=8 0 0 0`); with #2784's variadic classifier change on top it passes. That change is **independently justified** and its un-exemption belongs in #2784, proven there by executing.

## Found while here, filed not folded

**#2813** — every linked riscv64 image carries `e_flags = 0x0`, claiming a soft-float non-compressed ABI while its objects carry `0x5`. `llvm-objdump -d` therefore decodes compressed instructions in a linked image as `<unknown>`, which this PR's image scan works around with `--mattr=+c` (the workaround names the issue). Wrong metadata, not wrong execution; separate subsystem. ~1 hour.


octalide:
Merging with one failing check, justification recorded so this is auditable rather than a silent override.

`int linux` fails on `surface/spirv-ext-inst` only, at both profiles, with a **build** failure against `mach-shader@0e538ca`. That case declares `ref = "branch/main"` for mach-shader with no `mach.lock`, so it tracks another repository's moving main branch. mach-shader's main dropped `normalize`/`dot`/`mix4`, and the case now fails with zero mach changes.

Evidence it is unrelated to this PR:
- This PR touches no SPIR-V, shader, or frontend code. Its files are `riscv64/encode.mach`, `riscv64/reloc.mach`, `of/elf.mach`, `int/lib/cc.sh`, `int/lib/produce.sh`, and its own two cases.
- The identical failure is currently red on unrelated branches, including #2824.
- Two agents independently reproduced it against an unmodified checkout of this branch's base.

The riscv64 legs this PR exists to fix are **green**, including `surface/riscv-pcrel-pair` and the newly un-exempted `surface/narrow-stack-args`.

A pin for the floating dependency is in progress, along with an audit for other cases with the same shape. Until it lands, every PR carries this red — which is precisely why a test that can fail without a code change is worse than no test: it trains people to merge past red. Doing that here deliberately and in writing, once, rather than by habit.

Issue #2539, opened 2026-08-05
driver: mach build silently skips modules unreachable from the artifact entry, so a broken module can look green


octalide:
Split out of #2523, where this was a secondary note. It is independent of that typing bug and, unlike it, affects every project rather than one language corner.

## What

`mach build` compiles only the modules reachable from an artifact's `entry`. A module that sits under the project's `src` directory but is not imported by anything on that path is never compiled, so `mach build` reports success while the module is broken. `mach test` does compile it, because it collects `test` declarations across the whole project. The two commands therefore disagree about whether the project builds.

Confirmed directly while adding modules to mach-std (briar-systems/mach-std#417). `src/compress/inflate.mach` existed on disk and contained a genuine, hard compile error. With `src/lib/libstd.mach` unchanged:

```
$ mach build .
$                       # exit 0, silent
```

After adding one `fwd std.compress.inflate;` line to the artifact entry, the same source:

```
$ mach build .
error: constant typing: constant carries an incompatible type
```

Nothing about the module changed between those two runs — only its reachability from the entry. `mach test .` reported the error in both cases.

Why it is wrong: a green `mach build` is the signal developers act on, and it currently means "everything I happen to link is fine", not "the project is fine". The failure mode is silent and time-shifted. A module can be written, committed, and reviewed while `mach build` stays green, and the error surfaces later — at the unrelated moment someone first imports it, attributed to whoever wrote the import rather than whoever wrote the module. I hit exactly that: I wrote a module, built it clean, wrote more code against it, and only discovered the error much later when I wired it into the entry, at which point the failure looked like it belonged to the wiring change.

It also silently weakens the invariant "if it is in the tree, it compiles", which is the thing that makes a monorepo's `src` directory trustworthy. Dead-but-broken modules accumulate for free.

CI in mach-std happens to catch this because it runs `mach build .` and then `mach test .`. That is luck, not design: a project whose CI runs only `mach build`, or one whose broken module contains no `test` declarations, gets no signal at all from either command.

## Fix

The underlying tension is real — `build` should not have to generate code for modules nobody links, and skipping that work is a legitimate optimization. The mistake is conflating "do not emit code for it" with "do not check it".

Candidate designs:

[...]

Issue #2173, opened 2026-07-24
build: codegen/emit/link phases return success with errors on the diagnostic sink — fail-open by omission


octalide:
Found while fixing #2168. `run_lower_pass` in `src/lang/driver/passes.mach` returned `R.ok` while errors sat on the diagnostic sink — a rejected program lowered, linked and ran with its diagnostic merely printed. PR #2170 gates that one pass.

**The sibling phases in `src/lang/build/engine.mach` have the same omission:** `codegen_phase`, `emit_phase`, and `link_phase` do not check `diagnostic.has_errors` before returning success. `sema_phase` does.

## Why it matters

This is fail-open by omission, and #2168 is the proof that the class is exploitable rather than theoretical: the moment any routine downstream of sema reports a diagnostic and carries on instead of returning an error, the build proceeds and the user gets a binary built from a program the compiler rejected. Today the codegen constant-time checks return hard errors, so no live path is known to escape — but that is a property of the current callees, not of the phase contract, and it will silently stop holding the first time a codegen-side check is written in the report-and-continue style that monomorphization uses.

## Fix

Gate every phase on the sink the way `sema_phase` and (post-#2170) `run_lower_pass` do. Prefer making it structural rather than three more copies of the same `if` — a phase runner that checks the sink once after each phase cannot be forgotten by the next phase that gets added. Audit for any other phase or pass boundary with the same gap.

## Bar

A regression test per phase: a program whose only error is reported from inside that phase must fail the build, not produce an artifact. Verify no legitimate program starts failing because a non-fatal diagnostic sits on the sink.

Blocked on #2170 merging first (it touches `driver/passes.mach` and establishes the gate pattern).

PR #2826, opened 2026-08-08
build: make every number `mach build -v` prints a true claim

### LINKED ISSUE #2816: build: the `-v` phase readout times the wrong regions — `codegen` reports the adoption, `optimize` has no row
Part of #2288.

`mach build -v` reports a `codegen` row three orders of magnitude too small and prints no `optimize` row at all, because the readout's phase spans are opened around the wrong regions: they bracket bookkeeping while the parallel preludes that do the actual work run outside them.

## Repro

Clean self-build of the compiler, 16-core host, debug profile:

```
$ mach clean && mach build . -o /tmp/m -v
load      227 modules    456ms
resolve   227 modules    216ms
sema      227 modules    390ms
lower     222 modules     2.0s
codegen   222 modules    183us
link        1 binary     397ms
built /tmp/m  222 modules  6 MB  in 4.1s
```

The same build serialised, where the gap is impossible to read as noise:

```
$ mach clean && mach build . -o /tmp/m -v --jobs 1
load      227 modules    449ms
resolve   227 modules    218ms
sema      227 modules    407ms
lower     222 modules     2.6s
codegen   222 modules    223us
link        1 binary     377ms
built /tmp/m  222 modules  6 MB  in 8.0s
```

The rows sum to **4.05 s of an 8.0 s build**. Half the wall time of a single-threaded build has no row on it.

## Expected

Every phase the build ran gets a row, each row reports the wall time actually spent in that phase's work, and the rows account for the build's wall clock to within the plan-level overhead outside any phase.

## Actual

Two independent breakages, from two different commits:

**1. `optimize` has no row.** The phase id exists (`readout.PH_OPTIMIZE`, `readout.mach:34`) and `phase_label` renders it as `"optimize"` (`readout.mach:348`), but nothing has emitted it since `c785ca39` (2026-07-02, "bind Q_LOWER into the query engine") folded lowering and optimization into one `Q_LOWER` compute. Before that commit the row existed and was measured — it is the `optimize` row in #1825's audit table. The constant has been a dead phase for a month.

**2. `codegen` times the adoption, not the codegen.** `engine.codegen_phase` (`build/engine.mach:611`) runs `run_codegen_parallel` — the worker pool that actually generates every module's `ObjectImage` — and only *then* calls `driver.run_codegen_pass`, which opens `PH_CODEGEN` (`driver/passes.mach:1681`) around a loop that fetches already-staged images out of the query cache. The pool's work falls in the gap between `PH_LOWER`'s `phase_end` and `PH_CODEGEN`'s `phase_begin`, so it is attributed to nothing at all. This dates to `b558af4b` (2026-07-11, the parallel codegen pool).

The same shape is why `lower` is overstated: `run_lower_parallel` monomorphises, lowers, *and* optimizes inside `PH_LOWER`'s span (`driver/passes.mach:753`), so the missing `optimize` time is silently folded into the `lower` row rather than merely lost.

`223us` for 222 modules producing a 6 MB binary is not a number a reader discards as obviously broken — it is a number someone acts on. That is #2288's defect class exactly: the surface produces positive evidence about where a build's time goes, and it is wrong by orders of magnitude.

## Fix

Move the phase spans to where the work is dispatched.

- `build/engine.mach` owns the rows for the back half, since the engine is what dispatches the preludes. `lower_phase` brackets `PH_LOWER` and `PH_OPTIMIZE`; `codegen_phase` brackets `PH_CODEGEN` around the pool *and* the adoption pass. `run_lower_pass` and `run_codegen_pass` stop emitting rows.
- `lower` and `optimize` cannot both be plain wall spans: in the parallel path the two halves are separate regions of `run_lower_parallel`, but in the serial fallback they are fused per module inside `q_lower_compute`. So `readout` grows one carve-out primitive — `span_begin`/`span_end(ph, host_ph, start)` — that attributes a slice of wall to `ph` and deducts it from the enclosing phase `host_ph`. `PH_OPTIMIZE` is carved out of `PH_LOWER` at both call sites (the pool region, and the per-module `lower_optimize` call in `q_lower_compute`). The two rows are then exactly disjoint and together account for the whole of `lower_phase`.
- The `-vv` per-module `lower` items move to the prelude's serial half, where the lowering actually happens, instead of timing the adoption. `run_lower_one` records an item only for a module the prelude did not stage, so each module still contributes exactly one item.
- `driver/tests.mach`'s #2173 phase-gate injections construct `readout.Progress` literals by hand and must carry the new callback; the `PH_CODEGEN` `phase_end` injection point moves from `run_codegen_pass` to `codegen_phase` and its comment has to move with it.

`load` / `resolve` / `sema` / `link` stay where they are: each is one contiguous region inside its own pass, so its span is already honest.

## Acceptance

- A `-v` build of the compiler prints an `optimize` row, and the `codegen` row reports the pool's real time.
- The rows account for the build's wall clock. Asserted as a test, not as a readout: a test that merely checks `optimize` appears would pass against a row reading `0us`. The test must (a) drive a real multi-module build through the engine, (b) assert **every** phase the plan actually ran produced exactly one row, and (c) assert the rows sum to a substantial fraction of the measured wall rather than a token one — the failing observation here is a sum of 50 %, and a row of 223 us against seconds of work.
- **Emitted output byte-identical.** This is instrumentation and must not move a single emitted byte: a compiler built from the branch produces binaries byte-identical to one built from `dev`, across all three ELF ISAs at both profiles, and the self-host fixpoint holds (stage2 == stage3, each stage built into a freshly removed `out/`). If any byte moves, the spans are entangled with the work and the change is no longer instrumentation.
- #1803 and #1825 are told, on their own issues and in the PR, that their `lower` and `codegen` numbers were measured under the old attribution and are not comparable to the new rows.

## Environment

- `mach 4.16.0`, `dev` @ `da4cb95e`
- host: linux-x86_64, 16 cores
- both profiles; the readout is target-independent

octalide:
Closes #2816. Part of #2288.

## What was wrong

`mach build -v` reported `codegen 222 modules 343us` on a clean 227-module self-build producing a 6 MB binary, and printed no `optimize` row at all.

That first number is the point. It does not read as "codegen was not measured" — it reads as **"codegen is free"**, which is a conclusion someone acts on. An absent row invites suspicion; a confident wrong one does not. This is #2288's thesis in its purest form: the surface produced positive evidence for something false.

Two independent causes, from two different commits:

- **`codegen` timed the adoption, not the codegen.** `engine.codegen_phase` runs `run_codegen_parallel` — the pool that actually generates every module's `ObjectImage` — and only then calls `driver.run_codegen_pass`, which opened `PH_CODEGEN` around a loop that fetches already-staged images out of the cache. The pool's whole wall fell between `PH_LOWER`'s `phase_end` and `PH_CODEGEN`'s `phase_begin` and was attributed to nothing. From `b558af4b` (2026-07-11, the parallel codegen pool).
- **`optimize` had no row.** `readout.PH_OPTIMIZE` and its `"optimize"` label have existed since the readout landed, but nothing has emitted them since `c785ca39` (2026-07-02) folded lowering and optimization into one `Q_LOWER` compute. Before that commit the row existed — it is the `optimize` row in #1825's audit table. Its time has been silently inside the `lower` row for a month.

On a serialised build the gap is impossible to read as noise: the rows summed to **4.05 s of an 8.0 s wall**.

## The audit

The owner's position on scope: *"If we show a number, it better damn well be correct, 'conclusion' or not."* So this is an audit of every figure `mach build -v`/`-vv` prints, with a verdict for each — including the ones that were already right.

| number | before | verdict |
|---|---|---|
| `load` / `resolve` / `sema` row durations | wall of one contiguous pass region | **correct**, unchanged |
| `lower` row duration | wall of lower + optimize fused | **wrong** — overstated by the whole optimize phase. Fixed |
| `optimize` row | absent | **absent**. Added |
| `codegen` row duration | wall of the cache adoption | **wrong** by 3 orders of magnitude. Fixed |
| `emit` row | folded into `link` | **mislabelled** — `link 363ms` included writing 222 object files. Split out |
| `link` row duration | emit + link | **wrong** as labelled. Now the link alone |
| row counts (`227 modules`, `222 modules`, `1 binary`, …) | topo/emit set per phase | **correct** — the CLI always goes through cold `execute`, so every counted module really is processed |
| `-vv` `load` / `resolve` / `sema` items | per-module wall | **correct**, unchanged |
| `-vv` `lower` items | wall of the Q_LOWER adoption | **wrong** under the prelude. Moved to the work |
| `-vv` `optimize` items | absent | **absent**. Added |
| `-vv` `codegen` items | wall of the cache adoption — **185 items summing to 0.21 ms under a 640 ms row** | **wrong**. Moved to the work |
| `-vv` `emit` items | absent | **absent**. Added |
| `-vv` `link` item | the link's wall | **correct**, unchanged |
| `(slow)` marker | the max item | **correct** given correct items |
| items vs their row, under a pool | items sum to more than the row, unstated | **misleading whole**. The row now declares `(16 threads)` |
| rows vs the recap total | 65 %–98 %, no stated relationship | **misleading whole**. An `other` row makes it an exact partition |
| recap `in <total>` | measured from the invocation's start | **wrong on `--all-targets`** — unit 2 reported unit 1 + unit 2 as its own. Now per unit |
| recap module count | the emitted set (#2539) | **correct**, unchanged |
| recap size | rounded magnitude of the linked binary | **correct**, unchanged |
| recap `1 modules` | plural for a count of one | **grammar**. Fixed |
| the scalarization note's count | vector ops summed over modules | **correct**, unchanged |

Nothing was left approximate, and nothing is shown that could not be measured honestly. Two figures were considered for removal and kept because they are true as stated: the recap's rounded size, and the per-row module counts.

## The fix

Spans move to where the work is **dispatched**, which is the engine:

- `lower_phase` brackets `lower` **and** `optimize`; `codegen_phase` brackets the pool **and** the adoption.
- `run_lower_pass` / `run_codegen_pass` no longer emit rows. They emit the per-item events and the optimize carve-outs, at the places the work happens.

`lower` and `optimize` cannot both be plain wall spans — two halves of the prelude on the parallel path, one fused query compute on the serial one — so `readout` grows one primitive: `span_begin` / `span_end(ph, host, start)` attributes a slice to `ph` and deducts it from `host`. The two rows are then exactly disjoint and together cover the whole of `lower_phase`, bookkeeping included.

Four smaller seam changes fall out of the audit:

- An **item carries a `Duration`**, not a start instant. A worker can time its own module and the main thread replays the measurement after the join — the only place a pooled module's cost exists.
- **`phase_workers`** declares a phase's concurrency, and the row prints it. Under a pool the items deliberately sum to more than the row, and the output says so rather than letting a reader assume a partition.
- **`unit_begin`** starts the recap's clock per plan unit.
- The recap prints an **`other`** row — the unit's time no phase claimed — so rows + `other` = total exactly, and a region nobody times shows up here instead of vanishing.

`driver/tests.mach`'s #2173 phase-gate injections carry the new callbacks; the `PH_CODEGEN` injection point moved from `run_codegen_pass` to `codegen_phase` with the span, and its comment moved with it.

## Before and after

Same 227-module input, same host, `dev`-built compiler vs branch-built compiler.

**Default `--jobs` (16 threads):**

```
BEFORE                                AFTER
load      227 modules    460ms        load      227 modules    459ms
resolve   227 modules    226ms        resolve   227 modules    224ms
sema      227 modules    361ms        sema      227 modules    372ms
lower     222 modules     2.0s        lower     222 modules     1.9s
                                      optimize  222 modules     93ms  (16 threads)
codegen   222 modules    343us        codegen   222 modules    670ms  (16 threads)
                                      emit      222 objects    155ms
link        1 binary     363ms        link        1 binary     219ms
                                      other     -                1ms
built …  222 modules  6 MB  in 4.1s   built …  222 modules  6 MB  in 4.1s
```

**`--jobs 1`, where the old readout's gap was half the build:**

```
BEFORE                                AFTER
load      227 modules    452ms        load      227 modules    448ms
resolve   227 modules    222ms        resolve   227 modules    227ms
sema      227 modules    371ms        sema      227 modules    367ms
lower     222 modules     2.5s        lower     222 modules     1.9s
                                      optimize  222 modules    613ms
codegen   222 modules    177us        codegen   222 modules     3.7s
                                      emit      222 objects    150ms
link        1 binary     381ms        link        1 binary     205ms
                                      other     -                1ms
built …  222 modules  6 MB  in 7.7s   built …  222 modules  6 MB  in 7.7s
```

`codegen` goes from **177us to 3.7s** — a factor of 21,000 — and the rows go from covering 51 % of the wall to covering all of it. `optimize` appears for the first time since 2026-07-02.

## The historical numbers this invalidates

**#1803's and #1825's `lower` and `codegen` figures were measured under the old attribution and are not comparable to these rows.** Stated on both issues, not reinterpreted here.

Specifically:

- **#1803's audit table** (159 modules, serial, in #1825's body) predates `c785ca39`, so its `lower` / `optimize` split was real — the readout still emitted `optimize` then. Its `codegen 3.4s` also predates the pool, so it was a true serial measurement. **That table is sound as history**; what changed is that the tool that produced it stopped being able to reproduce it.
- **#1825's 2026-08-07 re-measurement** was taken with the broken readout. Its wall-time and CPU-utilisation figures (`6.7 s → 5.1 s`, `99 % → 221 %`) are wall-clock and `time(1)` measurements, not readout rows, so **they stand**. Its epic conclusion rests on those, so no conclusion in #1825 depended on the bogus attribution.
- No conclusion in either issue was drawn *from* the `codegen 231us` row — #1825 flagged it as a defect rather than reading it. The damage was potential, not realised.

## Testing

`mach.lang.readout:serial_build_rows_are_one_per_phase_and_bound_their_work` and `:pooled_build_rows_…` build a 16-module fixture through `execute_warm` and assert on `Readout.rows` / `.item_sum` / `.workers` — the **render record**, i.e. exactly what the printed lines carried, rather than re-deriving the arithmetic the test would then not be checking.

The assertions are structural, with no tolerance to tune:

1. every phase the unit runs prints **exactly one** row;
2. no row is zero and every phase recorded items;
3. **each row bounds the work its items describe** — a row times its declared thread count must cover the items' sum;
4. no row is wildly larger than the work under it (a span that swallowed a neighbour);
5. the rows plus `other` track the wall the test itself measured, and `other` is never negative.

Assertion 1 alone would pass against a row reading `0us`, which is exactly the trap this codebase keeps hitting — so it is not the test. Assertion 3 is: it is what the old attribution cannot satisfy, since the codegen row timed the adoption while the items describe the pool that built them.

**Mutation-tested.** Both mutations were applied to the fixed tree and the suite re-run:

| mutation | result |
|---|---|
| `PH_CODEGEN`'s span moved back to `run_codegen_pass` (the original defect) | `pooled_build_rows_…` **fails** |
| the optimize carve-out dropped, so `lower` and `optimize` double-count | **both** tests fail |

**Ran before marking ready:**

- `mach test .` — 1336 passed, 0 failed.
- `int/run.sh --target linux --deps pin` — all cases passed. (Unpinned/floating deps produce two `surface/spirv-ext-inst` build failures from `mach-shader` HEAD drift; identical on `dev` with a `dev`-built compiler, so pre-existing and unrelated.)
- **Byte-identity vs `dev`**: a compiler built from this branch and one built from `dev` compile the same source to byte-identical output — `linux-x86_64`, `linux-arm64`, `linux-riscv64` × `debug`, `release`, 223 files (222 objects + binary) identical in all six cells.
- **Self-host fixpoint**: stage2 == stage3 == stage4, byte-identical, each stage built into a freshly removed `out/`.

Not one emitted byte moves. This is instrumentation and stayed instrumentation.


octalide:
Follow-up commit `df1a1f60`: a latent wrong number in this branch's own mechanism, found reviewing it rather than by a test.

`ro_phase_end` inferred "this row is fed by carve-outs" from having received one. A carve-fed phase that received **none** — a warm build where every module's optimized IR came from the cache, so nothing was optimized — fell through to the wall between its own begin and end events. That wall spans the host's entire phase, so `optimize` would have reported the whole of `lower`'s time for work that did not happen. Same class of defect as the one this PR removes, one level in.

Not reachable from the CLI (`mach build` always goes through cold `execute`, so no module is ever warm-valid), but reachable through `execute_warm`, and reachable by omission the moment a second carve-fed phase is added — which is exactly the kind of contract gap that costs a rewrite later. The opener now declares the intent (`phase_carved`), so the fallback cannot be entered by accident and the empty case reports zero, which is true.

Re-verified on the final tree: `mach test .` 1336/0, byte-identity vs `dev` across all six ELF cells (223 files each), fixpoint stage2 == stage3 == stage4, `int --deps pin` green.

---

**Unrelated red leg, pre-existing and external.** `int linux` (the *unpinned* leg) fails on `surface/spirv-ext-inst` with 16 `no symbol ... exported by shader.math` errors. Not from this PR:

- `int pinned (linux)` — same case, `mach-shader@a058175` from `int/deps.lock` — **passes** here.
- Reproduced locally on `dev` with a `dev`-built compiler: identical 2 failures.
- `dev`'s own CI run `31234746933` at 02:20 today was green on this leg; the PR run at 05:53 is red. `mach-shader` HEAD moved to `0e538ca` in that window and dropped or renamed `shader.math`'s exports.

It will fail every PR opened from now until either `mach-shader` restores the exports or `int/surface/spirv-ext-inst` is updated to the new API. Which of those is right is a call for whoever owns the spirv track, so flagging rather than filing.

octalide:
Merged `dev` (`12d5e253`) in and re-established the whole bar against that base, since the byte-identity claim is only meaningful against current `dev`:

- **Byte-identical to `dev` @ `12d5e253`** — `linux-x86_64` / `linux-arm64` / `linux-riscv64` × `debug` / `release`, 223 files per cell, all six identical.
- **Fixpoint** — stage2 == stage3 == stage4, byte-identical, each into a freshly removed `out/`.
- `mach test .` — 1338 passed, 0 failed.
- `int/run.sh --target linux --deps pin` — all cases passed.

Re-measured on the new base, same source both sides:

```
BEFORE (dev 12d5e253)                 AFTER (this branch)
load      227 modules    457ms        load      227 modules    436ms
resolve   227 modules    228ms        resolve   227 modules    220ms
sema      227 modules    373ms        sema      227 modules    358ms
lower     222 modules     2.4s        lower     222 modules     1.9s
                                      optimize  222 modules    136ms  (16 threads)
codegen   222 modules    173us        codegen   222 modules    864ms  (16 threads)
                                      emit      222 objects    153ms
link        1 binary     368ms        link        1 binary     215ms
                                      other     -                1ms
built …  222 modules  6 MB  in 4.4s   built …  222 modules  6 MB  in 4.3s
```

and at `--jobs 1`, where nothing hides behind the pool:

```
BEFORE                                AFTER
lower     222 modules     2.5s        lower     222 modules     1.9s
                                      optimize  222 modules    813ms
codegen   222 modules    172us        codegen   222 modules     3.6s
```

`codegen` **172us → 3.6s**, a factor of 21,000, and the rows go from covering 48% of the wall to covering all of it.

The `int linux` leg is still red for the reason above — floating `mach-shader@0e538ca` dropped `shader.math`'s exports; `int pinned (linux)` is green, and `dev`'s own CI was green on that leg at 02:20 today before the dependency moved. Unrelated to this PR and will hit every PR until it is dealt with.

Issue #2592, opened 2026-08-07
int: no fixture pins mach-std, so any case (and the darwin release gate) can change verdict with zero mach changes


octalide:
## What

No `int/` fixture checks in a `mach.lock`. Every case declares

```toml
[dep.mach-std]
git = "https://github.com/briar-systems/mach-std"
ref = "branch/main"
```

and the harness resolves that ref **fresh at run time**, so each case builds against
whatever `mach-std/main` happens to be when the leg runs. The compiler in the same
checkout, by contrast, is pinned by the repo's own `mach.lock` (`a35a0c4` on current dev).

The harness already prints the resolved commit, which is where this is visible:

```
PASS surface/macho-abs-bind-x86_64 [linux/release] (mach-std@c500b97)
FAIL surface/macho-got-x86_64      [darwin-x86_64/debug] (mach-std@aa74840)
```

Two int-main runs hours apart resolved `c500b97` and `aa74840` for the same commit of this
repo.

## Why it matters

An integration case — and therefore the **darwin release gate** — can change verdict with
**zero changes in this repository**. Concretely:

- A red gate may be caused by a mach-std merge, not by anything in the release being cut.
- A bisect over mach commits is unsound, because the dependency moved underneath the
  window. Two builds of the same mach commit are not comparable if they straddle a
  mach-std merge.
- A green pre-merge run does not guarantee a green post-merge run of the same code.

This is not hypothetical: it produced a false lead during the v4.7.1 cut, where
"passed on 08-05, fails now" had to be checked against a floating dependency before the
real cause could be isolated.

Verified on current dev: `int/surface/macho-got-x86_64` and
`int/surface/macho-abs-bind-x86_64` contain no `mach.lock`; running `mach dep pull` in a
case directory reports `wrote mach.lock`, confirming none was tracked.

## Note on intent

Testing against current `mach-std/main` may well be deliberate — it catches downstream
breakage early. The defect is not "it floats", it is that **it floats silently and
irreproducibly**, including on a release gate, with no record in the repo of what was
actually tested.

## Possible resolutions

1. **Check in a `mach.lock` per case** and advance it deliberately. Fully reproducible;
   costs an explicit bump to pick up mach-std changes.
2. **One lock for the whole `int/` tree**, if the harness can share it — same benefit, one
   file to bump.
3. **Keep floating, but record it**: have the harness write the resolved commit into the
   run's artifacts/summary, and pin to the seed release for release-gate runs
[...]

PR #2829, opened 2026-08-08
fix(int): pin mach-shader instead of floating it, so spirv-ext-inst and spirv-shader-only test mach rather than mach-shader's main

### LINKED ISSUE #2827: int: spirv-ext-inst and spirv-shader-only floated mach-shader's branch/main despite int/deps.lock already pinning a working commit
mach-shader's 0.1.0 release dropped `shader.math`'s `normalize`/`length`/`distance` exports, and `int/surface/spirv-ext-inst` went red on every ordinary mach PR because it floats `branch/main` instead of using the working pin `int/deps.lock` already recorded.

## Repro
```
cd int
./run.sh --target linux --filter spirv-ext-inst <any from-source mach compiler>
```
against mach-shader's current `main` (commit `0e538caa6ff83bc7492cea1553fc0364381b28ae`, also tagged `0.1.0`).

## Expected
The case tests mach's own SPIR-V emission (GLSL.std.450 ext-inst selection and `OpDot`), which is what its golden actually asserts. It should not depend on whether an independent library's `main` branch still exports the same functions on any given day - that dependency is out of the PR's control entirely, unlike mach-std (the compiler's own runtime dependency, covered by the root `mach.lock` and deliberately floated in the PR lane per `ci.yml`'s own documented design, precedent set by #2592).

## Actual
```
error: no symbol `normalize` exported by `shader.math`
  --> ./src/main.mach:69:23
```
and similarly for `length` and `distance`. The case (and its sibling `spirv-shader-only`, which also declares `mach-shader`) both declared `ref = "branch/main"`, so the harness's default `--deps float` mode (used by every PR-lane CI run) resolved fresh against mach-shader's current `main` on every run, ignoring `int/deps.lock`'s existing pin (`a0581755753abb6d4ca99d2946b2b973af3d0ddf`) entirely - that lock is only consulted under `--deps pin`, the release-gate lane. `--deps pin` already passes against the locked commit; the default float lane is what broke.

This reproduces identically against a compiler built from unpatched `dev`, so it is not caused by any pending mach change - it is a property of the two cases' own dependency declaration, exactly the shape #2592 fixed for mach-std but left unaddressed for a dependency the root manifest does not cover at all.

## Fix
Both cases now pin `mach-shader` directly in their own `mach.toml` (`ref = "commit/a0581755753abb6d4ca99d2946b2b973af3d0ddf"`, the commit immediately before the `0.1.0` release, where the three exports still exist) rather than `branch/main`. `check-deps.sh` requires every case naming a dep to agree on its ref, so both cases had to move together even though `spirv-shader-only`'s own assertion (a link failure on a non-SPIR-V target) does not depend on which functions `shader.math` exports. `int/deps.lock` regenerated via `update-deps.sh` to match. Each case's `case.conf` documents the pin and notes that advancing it later may require updating `spirv-ext-inst`'s golden (and `mach-shader`'s own math module) together.

Audited every int case for the same shape: the whole suite declares exactly two distinct git dep names, `mach-std` (~46 cases, covered by the root `mach.lock`, deliberately floating per `ci.yml`'s own PR-lane design) and `mach-shader` (these two, now pinned). No other case has this exposure.

## Acceptance
- `check-target-matrix.sh` and `check-deps.sh` clean.
- Full `int/run.sh` on `linux` green under both default (`--deps float`) and `--deps pin`.
- `spirv-ext-inst` and `spirv-shader-only` resolve `mach-shader@a058175` under both modes, independent of what mach-shader's `main` currently is.

Related: #2592 (mach-std pin precedent), #2729 (mach-shader lock coverage gap in the pinned lane specifically - this issue is the same dependency's exposure in the float lane, which #2729 did not touch).

octalide:
## Summary
- `spirv-ext-inst`'s golden names three exports (`normalize`/`length`/`distance`) that mach-shader's `0.1.0` release dropped, so under the harness's default `--deps float` the case went red on every ordinary mach PR that touched none of it - the same shape #2592 fixed for `mach-std`, but for a dependency the root manifest does not cover at all.
- `int/deps.lock` already pinned a working commit (`a058175`), but the PR lane deliberately floats by default, so that pin was never consulted there.
- Both cases declaring `mach-shader` (`spirv-ext-inst`, `spirv-shader-only`) now pin it directly in their own `mach.toml` (`ref = "commit/a058175..."`, immediately before the `0.1.0` release) rather than `branch/main` - `check-deps.sh` requires every case naming a dep to agree on its ref, so both moved together even though `spirv-shader-only`'s own assertion does not depend on which functions `shader.math` exports.
- `int/deps.lock` regenerated via `update-deps.sh` to match.
- Audited every int case for the same shape: the whole suite declares exactly two distinct git dep names - `mach-std` (~46 cases, covered by the root `mach.lock`, deliberately floating per `ci.yml`'s own PR-lane design) and `mach-shader` (these two, now pinned). No other case has this exposure.
- Note for whoever deliberately advances this pin past `0.1.0`: `spirv-ext-inst`'s golden (and `mach-shader`'s own math module) may need updating together, or the case goes back to failing against a release that dropped what it asserts.

## Test plan
- [x] `check-target-matrix.sh` and `check-deps.sh` clean
- [x] `mach test .` - 1336/1336
- [x] Full `int/run.sh` on `linux`, default (`--deps float`) mode - all green, including `spirv-ext-inst`
- [x] Full `int/run.sh --deps pin` on `linux` (the release-gate lane) - all green

Closes #2827

### DIFF
--- a/int/deps.lock
+++ b/int/deps.lock
@@ -6,5 +6,5 @@
 
 [dep.mach-shader]
 url = "https://github.com/briar-systems/mach-shader"
-ref = "branch/main"
+ref = "commit/a0581755753abb6d4ca99d2946b2b973af3d0ddf"
 commit = "a0581755753abb6d4ca99d2946b2b973af3d0ddf"
--- a/int/surface/spirv-ext-inst/case.conf
+++ b/int/surface/spirv-ext-inst/case.conf
@@ -12,3 +12,15 @@
 run: spirv-shader
 targets: linux
 build-target: spirv
+
+# mach-shader is pinned in mach.toml (commit/, not branch/main), deliberately,
+# unlike every other int dependency (#2827). the golden here names three specific
+# exports - normalize, length, distance - and mach-shader's 0.1.0 release dropped
+# all three, so under the harness's default `--deps float` this case went red on
+# every ordinary mach PR that touched none of this: the exact shape #2592 fixed
+# for mach-std, applied to a dependency the root manifest does not cover. pinned
+# to a0581755753abb6d4ca99d2946b2b973af3d0ddf, the commit immediately before that
+# release, where the three exports still exist. if someone deliberately advances
+# this pin past 0.1.0, this case's golden (and mach-shader's own math module) may
+# need updating together - advancing it silently would reintroduce exactly the
+# gap this note exists to prevent.
--- a/int/surface/spirv-ext-inst/mach.toml
+++ b/int/surface/spirv-ext-inst/mach.toml
@@ -38,6 +38,13 @@ need = []
 # check is a call ACROSS A MODULE BOUNDARY into a real published library, which is
 # where the substitution has to happen, since a SPIR-V module is compiled per mach
 # module and a shader module may declare no linkage to call out through.
+#
+# pinned rather than branch/main (#2827): this case's own golden names specific
+# exports (normalize/length/distance) that mach-shader's 0.1.0 release dropped, so
+# floating turns this case red on an ordinary mach PR that touched none of this -
+# exactly the shape #2592 fixed for mach-std. see case.conf for the rationale and
+# spirv-shader-only's identical pin (check-deps.sh requires every case naming a dep
+# to agree on its ref).
 [dep.mach-shader]
 git = "https://github.com/briar-systems/mach-shader"
-ref = "branch/main"
+ref = "commit/a0581755753abb6d4ca99d2946b2b973af3d0ddf"
--- a/int/surface/spirv-shader-only/case.conf
+++ b/int/surface/spirv-shader-only/case.conf
@@ -14,3 +14,11 @@
 # decision until someone gives the library a body; this case is what makes it hold.
 run: build-fails
 targets: linux
+
+# mach-shader is pinned in mach.toml (commit/, not branch/main), matching
+# spirv-ext-inst (#2827): check-deps.sh requires every case naming a dep to agree
+# on its url and ref, and that sibling case's golden names exports mach-shader's
+# 0.1.0 release dropped, so it is pinned before that release. this case's own
+# assertion (a link failure on a non-SPIR-V target) does not depend on which
+# functions shader.math exports, so it would tolerate floating on its own - it is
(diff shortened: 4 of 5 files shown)


octalide:
Closing in favour of #2830, which fixes the same break but forwards rather than backwards. Two things here are worth keeping, and one is worth correcting.

**Kept.** Your audit is right and I re-ran it: the suite declares exactly two distinct *git* dep names, `mach-std` (covered by the root `mach.lock`, floating by `ci.yml`'s deliberate PR-lane design) and `mach-shader` (two cases). The other two dep names in `int/`, `prov` and `provider`, are `path` deps, so they carry no exposure. Also kept: your point that `check-deps.sh` forces both `mach-shader` cases to agree on a ref, so they move together regardless of whether `spirv-shader-only` cares.

**Corrected.** The body says the golden "names three exports (`normalize`/`length`/`distance`) that mach-shader's `0.1.0` release dropped". It does not. I renamed the call sites to the `0.1.0` spelling, ran the case against `0e538ca`, and the golden passed **unblessed**. The golden names the set, instruction number and operand count of every `OpExtInst` and `OpDot` - a callee rename moves none of those. That matters, because it removes the reason for the warning at the end of your body: advancing this pin does not drag the golden with it. If the golden ever *did* shift under a pure rename, that would mean the case was keying on the wrong thing.

**Why forwards.** `a058175` is the commit immediately before the `0.1.0` release, so pinning there freezes the case on a `shader.math` API that no longer exists at any published version. The case's stated premise is a cross-module call into a *real published library*, and a pre-release commit is the one thing that is not. #2830 instead follows the rename (`normalize` -> `normalize_4`; the scalar surface is unchanged, so only the vector call sites move) and pins to `tag/0.1.0`, which resolves to `0e538ca` - the same commit `branch/main` was already floating to, so `spirv-shader-only`'s resolution does not move at all.

That gets both halves of what you and I were each half-right about: deterministic CI, so an unrelated mach PR never goes red for this again, and a case that exercises a real release rather than a snapshot between two of them. Advancing the pin stays a deliberate act, which was your central point.

PR #2830, opened 2026-08-08
fix(int): follow mach-shader's width-carrying names in spirv-ext-inst

octalide:
Every open PR has been showing a red `int linux` leg that belongs to none of them:

```
FAIL surface/spirv-ext-inst [linux/debug] (mach-shader@0e538ca) (build)
    error: no symbol `normalize` exported by `shader.math`
    ... 16 errors / 0 warnings
```

## Cause

mach-shader `58e6444` ("f32x2 and f32x3 families, cross_3, and a width-carrying naming rule"), released as `0.1.0` in `0e538ca`, made the width part of the name: `normalize` became `normalize_4`, `mix4` became `mix_4`. The case pulls `ref = "branch/main"`, so it picked the rename up the moment it landed upstream.

The scalar surface (`max`, `sqrt`, `clamp`, `pow`, `smooth_step`, `fma`, `sin`, `radians`) kept its bare names, so only the vector call sites move.

## Pinning

The suite's pinning machinery already exists and works. `--deps float` is the **default on purpose** - `int/run.sh`'s own header says it "is what catches a downstream break early" - and here it did exactly that, on a first-party sibling repo, within a day of the upstream change. So the fix is to follow the rename, not to freeze the ref.

`int/deps.lock` advances to the released head (`a058175` -> `0e538ca`) via `int/lib/update-deps.sh`, so `--deps pin` and `--deps float` resolve to the same commit rather than disagreeing silently. The per-case `mach.lock` is untracked build output and is not part of this change.

## The golden did not move

Deliberately not reblessed, and it did not need to be. The case's observable is the emitted instruction number, set and operand count of every `OpExtInst` and `OpDot`. A callee rename changes none of those, so a golden that had shifted here would have meant the case was keying on the wrong thing. It held, which is a small piece of evidence that it tests what its header claims.

Two stale comment spellings (`sh.normalize(n)`, `` `abs4`/`sqrt4`/`floor4` ``) follow the code, and the note saying the maths library "gains a `cross` of its own separately" is updated - it has one now (`cross_3`), and the local `Cross` declaration stays local for the reason already given there.

## Verified

`int/run.sh --target linux --filter '*spirv-ext-inst*'` against the drifted upstream head:

```
PASS surface/spirv-ext-inst [linux/debug] (mach-shader@0e538ca)
PASS surface/spirv-ext-inst [linux/release] (mach-shader@0e538ca)
```

`check-deps.sh` reports 120 git dep declarations OK. No compiler source is touched.

### DIFF
--- a/int/deps.lock
+++ b/int/deps.lock
@@ -6,5 +6,5 @@
 
 [dep.mach-shader]
 url = "https://github.com/briar-systems/mach-shader"
-ref = "branch/main"
+ref = "tag/0.1.0"
 commit = "0e538caa6ff83bc7492cea1553fc0364381b28ae"
--- a/int/deps.lock
+++ b/int/deps.lock
@@ -7,4 +7,4 @@
 [dep.mach-shader]
 url = "https://github.com/briar-systems/mach-shader"
 ref = "branch/main"
-commit = "a0581755753abb6d4ca99d2946b2b973af3d0ddf"
+commit = "0e538caa6ff83bc7492cea1553fc0364381b28ae"
--- a/int/surface/spirv-ext-inst/mach.toml
+++ b/int/surface/spirv-ext-inst/mach.toml
@@ -40,4 +40,7 @@ need = []
 # module and a shader module may declare no linkage to call out through.
 [dep.mach-shader]
 git = "https://github.com/briar-systems/mach-shader"
-ref = "branch/main"
+# a release tag, not `branch/main`. the subject is mach's substitution, so the
+# library only has to be real and published, not current, and floating it made an
+# upstream rename turn every unrelated mach PR red. advancing this is deliberate.
+ref = "tag/0.1.0"
--- a/int/surface/spirv-ext-inst/src/main.mach
+++ b/int/surface/spirv-ext-inst/src/main.mach
@@ -7,7 +7,7 @@
 # call SPIR-V can make: each is one instruction, drawn from the GLSL.std.450
 # extended set or, for `dot`, from the core opcode space (#2688).
 #
-# So the subject here is the substitution. `sh.normalize(n)` is written as a call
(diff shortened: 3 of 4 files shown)


octalide:
Supersedes #2829, which fixed the same break by pinning **backwards** to `a058175`, the commit immediately before mach-shader's `0.1.0` release. Reasoning for the difference is on that PR. This branch now does both halves:

- follows the rename forward, so the case builds against `0.1.0`;
- pins `ref = "tag/0.1.0"` in both cases that declare `mach-shader`, so CI stops being hostage to a sibling repo's API churn.

`tag/0.1.0` resolves to `0e538ca`, which is the commit `branch/main` was already floating to, so `spirv-shader-only`'s resolution does not move. `check-deps.sh` requires both cases to agree on the ref, so they moved together.

This revises the "not a missing pin" framing in the body above. Floating **did** do its job here - it surfaced a first-party API change within a day - but the cost landed on every unrelated open PR, and for a dependency that now has real releases, a tag is the better place to take that signal. Advancing the pin is a deliberate act, and per the correction on #2829 it does not drag the golden with it.

Also carried over from #2829: the suite declares exactly two distinct git dep names, `mach-std` (root `mach.lock`, floating by `ci.yml`'s design) and `mach-shader` (now pinned). `prov` and `provider` are `path` deps. Re-verified here, so no other case has this exposure.

Closes #2827.

octalide:
Superseded by #2831, which removes the case instead. Repinning and renaming both worked, and neither answered why the case fetched a sibling repo over the network at all. It does not survive that question: positive `#[spirv_op]` substitution is already covered by `spirv-sampled-image` from a local declaration, the negative side by `spirv-op-refused`, and what remained was a name-to-number table check that belongs in a unit test.

Issue #2748, opened 2026-08-07
spirv: no module well-formedness gate, so recursion and a 15-parameter signature emit invalid modules with no diagnostic


octalide:
SPIR-V imposes structural rules on a module that mach's IR has no equivalent of, and the emitter checks none of them. Two are reachable from ordinary source today and both produce an invalid module with no diagnostic: a recursive call, and a signature of 15 or 16 parameters.

## Repro

**Recursion.** SPIR-V forbids a cycle in the static call graph outright — there is no stack and no recursion in the execution model.

```mach
#[builtin("position")] var position: f32x4;

fun fact(n: i32) i32 { if (n <= 1) { ret 1; } ret n * fact(n - 1); }

#[stage("vertex")]
fun v() { position = f32x4{fact(5)::f32, 0.0, 0.0, 1.0}; }

fun main() i32 { ret 0; }
```

```
$ mach build . --target spirv          # succeeds, no diagnostic
$ spirv-val --target-env vulkan1.3 out/int/bin/probe/main.spv
error: line 59: [VUID-StandaloneSpirv-None-04634] Entry points may not have a call
graph with cycles.
  %2 = OpFunction %void None %37
```

A library module (no `#[stage]`, so `Linkage` rather than entry points) with the same recursive function *passes* `spirv-val`, because the validator's cycle rule is scoped to entry points. It is still an unusable module: a consumer that links it into a pipeline gets the cycle then, far from the source.

**A 15- or 16-parameter signature.** Two limits in the emitter disagree by two.

```mach
fun f15(a0:i32,a1:i32,a2:i32,a3:i32,a4:i32,a5:i32,a6:i32,a7:i32,
        a8:i32,a9:i32,a10:i32,a11:i32,a12:i32,a13:i32,a14:i32) i32 { ret a0 + a14; }
fun main() i32 { ret 0; }
```

```
$ mach build . --target spirv          # succeeds, no diagnostic
$ spirv-val out/int/bin/probe/main.spv
error: line 10: Id is 0
```

| parameters | build | `spirv-val` |
|---|---|---|
| 14 | ok | valid |
| 15 | ok | **`Id is 0`** |
| 16 | ok | **`Id is 0`** |
| 17 | refused, "more than 16 parameters is not supported by the SPIR-V target" | — |

[...]

PR #2833, opened 2026-08-08
feat(spirv): emit the bodies a shader calls across a module boundary

### LINKED ISSUE #2823: spirv: a shader cannot call a function defined in another module
## What

A shader cannot call a function defined in another module. Any such call fails codegen:

```
error: a call to a function defined outside the module is not yet supported by the
SPIR-V target (MIR_CALL in vert_main, at ./src/mesh_vert.mach:37:16)
```

A function with a body in the *same* module works: it is emitted as an `OpFunction` and
called with `OpFunctionCall`, and the result validates. `mach-shader` works only because
none of its declarations are functions in this sense - each carries `#[spirv_op]` and is
substituted as an instruction at the call site, so there is no call to emit.

The consequence is that a shader project cannot factor anything with a body into a shared
module. Concretely, in briar-systems/boom:

```mach
# shaders/src/transform.mach
pub fun mul_point(c0: f32x4, c1: f32x4, c2: f32x4, c3: f32x4, p: f32x4) f32x4 {
    ret c0 * f32x4{p[0], p[0], p[0], p[0]}
      + c1 * f32x4{p[1], p[1], p[1], p[1]}
      + c2 * f32x4{p[2], p[2], p[2], p[2]}
      + c3 * f32x4{p[3], p[3], p[3], p[3]};
}
```

```mach
# shaders/src/mesh_vert.mach
use tf: shaders.transform;
...
position = tf.mul_point(m.mvp_0, m.mvp_1, m.mvp_2, m.mvp_3, p);   # fails here
```

Mach has no matrix type by design, so a 4x4 transform is written over vector columns.
That expression is needed by every vertex shader a project has, and today it must be
copied into each one. The same applies to any shared shader helper: a tone-map curve, a
noise function, a shadow lookup, a skinning palette blend.

## Why this is the wrong shape rather than a missing convenience

The SPIR-V module a shader compiles to is self-contained, so a cross-module call has no
link step to resolve it later; the callee has to be emitted into the same module. That is
a fine constraint - it just means the SPIR-V backend has to pull in the transitive set of
called functions rather than only the entry module's own, which is what it already does
for functions inside one module.

Nothing about the callee changes: it is the same MIR, emitted as the same `OpFunction`.
The gap is that the emitter only walks the entry module.

## Fix

When lowering to SPIR-V, collect the transitive closure of called functions across module
boundaries and emit each as an `OpFunction` in the single output module, mangled as it
already is within a module. Calls then lower to `OpFunctionCall` exactly as same-module
calls do today.

Two cases need a decision:

- A called function that is not reachable from any entry point should not be emitted, so
  importing a large shared module does not bloat every shader. This falls out of walking
  from the entries rather than over the module.
- A call to a function that is not expressible in SPIR-V (recursion, a pointer parameter
  in a storage class the shader cannot name) should keep failing, and should say which of
  those it is rather than reporting the current message.

## Acceptance

- A vertex shader calling `pub fun` in a second module of the same project builds for
  `spirv`, passes `spirv-val`, and the disassembly contains one `OpFunction` for the
  callee and an `OpFunctionCall` at the call site.
- The same call through a dependency (a `[dep.*]` module, not just a sibling file) builds.
- A `pub fun` in an imported module that no entry point reaches does not appear in the
  emitted module.
- Recursion across a module boundary still fails, with a diagnostic naming recursion.

octalide:
Closes #2823

## The shape of the fix

The issue says "the emitter only walks the entry module". It is narrower than that: `spirv/emit.mach:emit_module` was **given** one `ir.Module` and had no way to reach another, so widening it is a change to what the driver hands the whole-module back half, not a change to a loop.

`codegen.EmitModuleFn` now takes a `unit.Unit` (new `mach.lang.be.codegen.unit`): the module being compiled plus the modules it draws called bodies from, lowered. The split of responsibility is the point:

- **Membership is the driver's.** It owns the module graph. `passes.code_sources` walks a fixpoint over the symbols each member actually references — the middle end already emits a body-less `FN_FLAG_EXTERN` declaration only at a call site that needs one, so the extern declarations *are* the reference set. Closing over imports instead would pay for the whole import tree on every shader.
- **What is emitted from each member is the emitter's.** It owns the call graph.

`isa.ModuleEmitter` is the contract that grew; nothing is threaded through `RegMachine`. A register machine is handed the empty set and reads nothing from it — it has a linker, and lowering a module it will not emit would be waste.

Inside the emitter, every defined function in the unit is flattened into **one table**, and every id, reachability mark and call resolution keys on a table index. A same-module call and a cross-module call are then the same code path. The boundary stays visible in exactly one place, where it must: a callee's signature is read against the type table of the module that **defined** it — an `IrTypeId` means nothing against anyone else's table, and the referencing module carries only a signature-only shadow whose parameter list is empty. `ir_value_spv_type` / `global_spv_type` make "which table" explicit at every site rather than leaving it to a convention, because getting it wrong resolves silently to a different type of the same index.

## The stale comment

The comment above the refusal claimed `shader.math` "could not be called even if its body were emitted", because a shader module declares no linkage. **It does not hold, and it did not hold before either.** A function emitted as an `OpFunction` in the same module is named by its result id, not by a symbol; linkage is only about publishing a name *out* of the module. What is true of `shader.math` is something else entirely: every one of its declarations is `#[spirv_op]`, so its body is never emitted anywhere and its call sites become instructions. The comment is rewritten to say that.

Chasing it turned up the argument's *valid* corollary, which was **not** being honoured: a drawn-in body is emitted but must **never be exported**. Its owning module publishes that symbol from its own `.spv`, and a consumer linking both would meet two definitions of one name — a collision this module would have invented over a function it does not own. Fixed, and it is one of the six tests.

## The two decisions

**Reachability, not membership.** A shader module emits the closure of its **entry points**; a library module has no entry points, so its own functions are its roots and it publishes exactly what it did before. This falls out of walking from the entries, as the issue says. It also means a shader module now drops its *own* unreachable functions — one existing case (`spirv-unsupported-names-opcode`) relied on a refusal inside dead code and is repointed to reach its fixture from the entry, with `#[noinline]` so both profiles agree.

**Every whole-module rule runs over that same emitted set** — the #2748 recursion check and the per-stage interface rules — rather than over a second, quietly different one. That is what makes a cycle drawn in from another module the responsibility of the module that draws it in.

On recursion specifically: mach forbids import cycles, so a call cycle *spanning* two modules cannot be spelled at all. The only cross-module recursion that exists is a cycle living inside one module that another one reaches, and the fixture is built so the **importing** shader is what refuses it (the defining module has an entry point of its own and does not emit the recursive function) — the only arrangement where an emitter that never looked across the boundary would not have caught it first.

The residual refusals each say which they are instead of sharing one message:
- recursion, naming the chain **qualified by module** (`spvcase.helper.step -> spvcase.helper.step`);
- a foreign `ext` import — no mach module defines it and a self-contained module has nothing to bind at load time, so the fix is a `#[spirv_op]` mapping or a mach body, not "make it visible";
- a callee no module in the build defines, naming where the emitter looked.

## Verification

**In `mach test .`, over the emitted module in process.** `int/` is frozen, and rightly: an integration case earns its place only when the fact it asserts is owned by something outside this repo. A SPIR-V instruction sequence is ours and is produced in memory, so the assertions read the words off the `.spirv` section of the `ObjectImage` codegen returns. That localizes, needs no golden, and needs no spirv-tools on the runner.

A shader's internal helpers carry no name in the emitted module — no entry point, no export, nothing but an id that renumbers on any unrelated change — so the fixtures give **every function in the tree a distinct arity**, and an arity is then a name. "No five-parameter function in this module" is exactly "the unreachable helper did not come along", said in the only vocabulary the module has.

Six tests over a four-module fixture (root shader, sibling module, `[dep.*]` project of two modules):

| test | property |
|---|---|
| `spirv_emits_a_callee_from_another_module` | callee drawn in across a **sibling** and a `[dep.*]` boundary; a third module arriving **transitively**; every call resolves to a defined `OpFunction`; a shader exports nothing |
| `spirv_omits_what_no_entry_point_reaches` | the two unreachable `pub fun`s absent from the shader while their own modules still publish them; `main` dropped too |
| `spirv_shares_one_helper_between_two_entry_points` | two entry points, one copy of each helper, and both entries calling the **same id** |
| `spirv_does_not_export_a_drawn_in_body` | a library module that drew a body in carries it as internal — three functions, two exports |
| `spirv_refuses_recursion_drawn_in_from_another_module` | the defining module builds **clean**; the importing one refuses, naming the cycle qualified by module |
| `spirv_refuses_a_foreign_ext_call_by_name` | the foreign-import message, and that the superseded wording is gone |

**Each test is load-bearing**, verified by mutating the implementation it guards and confirming it and only it fails:

| mutation | fails |
|---|---|
| seed every root function instead of walking from entries | `omits_what_no_entry_point_reaches`, `refuses_recursion…` |
| export every emitted function (drop the `mi == root` guard) | `does_not_export_a_drawn_in_body` |
| root the cycle check on the root module instead of the emitted set | `refuses_recursion_drawn_in_from_another_module` |
| restore the pre-#2823 refusal for any non-root callee | the four positive tests |
| drop the foreign-import refusal | `refuses_a_foreign_ext_call_by_name` |

### Pre-fix evidence, run by hand

A compiler built from unpatched `dev` (`git archive origin/dev` → `mach build`), against the same fixtures the tests use. The tests themselves cannot be run there — they call `passes.code_sources`, which does not exist on `dev` — so the behaviour is demonstrated directly.

**Cross-module call.**
```
### BEFORE (unpatched dev)
error: a call to a function defined outside the module is not yet supported by the
SPIR-V target (MIR_CALL in curve, at ./dep/shlib/src/tone.mach:4:49)

### AFTER (this branch)
spirv-val: clean
OpFunction count: 6
OpFunctionCall count: 5
exports: 0
```
Six functions: two stages, one same-module helper, and three drawn in (the sibling's `scale`, the dep's `curve`, and `shlib.base.gain`, which the shader never names and which arrives only because the closure is a fixpoint). `main` and the two unreachable `pub fun`s are absent.

The library module that itself drew a body in:
```
OpDecorate %1 LinkageAttributes "shlib.tone.curve" Export
OpDecorate %2 LinkageAttributes "shlib.tone.never_used" Export
%1 = OpFunction %float None %5
%18 = OpFunctionCall %float %3 %15 %16 %17 %float_2
%2 = OpFunction %float None %21
%3 = OpFunction %float None %56        <- gain, drawn in, NOT exported
spirv-val: clean
```

**Recursion across a boundary.** The message is the same shape but comes from a different module, which is the whole point — before, the *defining* module refused it (unqualified `step`); now that module builds clean and the *importing* shader refuses it (qualified).
```
### BEFORE (unpatched dev)
error: this call graph is a cycle and the SPIR-V target cannot express it: step -> step. …

### AFTER (this branch)
error: this call graph is a cycle and the SPIR-V target cannot express it:
spvcase.helper.step -> spvcase.helper.step. …
```

**Foreign `ext` call.**
```
### BEFORE (unpatched dev)
error: a call to a function defined outside the module is not yet supported by the
SPIR-V target (MIR_CALL in vert_main, at ./src/main.mach:7:27)

### AFTER (this branch)
error: `host_sqrtf` is declared `ext`, and a SPIR-V module cannot call a foreign import:
the module a shader compiles to is self-contained, so there is no symbol to bind at load
time. give the operation a `#[spirv_op(set, name)]` mapping so the call becomes that
instruction, or write it in mach so its body can be emitted here
(MIR_CALL in vert_main, at ./src/main.mach:7:27)
```

### Suites

- `mach test .` — 1344 passed, 0 failed (+6)
- `int/run.sh --target linux` — all cases passed (float deps; green now that #2831 removed `spirv-ext-inst`)
- `int/run.sh --target linux --deps pin` — all cases passed
- self-host fixpoint — stage2 == stage3 byte-identical, each stage built into a freshly removed `out/`; stage2 re-runs both suites clean
- incremental correctness — `Q_CODEGEN` records a `Q_LOWER` edge on every module it draws a body from (serial and pooled paths both), verified by editing a helper and watching the importing shader's `.spv` change

### Diff hygiene

`git diff --diff-filter=D --name-only origin/dev...HEAD` is **empty**, and every file in the diff is one this change owns. `int/` is touched only by the `spirv-unsupported-names-opcode` repoint — `int/observability.md`, `int/run.sh` and `int/lib/produce.sh` are byte-identical to `dev`.

## Follow-up, filed not folded

#2843 — a shared module cannot declare an interface variable a drawing shader can use. Globals remain the root module's; a drawn-in function referencing one is refused rather than mis-emitted. The flat-table shape here extends to globals additively, and the issue says which three rules need deciding rather than merely moving.


octalide:
**CI: every leg green except `int linux`, which fails for a reason outside this PR.**

| leg | result |
|---|---|
| build / test / release / incremental, x86_64 + aarch64 + windows | pass |
| `int pinned (linux)` | **pass** — this is the leg that covers the three new cases |
| `int linux-arm64`, `int linux-riscv64`, `int windows` | pass |
| `int linux` (`--deps float`) | **fail**, on `surface/spirv-ext-inst` only |

`spirv-ext-inst` resolves `shader.math` at `ref = "branch/main"`, which tracks mach-shader's latest release. mach-shader released `0.1.0` (`0e538ca`) at 05:31Z today carrying a **breaking rename** — `feat: f32x2 and f32x3 families, cross_3, and a width-carrying naming rule` — so twelve symbols the fixture calls no longer exist:

```
error: no symbol `normalize` exported by `shader.math`      -> normalize_4
error: no symbol `length` exported by `shader.math`         -> length_4
error: no symbol `distance` exported by `shader.math`       -> distance_4
error: no symbol `abs4` / `sqrt4` / `max4` / `floor4` / `fract4` / `clamp4` / `mix4`
error: no symbol `dot` / `reflect`
```

This is not caused by anything here:

- the same failure reproduces on a compiler built from **unpatched `dev`** (`./int/run.sh --target linux --filter spirv-ext-inst <dev-build>` → identical `no symbol` errors);
- the same case passes on both compilers under `--deps pin` (`mach-shader@a058175`), which is why `int pinned (linux)` is green;
- `dev`'s last green run was 02:20Z, **before** the mach-shader release, so this PR is simply the first build to resolve the new one. `int linux` will be red on `dev` and on every open PR until the fixture is repointed.

The repair is mechanical — rename the twelve calls in `int/surface/spirv-ext-inst/src/main.mach` to the width-carrying names and re-bless the golden — but it belongs with whoever made the rename, and folding it into a boundary-change PR would mix two unrelated things and risk conflicting with their own follow-up. Flagged rather than taken.

octalide:
**Reworked for the `int/` freeze, and rebased.**

- The `spirv-calls` producer and the three new int cases are **gone**. `int/lib/produce.sh` is now byte-identical to `dev`.
- Every property they held moved into `mach test .` as six in-process checks over the emitted module, read off the `.spirv` section of the `ObjectImage` codegen returns. Each is pinned load-bearing by a mutation that breaks it and nothing else — the table is in the PR body.
- The pre-fix evidence is now a hand-run transcript in the PR body rather than a committed case. The tests cannot run against `dev` at all (they call `passes.code_sources`, which does not exist there), so the behaviour is demonstrated with a compiler built from `git archive origin/dev`.
- `spirv-unsupported-names-opcode` stays repointed — an existing case invalidated by the behaviour change, not a new one.

**On the `git add -A` hazard.** I had already rebased when your message landed, and I had used `git add -A` on the conflicts, so I checked immediately: `git diff --diff-filter=D --name-only origin/dev...HEAD` is **empty**, and `int/observability.md` is intact at 331 lines with its `int/run.sh` reference. Nothing was eaten. What I did find was leftover *untracked* harness dirt — `int/surface/spirv-ext-inst/` still existed on disk as `out/`, `dep/` and `mach.lock` that git could not remove when #2831 deleted the tracked files, which is why a bare `ls` still showed the directory. Removed; it was never staged.

I did hit the same class of hazard through a different door, worth recording: `git reset --soft origin/dev` to squash, after `dev` had moved again (#2837), leaves the *index* holding the old base's tree — so the commit silently reverted `of.mach`, `elf.mach`, `linker.mach`, `obj.mach`, `riscv64/register.mach` and `int/run.sh`. Caught by the same `git diff --name-only origin/dev...HEAD` check listing files I had never touched; recovered from the reflog and redone as a squash against my own parent followed by a real rebase. **`--soft` onto a moved remote ref is as dangerous as `add -A`, and the file-list check catches both.**

Final state: no deletions, every file in the diff owned by this change, `int/` touched only by the repoint. `mach test .` 1344/0, both int legs green, self-host fixpoint byte-identical.

Follow-up #2843 filed for interface globals in a shared module, not folded in.

Issue #2778, opened 2026-08-08
target(riscv32): RV32 is not supported, and it is the first target that would catch a baked-in 64-bit assumption


octalide:
> ## Status (2026-08-08)
>
> Survey complete on `feat/2778-riscv32`; no code committed yet. Three findings settle the shape of the work.
>
> **The encoder EXTENDS, it does not fork.** Settled, high confidence. `isa/riscv64/encode.mach` was written width-honest: the instruction-word machinery (`pack_r/i/s/b/u/j`), register fields, immediate bit-scatter, and branch/jump fixups carry no XLEN assumption. The RV64-specific choices are four localized branches — `alu_opcode(width)` (`:1040`, `width == 4` selects the `*W` word ALU, which on RV32 *is* the full register and must select `OPC_OP`), `emit_li32` (`:1101`, emits `addiw` which RV32I lacks), `mem_width_funct3` (`:1073`, `width == 8` → `ld`/`sd`), and the shift-amount field being 6 bits on RV64 and 5 on RV32. `EncodeFn` already receives `*resolved.Target`, so the encoder can read `tgt.model.gpr_width` today; it is not blocked on #2818.
>
> **The #2777 `AbiVTable` seam admits a 32-bit member cleanly.** `xlen_bytes` / `max_reg_ret` are read from the descriptor, and `abi/riscv.mach` already contains `descriptor_facts_are_read`, which builds a live `ilp32d` descriptor and asserts the 8-byte by-reference threshold and the two-word aggregate pair. Adding `ilp32` is a descriptor, three thunks, and a `register` call — no classifier edit. One correction needed: `callee_saved_v` (~`:550-560`) writes the float bank at `size = 8` while deriving the integer bank from `xlen_bytes(v)`. That 8 is the FLOAT register width, correct today only because all three shipped members are FLEN 64, and wrong for `ilp32f`.
>
[...]

PR #2836, opened 2026-08-08
codegen: wide integer divide and remainder in legalize

octalide:
Part of #2778. Prerequisite for the RV32 bring-up, but target-blind and useful on its own.

## Why

`be/codegen/legalize.mach` splits any scalar operation wider than the target's `max_alu_width` into native-width lanes. It handles add/sub carry chains, bitwise, shifts, multiply and comparisons. It does **not** handle divide or remainder — `legalize.mach:2388` rejects them outright:

```
"legalize: wide divide is unsupported on this target"
```

That is not a corner case. `mach-std`'s `format.write_u64` reaches every printed integer through `x % 10` and `x / 10` on a `u64`. On any target whose ALU is narrower than 64 bits that is a multi-lane operation, so **printing any number is a compile error**. It blocks `riscv32`/`ilp32` (#2778) before the first line of execution evidence, and it blocks mos6502 the same way today.

There is no compiler-rt equivalent in mach for the backend to call, so this must be an expansion in the pass, in the same shape as the multiply and the shift barrel already there.

## What

Restoring division over the operation's own width, straight-line, built from the primitives already in the file — the shift-by-one template, the carry/borrow chain, the ordered wide relation, and the mask-select the multiply uses. `MIR_DIV_S` / `MIR_DIV_U` / `MIR_MOD_S` / `MIR_MOD_U` are one expansion selecting a different result; signed is a branchless abs, the unsigned divide, and a sign fixup.

Cost lands in the same order as `expand_mul` already accepts, and the comment will say so plainly rather than imply it is cheap.

## Status

- [x] `emit_wide_rel` / `wide_rel_pieces` extracted so the divide's `rem >=u divisor` test uses the same lexicographic walk a wide `<` does, rather than a second copy of it
- [ ] the restoring-division expansion and its exact piece count
- [ ] signed divide and both remainder forms
- [ ] `mach test .` assertions against hand-computed 64-bit quotients and remainders
- [ ] before/after transcript demonstrating the failure against the unpatched compiler
- [ ] byte-identity across `linux-x86_64` / `linux-arm64` / `linux-riscv64` at debug and release
- [ ] self-host fixpoint, each stage into a freshly removed `out/`

## Acceptance

- Wide divide and remainder expand correctly at 2+ lanes, asserted against **hand-computed** values — not against mach's own 64-bit output, which is the self-consistency trap that hid #2797.
- Signed and unsigned, including the sign-fixup cases and division by a value larger than the dividend.
- No emitted byte changes on any shipped target. The pass is inert where `max_alu_width = 8`: `function_needs_legalize` pre-scans and returns having touched nothing, so this is structural rather than hoped for.
- `mach test .` green.
- A case that fails against the unpatched compiler, run by hand and pasted here. `int/` is frozen, so no integration case is added.

### DIFF
--- a/src/lang/be/codegen/legalize.mach
+++ b/src/lang/be/codegen/legalize.mach
@@ -681,6 +681,10 @@ fun expansion_count(alloc: *A.Allocator, plan: *LanePlan, mi: *mir.MirInstr) R.R
         ret R.ok[u32, str](dst_lane_count(plan, mi));
     }
     if (op == mir.MIR_MUL) { ret mul_piece_count(alloc, plan, mi); }
+    if (op == mir.MIR_DIV_S || op == mir.MIR_DIV_U ||
+        op == mir.MIR_REM_S || op == mir.MIR_REM_U) {
+        ret divmod_piece_count(alloc, plan, mi);
+    }
     if (mir.is_cmp_opcode(op)) { ret cmp_piece_count(alloc, plan, mi); }
     ret R.err[u32, str](unsupported_message(alloc, op, mi.width));
 }
@@ -885,6 +889,10 @@ fun expand_instr(alloc: *A.Allocator, plan: *LanePlan, f: *mir.MirFunction, mi: 
     }
     if (op == mir.MIR_MUL) {
         ret expand_mul(alloc, plan, f, mi, dst, cursor);
+    }
+    if (op == mir.MIR_DIV_S || op == mir.MIR_DIV_U ||
+        op == mir.MIR_REM_S || op == mir.MIR_REM_U) {
+        ret expand_divmod(alloc, plan, f, mi, dst, cursor);
     }
     if (mir.is_cmp_opcode(op)) {
         ret expand_cmp(alloc, plan, f, mi, dst, cursor);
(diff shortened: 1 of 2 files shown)


octalide:
## Hand-computed test vectors (independent of mach)

Computed in Python's arbitrary-precision integers, reduced to two's-complement 64-bit. Nothing here consults a mach binary or mach output. These go into the tests as literals when the file frees up.

**Unsigned (u64)**

| case | dividend | divisor | quotient | remainder |
|---|---|---|---|---|
| exact, remainder zero | `0x3E8` | `0x8` | `0x7D` | `0x0` |
| divisor larger than dividend | `0x7` | `0x3E8` | `0x0` | `0x7` |
| divide by one | `0xDEADBEEFCAFEF00D` | `0x1` | `0xDEADBEEFCAFEF00D` | `0x0` |
| both operands wide | `0xFFFFFFFFFFFFFFFF` | `0xFFFFFFFF` | `0x100000001` | `0x0` |
| crosses the lane boundary | `0x100000000` | `0x3` | `0x55555555` | `0x1` |
| max over max | `0xFFFFFFFFFFFFFFFF` | `0xFFFFFFFFFFFFFFFF` | `0x1` | `0x0` |
| the `write_u64` step | `0xFFFFFFFFFFFFFFFF` | `0xA` | `0x1999999999999999` | `0x5` |
| dividend zero | `0x0` | `0x3039` | `0x0` | `0x0` |

**Signed (i64), truncating — quotient toward zero, remainder takes the dividend's sign**

| case | dividend | divisor | quotient | remainder |
|---|---|---|---|---|
| pos / pos | 1000 | 7 | 142 | 6 |
| neg / pos, signs disagree | -1000 | 7 | -142 | -6 |
| pos / neg, signs disagree | 1000 | -7 | -142 | 6 |
| neg / neg | -1000 | -7 | 142 | -6 |
| exact negative, remainder 0 | -1000 | 8 | -125 | 0 |
| `INT_MIN / 1` | `0x8000000000000000` | 1 | `0x8000000000000000` | 0 |
| `INT_MAX / -1` | `0x7FFFFFFFFFFFFFFF` | -1 | `0x8000000000000001` | 0 |
| small neg / large pos | -7 | 1000 | 0 | -7 |

**`INT_MIN / -1`** is the two's-complement overflow: the true quotient `2^63` is unrepresentable. RISC-V's `div` defines the result as `INT_MIN` with remainder 0, and that is what this expansion produces — `abs(INT_MIN)` is `INT_MIN`, `abs(-1)` is 1, the unsigned core returns `0x8000000000000000`, and the two sign words cancel so no negation is applied. Worth stating because it falls out of the algorithm rather than being special-cased, and a reader checking the edge case should know it was verified rather than assumed.

**Division by zero is deliberately not normalized.** The core drives the divisor through the same compare either way and yields an all-ones quotient with a wrapped remainder, where RISC-V's hardware defines an all-ones quotient and the *dividend* as remainder. Matching that exactly would cost a post-loop select on every divide to service an input the language does not define. Documented in the expansion's comment rather than paid for.

octalide:
## Rebased onto `c6a3f64d` (#2841), and divide-by-zero is now normalized

**Rebase.** Clean onto #2841. Verified nothing was lost:

```
$ git diff --diff-filter=D --name-only origin/dev...HEAD
(empty)
$ git diff --stat origin/dev...HEAD
 src/lang/be/codegen/legalize.mach | 1 file changed
```

No `git add -A` anywhere; paths staged by name. Pushed with `--force-with-lease` rather than `--force`, since the rebase rewrote history and the lease is what makes a concurrent push refuse instead of vanish. mword's `instr_is_vector` → `instr_rides_fp_bank` rename landed under my dispatch arms without conflict, and it strictly helps: a wide float divide can no longer reach them at all.

## Divide by zero: normalized

Reversed the earlier "documented rather than paid for" call. The reason is not that the language demands a value — it doesn't — but that **riscv64 would otherwise disagree with itself**. A native-width divide there is a hardware `div` defining an all-ones quotient and the dividend as remainder; a wide one came through this expansion and produced an all-ones quotient with a *wrapped* remainder. Same target, same source-level operator, two different answers selected by a width the programmer never chose and which is invisible at the call site.

`emit_divmod_zero_fixup`: an OR reduction over the divisor's lanes, one zero test, one mask, and a per-lane branchless select — the same xor/and/xor the multiply's accumulate uses. Cost lands only on the wide path, which is already the thousand-piece one.

Two details worth flagging for review:

**It runs after the sign fixup, not before.** A divide by zero yields `-1` whatever the dividend's sign, so applying the sign afterwards would turn it into `1` on a negative dividend. Getting that order wrong would be invisible in unsigned tests.

**No per-signedness case.** The all-ones quotient is `-1` at every width, so the signed and unsigned hardware definitions agree on the bit pattern.

The per-lane select is also what writes the destination now, so the separate copy-out is gone and the piece count dropped by `nl` accordingly.

## Teeth check on the new assertion

Per the standing rule that a test passing either way is maintenance cost with no teeth, I broke the fixup deliberately and confirmed the test catches it — changed the zero test from `CMP_EQ acc, 0` to `CMP_NE acc, 1`:

```
failures:
  mach.lang.be.codegen.legalize:expands_a_wide_divide  (exit 1)
0 passed, 1 failed, 1 total
```

Restored, and `mach test .` is **1345 passed, 0 failed**.

## Still outstanding before ready

- hand-computed arithmetic executed on the reference 6502 core in `driver/tests.mach` (mword's file, now free) — the vectors are posted above
- byte-identity across `linux-x86_64` / `linux-arm64` / `linux-riscv64` at debug and release
- self-host fixpoint, each stage into a freshly removed `out/`

octalide:
## Everything the PR claims has now been run. Rebased onto `5e772dee` (#2833).

Rebase verified — nothing lost, one file touched:

```
$ git diff --diff-filter=D --name-only origin/dev...HEAD
(empty)
```

Paths staged by name throughout; `--force-with-lease`, never `--force`.

## 1. Execution against hand-computed values

New test `mach.lang.driver:wide_divide_executes_on_a_6502_core_mos6502` runs real emitted bytes on the reference 6502 core. **16 vectors × 4 opcodes = 64 independent checks**, all green.

The expected values were computed in Python's arbitrary-precision integers and written as literals. Nothing in the test consults mach's own arithmetic — checking the wide expansion against the native path it replaces would have both sides agree while both were wrong.

Vectors cover what a restoring loop actually gets wrong: divisor larger than dividend (no subtraction ever fires), quotient crossing a lane boundary, exact division, all four sign combinations **including the two where quotient and remainder signs disagree**, `INT_MIN / 1`, and `INT_MIN / -1`.

**One correction I owe you.** I first wrote that the probes run at two lanes. They do not. The probes are declared `u16`, so two bytes cross the ABI, but the divide promotes to four bytes and `alu_min_width` is 1 on mos6502 — so it is a **four-lane** expansion. The unpatched compiler's own refusal names the width it saw:

```
legalize: wide divide is unsupported on this target (MIR_DIV_U, 4 bytes wide)
```

Four lanes is not riscv32's shape (`i64` there is two lanes on a four-byte ALU) and is the more demanding of the two. I had written a comment asserting a lane count I had not verified — the same hazard I flagged in `rejects_unsupported_wide_op`, in my own new code. Corrected, with the evidence quoted in the comment.

**Eight lanes does not fit and that is not a defect.** One stage per bit makes a 64-bit divide on one-byte lanes 64 stages of an eight-lane borrow chain — past the 60,928 bytes between `UT_MOS_CODE` and `UT_MOS_ARGS`. The 64-bit-on-32-bit case belongs to #2778, where an rv32 image runs it natively.

## 2. Before / after against an unpatched compiler

Same probe module, two stage compilers built from `5e772dee` and from this branch:

```
=== BEFORE (dev @ 5e772dee, unpatched) ===
error: legalize: wide divide is unsupported on this target (MIR_DIV_U, 4 bytes wide)

=== AFTER (#2836) ===
error: a flat-image object format produces only executables; ... '--emit obj' ... unsupported
```

The refusal is gone; the remaining message is raw-format's unrelated `--emit obj` restriction.

## 3. Byte-identity — measured, not argued

Two compilers, same tree, object bytes hashed:

| target | profile | result |
|---|---|---|
| linux-x86_64 | debug | **IDENTICAL** |
| linux-x86_64 | release | **IDENTICAL** |
| linux-arm64 | debug | **IDENTICAL** |
| linux-arm64 | release | **IDENTICAL** |
| linux-riscv64 | debug | **IDENTICAL** |
| linux-riscv64 | release | **IDENTICAL** |

## 4. Self-host fixpoint — fresh `out/` per stage

```
stage2 built by stage1  cffa0b76...
stage3 built by stage2  cffa0b76...
stage4 built by stage3  cffa0b76...

FIXPOINT: stage3 == stage4
(stage2 == stage3 as well)
```

Each stage into a freshly removed `out/`. stage2 also matches, since stage1 was built from this same tree rather than a stale one.

## 5. `mach test .`

**1355 passed, 0 failed.**

Teeth confirmed on the structural assertion: breaking the zero test (`CMP_EQ acc, 0` → `CMP_NE acc, 1`) fails `expands_a_wide_divide`, restored after.

Marking ready.

Issue #2874, opened 2026-08-08
test(riscv32): no float path is executed by any test


octalide:
No float path on riscv32 is executed by any test.

## What exists now

The reference RV32IM core added for #2778 models RV32I plus M and **refuses anything else by name**, floats included. That refusal is deliberate — a harness that skipped what it did not model would report a pass for a program it never really ran — but it means the executed sweep covers integer work only.

rv32 codegen does select float instructions. A histogram of a broad rv32 corpus shows `fld`, `fsd`, `fadd.d`, `fsub.s`, `fmul.d`, `fdiv.d`, `fmv.w.x`, `fcvt.w.d`, `fcvt.s.d`, `fcvt.d.w`, `fcvt.d.s`. None of them has ever been executed.

## Why this matters more on rv32 than elsewhere

This is the same gap that was just closed for integers, one extension over, and rv32 is where it is most likely to bite:

- **FLEN and XLEN differ.** rv32d has 64-bit float registers on a 32-bit machine, which is the exact asymmetry that produced the `callee_saved_v` and `frame_reserved_top` findings under #2778 — a frame stride that was one literal serving three different widths.
- **The float-bank gate in `be.codegen.legalize` is load-bearing here.** An `f64` on rv32 must NOT be split into integer lanes; it is native at `lane_width` 4 because it rides the float bank. That gate (#2818) is the only thing standing between a correct `f64` and a silent miscompile, and nothing executes it.
- **The soft-float members are untested at their own convention.** `ilp32` and `ilp32f` pass an `f64` in an integer register pair — a path #2870 just had to fix for integers and which has an independent classifier arm for floats.
- **`fcvt` between a 64-bit float and a 64-bit integer is RV64-only**, which is why `legalize` refuses that conversion by name on rv32 (#2860). Nothing checks that the refusal fires where it should and not where it should not.

Four integer defects sat in this target through a green suite because nothing ran the emitted bytes. The float half is currently in exactly that state.

## Shape of the work

Extend `isa.riscv.refcore` with the F and D extensions and add float probes to the executed sweep. Specifically:

- The float register file (32 × 64-bit), `fld` / `fsd` / `flw` / `fsw`, the arithmetic ops, the compares, and the `fcvt` / `fmv` families rv32 actually selects.
[...]

PR #2905, opened 2026-08-08
feat(riscv32): open-code the conversion between a 64-bit integer and a float

### LINKED ISSUE #2860: target(riscv32): RV32 has no 64-bit integer to float conversion, and legalize has no arm for it
Depends on #2778.

## What

RV32 has no instruction that moves a 64-bit integer between the integer and float banks, and none that converts one. `fcvt.l.d`, `fcvt.lu.d`, `fcvt.d.l`, `fcvt.d.lu` and the `fmv.x.d` / `fmv.d.x` pair are all RV64-only: the `L` suffix names a 64-bit GP operand, and on RV32 no GP register is that wide. The RV32 target added by #2778 therefore has no encoding for `i64 -> f64`, `f64 -> i64`, or their unsigned twins, and none for a 64-bit bitcast across the banks.

**The shared lowering does not cover it either.** `be/codegen/legalize.mach` splits an integer value wider than `MachineModel.max_alu_width` into native-width lanes and threads carries between them, which is what makes `i64` arithmetic work on a 4-byte ALU. It has no arm for `MIR_SI_TO_FP`, `MIR_UI_TO_FP`, `MIR_FP_TO_SI` or `MIR_FP_TO_UI` - grep the file and the opcodes do not appear. A lane-split `i64` reaching a conversion is a GP-class pair with no rule to reassemble it into an operand the float unit can read, so the conversion arrives at the encoder whole and unencodable.

The conversion is also not a lane-wise operation, so it could not be one: `(hi, lo) -> f64` is a single rounding of the combined 64-bit value, not two independent conversions combined afterwards. The carry-threading model the pass is built on does not apply.

**What #2778 does instead.** The RV32 encoder refuses these four conversions with a loud error naming the operation and the target, rather than emitting an RV64 opcode that the hardware traps on. That keeps "rv32 works" and "rv32 is complete" able to fail separately, which is the same reason widening multiply is its own issue.

## Fix

Two candidate designs.

**A. Soft-float helper calls.** Lower the four conversions to calls into the compiler runtime (`__floatdidf`, `__floatundidf`, `__fixdfdi`, `__fixunsdfdi`, and the `sf` single-precision twins), which is what every C toolchain does on RV32. Cheap to state, correct by construction, and it introduces a runtime dependency the mach backend does not otherwise have on this target - which is the real cost, since a freestanding RV32 image would then need those symbols provided.

**B. Open-coded expansion.** Emit the conversion inline from the lane pair using the float unit's 32-bit conversions plus a scale and add, with the rounding and the unsigned/signed edge cases handled explicitly. No runtime dependency, and it keeps a freestanding image self-contained, but the rounding behaviour has to be got exactly right at the boundaries (values above 2^53, the signed minimum, and the unsigned range above 2^63) or it is a silent wrong answer rather than a failure.

**Preferred: B**, with A as the fallback if the edge cases prove not to be worth the code. mach's RISC-V targets are the ones most likely to be used freestanding, and a conversion that only works when a libgcc-shaped runtime is linked is a conversion that does not work on the target that most wanted it. The decision should be recorded in the code either way, since the tradeoff is the interesting part.

## Acceptance

- `i64 -> f64`, `u64 -> f64`, `f64 -> i64` and `f64 -> u64` compile on riscv32 and produce the same values as the riscv64 target for a shared vector of inputs, including 0, 1, -1, the signed minimum, `2^53`, `2^53 + 1`, `2^63 - 1`, and `2^64 - 1`.
- The f32 twins are covered by the same test vector.
- The refusal added by #2778 is removed, and no path can reach an RV64-only `fcvt` opcode on an RV32 target.
- Unit tests only. The integration-test freeze holds and nothing here needs a runner to observe: the encoder's bytes and the computed values are both checkable locally.

### LINKED ISSUE #2904: target(riscv32): a 64-bit cross-bank bitcast has no lowering, and refuses through the generic width message
Refs #2860.

A 64-bit reinterpret across the register banks (`f64 :~ i64`, `i64 :~ f64`) does not compile on riscv32, and the diagnostic names neither the cause nor the instruction that is missing.

## Repro

```
[target.rv32]
isa = "riscv32"
os = "freestanding"
abi = "ilp32d"
of = "elf"
```

```mach
fun bits(x: f64) i64 { ret x :~ i64; }
```

```
mach build . --target rv32
```

## Expected

Either the reinterpret is lowered, or it is refused by a message naming the reinterpret and the reason.

RV32 has no instruction for it. `fmv.x.d` and `fmv.d.x` move a 64-bit value between the float bank and a GP register, and no RV32 GP register is that wide, so the pair is RV64-only for exactly the reason `fcvt.d.l` is (#2860). The value can still be moved: a bit reinterpret through a frame slot is an `fsd` followed by two `lw`, or two `sw` followed by an `fld`, which is what a C toolchain emits on this target and what the `f32` twin already does when its float side is spilled.

## Actual

```
error: legalize: operation wider than the target ALU is unsupported (MIR_MOV, 8 bytes wide)
```

The reinterpret reaches `be.codegen.legalize` as a `MIR_BITCAST` that `mir.lower` has retagged, `instr_rides_fp_bank` excludes it from the lane split because one side is an FP-class vreg, and it falls out of the rewrite into the generic width refusal. The message names `MIR_MOV`, which is not what the user wrote, and gives no indication that the missing thing is a cross-bank transfer at a width this machine has no register for.

Two things are wrong and they are worth separating.

**The lowering is missing.** The transfer has an obvious realization on this target through a frame slot, and it is the realization the encoder already reaches for when the float side is a spill slot: `encode_fmov` handles a cross-bank bitcast whose float side is a `MIR_OP_MEM` as a bit-identical GP load or store of that slot, because the bits are the same in either bank. The register-to-register case on rv32 needs the same treatment with a slot the backend supplies, which is why this is not folded into #2860 - that change owns `legalize` and the conversion identities, and this one needs a stack slot materialized after lowering, which is frame and regalloc territory.

**The refusal is wrong even while the lowering is missing.** A conversion that this pass genuinely cannot expand should say so by name, the way the wide divide and the wide comparison do. Landing in the generic clause sends the reader to `MIR_MOV` and to a width, when the fact that explains it is the bank crossing.

## Environment

- target: `riscv32` / `freestanding` / `ilp32d`, `of = "elf"`
- toolchain: `dev` at d8a275d6, and reproduced identically on a `dev`-built baseline, so this predates #2860
- host: linux-x86_64

octalide:
RV32 has no instruction converting between a 64-bit integer and a float. `fcvt.d.l`, `fcvt.lu.d`, `fcvt.s.l` and the rest name a 64-bit GP operand through their `L` selector and no RV32 register is that wide, and `be.codegen.legalize` had no arm for the conversion either: it splits a wide integer into native lanes and threads carries, and a conversion is a single rounding of the combined value rather than two lane conversions recombined afterwards.

## The design decision, and why B

The issue offered soft-float helper calls (A) and open-coded expansion (B) and preferred B. This implements B.

A is correct by construction and is a fraction of the code. It was rejected because it introduces a runtime dependency this backend has nowhere else on this target: a freestanding RV32 image, which is the configuration mach's RISC-V targets exist for, would then need `__floatdidf` and friends linked before an `i64 -> f64` worked at all. The conversion would be absent on exactly the target that wanted it most. The argument is written into `legalize.mach` above the expansion, because the tradeoff is the interesting part and the price of choosing B is that a rounding mistake is a silent wrong answer rather than a failure.

Each direction is an identity that rounds exactly once, and each carries its exactness argument in the source:

- `i64 -> f64` is `f64(hi) * 2^32 + f64(lo)`. Each lane converts exactly, the scale is a power of two, so the closing add is a sum of two exact values.
- `f64 -> u64` is a truncating split whose two scalings and one subtraction are all exact. It inherits saturation and the NaN rule from the native 32-bit `fcvt` in each lane rather than testing for them, so NaN lands on the destination maximum and a negative input on zero with no compare at all.
- `f64 -> i64` converts the magnitude that way, then clamps and negates, which is what puts NaN and `+inf` on `INT64_MAX` and `-inf` on `INT64_MIN`.
- The `f32` forms may **not** narrow the `f64` result. That is a double rounding and it is not innocuous: a 64-bit integer can round in `f64` to exactly an `f32` midpoint without being one, and ties-to-even then breaks the wrong way. The value is rounded to odd at a fixed 11-bit granularity when it needs more than 53 bits, which keeps the `f64` carrier exact and makes the narrowing the only rounding.

The arms fire only where the conversion's float side rides a float register bank, so mos6502 keeps its refusal: the identities are built from native float instructions it does not have.

Counting and emitting are one code path (`ConvEmit` with a nil destination meaning count), because the pass sizes each rebuilt block before writing it and a forty-piece straight-line sequence with compile-time branches cannot be counted by hand reliably.

## Replacing the refusal

Both encoder refusals are gone. What stands in their place is an internal-error check, because after this change no conversion carrying a 64-bit integer reaches the rv32 encoder at all, so an 8-byte operand there means legalization was bypassed rather than that a case is unimplemented.

## Values

Values are the deliverable, not compilation. A probe exercising all eight conversions was built for `riscv32` / `freestanding` / `ilp32d`, linked, and the resulting ELF executed by an RV32IMFD interpreter written independently from the manual's encodings. References come from x86-64 through C (`(double)`, `(float)` on `long long` and `unsigned long long`), never from mach's own riscv64 target.

The full acceptance vector - 0, 1, -1, `INT64_MIN`, 2^53, 2^53 + 1, 2^63 - 1, 2^64 - 1, across `i64 -> f64`, `u64 -> f64`, `f64 -> i64`, `f64 -> u64` and the `f32` twins - is 64 of 64 exact. A wider sweep adds 3400 int-to-float and 1724 float-to-integer checks: randomized values at sixteen bit-lengths, every value within 3 of each power of two from 2^24 to 2^63, deliberately constructed `f32` double-rounding midpoints at every exponent from 2^54 to 2^63, and NaN, both infinities, `+/-2^63`, 2^64, -0.5, `+/-2.5` and a large finite on the float-to-integer side. Zero mismatches.

## Counterfactual

Each mutation applied alone, compiler rebuilt, probe rebuilt, sweep rerun, then reverted.

| Mutation | Result |
|---|---|
| Skip the round-to-odd reduction, narrowing the `f64` result directly | 103 mismatches, all `i64 -> f32` / `u64 -> f32` |
| Read the high lane unsigned in `i64 -> f64` | 3 mismatches on the acceptance vector: -1, `INT64_MIN`, 2^64 - 1 |
| Drop the magnitude clamp in `f64 -> i64` | 2 mismatches: `f64 -> i64` and `f32 -> i64` of 2^63 |
| Remove the `instr_needs_legalize` gate | `legalize:open_codes_a_wide_int_float_conversion` fails |

The first mutation is worth calling out. The original randomized and boundary sweep **did not catch it** - zero mismatches with the double rounding in place. The hazard needed values constructed to sit on the `f32` midpoint, and those were added before the mutation was caught. That is why the sweep is shaped the way it is.

## rv64 stays byte-identical

The compiler's own source tree compiled with a `dev`-built baseline and with this change, per target, objects compared byte for byte:

| Target | Objects | Identical | Differing |
|---|---|---|---|
| linux-x86_64 | 224 | 224 | 0 |
| linux-arm64 | 224 | 224 | 0 |
| linux-riscv64 | 224 | 224 | 0 |

## Tests

`mach build .` and `mach test .` green, 1412 passed. Two unit tests added in `legalize.mach`: one asserting no piece names a wide integer across a float conversion after legalization (the property the missing instruction turns on) with pinned piece counts and a no-float-bank control that must still refuse, and one asserting the arms are inert on a wide ALU. Nothing added under `int/`.

## Found and not fixed

#2904, a 64-bit cross-bank bitcast (`f64 :~ i64`) on rv32, which has no lowering and refuses through the generic width message naming `MIR_MOV`. It predates this change and is reproduced on a `dev`-built baseline.

The executed in-tree float sweep on the reference core is #2874 and lands separately, since the core has to learn F and D first.

Closes #2860.


octalide:
Ran the sweep against the other two ABI members as well, since where an `f64` rides differs across them and the conversion sits on both sides of that.

| ABI | int -> float checks | float -> int checks | mismatches |
|---|---|---|---|
| `ilp32d` | 3400 | 1724 | 0 |
| `ilp32f` | 2600 | 1324 | 0 |
| `ilp32` | 2600 | 1324 | 0 |

The selected float instruction set is identical across the three: `fadd.d`, `fsub.d`, `fmul.d`, `fdiv.d`, `flt.d`, `fcvt.d.w`, `fcvt.d.wu`, `fcvt.wu.d`, `fcvt.s.d`, `fcvt.d.s`, `fld`, `fsd`, `flw`, `fsw`. No `fmv.d.x`, no `fmv.x.d`, and no `L`-selector `fcvt` in any of them, which is the acceptance criterion that no path reaches an RV64-only opcode.

octalide:
Pushed a fix for a hole I found in my own gate after opening this.

A float CONSTANT converted to a wide integer (`ret (1.5)::i64;` on rv32) has an immediate on its float side and a GP destination, so there is no float vreg to read a register class from. It fell out of the gate and refused with `converting between a float and an integer wider than one register needs a float register bank this target does not have` - on a target that has one. A misleading refusal rather than a wrong answer, but the wrong shape either way.

The float bank is now read as a machine fact (`RC_FLOAT` off the ISA's own register classes, which is the fact mos6502 declines to declare) rather than off an operand's class, and a float side not already in a register is read into one first. `(1.5)::i64` gives 1 and `(1.0e300)::i64` gives `0x7fffffffffffffff`, both executed.

Every measurement above re-run against the new build: acceptance vector 64/64, wide sweep 3400 + 1724 with zero mismatches, and 224/224 objects byte-identical on x86-64, arm64 and riscv64.

PR #2927, opened 2026-08-08
fix(codegen): a spirv value is typed by what defines it, not by a MIR width

### LINKED ISSUE #2919: codegen(spirv): an 8- or 16-bit loop counter compares against a mismatched-width constant, producing an invalid module
Found by the codegen corpus (#2903), cases `scalar/arith_u8`, `arith_u16`, `arith_i8`,
`arith_i16`. A `for` loop whose induction variable is 8 or 16 bits is allocated as a
64-bit SPIR-V storage variable, while the loop-condition constant keeps the declared
narrow type. The resulting comparison mixes bit widths and `spirv-val` refuses the module.

mach reports success and writes the `.spv`, so nothing in-tree notices. The external
validator is what catches it.

## Repro

```mach
use corpus.lib.fold;

pub fun checksum(seed: u64) u64 {
    var h: u64 = fold.init();
    var i: u8 = 0;
    for (i < 32) { h = fold.mix_u8(h, i + seed::u8); i = i + 1; }
    ret h;
}
```

`mach build --target spirv --profile o0` exits 0. `spirv-val` then reports:

```
error: line 46: Expected both operands to have the same component bit width: ULessThan
```

The disassembled module shows the mismatch:

```
%_ptr_Function_ulong = OpTypePointer Function %ulong
   %uchar_32 = OpConstant %uchar 32
        %63 = OpVariable %_ptr_Function_ulong Function   ; i, declared u8, stored as 64-bit
              OpStore %63 %ulong_0
        %75 = OpLoad %ulong %63
        %77 = OpULessThan %bool %75 %uchar_32            ; 64-bit load vs 8-bit constant
```

## Expected

A valid module. Either the counter is allocated at its declared width and compared
against a constant of that width, or both sides are widened together.

## Actual

`OpULessThan` (and `OpSLessThan` for the signed widths) with operands of different
component bit widths, which the SPIR-V spec forbids and `spirv-val` rejects.

## Narrowing

Reproduces for `u8`, `i8`, `u16` and `i16` counters at every profile (`o0`, `o2`, `g`).

Positive control: the identical loop with a `u32` counter emits a valid module and passes
`spirv-val` cleanly, as do `u64`, `i32` and `i64`. The defect is isolated to sub-32-bit
induction variables.

## Environment

- compiler: `out/linux-x86_64/debug/bin/mach`, mtime 2026-08-08 17:04:11, `mach 4.17.0`
- target: spirv / freestanding / spirv
- validator: spirv-val, SPIRV-Tools 2026.3

### LINKED ISSUE #2920: codegen(spirv): a 16-byte vector bit reinterpret declares OpTypeInt 128, producing an invalid module
Found by the codegen corpus (#2903), case `vec/vec_cmp_select`. A bit reinterpret between
two 16-byte vector types emits a scalar `OpTypeInt 128 0` alongside the correct
`OpBitcast`. SPIR-V has no 128-bit integer type, so `spirv-val` refuses the module.

mach reports success and writes the `.spv`, so nothing in-tree notices.

## Repro

```mach
use corpus.lib.fold;

pub fun checksum(seed: u64) u64 {
    var fa: f32x4 = f32x4{1.5, -0.0, 3.0, -2.0};
    fa[0] = fa[0] + seed::f32;
    val u: u32x4 = fa:~u32x4;
    ret fold.mix_u32(fold.init(), u[0]);
}
```

`mach build --target spirv --profile o0` exits 0. `spirv-val` then reports:

```
error: line 19: Invalid number of bits (128) used for OpTypeInt.
```

## Expected

`OpBitcast` between the two four-component vector types, with no scalar integer type
declared at all.

## Actual

The module declares `%u128 = OpTypeInt 128 0` and also contains a correct
`OpBitcast %v4float`, so the 128-bit scalar integer type is pure dead weight that makes an
otherwise well-formed module invalid. It reads as the emitter reaching for a
same-size scalar integer to route the reinterpret through, sized from the vector's byte
width rather than from its component type.

## Narrowing

`f32x4 :~ u32x4` lowers correctly on x86_64 and riscv64: the corpus case passes layer C
differential execution on both, at both pipelines. Only the SPIR-V emitter declares the
128-bit type. Reproduces at every profile (`o0`, `o2`, `g`).

## Environment

- compiler: `out/linux-x86_64/debug/bin/mach`, mtime 2026-08-08 17:04:11, `mach 4.17.0`
- target: spirv / freestanding / spirv
- validator: spirv-val, SPIRV-Tools 2026.3

octalide:
Closes #2919
Closes #2920

Both issues are the same shape: a SPIR-V type reconstructed from a MIR byte width
at a point where the value's real type lives somewhere else. They are not one
lever, and neither fix would have exposed the other, but they belong together
because the rule they violate is one rule, and it is now stated once in the
emitter.

## #2919, a narrow loop counter

`imm_merge_width` materializes a phi merge's constant at the machine word, because
a sub-word write to a machine register leaves the upper bits unspecified. That is a
statement about a register file, not about the value. The SPIR-V emitter already
knew a copy is not the source of type truth and skipped register copies when typing
a vreg, but it did not skip immediate ones, so the entry edge of a `for` loop typed
the counter at the machine word. An `i8` counter became a 64-bit Function variable
while its increment and its bound stayed 8 bits wide:

```
%63 = OpVariable %_ptr_Function_ulong Function
%75 = OpLoad %ulong %63
%77 = OpULessThan %bool %75 %uchar_32
```

`u32` was unaffected only because `imm_merge_width` special-cases width 4.

An immediate source is now a copy too, so the destination is typed by the value
copied into it on the back edge, and falls back to the move's width only when
nothing else defines it.

## #2920, a vector reinterpret

Two defects, both in the one instruction.

`emit_convert` reconstructed its source type unconditionally, and interning a type
DECLARES it. A `:~` between two 16-byte vectors has a 16-byte source width, so
reading that as a scalar integer put `OpTypeInt 128` in the module even though the
vreg source never read it. The reconstruction now happens only for an operand that
carries no type of its own.

The reinterpret was also stamped with its OPERAND'S lane shape. `compute_vec_lane`
takes the first vector-typed operand, which is right for every instruction whose
two sides share a shape, and a `:~` is exactly the one whose operand shape is what
it discards. So `f32x4 :~ u32x4` emitted `OpBitcast %v4float`, and the lane read
that followed pulled a `u32` out of a float vector. `spirv-val` never got that far,
because the 128-bit type stopped it first. A reinterpret is now described by its
result.

## The in-tree check

`type_vector` already returns 0 for a component count no Shader module may declare,
and its callers name the offending type in a diagnostic. `type_int` and
`type_float` did not, so a width derived from a MIR byte size declared whatever it
was handed. They do now, which would have turned #2920 into a compiler refusal
rather than a `.spv` only CI looks at.

A width guard cannot see #2919, where every individual type was legal and only the
pairing was wrong. So the arithmetic and comparison forms, which SPIR-V holds to
one shared type, now refuse an operand whose own type is not the one its position
declares. That fires on the `OpIAdd` as well as the `OpULessThan` in the #2919
repro. It is deliberately not in `read_operand`: an access chain's index may
legally be any integer width, and only these positions carry the one-type rule.

## Evidence

spirv-val, SPIRV-Tools v2026.3, universal env (the fixture is a library module, so
Vulkan's rejection of `OpCapability Linkage` does not apply).

| | before | after |
|---|---|---|
| repro `.spv` | `error: Invalid number of bits (128) used for OpTypeInt` | clean, exit 0 |
| `OpULessThan` operands | `%ulong` load vs `%uchar` constant | both `%uchar` |
| `OpBitcast` result | `%v4float` from `%v4float` | `%v4uint` from `%v4float` |
| `OpCapability Int64` | declared | gone, nothing is 64 bits |

Unit tests: 1457 passed, 0 failed (1455 before, two added).

Counterfactual: with the three source files reverted to `dev` and the tests kept,
1455 passed / 2 failed, and `spirv-val` rejected the repro again. Restored, 1457
passed / 0 failed and `spirv-val` exits 0.

Machine target output, mach's own source built as the corpus at the merge base,
baseline kept outside the build tree:

| leg | objects | differences |
|---|---|---|
| determinism control (baseline compiler, twice) | 675 | 0 |
| linux-x86_64 + linux-arm64 + linux-riscv64 | 675 | 0 |
| vector-reinterpret fixture, three targets | 3 | 0 |

The last leg matters because `compute_vec_lane` is shared code and mach's own
source has no vector `:~` in it. The fixture is five reinterprets across four
vector shapes plus a scalar one, and its objects are byte-identical on all three.

int suite, linux leg, `--deps pin`: 40 spirv case runs pass, 0 fail.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,15 @@ A textured shader emits a **byte-identical module** across the migration: the sa
 
 ### Fixed
 
+#### A SPIR-V value is typed by what defines it, not by a MIR width (#2919, #2920)
+Two modules the compiler wrote while reporting success, and that only the external validator refused. A `for` loop whose counter is 8 or 16 bits allocated the counter as a 64-bit variable while its increment and its bound stayed narrow, so the comparison mixed bit widths. A `:~` between two 16-byte vector types declared an `OpTypeInt 128`, which SPIR-V has no form for, beside an `OpBitcast` that was itself typed as the vector it read rather than as the one it produces.
+
+Both reconstruct a type from a **MIR byte width** at a point where the value's real type lives somewhere else. A phi merge materializes its constant at the machine word because a sub-word write to a machine register leaves the upper bits unspecified, which is a statement about a register file SPIR-V does not have. The emitter already knew a copy is not the source of type truth and skipped register copies, but not immediate ones. A conversion reconstructed its source type unconditionally, and interning a type declares it, so a 16-byte source width put a scalar integer in the module the instruction never read. And `compute_vec_lane` describes an instruction by its first vector operand, which is right for every instruction whose two sides share a shape and wrong for the one whose operand shape is exactly what it discards.
+
+**The emission point now refuses rather than writing an invalid module.** `type_int` and `type_float` return 0 for a width SPIR-V has no type at, on the same terms `type_vector` already returns 0 for a component count no Shader module may declare, and the arithmetic and comparison forms refuse an operand whose own type is not the one its position declares. A width guard alone could not have seen the loop counter, where every individual type was legal and only the pairing was wrong.
+
+Every machine target's output is unchanged byte for byte: 675 objects across linux-x86_64, linux-arm64 and linux-riscv64, plus a vector-reinterpret fixture built for all three, which is the shape the shared lane descriptor touches and which mach's own source does not contain.
+
 #### An aggregate is passed and returned by value on the SPIR-V target (#2914, #2915)
 `fun sum(p: Pair) u32` was refused for `spirv` with an addressed-memory diagnostic, and `fun make() Pair` failed earlier still with a bare `spirv.emit: unreadable move source` that named no opcode, no function and no position. A shader API where a record cannot cross a call in either direction is not a finished one.
 
--- a/src/lang/be/codegen/mir/context.mach
+++ b/src/lang/be/codegen/mir/context.mach
@@ -388,11 +388,23 @@ pub fun vec_lane_for_type(ctx: *LowerCtx, ty: ir_type.IrTypeId) u32 {
 # the element kind/width its unsigned-mask result does not - falling back to a
 # vector result (a load or literal has no vector operand). VEC_LANE_NONE when
 # nothing vector-typed flows through, which is every non-vector instruction.
+#
+# A BIT REINTERPRET IS DESCRIBED BY ITS RESULT (#2920). Every other instruction's
+# operand and result share a lane shape, which is why the operand answers for both.
+# A `:~` is the one whose operand's shape is exactly what it discards: the two sides
+# share a byte width and nothing else, so a consumer reading the operand's shape as
+# the result's saw `f32x4 :~ u32x4` produce another `f32x4`, and every lane read off
+# that was then a type error. A register machine copies the whole 16-byte register
+# either way, so only a target that realizes a vector as a TYPE can see it.
 # ---
 # ctx:  the lowering context
 # inst: the IR instruction being lowered
 # ret:  the packed vec_lane descriptor to stamp on its MIR instructions
 pub fun compute_vec_lane(ctx: *LowerCtx, inst: *instruction.Instruction) u32 {
+    if (inst.kind == instruction.OP_BITCAST) {
+        val r: u32 = vec_lane_for_type(ctx, inst.ty);
+        if (mir.vec_lane_is_vector(r)) { ret r; }
+    }
     var i: u32 = 0;
     for (i < inst.operand_count) {
         val d: u32 = vec_lane_for_type(ctx, inst.operands[i].ty);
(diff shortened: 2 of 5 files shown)


octalide:
One extra finding worth recording, because it changes how #2920 reads.

The `OpTypeInt 128` was MASKING a second defect in the same instruction. With only the `emit_convert` half applied and `compute_vec_lane` left at `dev`, the module gets past the 128-bit type and `spirv-val` then reports:

```
error: line 189: Result type (OpTypeInt) does not match the type that results from indexing into the composite (OpTypeFloat).
  %120 = OpCompositeExtract %uint %119 0
```

That is the lane read pulling a `u32` out of the `f32x4` the bitcast wrongly produced. The validator had been stopping at the dead type declaration and never reaching it, so the issue's "the 128-bit scalar integer type is pure dead weight" was true of the type and not of the module. Both halves are needed.

Issue #2940, opened 2026-08-09
codegen(spirv): a pointer to a record is refused as a parameter type


octalide:
Mandatory Mach v5 release gate.

SPIR-V logical-addressing pointer-to-record parameter support and precise genuine refusals. Refresh obsolete debug acceptance against the target matrix.

Acceptance: implement legal logical-addressing pointer-to-record parameters and retain
  precise refusals for genuinely unsupported pointer/function-value operations.

The original reproduction and relevant acceptance controls below remain required, subject to this current scope. Historical site counts and time estimates are not the acceptance criterion.

Replace the historical unconditional debug-build expectation with the explicit support matrix in briar-systems/mach#3112. A current debug refusal is not evidence that SPIR-V intrinsically forbids debug information. Legal pointer-to-record support remains mandatory.

## Tracking

Parent: [mach#2212](https://github.com/briar-systems/mach/issues/2212).

Completion prerequisites:

- [mach#2963](https://github.com/briar-systems/mach/issues/2963) backend: implement a value-oriented ABI contract for module emitters

Work that does not depend on an unfinished prerequisite can proceed independently.

<details>
<summary>Historical report and reproduction, retained for provenance</summary>

A pointer to a record is refused as a parameter type on the SPIR-V target, though logical addressing expresses exactly that: a `Function` storage class pointer parameter whose fields are reached with `OpAccessChain`.

## Repro

```mach
rec Pair {
    a: u32;
    b: u32;
}

fun fill(p: *Pair, v: u32) {
    p.a = v;
    p.b = v + 1;
}

pub fun run(s: u32) u32 {
    var q: Pair;
    fill(?q, s);
    ret q.a + q.b;
}
```

```sh
mach build . --target spirv --profile o0
```

## Expected

A module. `fill` takes `%_ptr_Function_Pair`, the caller passes the `OpVariable` for `q` directly, and each field write is an `OpAccessChain` plus an `OpStore`. SPIR-V logical addressing permits a pointer function parameter: what it forbids is a pointer held in memory, a pointer selected between, and pointer arithmetic. None of those appear here.

## Actual

```
error: unsupported parameter type in the SPIR-V target (in fill)
```

The refusal carries no source location and does not name the parameter or its type, so on a real case it points at a function rather than at a line.

## Environment

- `feat/2903-codegen-corpus` at the merge of `origin/dev` 206861d5, built at `out/linux-x86_64/debug/bin/mach`, mach 4.17.0
- linux x86_64

[...]

Issue #2939, opened 2026-08-09
codegen(spirv): a narrow value merged across an if gets a machine-word variable


octalide:
A narrow value merged across an `if` is given a SPIR-V variable at the machine word, so the module stores a `%uint` into a `%_ptr_Function_ulong` and `spirv-val` rejects it. Regression from 46cd5558 (PR #2927).

## Repro

```mach
pub fun pick(s: u32) u32 {
    var r: u32 = 0;
    if (s < 4) {
        if (s < 2) { r = 10; } or { r = 11; }
    } or {
        if (s < 6) { r = 12; } or { r = 13; }
    }
    ret r;
}
```

```sh
mach build . --target spirv --profile o0
spirv-val out/spirv/o0/obj/min/min.spv
```

## Expected

A module `spirv-val` accepts. Every merge slot for `r` is `%_ptr_Function_uint`, because `r` is a `u32` and every value reaching the merge is a `u32`.

## Actual

```
error: line 52: OpStore Pointer <id> '25[%25]'s type does not match Object <id> '37[%37]'s type.
  OpStore %25 %37
```

The inner merge slots are typed correctly and the outer one is not:

```
%23 = OpVariable %_ptr_Function_uint Function
%25 = OpVariable %_ptr_Function_ulong Function
...
%14 = OpLabel
%37 = OpLoad %uint %23
      OpStore %25 %37
```

`mach build` reports success and writes the module, so the only thing that catches it is an external validator.

The mechanism is the ordering in `src/lang/target/isa/spirv/emit.mach:3127`:

```
type_computed_defs(e, mf, out, true)    # definite defs, register copies skipped
propagate_copy_types(mf, out)           # carry types across copies to a fixed point
type_computed_defs(e, mf, out, false)   # fall back to the move's width
```

The slots for `r` form a copy-connected component whose only definitions are immediate moves, so pass one types none of them and propagation has nothing to carry. Pass three then types every one of them from its own move width, and those widths disagree: the inner phi-edge moves carry 4 and the outer carries 0, which `type_computed_defs` turns into `gpr_width`. Propagation never runs again, so the disagreement stands. Before 46cd5558 every slot in the component was typed by its move width in one pass and the component came out uniformly `uint`.

Widening the fallback is not the fix and neither is narrowing it. The move width is a register-machine materialization convention, which is exactly what 46cd5558 established, so a component with no definite type inside it has no authority to read at all. The type has to come from where the value is actually typed.

## Environment

[...]

Issue #2934, opened 2026-08-09
sema: $size_of(usize) answers at 64 bits on a 32-bit target, so mach-std cannot compile for any ilp32 target


octalide:
`mach-std` does not compile for any 32-bit target. `dep/mach-std/src/memory.mach:23` reads

```mach
val step: usize = $size_of(usize);
```

and on an ilp32 target that reports `type mismatch: expected u32, found u64`, so `$size_of` answers at 64 bits whatever the target's pointer width is.

## Repro

Build any one-function project for a 32-bit target with a `mach-std` dependency declared. It reproduces with a single function, because mach pulls `std.memory` in implicitly whenever a mach-std dep is present, so no source needs to reference it.

## Expected

`$size_of(usize)` answers at the target's pointer width, and `mach-std` compiles for riscv32 and any other ilp32 target.

## Actual

`type mismatch: expected u32, found u64` from inside the dependency, before any user code is reached.

## Why it matters

It blocks every 32-bit target at the standard library, so no riscv32 program that uses mach-std can build at all. It was found by the #2903 corpus, whose riscv32 column is entirely unreachable partly for this reason.

## Environment

riscv32, ilp32. mach 4.17.0 built from `dev`. Found by the #2903 corpus build.

Issue #2923, opened 2026-08-08
codegen(spirv): the emitter's fixed 16-parameter buffer is presented as a target capability


octalide:
Superseded by [mach#2963](https://github.com/briar-systems/mach/issues/2963). Remaining obligations and acceptance are now carried by those issues in the mandatory v5 set. Closed as superseded, not fixed.

<details>
<summary>Historical report and reproduction, retained for provenance</summary>

Found by the codegen corpus (#2903), case `call/call_mixed`. The SPIR-V emitter refuses any
signature with more than sixteen parameters. SPIR-V itself allows 255, so the bound is the
emitter's own fixed operand buffer surfacing as a language-level refusal.

## What

`src/lang/target/isa/spirv/types.mach:129` states the bound and says plainly where it comes
from:

```mach
# SPIR-V itself imposes no small bound; sixteen is here because the emitter
# assembles a signature's operands into fixed arrays. It is stated ONCE and both
# the refusal in `emit_function` and the operand buffer in `type_fn` are sized from
# it.
pub val MAX_FN_PARAMS: u32 = 16;
```

`emit.mach:2451` and `emit.mach:4162` then refuse with
`more than 16 parameters is not supported by the SPIR-V target`.

Stating the bound once was the right fix for #2748, where two independent literals
disagreed at fifteen and sixteen. The remaining problem is the bound itself: it is an
implementation detail of the buffer, presented to the user as a target capability.

Confirmed against the current compiler: a 17-parameter function refuses, a 16-parameter
function builds and validates cleanly.

The corpus case that hits it is `call/call_mixed`, which exists to exhaust the integer and
float argument banks of every ABI in one signature. That needs at least eighteen
parameters, so the case cannot be shortened without giving up the thing it tests. Two
neighbouring cases, `call_int_regs` and `call_float_regs`, were written to sixteen
parameters specifically to stay inside this bound, which still spills eight arguments on
aarch64 and riscv and ten on x86_64 but is a corpus shape chosen around an emitter limit
rather than around an ABI boundary.

## Fix

Size the signature operand buffers from the function being emitted rather than from a
constant, and delete `MAX_FN_PARAMS` along with both refusals. If a bound is still wanted
as a guard, it belongs at SPIR-V's own 255 and should name the spec limit rather than the
buffer.

## Acceptance

- A function with more than sixteen parameters builds for `--target spirv` and passes
  `spirv-val`.
[...]

PR #2943, opened 2026-08-09
infra: the unified codegen corpus replaces the integration suite

### LINKED ISSUE #2903: infra: unified codegen corpus (differential execution, golden disassembly, structural validation)
## What

A single unified codegen test suite, designed from a blank sheet, that answers one question with direct logic: **is the compiler generating the code we told it to generate, on every registered target?** It replaces the project-per-case integration model for everything except cross-compilation and linking, and it is built so hand-written regression projects become largely unnecessary: a past miscompile is pinned as one small case file that executes on every target forever.

The suite is a flat corpus of single-file mach programs, one codegen surface per file, run by one driver through three independent oracle layers. Nothing in the suite restates a constant the compiler owns. Every oracle is grounded outside mach: an external disassembler, an external validator, or execution compared against a C reference compiled by the host's own C compiler.

This issue is design-complete. It specifies the layout, the case contract, the oracle layers, the engine table, the tool pins, and the full initial corpus by name and scope. Implementation choices not specified here (driver language, internal helper structure) are the implementer's call.

### Why the current shapes are not enough

- The int suite's oracle is mostly "it built and exited zero." Green matrices have shipped over miscompiled 3-lane vectors and over an rv32 target on which no test ever executed wide arithmetic.
- Nothing checks mach's instruction encodings against an external decoder. An RV64-only instruction assembled under an rv32 tag traps at run time instead of failing in CI.
- Nothing validates object structure externally. An ELF writer stamped ELFCLASS32 onto Elf64 geometry and nothing in-tree noticed.
- Each int case is a whole project. That cost is why the int suite is frozen. Pinning a regression must cost one small file, not a directory tree and minutes of CI.

### Layout

```
corpus/
  run.sh                     driver entry point (CLI contract below)
  tools.lock                 pinned external tool versions, checked by the driver at startup
  engines.conf               one row per target: name, execution engine, disassembler
  cases/
    <group>/<case>.mach      one case, one codegen surface, target-independent source
  ref/
    <group>/<case>.c         the C reference translation of the case
  golden/
    <target>/<group>/<case>.dis   blessed external-disassembler text (release pipeline)
    <target>/SKIPS                declared per-target skips, one per line with a reason
  lib/                       driver internals
  out/                       all build products, gitignored
```

No per-case projects, no per-case manifests. The driver compiles `cases/<group>/<case>.mach` directly for each target under a single generated manifest it owns.

### Case contract

Every case file obeys the same contract, and the driver enforces it rather than trusting it:

1. Target-independent source. No target conditionals, no inline asm, no OS-specific calls. The same source compiles for every target (or is declared in that target's `SKIPS` with a reason).
2. Deterministic. No input, no time, no addresses observed as values, no float printing. Terminates in well under one second on the slowest engine.
3. Computes a `u64` checksum by folding every intermediate result with FNV-1a. Float results are folded by bit pattern (`f32`→`u32`, `f64`→`u64` bitcast), never by value, so NaN and -0.0 are exact.
4. Prints exactly one line: the 16-hex-digit checksum. Exits 0. Any other exit or output shape is a failure of the case, loudly reported.
5. Operands are chosen inside mach's defined semantics AND the defined-behavior C mapping below. No case may depend on behavior that is UB in the C translation.
6. Every value that matters flows through an opaque boundary (function argument or global read) at least once, so the optimizer cannot fold the entire case to a constant. A separate comptime group deliberately does the opposite.

### The C reference, and its rules

`ref/<group>/<case>.c` is a line-for-line translation of the case. It is the semantic anchor: the checksum's truth comes from the host platform's C toolchain, not from mach.

Rules that make the anchor sound:

- All arithmetic in the C translation is performed on unsigned fixed-width types (`uint8_t`..`uint64_t`) with explicit casts at every width change. Signed mach operations are expressed via two's-complement identities on unsigned types. **No operation in any reference file may be UB in C.** An oracle disagreement must always indict mach, never the reference.
- The reference is compiled by the pinned host C compiler at **both -O0 and -O2**, and the two binaries must agree with each other before either is compared against mach. Disagreement between the two reference builds fails the run as a harness defect, not a mach defect.
- Checksum printing in C uses `PRIx64` with zero padding to match the case contract byte for byte.

The reference runs on the host only. Cross-target agreement is transitive: every target's output equals the host C answer independently, so no cross-leg communication is needed and every leg is self-contained on one machine.

### Oracle layers

Every artifact the driver builds passes through three independent layers. A layer never consumes another layer's output.

**Layer A, structural validation.** Every object and executable, every target, both pipelines.
- ELF: `llvm-readobj --file-header --sections --relocs` must parse cleanly, plus assertions that class/machine/endianness match the target.
- COFF and Mach-O: same via `llvm-readobj`.
- spv: `spirv-val` with the target environment mach claims.
- With `-g`: `llvm-dwarfdump --verify` must pass.
- raw: size and entry sanity assertions only.

**Layer B, golden disassembly.** The "code we told it to generate" check. The artifact is disassembled by an **external** decoder and diffed against a checked-in text golden.
- x86_64, aarch64, riscv64, riscv32: `llvm-objdump -d --no-show-raw-address` (exact flag set fixed in `tools.lock` commentary so goldens are reproducible).
- spirv: `spirv-dis`.
- mos6502: `da65` from the cc65 suite (LLVM has no 6502 disassembler).
- Goldens are recorded for the release pipeline (`-O2`) only, to bound churn. They are text, tiny, and reviewable. `run.sh --bless` regenerates them, prints the full diff, and never runs in CI. A golden change lands in the same PR as the compiler change that caused it, with the diff reviewed as part of the PR.
- The point of the external decoder: an instruction mach encodes wrong is *displayed* wrong or refused, so the golden cannot agree with a bad encoding by construction.

**Layer C, differential execution.** The miscompile net.
- Every case is compiled at **both -O0 and -O2** for every target with an execution engine, run, and its checksum compared against the C reference answer. All four numbers (mach -O0, mach -O2, ref -O0, ref -O2) must be identical. -O0 vs -O2 disagreement inside mach is a pipeline miscompile caught with no golden and no reference needed.
- This layer's expected values never change unless language semantics change. That is the property that retires regression maintenance.

### Engine table

`engines.conf` is the single source of truth for how each target executes, and the coverage report names the engine per cell so a qemu result can never be mistaken for native ABI evidence.

| target | engine | disassembler | note |
|---|---|---|---|
| x86_64-linux | native | llvm-objdump | |
| aarch64-linux | native (arm runner) | llvm-objdump | |
| riscv64-linux | qemu-user | llvm-objdump | compute evidence only, not ABI evidence |
| riscv32 | qemu-user (`qemu-riscv32`) | llvm-objdump | first standing execution of rv32 wide arithmetic |
| x86_64-windows | native (windows runner) | llvm-objdump | **no wine, anywhere, ever** |
| darwin (both) | native, main cadence | llvm-objdump | |
| spirv | none (declared) | spirv-dis | layers A and B only |
| mos6502 | none (declared) | da65 | layers A and B only; promotion to an execution engine is a follow-up |

Registering a new target requires a row here. `none` is legal but it is a public label, not a default.

### Honesty mechanisms

The dominant historical defect class is a test reporting more than it verifies. The driver is built against that:

- After every run it emits the **coverage matrix**: case × target × layer × pipeline, with the engine that produced each layer C cell. The matrix is printed locally and uploaded as a CI artifact.
- The run **fails** if any target with a declared engine has an empty layer C column, or any target has an empty layer A or B column, unless each missing cell is covered by a `SKIPS` entry. There are no silent skips: a skip exists in a reviewed file with a reason, and the matrix counts skips separately from passes.
- A case that produces no output, wrong-shaped output, or a nonzero exit is a **failure**, never a skip.
- The driver verifies `tools.lock` against the actual tool versions on the machine before running, and refuses to run layer B with an unpinned disassembler. Golden diffs caused by tool drift are impossible by construction.
- The driver prints the mach binary's version and path at startup, so it is never ambiguous which compiler was measured.

### Tool pins (`tools.lock`)

- LLVM tools (`llvm-objdump`, `llvm-readobj`, `llvm-dwarfdump`): one pinned major version. Upgrading the pin is an explicit PR that re-blesses goldens with a reviewed diff.
- SPIRV-Tools (`spirv-val`, `spirv-dis`): pinned version.
- cc65 (`da65`): pinned version.
- Host C reference compiler: pinned per platform (gcc on linux, clang on darwin, mingw-gcc on windows), floor version recorded.
- `qemu-user` (`qemu-riscv64`, `qemu-riscv32`): floor version recorded.

### Driver CLI contract

```
run.sh                          # everything: all cases, all targets, all layers
run.sh --target <t>             # one target
run.sh --case <group>/<name>    # one case (repeatable), still all layers
run.sh --layer a|b|c            # one layer
run.sh --bless [--target <t>]   # regenerate goldens, print diffs, never in CI
run.sh --matrix                 # print the coverage matrix from the last run
```

Exit nonzero on any failure, any unexplained empty cell, or any tool-pin mismatch.

### The initial corpus

One case per line. Each is one small `.mach` file plus its `ref/` twin. Names are binding, scope lines say exactly what the case must exercise. Roughly 90 cases.

**`scalar/`, per-width integer arithmetic.** Add/sub chains crossing the wrap boundary, negation, mixed opaque values and constants.
- `arith_u8` `arith_u16` `arith_u32` `arith_u64` `arith_i8` `arith_i16` `arith_i32` `arith_i64` — on riscv32 the 64-bit cases are the register-pair lowering that shipped untested once, on mos6502 everything above 8 bits is
- `mul_u32` `mul_u64` `mul_i32` `mul_i64` — products exceeding the width, wrap behavior
- `divrem_u32` `divrem_u64` `divrem_i32` `divrem_i64` — fold `q`, `r`, and the identity `d*q + r` per case, near-boundary dividends, operands inside mach's defined domain only

**`bits/`, bitwise and shifts.**
- `logic_u32` `logic_u64` — and, or, xor, not over boundary patterns (0, ~0, alternating masks)
- `shift_u32` `shift_u64` `shift_i32` `shift_i64` — logical and arithmetic shifts at amounts 0, 1, w-1, and mach's defined behavior at the masking boundary
- `bitintrin` — one folded case per bit-manipulation intrinsic the language ships (count-ones, leading/trailing zeros, byte swap, as applicable); if none ship, the case is omitted and this line records why

**`convert/`, every width and domain crossing.**
- `ext_zero` `ext_sign` — every widening pair, values with the top bit set
- `trunc` — every narrowing pair, values that differ above the cut
- `bitcast` — f32↔u32, f64↔u64 round trips
- `f2i` `i2f` — every float↔int pair, in-range values plus mach's defined edge behavior
- `f2f` — f32↔f64 both directions, precision-losing values

**`float/`, folded by bit pattern.**
- `arith_f32` `arith_f64` — add, sub, mul, div over ordinary and boundary magnitudes
- `cmp_f32` `cmp_f64` — every predicate, including NaN operands on both sides
- `special_f32` `special_f64` — infinities, -0.0, NaN propagation, denormal handling per mach's semantics

**`cmp/`, comparisons and control flow.**
- `cmp_u32` `cmp_u64` `cmp_i32` `cmp_i64` — all six predicates over boundary pairs (0x7fff…/0x8000…, 0/~0)
- `branch_nest` — a nested if/else decision tree wide and deep enough to force branch chain layout
- `loop_shapes` — counted, while, nested, early-exit, loop-carried dependency
- `short_circuit` — proves both evaluation order and non-evaluation via side-effect counters folded into the checksum

**`mem/`, memory and layout.**
- `loadstore` — every scalar width through stack and global memory
- `rec_layout` — nested records with deliberate padding, field write/read round trips, offset-sensitive checksums
- `uni_layout` — union write-as-one-read-as-another per mach's defined semantics
- `array_index` — constant and dynamic indices, interior pointers
- `ptr_arith` — pointer add/sub/difference, writes through aliased pointers
- `global_data` `global_zero` — initialized and zeroed global round trips (data and bss section correctness)
- `addr_modes` — expression shapes that should exercise base+offset and base+index×scale selection

**`imm/`, immediate-encoding boundaries.** One source, all targets. The constants are chosen to straddle every encoder's boundaries at once: ±2^11 (riscv I-type), aarch64 shifted and bitmask immediates (0x5555…, 0x0f0f…, 0xffff0000), ±2^31 and 2^32 (x86 sign-extension), 2^63.
- `imm_add` `imm_logic` `imm_mov` `imm_addr` — the same boundary constant set flowing through adds, logic ops, materialization, and address arithmetic

**`call/`, calling convention corners.**
- `call_int_regs` — enough integer args to exhaust every ABI's argument registers plus several stack args
- `call_float_regs` — same for float registers
- `call_mixed` — interleaved int, float, and vector args
- `call_rec_small` `call_rec_edge` `call_rec_large` — record args at sizes 1, 2, 4, 8, then 9, 12, 16, 17, then 24, 32, 48 bytes, straddling every ABI's classification boundaries
- `call_ret_small` `call_ret_large` — record returns in registers and via sret
- `call_indirect` — calls through function values
- `call_variadic` — per `doc/language/variadics.md`
- `call_recursive` — recursion deep enough to prove per-frame locals don't alias
- `call_chain` — a long chain clobbering and restoring callee-saved registers

**`vec/`, SIMD vectors.** The historical miscompile home. Each arith case runs every lane through add, sub, mul, and lane-position-sensitive folding, so a wrong lane changes the checksum.
- The seeded 128-bit set, one case each: `vec_f32x4` `vec_f64x2` `vec_i8x16` `vec_i16x8` `vec_i32x4` `vec_i64x2` `vec_u8x16` `vec_u16x8` `vec_u32x4` `vec_u64x2`
- The awkward widths: `vec_f32x2` (half register), `vec_f32x3` (12-byte, the shipped miscompile), `vec_f32x5` (20-byte, reg+tail), `vec_f32x8` (two registers), `vec_i16x4`
- `vec_lane_ops` — insert and extract at every lane index of several types
- `vec_cmp_select` — vector comparisons and select/blend shapes
- `vec_mem` — vector loads and stores through records and arrays, including the 12- and 20-byte types whose size and alignment disagree
- `vec_scalar_mix` — vector-scalar operation combinations
- `autovec_loop` — a loop the auto-vectorizer takes at -O2; the -O0/-O2 checksum equality is the oracle that vectorization preserved semantics

**`frame/`, frames and register allocation.**
- `frame_large` — locals pushing the frame past 4 KiB so offsets exceed short immediate ranges
- `frame_spill` — enough simultaneously-live values to force spills, checksum sensitive to every one
- `frame_align` — over-aligned locals including vector locals, alignment proven by masking addresses is forbidden (addresses are not values), proven instead by correct results of aligned vector ops

**`comptime/`.**
- `ct_runtime_agree` — the same computation performed once through comptime evaluation and once through opaque runtime arguments, both folded, both required to equal each other and the C reference: the comptime evaluator and the codegen path must implement the same semantics

### CI

One leg per `engines.conf` row with a runner, replacing the int legs for everything except the residual cross/link projects. Each leg is one machine, self-contained, no cross-leg joins: checkout, verify tools, build mach (or take the artifact from the build job), `run.sh --target <t>`, upload the matrix. Budget: under 2 minutes per leg, driven by the fact that every case is a single small file. Cadence mirrors `targets.conf` today: linux x86_64/aarch64, riscv64 and riscv32 under qemu, windows on PR; darwin legs on main. The spirv and mos6502 columns (layers A and B) run on the linux x86_64 leg since they need no execution host.

### Explicit non-goals

- Diagnostics, driver behavior, and refusal testing: unit tests own these. They proved this week they can.
- Debug info *quality*: layer A gets well-formedness via `llvm-dwarfdump --verify` only. The debug info milestone owns goodness.
- Optimization quality (how fast the code is): the benchmark program owns that. This suite only proves optimized code is *correct*.
- Frontend fuzzing: not a priority by owner decision. Generated programs transit the parser as a side effect and that is all.
- Inline asm: target-specific by nature, violates the case contract. Excluded from v1.

## Acceptance

- `corpus/` exists with the layout above, all ~90 cases and their C references present, goldens blessed for all six targets.
- `run.sh` on a linux x86_64 host with the pinned tools completes all layers for x86_64 (native), riscv64 and riscv32 (qemu), spirv and mos6502 (A and B), in under 3 minutes total, and prints the coverage matrix.
- Every case's four checksums (mach -O0, mach -O2, ref -O0, ref -O2) agree on every executing target. Any historical still-reproducible miscompile class (3-lane vectors, rv32 wide arithmetic) is demonstrably covered: reverting one such fix locally makes the corresponding case fail.
- Deleting a golden file or emptying a target's layer C column without a `SKIPS` entry makes the run fail. Running with a wrong-version `llvm-objdump` makes the run refuse to start layer B.
- `--bless` prints a reviewable diff and CI has no path that invokes it.
- CI legs exist per the engine table, green, each under budget, uploading matrices as artifacts.
- The int suite is untouched by this issue. Retiring `int/regression` into corpus cases and shrinking `int/` to cross-compilation and linking is follow-up work, as are the cross-compile byte-identity join, mixed-toolchain interop link cases, mos6502 engine promotion, and the seeded case generator (fuzzer) that emits contract-conforming case files into the same driver.

octalide:
Closes #2903.

A single unified codegen corpus at `test/`: 91 single-file mach programs, one codegen surface per file, run by one driver through three independent oracle layers. `int/` and `tools/` are deleted, and the 65 integration cases whose property cannot be observed by building and looking on a developer's own machine move to `test/link/`.

Every oracle is grounded outside mach: `llvm-readobj` and `spirv-val` and `llvm-dwarfdump --verify` for structure, `llvm-objdump` and `spirv-dis` against checked-in goldens for encoding, and execution compared against a C reference the host's own C compiler builds at both `-O0` and `-O2`.

## It catches the bugs it was built for

`convert/f2u_high` is new. The float group kept every float-to-int value below 2^31 and 2^63 while #2909 was open, so the arm of the x86-64 expansion above the boundary had no case. The boundary itself proves nothing and is present only as a labelled control: a bare signed convert of exactly 2^31 or 2^63 returns the x86 integer-indefinite value, which is the correct unsigned answer, so a probe there agrees with the lowering it is supposed to catch. Every probe is strictly above, and the values below stay as the positive control.

Reverting c02c91e9's encoder change and rebuilding:

```
FAIL convert/f2u_high x86_64-linux layer c o0: mach o0 says 738cf2bc869ea360, C reference says d356b592966ce737
FAIL convert/f2u_high x86_64-linux layer c o2: mach o2 says 738cf2bc869ea360, C reference says d356b592966ce737
layer C
  case              x86_64-linux
  convert/f2i       o0+o2(native)
  convert/f2u_high  FAIL
cells: 2 pass, 2 fail, 0 skip
```

`convert/f2i` stays green through the same revert, which is the coverage gap and its closure in one measurement.

## It found two defects on the way in

Merging current `dev` into the branch turned up two things nothing else had, both filed and both declared in `test/golden/spirv/SKIPS` naming their issue.

**#2939**, a regression. A narrow value merged across an `if` is given a machine-word SPIR-V variable while the value stored into it keeps its own width, so `mach build` reports success and writes a module `spirv-val` rejects. Bisected to 46cd5558, the #2919 fix: 46cd5558 BAD, 46cd5558^ GOOD. `cmp/branch_nest` built and validated at 2efe5ce8 and does not now. `float/cmp_f32` and `float/cmp_f64` clear the #2910 boundary they had been skipped for only to stop at the same fault reaching an add instead of a store. Filed `critical` because it ships silently.

**#2940**, the residue of #2914. A pointer to a record is refused as a parameter type on spirv, which logical addressing does express as a `Function` pointer parameter reached by access chain. `mem/rec_layout` now gets all the way there.

## Twenty spirv skips came out

Each of the 26 mach-defect entries was rerun with the entry removed. Twenty build and validate and their entries are deleted: #2919 (`scalar/arith_u8` `arith_u16` `arith_i8` `arith_i16`), #2910 (`cmp/cmp_u32` `cmp_u64` `cmp_i32` `cmp_i64`, `float/special_f32` `special_f64`), #2914 (`call/call_rec_small` `call_rec_edge` `call_rec_large` `call_ret_large`, `mem/loadstore` `array_index`), #2915 (`call/call_ret_small`), #2920 (`vec/vec_cmp_select`), #2921 (`vec/autovec_loop`), #2922 (`vec/vec_mem`).

Two are reclassified rather than deleted, because they were never mach defects: `call/call_indirect` passes a function value and `mem/ptr_arith` indexes off a pointer to a scalar, and logical addressing has neither, so both move to target limitation. `call/call_mixed` keeps its #2923 entry.

78 of the 91 cases now carry the spirv column, against 58 before.

The four `cmp/cmp_*` cases go back to brace-literal arrays. They had been written around #2911 with a comment saying so, and that comment was the last thing keeping the mach source and its C reference from being line for line. The values are unchanged, so the checksums are unchanged.

## Carried forward honestly

- **riscv32 builds nothing and executes nothing.** All 91 cases refuse with `legalize: operation wider than the target ALU is unsupported (MIR_MOV, 8 bytes wide)`, reverified this run with the blanket entry removed. The issue's "first standing execution of rv32 wide arithmetic" is not reachable here and is not claimed. The column and an execution engine open together or not at all, because the register-pair lowering is the same work.
- **mos6502 builds nothing.** All 91 refuse with `mir.lower_ir: target has no float/vector register bank`. `da65` gets a `tools.lock` pin when something here has run against it, not before.
- **`mach-std` does not compile on any 32-bit target**, filed as #2934.
- **Six link cells fail on this host only.** `c-variadic`, `narrow-stack-args` and `riscv-pcrel-pair` on `riscv64-linux` at both profiles need riscv64 libc cross headers, which this Arch machine does not have and `ci.yml` installs as `libc6-dev-riscv64-cross` on that leg. The refusal is `cc.sh`'s own, by design, rather than a silent build against host headers.

## Acceptance, each with its output

**Full run, exit 0, under budget.** `mach 4.18.0` at `out/linux-x86_64/debug/bin/mach`:

```
cases:    91   targets: x86_64-linux,riscv64-linux,riscv32,spirv,mos6502   layers: abc
cells: 1404 pass, 0 fail, 663 skip
./test/run.sh  180.75s user 33.27s system 139% cpu 2:33.07 total
```

**A deleted golden fails the run**, exit 1:

```
FAIL scalar/mul_u64 x86_64-linux layer b o2: no golden at .../scalar/mul_u64.dis; run --bless and review the diff
EMPTY scalar/mul_u64 x86_64-linux layer b has no passing cell and no SKIPS entry
```

**An empty layer C column with no SKIPS entry fails the run**, exit 1, with no failing cell anywhere. Declaring `c:o0` and `c:o2` away individually leaves the column empty, and the gate refuses that rather than reading two skips as coverage:

```
EMPTY scalar/mul_u64 x86_64-linux layer c has no passing cell and no SKIPS entry
cells: 0 pass, 0 fail, 2 skip
```

**A wrong-version `llvm-objdump` refuses to start layer B**, exit 2:

```
corpus: refusing to run; tools.lock is not satisfied on this host:
  llvm-objdump is version 19.1.0, tools.lock pins major 22
```

**`--bless` prints a reviewable diff and no CI path invokes it.** The run that reblessed this branch printed a 44,400-line unified diff and `blessed 37 golden file(s)`. `grep -rn bless .github/` returns nothing, exit 1. `CI=1 ./test/run.sh --bless` exits 2 with `--bless regenerates goldens and must never run in CI`.

**The link suite still carries cross-compilation and linking**: 133 pass / 6 fail / 0 skip over 139 cells in 45s, the 6 being the host header gap above.

## Goldens that moved, and why each is correct

Six move for a compiler change on `dev` that owns them: #2909 in `convert/f2i`, `float/special_f32`, `float/special_f64`; #2917 in `vec/vec_scalar_mix` on riscv64 only, which is where the fix was; #2918 in `comptime/ct_runtime_agree` and `convert/f2f`. Eight move because the `cmp/cmp_*` cases went back to array literals. Twenty three are new: the spirv cases the closed defects reopened, plus `convert/f2u_high` on every column that builds it.

Nothing in the corpus named `#[spirv_op]`, an image or a sampler, so #2888's rename reaches none of it.

---

## Where all 176 integration cases went

The suite held **176 case directories under 174 distinct names** (`shell` and `step`
each existed once under `regression` and once under `surface`). 65 moved to
`test/link/`. The other 111 are classified below.

| disposition | count | what it means |
|---|---|---|
| `link` | 65 | moved to `test/link/`, unchanged in what it asks |
| `corpus` | 57 | a codegen fact. The corpus asks it by construction, through structural validation, golden disassembly and differential execution against a C reference. |
| `unit` | 37 | a front-end or middle-end fact: a refusal, a comptime fold, a reflection walk, a diagnostic. The compiler's own tests assert it, and a target adds nothing because the decision is made before lowering. |
| `gone` | 14 | a driver or toolchain check. Building and looking on a developer's machine observes it, which is the rule the split was made against, so no case replaces it. |
| `LOST` | 3 | coverage genuinely lost. See below. |

### The three that are lost, and the harness that went with them

These are not reclassified. Each measured something that only means anything on the
machine it executes on, which is why it was an integration case and why the corpus
cannot carry it: the corpus builds for a target and reads what came out, and none of
these is a property of the bytes.

- **`int/surface/ct-flags-hardware`** — the eighteen-row flags table in
  `x64/encode.mach` is transcribed from this probe, and that table is the security
  surface the `#[oblivious]` model rests on. Nothing re-runs it, so a CPU that stops
  matching diverges silently. The comment in `x64/encode.mach` now says so.
- **`int/regression/2791-asm-noreturn`** — the linux and darwin kernel-entry
  conventions an inline-asm `noreturn` body ends control with. #2942 fixed the darwin
  arms one day before the suite was retired.
- **`int/surface/syscall-inline-asm`** — the same shape for a console write, carrying
  the windows non-applicability with it.
- **`int/ct/`** — the dudect-style timing harness behind the empirical claims in
  `doc/language/secrecy.md`. Never a case, ran on demand and never in CI, and gone.
  Those claims are now marked as measurements taken once.

**#2944** restores all four as host-executed unit tests, which is where a measurement
that only means anything on the executing ISA belongs.

### Per-case

| case | disposition |
|---|---|
| `int/regression/1750-i32-signext` | corpus |
| `int/regression/1762-rv64-float-vararg` | corpus |
| `int/regression/1851-rv64-phi-signext` | corpus |
| `int/regression/1852-rv64-self-host` | link |
| `int/regression/1923-comptime-field-gate` | unit |
| `int/regression/1925-global-aggregate-byval` | corpus |
| `int/regression/1935-fused-cmp-branch` | corpus |
| `int/regression/1936-block-fallthrough` | corpus |
| `int/regression/1937-loop-rotation` | corpus |
| `int/regression/2092-inline-result-agg-copy` | corpus |
| `int/regression/2163-instantiation-breadth` | unit |
| `int/regression/2163-instantiation-depth` | unit |
| `int/regression/2163-instantiation-mutual` | unit |
| `int/regression/2163-revalidation-breadth` | unit |
| `int/regression/2163-revalidation-depth` | unit |
| `int/regression/2178-intrinsic-type-operand` | unit |
| `int/regression/2204-licm` | corpus |
| `int/regression/2211-loop-rotation-scope` | corpus |
| `int/regression/2231-cross-module-inline-asm` | gone |
| `int/regression/2231-cross-module-inline-shape` | gone |
| `int/regression/2248-constant-pool` | corpus |
| `int/regression/2254-pack-enclosing-body` | unit |
| `int/regression/2257-generic-instance-fold` | unit |
| `int/regression/2274-negative-zero` | corpus |
| `int/regression/2342-generic-instance-value` | unit |
| `int/regression/2355-linked-list` | corpus |
| `int/regression/2355-recursive-type` | unit |
| `int/regression/2357-narrow-truncation` | corpus |
| `int/regression/2368-deep-nest` | corpus |
| `int/regression/2369-panic-exit` | gone |
| `int/regression/2373-secret-repr-classify` | unit |
| `int/regression/2376-fields-each-per-element` | corpus |
| `int/regression/2402-raw-bss-fill` | link |
| `int/regression/2509-resource-missing` | link |
| `int/regression/2509-resources-nonwindows-gated` | gone |
| `int/regression/2523-unreachable-fallthrough-return` | corpus |
| `int/regression/2532-step-clobbers-module-obj` | gone |
| `int/regression/2536-length-of` | unit |
| `int/regression/2543-bsd-ar-extended-name` | link |
| `int/regression/2558-dep-step-home` | link |
| `int/regression/2654-vector-bitcast-scalarize` | corpus |
| `int/regression/2670-bitcast-value-address` | corpus |
| `int/regression/2672-record-reinterpret-mismatch` | unit |
| `int/regression/2672-record-reinterpret` | unit |
| `int/regression/2689-asm-operand-reach` | unit |
| `int/regression/2691-reflection-recursion` | unit |
| `int/regression/2693-pointee-of` | unit |
| `int/regression/2694-is-secret` | unit |
| `int/regression/2725-frame-slot-packing` | corpus |
| `int/regression/2735-overaligned-stack-local` | corpus |
| `int/regression/2749-gap-op-loop-carried` | corpus |
| `int/regression/2751-const-index-bounds` | unit |
| `int/regression/2759-riscv64-fbreg-bias` | link |
| `int/regression/2764-comptime-ident-identity` | unit |
| `int/regression/2764-gate-runtime-binding` | unit |
| `int/regression/2764-shadowed-comptime-binding` | unit |
| `int/regression/2777-riscv-abi-lp64` | link |
| `int/regression/2777-riscv-abi-lp64d` | link |
| `int/regression/2777-riscv-abi-lp64f` | link |
| `int/regression/2791-asm-noreturn` | LOST |
| `int/regression/2800-cascade-library-attribution` | link |
| `int/regression/2800-pin-to-absent-library` | link |
| `int/regression/2800-pin-to-static-entry-with-dynlib` | link |
| `int/regression/2800-pin-to-static-entry` | link |
| `int/regression/2812-int-to-fp-immediate` | corpus |
| `int/regression/constagg-fold-leaves` | corpus |
| `int/regression/each-const-array` | corpus |
| `int/regression/fin-block-scope` | unit |
| `int/regression/fin-ret-order` | unit |
| `int/surface/aggregate-abi` | corpus |
| `int/surface/asm-name-large-frame` | gone |
| `int/surface/asm-symbol-operands` | gone |
| `int/surface/bmos` | link |
| `int/surface/c-variadic` | link |
| `int/surface/const-pool` | corpus |
| `int/surface/control-flow` | corpus |
| `int/surface/ct-flags-hardware` | LOST |
| `int/surface/debugger-gdb` | link |
| `int/surface/debuginfo` | link |
| `int/surface/declassify` | unit |
| `int/surface/elf-gotpcrelx` | link |
| `int/surface/elf-pie` | link |
| `int/surface/elf-relro` | link |
| `int/surface/embed-dedup` | link |
| `int/surface/embed-freestanding` | link |
| `int/surface/embed` | gone |
| `int/surface/float-emit` | corpus |
| `int/surface/float` | corpus |
| `int/surface/frame-elision` | corpus |
| `int/surface/freestanding-aarch64` | link |
| `int/surface/freestanding-riscv64` | link |
| `int/surface/freestanding` | link |
| `int/surface/macho-abs-bind-x86_64` | link |
| `int/surface/macho-entry-pie-x86_64` | link |
| `int/surface/macho-entry-static-x86_64` | link |
| `int/surface/macho-foreign-aarch64` | link |
| `int/surface/macho-foreign-x86_64` | link |
| `int/surface/macho-framing-pie` | link |
| `int/surface/macho-framing-static` | link |
| `int/surface/macho-got-x86_64` | link |
| `int/surface/macho-header-span-aarch64` | link |
| `int/surface/macho-header-span-x86_64` | link |
| `int/surface/macho-import-claim` | link |
| `int/surface/macho-mod-init` | link |
| `int/surface/macho-named-sections` | link |
| `int/surface/macho-pie-aarch64` | link |
| `int/surface/macho-pie-x86_64` | link |
| `int/surface/narrow-stack-args` | link |
| `int/surface/oblivious` | unit |
| `int/surface/packed-atomic-refused` | unit |
| `int/surface/packed-byte-image` | corpus |
| `int/surface/pe-aslr` | link |
| `int/surface/pe-dllimport-indirection` | link |
| `int/surface/pe-foreign-codeview` | link |
| `int/surface/pe-foreign-unwind` | link |
| `int/surface/pe-imp-alias-refused` | link |
| `int/surface/pe-import-attribution` | link |
| `int/surface/pe-import-claim-cascade` | link |
| `int/surface/pe-import-claim` | link |
| `int/surface/pe-import-library` | link |
| `int/surface/pe-local-import-pointer` | link |
| `int/surface/pe-mingw-static-crt` | link |
| `int/surface/pe-resources-native` | link |
| `int/surface/pe-resources` | link |
| `int/surface/pie-reloc` | link |
| `int/surface/pie-relro-fault` | link |
| `int/surface/recursion-peel` | corpus |
| `int/surface/reduce` | corpus |
| `int/surface/riscv-pcrel-pair` | link |
| `int/surface/sign-extension` | corpus |
| `int/surface/spirv-builtin-type` | unit |
| `int/surface/spirv-entrypoint` | link |
| `int/surface/spirv-flat-conflict` | unit |
| `int/surface/spirv-flat` | link |
| `int/surface/spirv-float` | corpus |
| `int/surface/spirv-interface` | link |
| `int/surface/spirv-location-overlap` | unit |
| `int/surface/spirv-module` | corpus |
| `int/surface/spirv-recursion-refused` | unit |
| `int/surface/spirv-sampled-image` | link |
| `int/surface/spirv-sampler-binding-refused` | unit |
| `int/surface/spirv-shader-only` | unit |
| `int/surface/spirv-signature-width` | unit |
| `int/surface/spirv-storage` | link |
| `int/surface/spirv-uniform-layout` | unit |
| `int/surface/spirv-unsupported-names-opcode` | unit |
| `int/surface/spirv-vector-literal` | corpus |
| `int/surface/spirv-vector-narrow` | corpus |
| `int/surface/spirv-vector-wide` | corpus |
| `int/surface/spirv-vector` | corpus |
| `int/surface/step` | gone |
| `int/surface/shell` | gone |
| `int/surface/posix` | gone |
| `int/surface/step` | gone |
| `int/surface/shell` | gone |
| `int/surface/windows` | gone |
| `int/surface/symtab-pie` | link |
| `int/surface/symtab-shared` | link |
| `int/surface/symtab` | link |
| `int/surface/syscall-inline-asm` | LOST |
| `int/surface/unaligned-access` | corpus |
| `int/surface/unaligned-vector` | corpus |
| `int/surface/vector-abi` | corpus |
| `int/surface/vector-carry-and-boundary` | corpus |
| `int/surface/vector-lanes` | corpus |
| `int/surface/vector-mul-packed-4s` | corpus |
| `int/surface/vector-mul-scalar-shortfall` | corpus |
| `int/surface/vector-mul-widths` | corpus |
| `int/surface/vector-subwidth-direct` | corpus |
| `int/surface/vector-subwidth` | corpus |
| `int/surface/vector-wide` | corpus |
| `int/surface/vectorize-emit` | corpus |
| `int/surface/vectorize-reassoc-emit` | corpus |
| `int/surface/vectorize-reassoc` | corpus |
| `int/surface/vectorize` | corpus |
| `int/surface/vectors` | corpus |


octalide:
Reviewed and merged `dev` in. Locally green: 1470 unit tests, and the corpus runs 1404 cells pass / 0 fail on this host.

Two things fixed on the branch:

- The changelog claimed "the classification of all 174 is in the pull request" and no such classification existed. The counts were also wrong: the suite held **176 case directories under 174 distinct names** (`shell` and `step` each existed once under `regression` and once under `surface`), and the split is 65 moved, 57 corpus, 37 unit, 14 gone, 3 lost. The per-case ledger is now in the PR body and the changelog numbers match it.
- #2944 files the three lost cases plus the `int/ct/` timing harness for restoration as host-executed unit tests.

**Blocked on #2948.** All three `targets on <runner>` legs refuse to start because no runner satisfies `test/tools.lock`: the pins are llvm major 22 and spirv-tools 2026.3, ubuntu-latest ships llvm 18.1.3 and spirv-tools 2025.1, and windows-latest carries llvm 20.1.8 and no spirv-tools. The refusal is the driver working as designed. Relaxing the pin is the wrong direction, since the goldens were blessed against the pinned decoders. CI has to install them.

Issue #2957, opened 2026-08-09
link: mach-o has no external reader on the pr lane, so mach is its own only reader between releases


octalide:
Mach-O is written by mach and read back only by mach between releases, because the only external reader of a mach Mach-O runs at `main` cadence.

## Repro

1. `test/engines.conf` gives `x86_64-darwin` the runner `macos-15-intel` and `aarch64-darwin` the runner `macos-15`, both with cadence `main`.
2. `test/ci-legs.sh pr` therefore emits no darwin leg, and `.github/workflows/ci.yml` schedules none on a pull request.
3. Corpus layer A is what runs `llvm-readobj --file-header --sections --relocs` over a mach-produced object. On a pull request it runs over ELF and COFF only.
4. So a change to `src/lang/target/of/macho.mach` can be merged to `dev` with no external tool having parsed its output. The first external read is the `dev` to `main` release merge.

## Expected

Any object format mach writes has an external structural reader on the lane that gates merges into `dev`, since that is where a format defect is introduced and where it is cheap to find.

## Actual

Mach-O's external reader runs only at release-merge cadence. Between releases mach is effectively its own only reader of the format it emits.

**This is a statement about the cadence column, not about Mach-O.** Any surface whose sole external oracle sits on a `main`-cadence row has the same property, and the cadence column is the thing that decides it.

COFF is the demonstrated instance and the reason to take this seriously rather than treat it as theoretical. `x86_64-windows` had a leg, but that leg had never once started (#2948), and a linux developer host declines the target so nothing local read it either. The very first layer A run over it found #2952: mach wrote the PE/COFF long section-name form right-justified and space-padded, `"/      4"`, where the form is a slash followed by the decimal and nothing else. binutils tolerates the padding, so mach's own tests and every local `objdump` were happy; `llvm-readobj` and `llvm-objdump` both answer `invalid section name` and decode nothing. Every windows object carrying a section name over eight bytes was unreadable, and `.rodata.cst` is eleven, so that is every object with a constant pool. It was invisible for as long as it was because no external reader had ever looked. Mach-O is in that state today and darwin is a shipping target.

Related to #2956, which is the same shape in the third direction: a format with no external reader configured at all.

## Environment

- Branch `dev` at d15c5c28, corpus as merged in #2943.
- `test/engines.conf` rows `x86_64-darwin` and `aarch64-darwin`, cadence `main`.
- The metered-runner reason for that cadence is real and this issue does not propose ignoring it.

[...]

Issue #2956, opened 2026-08-09
test: corpus layer A passes a raw object format on a check that cannot fail


octalide:
## What

Corpus layer A reports `PASS` for a `raw` object format on a check that cannot fail.

`test/lib/layers.py`:

```python
if fmt == "raw":
    return Outcome(True, "raw image, %d byte(s)" % os.path.getsize(paths[-1]))
```

Every other format is handed to an external oracle: ELF, COFF and Mach-O go to `llvm-readobj --file-header --sections --relocs` and then have their class, machine and endianness compared against the registry row, and SPIR-V goes to `spirv-val`. The `raw` branch asserts only that the file exists and is non-empty, which the caller already checked immediately above it, and then records a passing cell.

That verdict overstates what was checked. The corpus's own rule, in `test/README.md`, is that a skip is a reviewed statement about coverage and a cell that nothing verifies says so rather than passing quietly. A green cell that cannot go red is worse than no cell, because the matrix is what the project reads to answer which surfaces are covered.

Today the only `raw` row is `mos6502`, whose SKIPS file declares `* ab` for a compiler reason, so no run currently reaches this branch. That is exactly why it should be fixed now rather than when it starts being reached: the day mos6502 builds, the layer A column turns green across 91 cases while nothing has inspected a single byte.

This was found while adding the `x86_64-windows` column, whose first layer A run turned up #2952: a COFF long section name written right-justified and space-padded, which every llvm tool refused to decode. The general shape is a format whose only reader is mach.

## Fix

The defect is the verdict, not the missing decoder. Whether mos6502 ever gets an external decoder is a separate and lower-priority question.

Preferred: layer A returns a skip verdict for a format it has no oracle for, with the reason, so the cell reads `SKIP` and the matrix counts it as uncovered rather than as covered.

That means `layer_a` needs a third outcome besides ok and not-ok. `Outcome` gains a state (or a `SKIP` sentinel the caller maps to `m.add(..., "SKIP", ...)`), and the `raw` branch produces it with a reason naming the format and that no external structural oracle is wired for it. `driver.gate` already treats an uncovered cell as a hole unless a `SKIPS` entry covers it, so a target that grows a `raw` layer A column with no declared skip fails the run, which is the correct pressure.

Alternative considered and rejected: declaring `a` in `golden/mos6502/SKIPS` alongside the existing entry. That hides the problem for exactly one target rather than fixing the layer, and the next `raw` row inherits the false pass.

## Acceptance

[...]

Issue #2952, opened 2026-08-09
link(coff): a long section name is space-padded, and llvm refuses the object


octalide:
Every `x86_64-windows` object carrying a section name longer than eight bytes is written with the long-name field padded the wrong way, and llvm refuses to read it. `mach build` reports success.

## Repro

Any build that emits a constant pool, since the pool's section is `.rodata.cst`, eleven bytes:

```sh
mach build . --target windows-x86_64
llvm-readobj --sections <any object with a float or vector constant>
llvm-objdump -h <same>
```

## Expected

The object parses. The PE/COFF long section-name form is a slash followed by an ASCII decimal offset into the string table, so a name at offset 4 is the eight bytes `/4` and six NULs.

## Actual

```
llvm-readobj: error: 'vec_f32x3.o': invalid section name
llvm-objdump: error: 'vec_f32x3.o': invalid section name
```

The field holds `"/      4"`: the decimal is right-justified in the seven bytes after the slash and space-padded on the left. `write_decimal` in `src/lang/target/of/coff.mach` was written that way and is used by both section-name call sites, the object writer and the PE image writer.

binutils reads it anyway, which is why nothing noticed:

```
$ objdump -h vec_f32x3.o
  1 .rodata.cst   00000004  ...
$ llvm-objdump -h vec_f32x3.o
llvm-objdump: error: 'vec_f32x3.o': invalid section name
```

llvm parses the bytes after the slash as a decimal integer with no leading-space allowance, so the whole object is rejected rather than the one field.

## Environment

- `feat/2903-codegen-corpus`, mach 4.18.0 built at `out/linux-x86_64/debug/bin/mach`
- llvm 22.1.8, binutils 2.46, linux x86_64 cross-building `x86_64-windows`

## How it was found, and what it means for coverage

The codegen corpus (#2903) has no `x86_64-windows` goldens because the developer host declines the target, and its CI leg has never started, so nothing had ever run an external reader over a mach COFF object. The first layer A run on that target refused all 91 cases on this. It is the same shape as the ELFCLASS32-on-Elf64 defect that motivated layer A: a writer that only its own reader ever checked.

## Acceptance

- `llvm-readobj --sections` and `llvm-objdump -d` parse an object whose section name exceeds eight bytes.
- The field is pinned by byte in a unit test, since both readers disagreeing about padding is exactly what went unnoticed.
- The corpus's `x86_64-windows` column reports layers A and B rather than refusing.

[...]

PR #2958, opened 2026-08-09
ci: install the pinned oracles, and let the registry assign corpus legs

### LINKED ISSUE #2948: infra(ci): no runner satisfies test/tools.lock, so every corpus leg refuses to start
Blocks #2943.

## What

`test/tools.lock` pins the external oracles the corpus is grounded in, and the driver
refuses to start rather than produce a result an unpinned tool could have shaped. That
refusal is correct and is working. The problem is that no CI runner can satisfy the
pins, so all three `targets on <runner>` legs refuse to start:

| runner | has | tools.lock wants |
|---|---|---|
| ubuntu-latest | llvm 18.1.3, spirv-tools 2025.1 | llvm major 22, spirv-tools exact 2026.3 |
| ubuntu-24.04-arm | llvm 18.1.3, spirv-tools 2025.1 | same |
| windows-latest | llvm 20.1.8, no spirv-tools at all | same |

`ci.yml` installs `llvm binutils spirv-tools gdb` from apt, which is whatever the
runner image ships, and installs nothing at all on Windows. The lock was recorded on a
developer host carrying llvm 22.1.8 and SPIRV-Tools v2026.3, and its own header says
so.

## Why it is not fixed by relaxing the pin

The pin is what makes a golden reproducible. 241 goldens were blessed against llvm 22
disassembly and spirv-dis 2026.3 output, and a decoder several majors older prints
different text for the same bytes, so relaxing the rule does not make the legs pass, it
makes them fail differently. Re-blessing against ubuntu's llvm 18 would then make every
developer running a current toolchain mismatch locally, which inverts the problem
rather than solving it.

## Fix

CI installs the pinned versions, reading the wanted version from `test/tools.lock` so
there is one source of truth, as the file's header already promises.

- **LLVM on Linux**: the runner images do not carry the pinned major, so install it
  from `apt.llvm.org` (`llvm.sh <major>`) and put the versioned binaries on PATH under
  their unsuffixed names. Both ubuntu runners.
- **SPIRV-Tools**: ubuntu ships 2025.1 and the pin is 2026.3, so apt cannot serve it.
  It needs a pinned release download. Decide the source deliberately rather than
  reaching for whichever URL responds: the Vulkan SDK and the SPIRV-Tools CI artifact
  bucket are the two candidates, and one of them has to be the recorded one.
- **Windows**: carries neither. Either install both there too, or decide that the
  windows leg builds the `x86_64-windows` column only and the `spirv` column belongs to
  a Linux leg. `engines.conf` already assigns `spirv` the `ubuntu-latest` runner, and
  the corpus driver selects it on windows-latest anyway because it selects by host
  capability rather than by the row's runner column. Those two rules disagree, and the
  registry is supposed to be the single source of truth, so settle which one decides.

## Acceptance

- All three `targets on <runner>` legs run the corpus rather than refusing to start.
- The wanted versions come from `test/tools.lock`, not from a second copy in the
  workflow.
- A runner that cannot be brought to the pin fails loudly with what is missing, which
  is today's behaviour and should stay.
- Whichever rule decides the `spirv` column is the only rule that decides it.

octalide:
Closes #2948.

The oracle-install half of #2948 merged with #2943 (`source`, `alias` and `no-cflags` rows in `tools.lock`, `test/ci-tools.sh`, `run.sh --tools`), and all three `targets on <runner>` legs now run. **This PR is the other half**: which rule decides what a leg covers, plus writing down the one constraint that decision rests on.

## Which rule decides what a leg covers

**The registry decides, and it is now the only rule that does.**

The driver selected by host capability. A target with `engine none` needs no execution host, so `spirv`, `riscv32` and `mos6502` passed that test on every runner and were recomputed on windows and arm as well as on the one `engines.conf` assigns them. Two rules read one file and disagreed, and for spirv the pinned decoder has no build for either of those hosts at all.

Assignment and capability are now separate questions and only one of them selects:

- The `runner` column is the **assignment**. `run.sh --runner <label>` covers exactly the rows the registry gives that label.
- Host capability is a **check**, not a second selector. A row assigned to a host that cannot serve it is a refusal naming the row and the reason, never a silent substitution of a set the host can serve.
- A row the runner does not own is **named** in the run header as `not assigned`, so a leg a run was never applicable to and a leg it was quietly dropped from do not look alike. That is the distinction `test/README.md` already claims the corpus keeps, and the driver was on the wrong side of it for spirv on windows: it added a column the registry never assigned.

Capability is checked against layer C only, since that is the layer needing an engine. That is what keeps one host able to bless every column's goldens, which is how the four columns no developer host can execute got theirs in #2943.

Each CI job passes **the label of the runner it is already executing on** and no target list at all. A workflow that names a leg is a workflow that can claim a leg it never ran, so the corpus step is `run.sh --runner "${{ matrix.runner }}"`, `ci-tools.sh` takes the same label, and the link suite still takes no `--leg`. `link-pin` loses its hand-written `x86_64-linux spirv` for the same reason: `ubuntu-latest` is what gives it the spirv row.

**This removes the need for spirv-tools on windows and darwin entirely**, and that is the reason for the change rather than a side effect of it. What it costs is an incidental cross-host check that a compiler built on arm or windows emits the same spirv module as one built on x86_64. The corpus never claimed that, and if it is wanted it belongs in an explicit comparison rather than in a selection accident.

`run.sh` with no assignment still runs everything the host can serve. That is the right question on a developer machine.

## The spirv constraint is now recorded where the registry lives

The spirv row's runner is a **constraint, not a preference**, and that was knowledge held in a pull request body:

- SPIRV-Tools' newest release tag is v2026.2 and it attaches no binaries to any release. Nothing upstream serves 2026.3 as a release; it is what the tree reports between v2026.2 and the next tag. `tools.lock` now says so beside the pin, because a reader will otherwise go looking for a v2026.3 release and not find one.
- LunarG's apt repository is not a source either: for noble it serves amd64 only, and its newest `spirv-tools` is `2024.4+1.4.313.0`, several SDKs behind.
- Vulkan SDK 1.4.357.0 ships SPIRV-Tools at git tag `vulkan-sdk-1.4.357.0`, commit `9a49b088`, whose binaries report `SPIRV-Tools v2026.3` and reproduce **19 of 19** blessed spirv goldens byte for byte. That reproduction is what makes an exact pin on a between-tags version correct rather than sloppy, and the pin is not widened.
- **The linux SDK is x86_64 only.** Moving the spirv row to `ubuntu-24.04-arm`, or to windows or darwin, needs a source build of SPIRV-Tools at that tag first, and without one the leg refuses at install time. That now sits beside the row in `engines.conf` and beside the pins in `tools.lock`, so it is a statement rather than a mystery install failure.

## Also here

Two findings from the same surface class are filed rather than carried: **#2956** (corpus layer A reports PASS for a `raw` object format on a check that cannot fail) and **#2957** (Mach-O has no external reader on the pr lane, so mach is its own only reader between releases, which is exactly why #2952 sat as long as it did).

## Verification

Local, on a linux x86_64 host, before CI:

```
$ bash test/run.sh --tools --runner windows-latest
cc cc min 9 - -
llvm-objdump llvm-objdump major 22 llvm 22.1.8
llvm-readobj llvm-readobj major 22 llvm 22.1.8
```

No `spirv-dis` and no `spirv-val`, which is the cheapest correct outcome stated explicitly.

```
$ bash test/run.sh --runner ubuntu-24.04-arm ; echo "exit=$?"
corpus: this host cannot serve 1 row(s) it was assigned, and layer C needs an engine:
  aarch64-linux    engines.conf names engine native, which needs a linux/aarch64 host, and this is linux/x86_64
install what each row names, or take the row off this runner in engines.conf. the driver does not quietly run a different set.
exit=2
```

It refuses naming the row rather than running the other rows it could serve. `--runner` together with `--target` is also refused, since two assignments over one registry is the same defect in miniature.

A full local run under the new selection reproduces the merged numbers exactly:

```
cases:    91   targets: x86_64-linux,riscv64-linux,riscv32,spirv,mos6502   layers: abc
runner:   ubuntu-latest    assignment from engines.conf, not host capability
not assigned: aarch64-linux    engines.conf assigns it to ubuntu-24.04-arm
not assigned: x86_64-windows   engines.conf assigns it to windows-latest
not assigned: x86_64-darwin    engines.conf assigns it to macos-15-intel
not assigned: aarch64-darwin   engines.conf assigns it to macos-15
cells: 1404 pass, 0 fail, 663 skip
```

## Leg output

Filled in below from the CI run on this branch.


octalide:
Leg outputs from run 31292210313, pulled from the logs rather than reported:

```
targets on ubuntu-latest
  cases: 91  targets: x86_64-linux,riscv64-linux,riscv32,spirv,mos6502  layers: abc
  not assigned: aarch64-linux    engines.conf assigns it to ubuntu-24.04-arm
  not assigned: x86_64-windows   engines.conf assigns it to windows-latest
  not assigned: x86_64-darwin    engines.conf assigns it to macos-15-intel
  not assigned: aarch64-darwin   engines.conf assigns it to macos-15
  cells: 1404 pass, 0 fail, 663 skip

targets on ubuntu-24.04-arm
  cases: 91  targets: aarch64-linux  layers: abc
  not assigned: x86_64-linux, riscv64-linux, riscv32 -> ubuntu-latest
  not assigned: x86_64-windows -> windows-latest, both darwin rows -> macos runners

targets on windows-latest
  cases: 91  targets: x86_64-windows  layers: abc
  not assigned: x86_64-linux, riscv64-linux, riscv32 -> ubuntu-latest
  not assigned: aarch64-linux -> ubuntu-24.04-arm, both darwin rows -> macos runners
```

1404 / 546 / 546 across the three legs, zero fail, and the ubuntu-latest numbers reproduce the merged #2943 run exactly, which is the control that says the assignment rule changed who runs what and not what any leg computes.

The line that matters is `not assigned: <row>  engines.conf assigns it to <runner>`. A row a runner does not own now states which runner does, so "never applicable here" and "quietly dropped" no longer look alike from the log. That was the defect the old host-capability selection had in the other direction: it silently *added* the spirv column on windows, a row the registry never assigned and whose pinned decoder has no windows build at all.

Merged `dev` in for #2951. Marking ready.

PR #2960, opened 2026-08-09
docs(manifest): a cross-platform binary needs one artifact per extension convention

### LINKED ISSUE #2959: docs(manifest): an executable extension is literal, so a cross-platform binary cannot use targets = ["*"]
Raised by #2955.

## What

`[artifact].out` is one literal string shared across every target the artifact builds
for, and `doc/manifest.md` says an executable extension "is written literally here".
Both are true and deliberate. What the doc does not say is the consequence: **a binary
that ships on Windows and anywhere else cannot use `targets = ["*"]`.**

`bin/app` produces a Windows file with no extension, which Windows will not execute
until it is renamed by hand. `bin/app.exe` gives every non-Windows target a binary
called `app.exe`. There is no single string that is right for both, so the author must
split into two artifacts with disjoint `targets` lists.

That is the correct answer and it works today with no compiler change. It is just
undocumented, so the first time anyone meets it is when a Windows user reports a binary
that will not run.

**mach's own repository has the bug and cannot see it.** `mach.toml` declares
`out = "bin/mach"` with `targets = ["*"]`, so `mach build .` on Windows produces
`bin/mach`. CI never notices because every Windows build passes `-o` explicitly
(`./b.exe build . -o c.exe` in `ci.yml`), so the manifest's own `out` is never what
produces the Windows binary. The project works around its own defect, which is why it
took an outside report to surface it.

## Fix

Documentation and the repository's own manifest. **No compiler change, and no automatic
extension**: mach must not append `.exe` behind the author's back, and a template
variable for the extension was considered and rejected as too magic for what it buys.

1. `doc/manifest.md`, on the `out` row: state that the extension is literal AND that
   this means a cross-platform executable needs one artifact per extension convention,
   so `targets = ["*"]` does not work for it. Show the two-stanza form:

   ```toml
   [artifact.app]
   kind = "bin"
   entry = "main.mach"
   out = "bin/app"
   targets = ["linux-x86_64", "darwin-aarch64"]
   link = []
   need = []

   [artifact.app-windows]
   kind = "bin"
   entry = "main.mach"
   out = "bin/app.exe"
   targets = ["windows-x86_64"]
   link = []
   need = []
   ```

2. Say plainly what it costs, because a reader who hits it later will think it is a bug:
   the two stanzas differ only in `out` and `targets` and are otherwise duplicates, so a
   `link` or `need` added to one and not the other diverges on Windows only. And every
   new target must be added to the right list by hand, since neither can use `*`.

3. Note that the artifact **name** differs between the two, so `$bin.name` differs by
   platform. Nothing in mach's source reads it, but a project that does will see
   `app-windows` on Windows.

4. Fix `mach.toml` the same way, so the repository demonstrates its own documented
   answer instead of relying on CI's `-o`.

## Acceptance

- `doc/manifest.md` states the `["*"]` restriction on the `out` row, not only that the
  extension is literal.
- The two-stanza form appears in the doc with its drift and `$bin.name` caveats.
- `mach build .` on Windows, with no `-o`, produces a runnable `mach.exe` from the
  repository's own manifest.
- No compiler change: no automatic extension, no new template variable.

octalide:
Closes #2959

`[artifact].out` is one literal string shared by every target in `targets`, so an executable that ships on Windows and anywhere else cannot use `targets = ["*"]`. `doc/manifest.md` now states that on the `out` row and adds a section showing the two-stanza form with what it costs: the stanzas are duplicates apart from `out` and `targets` so they drift on Windows only, every new target must be listed by hand since neither can use `*`, and the artifact name differs so `$bin.name` differs by platform.

`mach.toml` is split the same way, so the repository demonstrates its own documented answer instead of relying on CI passing `-o`. The non-Windows stanza lists the five non-Windows targets explicitly and `[artifact.mach-windows]` carries `out = "bin/mach.exe"` with `targets = ["windows-x86_64"]`. `entry`, `link`, `need` and `kind` are identical between them.

No compiler change. Nothing under `src/` is touched: no automatic `.exe`, no new template variable.

## Verification

Binary under test, built from this branch:

```
-rwxr-xr-x 1 octalide octalide 6838192 2026-08-08 23:56:01 out/linux-x86_64/debug/bin/mach
```

```
$ ./out/linux-x86_64/debug/bin/mach test .
1473 passed, 0 failed, 1473 total  (753ms)
```

Both artifacts build and land where the split says they should, with no `-o`:

```
$ mach build . --target linux-x86_64
built ./out/linux-x86_64/debug/bin/mach  225 modules  7 MB  in 4.2s

$ mach build . --target windows-x86_64
built ./out/windows-x86_64/debug/bin/mach.exe  225 modules  6 MB  in 4.0s
```

The only CI-side reference to the built compiler path is `test/link/run.sh`, which resolves `out/$hd/debug/bin/mach` for the host. The host is never Windows there, so the non-Windows stanza still serves it and no CI config needed changing.


octalide:
The split surfaced a real defect that the single-artifact manifest hid: **`mach test` on Windows refuses**, filed as #2961 and worked around here.

```
error: mach.toml: artifact "mach" declares no target for this host; select one with --target (linux-x86_64, linux-arm64, linux-riscv64, darwin-x86_64, darwin-aarch64)
```

`build_config` in `src/lang/driver/config.mach` takes `m.artifacts[0]` as the primary artifact for a test build, because a test links the whole source tree rather than one artifact. After the split the first declared stanza is the non-Windows one, so declaration order became load-bearing and no order is right for both platforms. `mach build` is unaffected: it enumerates artifact-by-target cells and filters through `cell_supported`, which is why `build x86_64-windows` passed and `test x86_64-windows` did not.

Fixing that is a driver change, which this pull request is scoped out of, so both Windows `self test` steps pass `--bin mach-windows` with a comment pointing at #2961, and `doc/manifest.md` says so where it recommends the split. Those two flags come out when #2961 lands.

octalide:
Dropped the `mach.toml` split and the two `--bin mach-windows` flags from `ci.yml`. **The documentation ships, the split does not.**

Reason: #2961 is worse than diagnosed, and I verified that rather than trusting the diagnosis. The issue names `load_config_manifest` taking `m.artifacts[0]`, and fixing exactly that produces **byte-identical behaviour** -- the same `artifact 'mach' does not support target 'windows-x86_64'` -- from a compiler I confirmed contains the fix by checking the built binary for the new string rather than assuming the build took.

The refusal actually comes from `config.mach:73 resolve_artifact_path`, which resolves the build unit from the raw selection with no primary applied at all. So "which artifact is primary" is answered independently in at least three places, and only the planner asks whether the artifact supports the target. Fixing one of three is how you get a change that looks right and does nothing.

That makes the split a shape whose own `mach test` does not run on Windows. Shipping it here would mean shipping CI flags whose only purpose is to hide that, in the same change whose documentation recommends the shape to users. The docs keep the limitation and say plainly that mach's own manifest stays single until #2961 lands, and why.

1473 unit tests pass. Ready to merge as a docs-only change.

Issue #3001, opened 2026-08-19
driver: retained analyze_project/dnit_project cycle leaks ~7.4 MiB per call


octalide:
`driver.analyze_project` + `driver.dnit_project` on one retained `Session` grows RSS by ~7.4 MiB per cycle, without bound, whether or not any source changed.

This is the retained-analysis path #2997 added for editor tooling. A language server calls it once per edit, so the leak is proportional to how long an editing session lasts: mach-lsp reaches ~550 MiB after 40 edits of one file and keeps climbing.

## Repro

`repro.mach` drives the same sequence an editor integration does — load the manifest, compose a `GOAL_OBJECTS` request for the primary artifact, `analyze_project`, `dnit_project` — in a loop on one `Session`, against a project whose closure is 222 modules (mach-lsp, which vendors the compiler).

```mach
var s: sess.Session = unwrap_ok[...](sess.init(alloc));
var i: i64 = 0;
for (i < N) {
    val mr: Result[manifest.Manifest, str] = driver.load_manifest(?s, ".");
    var m: manifest.Manifest = unwrap_ok[...](mr);
    var sel: manifest.Selection;          # primary artifact
    manifest.select_primary_artifact(alloc, ?s.interner, ?m, ?sel);
    var cli: request.CliArgs = empty_cli();
    var req: request.BuildRequest = unwrap_ok[...](
        request.compose(alloc, ?s.interner, ?m, ?cli, ".", request.GOAL_OBJECTS, ?sel));

    val pr: Result[driver.Project, outcome.Fail] =
        driver.analyze_project(?s, ?m, ?req, nil, 0);
    var p: driver.Project = unwrap_ok[...](pr);
    driver.dnit_project(?p);              # the documented teardown

    vector.dnit[request.LinkToken](?req.link_tokens);
    vector.dnit[str](?req.lib_dirs);
    manifest.dnit(?m, s.build_alloc);
    i = i + 1;
}
sess.dnit(?s);
```

Peak RSS against round count, release profile, **nothing invalidated between rounds**:

| rounds | peak RSS |
|---|---|
| 1  | 229 MiB |
| 3  | 244 MiB |
| 10 | 294 MiB |
| 25 | 406 MiB |

Slope ~7.4 MiB per cycle, flat across the range. Over 222 modules that is ~34 KB per module per cycle.

[...]

PR #3036, opened 2026-08-21
fix(driver): normalize a module path before overlay and source lookup

### LINKED ISSUE #2998: driver: canonicalize module paths before overlay and source lookup
The retained frontend API accepts filesystem-path overlays, but module loading currently depends on exact raw path spelling. For example, a manifest with `src = "./src"` may produce `<root>/./src/file.mach`, while an editor supplies canonical `<root>/src/file.mach`; the unsaved overlay is silently missed.

Normalize the owned path produced by module FQN/path composition before `overlay_get`, disk read, `load_source`, and SourceFile path registration, using the portable lexical normalizer tracked by briar-systems/mach-std#472. Add regressions for `./src`, `..` segments, Windows drive paths, and retained `analyze_project` overlays.

Blocks briar-systems/mach-lsp#141.

octalide:
Closes #2998. Unblocks briar-systems/mach-lsp#141. Advances the mach-std pin to **0.28.1** for `path.clean` (briar-systems/mach-std#472).

Module loading composed `<root>/<src>/<name>.mach` verbatim, so a manifest carrying `src = "./src"` produced `<root>/./src/beta.mach` while an editor supplies the canonical `<root>/src/beta.mach`. The same file, two strings — so the overlay lookup missed, **the stale on-disk text was parsed instead**, and the unsaved buffer was silently ignored. No error, just yesterday's source.

## Shape

Normalized **once, at the single place the path is composed**. That one string is what the overlay lookup, the disk read, and `load_source`'s SourceFile registration all take, so the three agree by construction rather than by three call sites remembering to call the same helper.

Lexical only — no filesystem I/O — which is the right strength for comparing two spellings of one path, and keeps it off the syscall path in a loop that runs once per module.

## A defect this found in mach-std, immediately

The first build with normalization turned **#3001's steady-state leak tests red**.  worked in a buffer sized for its *input* and returned it untrimmed, so a caller freeing with `str_free` (which releases `str_len + 1`) returned less than was taken. Every other allocating function in that module reserves exactly `str_len + 1`, so `clean` was the odd one out. Fixed as mach-std#478 and released in 0.28.1; the leak tests are flat again.

Worth noting because it is the second time today those assertions have caught something a repo away — a size mismatch that no path test would have seen.

## Verification

- **1500 passed, 0 failed** (1499 before, 1 new).
- **Codegen corpus 1416 pass, 0 fail.**
- The test puts a body that **fails sema** on disk and a clean one in the overlay, so a clean analysis is proof the buffer was read. Asserting the absence of an error is what lets it fail: before the fix the disk text was parsed and it reported.
- **Mutation-checked:** removing the normalization fails it.
- **Fixpoint:** three generations byte-identical. `tests.mach` diff purely additive.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -41,6 +41,29 @@ pack's elements on any contract.
 
 Measured at no cost: 48.95s of build CPU against a 49.14s baseline.
 
+## [Unreleased]
+
+### Fixed
+
+#### driver: a module path is normalized before overlay and source lookup (#2998)
+
+Module loading composed `<root>/<src>/<name>.mach` verbatim, so a manifest carrying
+`src = "./src"` produced `<root>/./src/beta.mach` while an editor supplies the
+canonical `<root>/src/beta.mach`. The same file, two strings - so the overlay lookup
+missed, the **stale on-disk text was parsed instead**, and the unsaved buffer was
+silently ignored. No error; yesterday's source.
+
+The composed path is normalized once at the single place it is built, using
+`path.clean` from mach-std 0.28.1. That one string is what the overlay lookup, the
+disk read, and `load_source`'s SourceFile registration all take, so the three agree by
+construction rather than by three call sites remembering to.
+
+Normalization is purely lexical - no filesystem I/O - which is the right strength for
+comparing two spellings of one path, and keeps it off the syscall path in a loop that
+runs once per module.
+
+Unblocks briar-systems/mach-lsp#141.
+
 ## [4.23.0] - 2026-08-21
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,9 +54,16 @@ missed, the **stale on-disk text was parsed instead**, and the unsaved buffer wa
 silently ignored. No error; yesterday's source.
 
 The composed path is normalized once at the single place it is built, using
-`path.clean` from mach-std 0.28.1. That one string is what the overlay lookup, the
-disk read, and `load_source`'s SourceFile registration all take, so the three agree by
-construction rather than by three call sites remembering to.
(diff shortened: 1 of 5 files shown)


octalide:
## Windows CI found the half-fix

Both windows legs failed on `self test` — the new regression test, on a separator mismatch. That is worth more than the red check: it says the first commit **moved** the defect instead of removing it.

Normalizing in `build_module_path` canonicalizes the half the compiler controls. The overlay table compares keys with `str_equals`, and **the two sides that meet in it pick their spelling independently** — an editor keys a buffer from a URI it decoded, the driver composes a path out of manifest text. So the compiler emitting a canonical `src\\a.mach` and the editor supplying `src/a.mach` is the same miss as before, wearing different clothes. One producer agreeing with itself is not agreement.

The key is now canonicalized **at the table's own boundary** — `set_overlay`, `clear_overlay` and `overlay_get` alike — so it is canonical by construction rather than by every producer remembering to. `build_module_path` keeps its normalization for a different reason: the disk read and `load_source`'s SourceFile registration need one spelling so one file gets one FileId.

The new test asserts the property directly and uses the equivalences an editor and a manifest actually produce between them, not synthetic ones: a `./` from `src = "./src"`, a doubled separator from joining a root that already ended in one, and a `..` from a relative include. It also pins the two things that must *not* collapse — a genuinely different file, and a `..` that climbs out of the root it started in — and that setting through a second spelling **replaces** rather than shadowing, since two entries for one file would make which text wins depend on scan order.

Mutation-checked twice: removing canonicalization entirely fails it, and canonicalizing on `set` but not on `get` — precisely the half-fix the first commit was — fails it too.

1501/0, corpus 1416/0, fixpoint byte-identical.

Issue #2992, opened 2026-08-11
Calling cfltk+fltk libs


LukyGuyLucky:
As said prevously,calling sqlite3 in mach is great;
Calling raylib in mach is great as well.

The cfltk bindings from [https://github.com/MoAlyousef/cfltk](url) is workable in other langauges.When I tried in mach,got below error message,I have no clue what it says:
ms vc2022 libs:
```
error: coff: unsupported relocation type 8
```
msys64/ucrt libs:
```
E:\Learning\machpack\ex\cfltk>mach build . -L . -l cfltk2 -l fltk -l fltk_gl -l fltk_images -l fltk_forms -l fltk_jpeg -l fltk_png -l fltk_z -L e:/learning/codeblocks/msys64/ucrt64/lib  -l user32 
-l shell32 -l gdi32 -l gdiplus -l comdlg32 -l comctl32 -l opengl32 -l glu32 -l imm32 -l msimg32 -l uuid -l winspool -l comctl32
error: coff: unsupported relocation type 5
```
main.mach:
```
use std.runtime;
use p: std.print;

use fl:cfltk.cfltk;
use cfltk.cfltk.Fl_Window;
use cfltk.cfltk.Fl_Button;
use cfltk.cfltk.Fl_Widget;

#[symbol("main")]
fun main(argc: i64, argv: **u8) i64 {
    p.println("Hello, World!");

    fl.init_all();
    fl.register_images();
    fl.lock();
    
    var win:*Fl_Window=fl.window_new(100,100,400,300,"नमस्ते दुनिया!👋🌎!");
    var btn1:*Fl_Button=fl.button_new(160,210,80,40,"(&O)K");

    fl.window_end(win);
    fl.window_show(win);

    fl.button_set_callback(btn1,cb,nil);
    
    fl.run();

    ret 0;
}

fun cb(w:*Fl_Widget,data:*u8){
    fl.widget_set_label(w,"Label changed");
}
```
cfltk.mach:
```
#pub def Fl_Widget:*u8;
#pub def Fl_Window:*u8;
#pub def Fl_Button:*u8;
pub rec Fl_Widget{_data_:*u8;}
pub rec Fl_Window{_data_:*u8;}
pub rec Fl_Button{_data:*u8;}

#[symbol("Fl_init_all")]
pub ext fun init_all();
#[symbol("Fl_register_images")]
pub ext fun register_images();
#[symbol("Fl_lock")]
pub ext fun lock() i32;
[...]

Issue #2968, opened 2026-08-09
user32 api and libc  varidic api like printf


LukyGuyLucky:
After a while I tried this again but no lucky.
```
E:\Learning\machpack\ex\b07>mach build . e:/sqlite3/sqlite3.lib e:/learning/codeblocks/msys64/ucrt64/lib/libuser32.a api-ms-win-crt-heap-l1-1-0.dll api-ms-win-crt-string-l1-1-0.dll user32.dll api-ms-win-crt-stdio-l1-1-0.dll

E:\Learning\machpack\ex\b07>mach run .                       <-----------------!!
E:\Learning\machpack\ex\b07>                                        <-----------------!!
```
The significant improvements are:
1. strlen,malloc,free works;
2. sqlite3 libs work great;

What not works well:
1.libc printf ,no matter the signature is``` ext fun printf(msg:*u8) i32``` or ```ext fun printf(msg:*u32,...) i32``` ,compiled successfully but no output/or crashed when running
2.Win32 api like MessageBoxA,same as p1,no output/or crashed when running.

```
use std.runtime;
use p: std.print;
use std.types.size.usize;
use std.types.string.str;
use std.types.result.Result;
use std.types.result.is_err;
use std.types.result.is_ok;

use std.filesystem;
use std.filesystem.STDIN;
use std.input;
use std.input.read_line;

use std.io.reader;
use std.io.reader.Reader;
use std.io.writer.Writer;

use std.system.panic.panic;

use b07.sqlite3;
use b07.sqlite3.Sqlite3;
use b07.sqlite3.sqlite3_stmt;

#[library("user32.dll")]
pub ext fun MessageBoxA(handle:ptr,msg:*u8,caption:*u8,style:i32) i32;
##[symbol("write")]
#pub ext fun libc_write(fd:i64,buf:*u8,n:i64) i64;

#[library("api-ms-win-crt-string-l1-1-0.dll")]
ext fun strlen(s:*u8) i64;
#[library("api-ms-win-crt-heap-l1-1-0.dll")]
ext fun malloc(size:usize) ptr;
#[library("api-ms-win-crt-heap-l1-1-0.dll")]
ext fun free(data:ptr);

#[library("api-ms-win-crt-stdio-l1-1-0.dll")]
ext fun printf(s:*u8) i32;

#[library("sqlite3.dll")]
ext fun sqlite3_libversion() *u8;

#[symbol("main")]
fun main(argc: i64, argv: **u8) i64 {
  
[...]

Issue #3003, opened 2026-08-20
frontend: a function type cannot carry the C-variadic `...` its `ext fun` declaration accepts


octalide:
Related to #3002. Motivated by #2993.

## What

A function *type* cannot carry the trailing `...` that the identical `ext fun` declaration
accepts, so a C-variadic function cannot be named as a value, stored in a variable, or
passed as a callback.

This compiles:

```mach
ext fun printf(fmt: *u8, ...) i32;
```

This does not:

```mach
def Printf: fun(*u8, ...) i32;
```

```
error: C-style variadic `...` was removed in v2.0.0
```

The message is a leftover. The C-variadic form was not removed, it was *narrowed* to `ext`
declarations and is documented in `doc/language/ext-fun.md`. Only the type parser was never
updated, so the diagnostic now tells the user a true-in-2.0 thing that is false today, and
tells it to them at the one spelling that ought to work.

### Why this is small

The flag is already modeled everywhere except the parser that would set it.

- `type.mach:470` carries `c_variadic` on the interned function type, and `type.mach:1442`
  takes it as an `intern_function` parameter.
- It is part of the type's structural identity, hashed at `type.mach:1986` and compared at
  `type.mach:2013`, with `type.mach:1570` stating the rule outright: a variadic and a
  fixed-arity signature of the same shape are different types.
- `me/ir/type.mach:159` carries it through to the IR fn type, interned at `:531` and
  compared at `:955`.
- `fe/sema/check.mach:143` reads it off the *type* rather than off the declaration when
  checking a call, so an indirect call through a C-variadic function type would already be
  checked correctly.
- The ABI layer classifies from the signature, not the callee decl, so an indirect
  C-variadic call needs no backend work.

`fe/parser/expr.mach:1485` and `:1493` are the only two sites that refuse it, and
`fe/parser/decl.mach:1060` is the working model to copy: accept a trailing bare `...` after
at least one fixed parameter, reject a leading bare `...`, and pass `c_variadic: true` into
the intern call.

### What this does not add

No new fact on any target contract, no new ABI axis, no `va_list`, no callee lowering. The
type identity and comparison paths already account for the flag, so the long-term
maintenance surface is zero.

### Note on the motivating report

#2993 reached for this while transcribing raylib's

```c
typedef void (*TraceLogCallback)(int logLevel, const char *text, va_list args);
```

which is a **fixed 3-arity** function, not a variadic one. That binding is #3004's subject
[...]

Issue #2598, opened 2026-08-07
abi: Apple arm64 gives a stack argument its natural size, but mach 8-byte-rounds every one — silent wrong values past eight GP args


octalide:
Found while implementing #2575 (C-variadic `ext fun`). Unrelated to the variadic tail, which is correct — this is the **fixed** argument rule on the same target.

## What

Apple's arm64 ABI gives each stack-passed argument its **natural size and alignment**. The AAPCS64 base standard rounds every stack argument up to an 8-byte slot. Mach applies the AAPCS64 rule on every aarch64 target, including darwin, so a `darwin-aarch64` call whose stack arguments are narrower than 8 bytes places them at the wrong offsets.

Like #2575, the failure is a **silent wrong value**: the callee reads a live offset that the caller wrote a *different argument* into. No link error, no crash.

## Reproduction

The trigger is a call that (a) exhausts the eight GP argument registers and (b) spills at least one argument narrower than 8 bytes. Below, three `i32`s spill:

```mach
ext fun many(a: i64, b: i64, c: i64, d: i64, e: i64, f: i64, g: i64, h: i64,
             i: i32, j: i32, k: i32) i64;

fun main() i64 {
    ret many(1, 2, 3, 4, 5, 6, 7, 8, 9::i32, 10::i32, 11::i32);
}
```

`mach build . --target darwin-aarch64 --emit-asm`:

```
  mov x7, #0x8
  mov x17, #0x9
  str x17, [sp]          # i -> [sp+0]
  mov x17, #0xa
  str x17, [sp, #0x8]    # j -> [sp+8]
  mov x17, #0xb
  str x17, [sp, #0x10]   # k -> [sp+16]
  bl _many
```

## Ground truth

The same call, from C, through each toolchain's own backend:

```c
long long many(long long a, long long b, long long c, long long d,
               long long e, long long f, long long g, long long h,
               int i, int j, int k);
long long call_it(void) { return many(1,2,3,4,5,6,7,8, 9,10,11); }
```

`clang -O1 --target=arm64-apple-macos11` — **4-byte slots, packed**:

```
  mov  w8, #11
  mov  x9, #9
  movk x9, #10, lsl #32   ; 9 and 10 fused into one 8-byte store
  str  w8, [sp, #8]       ; k -> [sp+8]
  str  x9, [sp]           ; i -> [sp+0], j -> [sp+4]
```

`clang -O1 --target=aarch64-linux-gnu` — **8-byte slots**, matching what mach emits:

[...]

PR #3040, opened 2026-08-21
fix(test/link): resolve the host C compiler by name, and run c-variadic on windows

### LINKED ISSUE #3005: test(link): the c-variadic case skips windows, so win64's tail-float rule is never executed against a real C callee
## What

`test/link/cases/c-variadic/case.conf` carries `skip: x86_64-windows`, so the one case that
executes a mach C-variadic call against a **real C callee** never runs on Windows. Win64's
variadic rule is therefore exercised only by the compiler's own unit tests, which check
mach's model of the ABI against itself and cannot catch the model being wrong.

That rule is not a shared one. Win64 duplicates a tail float into the integer register of
the same positional slot, because a `va_arg` reader walks only the integer slots
(`target/abi.mach:660`, `VA_MODEL_HOME`, `float_dup_gp`). If that duplication is wrong or
missing, a `printf("%f", x)` returns garbage and nothing in CI notices.

The skip is honest about why it is there, and the reason has expired:

> exempt windows: `[step]` now runs there (#2578/#2587 fixed it), but this case has not yet
> been verified against the windows-latest runner's own C toolchain (does it expose a bare
> `cc`, and does it target win64 the way this case needs?). Re-enable once that's confirmed
> rather than guessing at it here.

So the question was deferred rather than answered, and it is answerable.

This matters more than the usual coverage gap because the two most recent FFI reports from
users are both Windows: #2968 (libc `printf` producing no output) and #2992 (COFF
relocations). Windows is where mach's foreign-code surface is least verified and most
exercised.

## Fix

Answer the deferred question and re-enable the leg.

`windows-latest` does not expose a bare `cc`. It does have MSVC via the Visual Studio build
tools, and MinGW-w64 `gcc` on PATH. Two candidates:

1. **Teach `test/link/lib/cc.sh` a Windows branch** that resolves a host C compiler in
   preference order (`cc`, `gcc`, `clang`, then `cl` with its different flag spelling), and
   drop the skip. Preferred: the case's whole value is executing against a toolchain that is
   not mach, and MinGW's `printf` is a real `va_arg` consumer.
2. Cross-compile the probe object for win64 from a Linux leg and execute under Wine.
   Rejected: it verifies the object format but not the runner's real libc, and adds a Wine
   dependency to a lane that has none.

Prefer MSVC `cl` if both are present, since UCRT's `printf` is what a mach user on Windows
actually links against, and MinGW's differs.

## Acceptance

- `test/link/cases/c-variadic/case.conf` no longer carries `skip: x86_64-windows`.
- The case builds its probe with the runner's own C toolchain and executes on the
  `windows-latest` leg, matching the golden.
- The comment block records which compiler was resolved and why, replacing the deferred
  question rather than deleting it.
- A deliberate break of `float_dup_gp` in `abi/win64.mach` makes this case fail. Without
  that check the leg proves only that it runs, not that it discriminates.

octalide:
Closes #3005.

`test/link/cases/c-variadic/case.conf` carried `skip: x86_64-windows`, so the one case that executes a mach C-variadic call against a **real C callee** never ran on Windows. Win64's variadic rule was therefore checked only by the compiler's own unit tests — mach's model of the ABI compared against itself, which cannot catch the model being wrong.

That rule is not shared with any other target: win64 duplicates a tail float into the integer register of the same positional slot, because a `va_arg` reader walks only the integer slots (`target/abi.mach`, `VA_MODEL_HOME`, `float_dup_gp`). If that duplication is wrong or missing, `printf("%f", x)` returns garbage and nothing in CI notices.

## The deferred question, answered

The skip was honest about being a deferral, not a decision:

> does it expose a bare `cc`, and does it target win64 the way this case needs?

**No bare `cc`** — and that single missing *name* was the entire obstacle. Neither Git Bash nor the image's Visual Studio install provides one. `cc.sh` assumed the POSIX spelling, which every unix runner has.

So it now resolves the host compiler by name: `cc`, then `gcc`, then `clang`. All three accept the same flags a bare `cc` does, so no case has to know which one answered and no argv is translated. MSVC's `cl` is **named in the failure** rather than half-supported — its `/c` and `/Fo:` are a second flag dialect that would have to be maintained forever for one call site, and every candidate ahead of it produces the same object for this purpose.

The comment block records the answer in place of the question, rather than being deleted.

## Why this one is worth the leg

The FFI reports that reach us are disproportionately Windows — #2968 (libc `printf` producing no output) and #2992 (COFF relocations). Windows is where mach's foreign-code surface is least verified and most exercised.

## Verification

`bash -n` clean. The resolver's branches checked directly: falls through to `gcc` with `cc` absent from PATH, honors `CC`, and exits 1 with a message naming what it looked for when nothing is there.

Locally on `x86_64-linux` the case still passes both profiles and still resolves `cc`, so the unix path is unchanged. Preflight went 92 → 93 case×leg pairs, which is the windows cell now being declared.

**The windows leg's own result is the acceptance, and it can only come from CI** — I have no Windows host. Two things follow, and I'd rather state them than let a green check imply more than it proves:

1. If `gcc`/`clang` turn out not to be on that image's PATH, the failure will say so precisely rather than fail somewhere downstream.
2. A green leg proves it *runs*. #3005 also asks that it *discriminates* — that a deliberate break of `float_dup_gp` makes it fail. The golden reads every tail double back through the runner's real `va_arg` (`mixed n=3 7 8 2500`, `floats n=8 1000 4000 8000`), so it discriminates by construction; I'll confirm it empirically with a throwaway mutation run once this leg is green, and report the result here before merging.

### DIFF
--- a/test/link/cases/c-variadic/case.conf
+++ b/test/link/cases/c-variadic/case.conf
@@ -16,11 +16,21 @@
 # the a bank the probe's `va_arg` actually reads. before that the caller placed
 # them in fa0-fa7 and every float came back as flat 0.
 #
-# exempt windows: `[step]` now runs there (#2578/#2587 fixed it), but this case has
-# not yet been verified against the windows-latest runner's own C toolchain (does it
-# expose a bare `cc`, and does it target win64 the way this case needs?). Re-enable
-# once that's confirmed rather than guessing at it here; win64's variadic rule (a
-# tail float duplicated into its integer register) is otherwise exercised only by
-# the compiler's own unit tests.
+# windows runs since #3005, and the deferred question was answered rather than
+# deleted: windows-latest exposes no bare `cc`, which is the ONLY reason this was
+# exempt. It does carry gcc and clang, both of which take a bare `cc`'s flags, so
+# test/link/lib/cc.sh resolves the host compiler by name instead of assuming one -
+# see host_cc there for why MSVC's `cl` is named in a failure rather than translated.
+#
+# What the leg buys: win64 duplicates a tail float into the integer register of the
+# same positional slot, because a `va_arg` reader walks only the integer slots
+# (target/abi.mach, VA_MODEL_HOME, float_dup_gp). That rule is not shared with any
+# other target, and until this ran it was checked only by the compiler's own unit
+# tests - mach's model of the ABI compared against itself, which cannot catch the
+# model being wrong. The `mixed` and `floats` lines read every tail double back
+# through the runner's real `va_arg`, so a broken duplication is a wrong VALUE here.
+#
+# This matters more than an ordinary coverage gap: the FFI reports that reach us are
+# disproportionately from Windows (#2968, #2992), which is where mach's foreign-code
+# surface is least verified and most exercised.
 run: exec
-skip: x86_64-windows
--- a/test/link/cases/c-variadic/case.conf
+++ b/test/link/cases/c-variadic/case.conf
@@ -17,10 +17,17 @@
 # them in fa0-fa7 and every float came back as flat 0.
 #
 # windows runs since #3005, and the deferred question was answered rather than
-# deleted: windows-latest exposes no bare `cc`, which is the ONLY reason this was
-# exempt. It does carry gcc and clang, both of which take a bare `cc`'s flags, so
-# test/link/lib/cc.sh resolves the host compiler by name instead of assuming one -
-# see host_cc there for why MSVC's `cl` is named in a failure rather than translated.
+# deleted. The answer was not the one the exemption guessed at: the runner's C
+# toolchain was never the obstacle. CI already exports `CC=gcc` (ci-tools.sh, "cc is
+# spelled gcc on this host, per tools.lock"), so a compiler was there all along.
+#
+# WHAT ACTUALLY FAILED WAS THE STEP'S OWN COMMAND. A `[step]` cmd runs through the
+# PLATFORM shell, which on windows is `cmd.exe /s /c` - and cmd.exe cannot execute
+# `../../lib/cc.sh`, so every windows run died with "'..' is not recognized as an
+# internal or external command" before any compiler was consulted. That is a fact
+# about how a step is spelled, not about what the runner has installed, and the rest
+# of this suite already had it right: every `pe-*` case invokes its script as
+# `sh <script>`. These three were the outliers, and now match.
 #
 # What the leg buys: win64 duplicates a tail float into the integer register of the
 # same positional slot, because a `va_arg` reader walks only the integer slots
--- a/test/link/cases/c-variadic/mach.toml
+++ b/test/link/cases/c-variadic/mach.toml
@@ -45,7 +45,7 @@ debug = false
 simd = "scalarize"
 
 [step.probe]
-cmd = "../../lib/cc.sh -c -O1 -fno-stack-protector probe.c -o {project.out}/obj/probe.o"
+cmd = "sh ../../lib/cc.sh -c -O1 -fno-stack-protector probe.c -o {project.out}/obj/probe.o"
 in = ["probe.c"]
 out = ["{project.out}/obj/probe.o"]
 need = []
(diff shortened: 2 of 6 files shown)


octalide:
## CI answered the question, and my diagnosis was wrong

The windows leg went red, and the log says the premise of this PR was mistaken. Worth writing down, because the wrong lesson is easy to draw here.

**The runner's C toolchain was never the obstacle.** CI already exports `CC=gcc`:

```
ci-tools.sh: cc is spelled gcc on this host, per tools.lock
ci-tools.sh: cc (min 9) comes with the runner as gcc
```

So a compiler had been sitting there the whole time. The exemption's guess — "does it expose a bare `cc`?" — was answerable and the answer did not matter.

**What actually failed was the step's own command:**

```
step probe: ../../lib/cc.sh -c -O1 -fno-stack-protector probe.c -o out/.../probe.o
'..' is not recognized as an internal or external command,
operable program or batch file.
```

A `[step]` cmd runs through the **platform** shell, which on windows is `cmd.exe /s /c` (mach-std's `spawn_shell` via `%COMSPEC%`). cmd.exe cannot execute `../../lib/cc.sh`, so the run died before any compiler was consulted. That is a fact about how the step is *spelled*, not about what the runner has installed — and the rest of this suite already had it right: every `pe-*` case invokes its script as `sh <script>`. These three cc.sh cases were the outliers.

Fixed by spelling them the same way. No mach change: a `[step]` running through the platform shell is correct and deliberate (#2578/#2587 landed exactly that), and a Windows user writes cmd there.

## One wrong guess had exempted two cases

`narrow-stack-args` carried the identical exemption, word for word, for the identical false reason. Answering the question once restores both, so this un-skips both. That leg is worth having independently: win64 gives a stack-passed fixed argument its natural size and alignment rather than an eight-byte slot — the same axis Apple arm64 diverges on (#2598) — and the win64 rule was previously taken on trust from the compiler's own unit tests.

Leaving one case fixed and its twin skipped for a reason I had just proven false would have been the worse outcome.

## Kept, with an honest comment

`cc.sh` still resolves `cc`/`gcc`/`clang` by name. It is not what unblocked CI, and the comment now says so: it covers a host where *nothing* sets `CC` — a developer's own Windows machine, which is where a case is most likely to be run by hand and least likely to have the environment CI builds.

Local: c-variadic and narrow-stack-args both pass on `x86_64-linux`, both profiles, 4/4 cells. The windows result is again CI's to give.

The discrimination check I promised is still outstanding and still owed before merge.

octalide:
## The discrimination check: it discriminates, and the result is more interesting than pass/fail

Ran on a throwaway branch (#3042, now closed and deleted) that flipped `float_dup_gp` to `false` in `abi/win64.mach`, with the unit assertion on the flag commented out in the same commit so the **link leg** was the thing being read rather than a predictable unit failure.

The windows leg failed with a golden **diff**, not a build error, on both profiles:

```diff
 none=0
 mixed n=3 7 8 2500
 ptr n=1 4242
-floats n=8 1000 4000 8000
+floats n=8 0 4000 8000
 pair n=1 309
 wide n=1 1234
-indirect n=2 11 500
+indirect n=2 11 0
```

So the case catches a broken win64 tail-float duplication against a real MinGW `va_arg`. #3005's last acceptance criterion is met.

**`narrow-stack-args` passed under the same mutation**, which is the control I wanted: the break hit exactly the case whose rule it belongs to and not its neighbour. A mutation that reddens everything is an apparatus failure; a selective one is a real signal.

## But only two values moved, and that is worth recording

`mixed`'s 2500 did not change, nor `floats` positions 4 and 8. That is correct behavior, not a weak mutation: win64 gives only the **first four positional slots** registers, and there is no duplication to do for a tail float that rides the stack. Counting fixed parameters (`spec`, `out`) into the slot numbering, `mixed`'s double lands past the register slots, and so do most of `floats`'.

The consequence is that this case's sensitivity to `float_dup_gp` rests on **exactly two observed values**, and the golden's printed selection (`out[0]`, `out[3]`, `out[7]`) samples only one of the two register-slot floats in the `floats` line — `out[1]` is equally broken and never printed. The case discriminates, but with less margin than its eight-double line suggests. Filing that as a follow-up rather than widening this PR.

Worth noting the `indirect` line — new in #3003, going through a C-variadic **type** — independently catches it too. It is not load-bearing here (`floats` alone would have failed the run) but it doubles the signal, and it is the only line proving an indirect call places its tail by the variadic rule on win64.

Merging: 17/17 with the windows link leg green, 6 cells and 0 skips.

Issue #3039, opened 2026-08-21
chore(changelog): stray [Unreleased] headings are stranded inside released sections


octalide:
## What

`CHANGELOG.md` carries `## [Unreleased]` headings *inside* already-released sections.
As of `dev` today, `## [Unreleased]` appears between `## [4.23.0]` and `## [4.22.0]`,
and again between `## [4.22.0]` and `## [4.21.1]`. A reader of the published 4.23.0
section sees an "Unreleased" heading in the middle of it, above entries that shipped in
4.23.0.

The cause is the merge workflow, not the release step. Each PR prepends its own
`## [Unreleased]` block rather than appending its `####` entry under the existing one,
so several accumulate at the top between releases. The release step renames only the
**first** one to the version, and every other becomes an orphan heading stranded inside
that release's section.

This is a published artifact - it is what a user reads to find out what changed - so a
heading that says content is unreleased when it shipped two versions ago is wrong in
the way that matters most for the file's whole purpose.

## Fix

Two parts, and the second is what stops it recurring.

1. **Repair the file.** Delete every `## [Unreleased]` heading that sits inside a
   released section, folding its `###` subsections into the enclosing release's (so a
   release ends up with one `### Added`, one `### Fixed`, and so on rather than
   repeats). Collapse any run at the top into a single `## [Unreleased]`. Content is
   preserved verbatim; only headings move.

2. **Stop producing them.** A changelog entry is a `####` section appended under the
   existing `## [Unreleased]`'s matching `###`. Creating a second `## [Unreleased]` is
   the defect. Worth a check in CI, since this is exactly the class of mistake that is
   invisible in a diff (each PR's own diff looks correct in isolation - the damage is
   only visible in the merged file) and that nobody re-reads afterwards.

The CI check is one rule: at most one `## [Unreleased]` in the file, and it must be the
first `## [` heading. That is cheap, has no false positives, and fails on the PR that
introduces the second one rather than at release time when the two have already merged.

## Acceptance

- No `## [Unreleased]` heading appears after the first `## [X.Y.Z]` heading.
- Each release section has at most one heading of each `###` kind.
- No entry text is lost or reworded by the repair; `git diff` shows heading moves only.
- CI fails a PR that leaves more than one `## [Unreleased]` in the file, or that places
  it after a versioned heading.

PR #3463, opened 2026-09-16
ci(#3447): one gate job, and restore changelog headings, incremental and release

### LINKED ISSUE #3447: ci: eleven of dev's seventeen required checks are produced by no job, so every merge is an admin bypass
## What

`dev`'s ruleset requires seventeen status checks. Eleven of them are produced by no job in `.github/workflows/ci.yml`, so no pull request can ever satisfy them:

```
release x86_64-linux          incremental x86_64-linux      targets on ubuntu-latest
release aarch64-linux         incremental aarch64-linux     targets on ubuntu-24.04-arm
release x86_64-windows        test legs                     targets on windows-latest
link pinned (linux)           changelog headings
```

The workflow's jobs are `build`, `test`, `codegen`, `qemu`, `spirv`, `riscv32`, `link`, `dwarf`, `darwin`. Only `build ${{ matrix.target }}` and `test ${{ matrix.target }}` appear in the required list at all.

## Why it matters

Two consequences, and the second is the worse one.

**The gate never gates.** Every merge into `dev` is an `OrganizationAdmin` bypass, because there is no other way through. The last five merges (#3435, #3436, #3437, #3438, #3441) all went that way. A required check that can never pass is not a stricter gate than none, it is a gate that has been routed around by default, and it trains everyone to pass `--admin` without reading what is red.

**The checks that would catch a real regression are not required.** `codegen`, `link`, `dwarf`, `spirv`, `qemu riscv64`, `riscv32` and `darwin` all run and none of them is required. Those are the legs that catch backend and object-format regressions. So the one gate in the family is simultaneously impossible to satisfy and pointed at the wrong jobs.

This is the same defect as a step that cannot go red, inverted: a check that cannot go green is equally uninformative, and it removes the meaning from every merge that bypasses it.

## Family-wide state

| repo | dev ruleset | required checks | jobs that run but are not required |
| --- | --- | --- | --- |
| mach | yes | 17, of which 11 are phantom | codegen, qemu, spirv, riscv32, dwarf, darwin |
| mach-std | yes | 0 | linux x86_64, windows x86_64, darwin |
| mach-lsp | yes | 0 | all |
| hedge, laurel, mach-crypto, mach-acme, mach-tls, mach-http, mach-quic | none | n/a | all |

Seven repos have no `dev` ruleset at all.

## Wanted

- The required-check list on `mach` names contexts that the workflow actually produces, and names the legs whose failure should stop a merge. If `release`, `incremental`, `targets`, `link pinned` and `changelog headings` describe jobs that were intended and never built, that is a separate issue; the ruleset should not assert them until they exist.
- A merge into `dev` that needs `--admin` is a signal that something is wrong, not the normal path.
- The other repos get the same treatment, so "green" means the same thing across the family. Related: briar-systems/mach-std#664, which found 43 fmt-drifted files behind the same absence of enforcement.

## Acceptance

- [ ] Every context in `dev`'s required list is produced by a job in `ci.yml`.
- [ ] `codegen`, `link` and `dwarf` are required on the targets they run on.
- [ ] A pull request with all checks green merges without `--admin`, demonstrated once.
- [ ] Any required context that was removed rather than implemented is recorded here with the reason.

octalide:
Closes #3447

## What

- `gate` needs every job and runs with `if: always()`. It fails when any needed job's result is `failure` or `cancelled`, and a `skipped` job counts as a pass. Per the owner's ruling it becomes the only required check. The repo rulesets are not changed here.
- `changelog headings` is restored for every PR. It enforces the #3039 invariant (one `## [Unreleased]`, first, no repeated `###` within a release), with the check inline in `ci.yml`.
- `incremental <target>` is restored for every PR on both linux hosts, as `test/run.sh --incremental`. Warm no-op and post-edit rebuilds of the compiler, and a manifest-only fixture, must match clean builds.
- `release <target>` is restored for PRs into main and in the heavy picker. It runs the release-profile fixpoint and the unit suite of the binary `cd.yml` ships.
- No schedule trigger anywhere.

The #3447 comment records why the other six phantom contexts were dropped.

## Verified locally

- `run.sh --incremental` passes 3/3 with a from-source dev compiler. A mutation that feeds the clean binary in as the warm one fails both edit checks.
- The changelog check passes on the current file and fails on each of three mutations: a second `[Unreleased]`, `[Unreleased]` below a release, and a repeated `### Fixed`.
- The gate script passes on success+skipped and fails on failure and on cancelled.


octalide:
## CI evidence

- This PR, run 35121449957 (PR into dev): `changelog headings`, `build` ×3, `test` ×3, `codegen` ×2 and `incremental` ×2 passed. The main-tier jobs (`qemu`, `spirv`, `riscv32`, `link`, `dwarf`, `darwin`, `release`) were skipped. **`gate` passed.**
- Counterfactual #3464 (closed), run 35121465916: a second `## [Unreleased]` heading. `changelog headings` failed, everything else passed or was skipped, and **`gate` failed** with `##[error]gate: not passed: changelog`.

Issue #3532, opened 2026-09-17
refactor(config): one descriptor walker for the three physical-path walks


octalide:
## What

`src/lang/driver/config.mach` has three near-identical walks that open a path one component at a time by descriptor:
- `validate_physical_path`
- `validate_contained_ancestors`
- `ensure_physical_directory_path`

Each one reimplements the same steps: splitting components, `openat` with no-follow, the symlink and loop refusals, and closing the handles. So every fix has to be made three times. The std 4.0.0 migration had to edit all three, and it found a use-after-free and a format-argument mismatch in just one of them.

## Wanted

- [ ] One component walker. It takes a per-component policy for what to do at each step: validate only, require containment, or create missing directories. The three functions become thin callers of it.
- [ ] The existing tests stay green, plus one test per policy for a symlinked component and for a loop.

Issue #3469, opened 2026-09-16
resolve: a re-resolved positional case payload is left unbound, so a def-typed gate breaks err[Cause].err{Cause.missing{}}


octalide:
## What

A module that constructs a tag case with a positional payload that is itself a case construction, such as `err[Cause].err{Cause.missing{}}`, fails to type when **any** module in the build has a top-level `$if` whose gate is only decided in the gate fixpoint (for example a constant typed through a `def` alias) and whose branch declares something:

```
error: internal: this identifier reached type inference unbound; the name resolver neither bound it nor reported it
error: `<error>` is not a tag, so it has no cases to construct
  --> dep/std/src/types/error.mach:76:40
```

Found by mach-std (briar-systems/mach-std#685, PR briar-systems/mach-std#701), where it blocks the layering work from briar-systems/mach-std#686 onward. On 5.0.4 it only showed under `mach test`. On dev `c526fbbee` it breaks `mach build` too, as soon as the build reaches `std.types.error`.

## Reproduction (no std)

```mach
# cap.mach
pub def flag: u8;
pub val ON: flag = 1;

# gate.mach
use cap: dg.cap;
$if (cap.ON) { fun selected() i64 { ret 1; } }

# errc.mach
pub tag err[E]: u8 { err: E; ok; }
tag Cause: u8 { missing; refused: i32; }
pub fun check(key: i64) err[Cause] {
    if (key == 0) { ret err[Cause].err{Cause.missing{}}; }
    ret err[Cause].ok{};
}
```

With `pub val ON: u8 = 1;` it builds. It also builds with a gate that declares nothing, or without the outer generic tag.

## Cause

Resolve is not idempotent over its own AST rewrite. `bind_expr` turns a positional literal `T{...}` whose head names a case from `EXPR_KIND_VECTOR_LIT` into `EXPR_KIND_STRUCT_LIT` in place, keeping the payload in `elems_start/elems_len`, and binds those elems on that first pass. The `STRUCT_LIT` branch, however, binds only the named `fields`. A `def`-typed gate sends the build through the gate fixpoint, which resolves every module more than once over the same AST. On the second pass the node is already a `STRUCT_LIT`, so its positional payload is never bound and nothing is reported. Sema then sees an unbound head and types it `<error>`.

## Fix

The `STRUCT_LIT` binder binds the literal's positional elems as well as its fields, so a literal binds the same way whichever form it is in when resolve reaches it.

## Acceptance

[...]

PR #3533, opened 2026-09-17
chore(dep): build against mach-std 4.0.0

octalide:
Moves the compiler to mach-std v4.0.0: `[dep.std].ref` becomes `tag/v4.0.0`, and the gitlink moves to 90b2f11.

**Migration** (26 files, following std's MIGRATION.md)
- `os.E*` comparisons become `os.error_kind(n) == io_error.KIND`.
- `os.message`/`io_error.message`/`from_code` become `os.error_message`/`os.message`/`os.error`.
- `AT_FDCWD` becomes `working_dir()`, `STDERR_FD` becomes `stderr()`, and `stat_mode` moves to `std.filesystem.native`.
- Descriptors are `usize` handles with `os.INVALID_HANDLE`.

**Behaviour fixes a plain rename would have broken**
- **Removal refusals.** std 4 gives a removal's own refusals (a name too long, a final INVALID component) code 0.
  - `mach clean` and `mach dep remove` tested `code < 0`, so they would have reported those refusals as success.
  - `Removal` now carries `failed` and `kind`, and callers test `failed`.
  - The local ELOOP constants are gone.
  - `fmt` refuses a non-regular source with "source must be a regular file".
- **Out of memory vs a full disk.** std 4 files ENOMEM under `RESOURCE_EXHAUSTED`, together with ENOSPC, EMFILE and ENFILE.
  - The cache store treats ENOMEM as an internal failure but a full disk as "store unavailable". A straight swap would have made a full disk fail the build.
  - `store.memory_exhausted` now reads ENOMEM from the host OS module, and `driver/cache` uses it.
- **Two pre-existing bugs** in `validate_contained_ancestors` (`config.mach`): a component used after `str_free`, and a format string with 4 holes but 3 arguments.

**Tests**
- `output.remove.relative:refusals_fail_with_their_kind`
- `cli.cmd.clean.execute_entry:a_removal_refusal_is_an_error`
- `cli.cmd.dep.purge_beneath:a_removal_refusal_is_an_error`

Mutation-checked: reverting to `code < 0`, or dropping `failed: true`, fails them.

**Verified**
- The seed (5.2.1) builds the host, windows-x86_64, darwin-x86_64, darwin-aarch64, linux-arm64 and linux-riscv64. So the seed satisfies std 4.0.0's `mach 5.2.0+`.
- Unit suite: 3209 passed.

The three near-identical descriptor walks in `config.mach` were migrated in place. Consolidating them is #3532.


octalide:
The fixpoint failure came from the CI seed, not from dev. `seed-mach` defaulted to v5.0.0, which still has the #3469 bug, and std 4.0.0 requires mach 5.2.0 or later. 9686f6ddc moves the default seed to v5.2.1. Locally, 5.2.1 builds a, a builds b, b builds c, and `cmp b c` converges.

PR #3712, opened 2026-09-19
fix(#3711): substitute a type parameter by its declaration, not by position

### LINKED ISSUE #3711: sema: a generic instance substitutes type parameters by position, so a caller with a reordered or longer parameter list has its callee's body checked against the wrong types
## What

A generic that calls another generic with a type-argument list that is not position-for-position its own has the callee's body checked against the wrong types: `find[K, V, M, P]` calling `hk[K, M, P]` hands hk's `(K, M, P)` the caller's `(K, V, M)`. A `$type_of` gate inside `hk` then selects the wrong arm, and a field access in what should be the dead arm is refused against the wrong record. Reported by mach-std on 5.8.1 (its `find_extra` fixture); reproduced on 5.9.0. Wrong-code class: a body is judged, and would be lowered, against types the source did not name.

```mach
rec Natural {}
rec Custom {}
pub rec A[K, V] { k: K; v: V; policy: Natural; }
pub rec B[K, V] { k: K; v: V; policy: Custom; f: fun(*K) u64; }

#[inline]
fun hk[K, M, P](m: *M, p: P, key: *K) u64 {
    $if ($type_of(p) == Natural) { ret (@key)::u64; }
    $or ($type_of(p) == Custom)  { ret m.f(key); }
    $or { $error("bad"); }
}

fun find_same[K, M, P](m: *M, p: P, key: *K) u64 { ret hk[K, M, P](m, p, key); }      # compiles, runs
fun find_last[K, M, P, V](m: *M, p: P, key: *K) u64 { ret hk[K, M, P](m, p, key); }   # compiles, runs
fun find_extra[K, V, M, P](m: *M, p: P, key: *K) u64 { ret hk[K, M, P](m, p, key); }  # refused
```

```
error: no such field `f` on type M
  --> src/m.mach:16:60
   |
16 | fun find_extra[K, V, M, P](m: *M, p: P, key: *K) u64 { ret hk[K, M, P](m, p, key); }
   |                                                            ^^^^^^^^^^^^^^^^^^^^^^
   |
  --> src/m.mach:9:40
   |
 9 |     $or ($type_of(p) == Custom)  { ret m.f(key); }
   |                                        --- in this generic body, checked against this instance's type arguments
```

The tell is which callers work: any caller whose parameters sit at the positions hk's do (`find_same`, `find_last`, and `find_ok[K, M, P, V]` in my own repro) compiles, and only a list where a parameter of the callee maps to a caller parameter at a different index fails. A variant with the extra parameter first, `find[V, K, M, P]`, fails the same way.

## Where

- `src/lang/fe/sema/generics.mach:111` `substitute_inner`: a `TYPE_GENERIC_PARAM` is replaced by `args[index]`, the parameter's index alone; nothing ties the parameter to the declaration being instantiated.
- `src/lang/fe/sema.mach:520` `revalidate_one` sets `tsc.subst = args` for the instance, and `context.apply_subst` (`context.mach:529`) runs on every type the checker reads: annotations (`resolved_type_of`, `context.mach:510`), expression types (`context.mach:1314` `resolve_type_comparison_operand`, `sema.mach:96`) and freshly interned parameters (`context.mach:1496`).
- An open instance (the caller's own parameters as arguments) therefore holds types that are the *caller's* parameters after one substitution, and every later `apply_subst` maps those by index into the callee's argument list again: with `find_extra`, `find.M` (index 2) becomes `args[2] = find.P` and `find.P` (index 3) falls off the end (`TYPE_ERROR`), so the `$type_of(p)` verdict is not "awaiting an instance" and the Custom arm is inferred against a type that is not `p`'s. When the caller's indices coincide with the callee's the double substitution is the identity, which is why every same-position caller works and hid this.

The instance record and the substitution need to identify the parameter being replaced with the declaration whose instance it is (a parameter type already carries `owner` and `owner_name`, `type.mach:482`), or substitution must run exactly once per type. Sibling: #3710 (an open instance is checked loudly at all), found in the same investigation; separate mechanism, and fixing #3710 by not checking open instances would hide this one rather than fix it, since a fully substituted instance reached through two generic layers goes through the same double mapping.

## Acceptance

- [ ] `find_extra` compiles and `get_extra[i64, u8]` runs and returns the same value as `get_same` and `get_last` (mach-std's fixture).
- [ ] A driver test that a callee parameter maps to the caller's argument at the callee's position and never to the caller's parameter of the same index: a caller `[V, K, M, P]` and a caller `[K, V, M, P]`, red on dev first.
- [ ] A concrete two-layer instance (`get_extra[i64, u8]` → `find_extra` → `hk`) lowers `hk` against `(i64, A[i64, u8], Natural)`, asserted on the selected arm.

octalide:
Closes #3711

## What

A generic instance replaced every `TYPE_GENERIC_PARAM` by `args[index]`, whichever declaration the parameter belonged to. In an open instance (a callee called with the caller's own parameters) the callee's body held the caller's parameters after one substitution, and every later `apply_subst` mapped those by index into the callee's argument list again: with `find_extra[K, V, M, P]` calling `hk[K, M, P]`, `find.M` (index 2) became `args[2] = find.P` and `find.P` fell off the end. Same-position callers made the second pass the identity, which hid it.

Substitution now carries the declaration whose parameters it replaces (`type.GenericOwner`, the owner key a parameter type already records) and leaves a parameter of any other declaration alone, so a second pass is the identity. Every scope that substitutes names its owner from one derivation, `context.generic_owner_of_decl`, which also interns the parameters:

- sema: `revalidate_one`, `reinfer_instance_body`, `reinfer_pack_each_body` (owner threaded from the lowering scope), the explicit type-argument call path (`compute_generic_signature`, owner = the callee's definition), the union-secrecy walk (owner switches to the nominal at each instance level).
- lowering: `ModuleScope.subst_owner` set beside `subst` at the four instance sites.
- An anonymous `rec`/`uni` declared inside a generic mentions the enclosing declaration's parameters; it is instantiated exactly when the pass's owner is that declaration (`nominal_param_owner`, read from the parameters its fields mention). A named generic nominal never appears bare (resolve refuses it).

## Verification

- `mach.lang.fe.sema.generics.substitute:replaces_only_the_owners_parameters`: foreign parameter left alone, second pass identity, caller's own pass resolves by the caller's position, anonymous aggregate instantiated by its enclosing owner only.
- `mach.lang.driver:generic_arguments_map_to_the_callee_parameters_not_the_caller_positions`: `[K, V, M, P]` and `[V, K, M, P]` callers through concrete two-layer instances lower the arm the policy names (constant use counts in the IR), and a concrete instance whose arm reads a missing field is still refused with the trail. Red on the seed (5.9.0 = dev): `no such field`.
- Corpus case `comptime/generic_reorder` (C reference, all native targets blessed, differential green under qemu too; spirv skip: empty marker records have no SPIR-V struct type). Red on the seed.
- Mutation: dropping the `param_owned_by` test makes both new tests fail.
- Full unit suite through the built compiler (`out/linux-x86_64/<profile>/bin/mach test .`): 3536/3536 debug, 3536/3536 release. Corpus `test/run.sh --qemu` with the built compiler: 985 pass, 0 fail, 167 skip. `mach fmt --check .` clean.
- The merge commit from `origin/dev` is the repo's usual integration shape for a branch PR (as on #3705), not a conventional commit by design.


octalide:
Local run on 01b6f345c (same source, format-only last commit): unit suite 3536/3536 debug and 3536/3536 release; corpus `test/run.sh --qemu` 985 pass, 0 fail, 167 skip (the new `comptime/generic_reorder` blessed on every native target and run under qemu on riscv32/riscv64; spirv skip recorded in SKIPS).

